Issue #1, opened 2016-01-21
Input/Output module ...


kif:
HDF5 is part of the core of the data policy at ESRF. The natural binding for hdf5 is h5py.
The idea is to provide an API similar to h5py in silx when addressing images and specfiles.

Today: there is 1 binding for SpecFile and 2 for images: EdfImage and fabio.
H5py offers a Dataset class with _attrs_ which matches pretty well with the data/header dichotomy in image formats. Re-implementing the behavior (or subclassing?) Dataset solves most of the discussion about the API. The second advantage is about the formation to silx as trainee have to learn only one API.
# Difference between HDF5/other format

The main difference between HDF5 and  other is the hierarchical structure which is exposed as a Python dictionary where most of the other formats are using a linear structure.  The natural design would then be to use (/expose?)  a Python list as container to Dataset instances in a File object.  
# About file/file-series

The h5py File module provides a "driver" option to select the way data are saved. Moreover, file series would be represented as "prefix_%04d.img" to create a unified view, like what is done in h5py.
# Work to be performed:
- Implement Dataset, Attribute, and File
- Provide drivers for the various file formats (re-using the code existing in the pre-existing libraries)
# Advantages:

The transition to HDF5 over the period 2010-2015 has been very slow, most scientists being:
- reluctant to change something which already works
- don't want to learn yet another API
  If one wants actually to promote HDF5, offering the same API between HDF5/legacy would allow to bridge the gap.

vasole:
- I like the idea of having the API used for the most complete file format applicable to the other simpler format.

- I would leave the file-series for later discussion. HDF5 (not h5py) does that not to select a series of files but to split a single file. The concept of a stack of files falls on a different use case. You can have stacks of files using the family driver for instance.

- I do not see the need to implement Dataset, Attribute and File. When converting from specfile or EDF to HDF5 one does not need Attributes for instance, but one certainly needs Groups. If you mean not to raise an exception when trying to access the attributes instead of returning the same thing h5py returns when there are no attributes then I agree.

vasole:
Slides of presentation: [silx-io.pdf](https://github.com/silx-kit/silx/files/107826/silx-io.pdf)

vasole:
Summary
=======

- The proposal to access SPEC files in the same way as HDF5 files was well received

- The most repeated word from the meeting participants was "official":

    - Official SPEC to NeXus converter
    - Official layout of ESRF NeXus files

vasole:
[...]

PR #4, opened 2016-02-24
SpecFile cython binding

PiRK:
This new `silx.io.specfile` module provides 3 classes for reading specfiles: `SpecFile`, `Scan` and `MCA`. 

`SpecFile` provides a low level binding to the main C SpecFile functions, as well as a `__getitem__` method to get a scan and a `keys` method to get a list of scan keys (e.g. `['1.1', '2.1', …]`). It also works as an iterator to loop through all scans.

`Scan` provides properties and methods to retrieve scan data and metadata/headers.  In particular, `Scan.data` is a property returning a 2D numpy array of scan data (as doubles) on request,  `Scan.mca` is a property returning a `MCA` object and `Scan.header` returns a list of raw headers lines (file, scan and mca headers).

`MCA` is the interface to access MCA data. It works as an iterator to loop through MCA data lines (1D numpy arrays of doubles), and  can return a single MCA line through its `__getitem__` special method.


t20100:
PR related to #1.

Looks good to me.
First real code of the project!

PR #8, opened 2016-03-02
Histogramnd + small math module fix.

dnaudet:
**There is small fix to the math module included in this PR (silx/setup.py), let me know if you want me to do a separate PR for it.**

Pull request for the histogramnd function.

* 3 supported types for the sample and the weights -> 6 different functions
* 3 test suites (+ some benchmark code) :
  * nominal (small "hand written" data sets)
  * error cases (invalid arguments : shape, type, dtype, ....)
  * test vs numpy (large random data sets)
  * a simple benchmark function (not included in the tests)
  
**For some reason the doc doesnt display the full signature of the histogramnd function...**

Infos/questions/RFC :
* I used np.float32 in numpy and float in C, but I'm not sure the two will always be compatible...
* I can change the argument names if you have better ideas.
* in the C code I used some macro to avoid duplication of code when adding support for more data type. This isn't pretty, but I haven't got much choice (since there are no templates support in C)...
* I added a short README describing how to add support for other types, I'm not sure this is the right way to do it... (but I don´t think this should go into the public documentation. Or should it?)
* when building I see there's a warning about a "missing cimport in module...". I couldn't figure out how to solve that.
* I could have used a macro to declare each functions in the C header file (like I do in the source file), but I think it's better to have them expliclitly declared for the sake of readability.
* I'd like someone to review the cython code, please... I probably didn't do things in the best possible way...


PiRK:
Sphinx cannot detect function/method signatures in cython modules. You can include the signature as the first line of your docstring for Sphinx.
Example: 
https://github.com/silx-kit/silx/blob/master/silx/io/specfile/specfile.pyx#L227
https://github.com/silx-kit/silx/commit/4703edf9636d114607043cf0a802d78264eb7891

vasole:
np.float32 and C float will always be compatible.

dnaudet:
> Sphinx cannot detect function/method signatures in cython modules. You can include the signature as the first line of your docstring for Sphinx.

Thanks I had tried that already, but I just figured out that you can't split it several lines.

kif:
A few comments ...
* Why not using cython fused types to write a single function for all input types ? 
* using array.astype(dtype) enforces a copy, even if it already the right type. Using array.ascontiguousarray(dtype) will only make a view which is much faster (but don't modify it inplace !)
* Why weights is not None by default to have unit weights ?
* using cython typed memory-views would prevent the use of manual malloc & free. Leaving the memory-management this is proven to be less error prone.

vasole:
- The malloc functions should be checked at the C level and in case of failure any previously allocated memory should be freed and the function should return a different value than 0.

dnaudet:
> Why not using cython fused types to write a single function for all input types ?

Great, this is what I've been looking for, thanks!

> using array.astype(dtype) enforces a copy, even if it already the right type. Using array.ascontiguousarray(dtype) will only make a view which is much faster (but don't modify it inplace !)

I have tried ascontiguouasarray before but iirc it gave me an error, I'll give it another try.

> Why weights is not None by default to have unit weights ?

unit weights == bin counts? histogramnd returns both the bin counts and the "weighted histogram" (btw you have to call numpy.histogramdd twice to get the same result). I can turn the *weight* argument into a keyword parameter so that the user doesn't have to provide a *weight* array, and I will change the C code as well so that we don't have to allocate an array at all.

> using cython typed memory-views would prevent the use of manual malloc & free. Leaving the memory-management to Python is proven to be less error prone.

I did it that way to allow direct access to the C function. There is no C allocated memory passed to python (if that's what you had in mind).

dnaudet:
@vasole 
> The malloc functions should be checked at the C level and in case of failure any previously allocated memory should be freed and the function should return a different value than 0.

OK.

dnaudet:
- [x] remove unused imports
- [x] turn *weights* argument into a keyword parameter  (default=None).
  * if weights is provided : histogramnd returns a tuple (bin counts, weighted histogram)
  * if weights is None : histogramnd returns a tuple (bin counts, None)
- [x] add a specific return code for the memory allocation errors
- [x] replace stdint's int32_t and uint32_t by their numpy equivalent.
- [ ] investigate cython fuse types
- [ ] add a "known limitations" test (expected failures around the bin edges)? (just for the sake of documentation)

vasole:
Build fails with windows python 2.7 due to stdint.h not being available in VS 2008.

You can take a look at http://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio where a public domain alternative is suggested.

http://snipplr.com/view/18199/stdinth/

Do we need the full header? Can we just add the missing defines?

kif:
On Sat, 05 Mar 2016 07:49:53 -0800
V. Armando Solé <notifications@github.com> wrote:

> Build fails with windows python 2.7 due to stdint.h not being available in VS 2008.
> 
> You can take a look at http://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio where a public domain alternative is suggested.
> 
> http://snipplr.com/view/18199/stdinth/
> 
> Do we need the full header? Can we just add the missing defines?
> 
There is a solution in fabio about this ...

dnaudet:
**About the fused types**
Unless I missed something important I don't think the fused types would help here since in the end we still have to choose the c function to call depending on the sample and weights type (so we still end up with all those ugly if/else statements).

kif:
On Sun, 06 Mar 2016 13:19:12 -0800
dnaudet <notifications@github.com> wrote:

> **About the fused types**
> Unless I missed something important I don't think the fused types would help here since in the end we still have to choose the c function to call depending on the sample and weights type (so we still end up with all those ugly if/else statements).

If you want to go down to C, then they will not be of great help.
Cheers,
J.

dnaudet:
**stdint/inttypes.h**
Available implementations :
* https://msinttypes.googlecode.com/svn/trunk/stdint.h.
There is a previous version of this one used in Fabio (from 2008 judging by the header, @kif am i right?)
* http://snipplr.com/view/18199/stdinth/

Any preferences?

kif:
On Mon, 07 Mar 2016 04:36:04 -0800
dnaudet <notifications@github.com> wrote:

> **stdint/inttypes.h**
> Available implementations :
> * https://msinttypes.googlecode.com/svn/trunk/stdint.h.
> There is a previous version of this one used in Fabio (from 2008 judging by the header, @kif am i right?)
> * http://snipplr.com/view/18199/stdinth/
> 
> Any preferences?

The second one has no copyrights -> trouble ahead

the first one is BSD-3clauses which is close to MIT. 

Declare in the copyright.txt:

Files: .../stdint.h
Copyright: 2006-2008 Alexander Chemeris
License: BSD-3

-- 
Jérôme Kieffer
tel +33 476 882 445

dnaudet:
Looks like the tests are OK on windows/2.7 at last.
Had to add a little exception to the error cases tests because numpy shapes are long integers in that configuration. I can´t say I´m very happy with that...

silx/math/histogramnd/histogramnd.pyx
```diff
@@ -0,0 +1,710 @@
...
+        cumul = None
+    elif cumul is None:
+        cumul = np.zeros(output_shape, dtype=np.double)
+    else:
+        if cumul.shape != output_shape:
+            raise ValueError('Provided <cumul> array doesn\'t have '
+                             'a shape compatible with <n_bins> '
+                             ': should be {0} instead of {1}.'
+                             ''.format(output_shape, cumul.shape))
+        if cumul.dtype != np.float:
```
t20100: Maybe replace np.float by np.float64

Issue #9, opened 2016-03-05
Build not working under windows


vasole:
Problem
-----------

- Specfile build crashes under windows

Solution
-----------

- When dealing with multiplatform developments, one has to take care and systematically use os.path.join to build paths.

vasole:
@PiRK Review required.

vasole:
Not sure if it is an issue or not.

- If Cython is installed on the machine, the generated code works.

- If Cython is not installed on the machine, the supplied code does not work.

PiRK:
I'll have a look at the cython issue.  I probably need to adjust the list of source files in io/setup.py if ``os.environ["WITH_CYTHON"] == "False"`` is set in the general setup.py, to remove the pyx file and include the .c file.

PiRK:
File extensions  don't seem to be the problem, as ``fake_cythonize`` (https://github.com/silx-kit/silx/blob/master/setup.py#L296)  already takes care of replacing ``.pyx`` with ``.c`` in the absence of cython.

kif:
On Sun, 06 Mar 2016 23:43:17 -0800
Pierre Knobel <notifications@github.com> wrote:

> File extensions  don't seem to be the problem, as ``fake_cythonize`` (https://github.com/silx-kit/silx/blob/master/setup.py#L296)  already takes care of replacing ``.pyx`` with ``.c`` in the absence of cython.

Look at the version of cython used and enforce the minimum version.

Armando, which version have you been using ?

vasole:
- If I am not mistaken, I have been using cython > 0.17 in order to use memory views.

- FreeART wraps more advanced C++ features and I think it needs 0.22 minimum.

- I do not anticipate problems with the Cython version requirements if we make sure we distribute generated code.

kif:
is this issue fixed ?

PiRK:
We fixed the problem preventing building when cython isn't installed (#13). The previous  issue related to not using ``os.path`` has also been also fixed (#10).

PR #10, opened 2016-03-05
Allow build under windows.

vasole:
Addresses #9

### DIFF
--- a/silx/io/setup.py
+++ b/silx/io/setup.py
@@ -58,8 +58,8 @@ elif os.name.lower().startswith('posix'):
 else:
     define_macros = []
 
-current_dir = os.path.dirname(__file__)
 def configuration(parent_package='', top_path=None):
+    current_dir = os.path.dirname(__file__)
     config = Configuration('io', parent_package, top_path)
     config.add_subpackage('test')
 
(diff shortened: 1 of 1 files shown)


kif:
2 comments: 

1. one should probably use "os.path.dirname(os.path.abspath(__file__) )" instead of "os.path.dirname(__file__)", especially in setup.py which under certain circumstances results in "".

2. this "urgent pull request" highlights the lack of test for the given feature. The continuous integration tool did not spot it.

vasole:
On 06.03.2016 09:40, Jerome Kieffer wrote:
> 2 comments:
> 
> one should probably use "os.path.dirname(os.path.abspath(FILE) )"
> instead of "os.path.dirname(FILE)", especially in setup.py which under
> certain circumstances results in "".
> 

- Yes

> 
> this "urgent pull request" highlights the lack of test for the given
> feature. The continuous integration tool did not spot it.
> 

Things to be solved {are they worth a specific issue?):

- How can we test build without Cython besides build without Cython? Can 
we specify two builds?

- Same as above but running the build of the sources generated with 
sdist.

vasole:
- If possible, I would aim for the continuous integration tests to be run on the sources generated by sdist and not on a repository copy.

- I guess the build from the git sources is tested almost every day :-)

t20100:
+1 for test from sdist and without Cython in at least 1 CI.
See issue #12

Issue #13, opened 2016-03-07
C source files not included when building without cython


PiRK:
When running ``python setup.py build`` on a system (or virtualenv) without cython being installed, C source files ended up not being linked during compilation. This caused following error when trying to run tests: 
```
from silx.io.specfile import SpecFile, Scan
ImportError:
/users/knobel/nocython/local/lib/python2.7/site-packages/silx-0.0.1a2-py2.7-linux-x86_64.egg/silx/io/specfile.so:
undefined symbol: SfError
```

PiRK:
Fixed in pull request #14

Issue #12, opened 2016-03-07
CI test sdist cython not installed


t20100:
In at least 1 CI system:
- [x] Building using sdist generated tarball.
- [x] Building without Cython installed.

See PR #10

PR #14, opened 2016-03-07
Fix issue #13: C source files not included in extension

PiRK:
Issue #13: Function ``fake_cythonize`` rejected ``.c`` files at the same time as it was renaming ``*.pyx`` into ``*.c``. (found by @kif)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -313,7 +313,7 @@ def fake_cythonize(extensions):
                     source = base + '.c'
                 if not os.path.isfile(source):
                     raise RuntimeError("Source file not found: %s" % source)
-                new_sources.append(source)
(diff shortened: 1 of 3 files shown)


PiRK:
Two additional changes in the same commit:
- I removed h5py and cython requirement in the general ``setup.py``
- I changed ```setup(configuration=configuration)``` into ```setup(configuration=configuration())``` in ``silx/io/setup.py`` (```configuration``` is a function that returns a ```numpy.distutils.misc_util.Configuration``` object)

kif:
Looks good to me.
Nevertheless issue #12 looks of prime importance to prevent such regression !

kif:
now close :)

PR #16, opened 2016-03-08
Fix #15 : WITH_CYTHON environment variable check

### LINKED ISSUE #15: Setup uses cython even if WITH_CYTHON is False
Currently, if setup.py is called with the environment variable WITH_CYTHON set to
* True : setup doesn't use cython, the pre-generated C files are used.
* False : setup uses cython

Judging by the code (the other lines all set WITH_CYTHON to False when not using cython) this should probably be the opposite.

dnaudet:
Fix for issue #15.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -268,7 +268,9 @@ def check_cython():
     """
 
     if "WITH_CYTHON" in os.environ:
-        return os.environ["WITH_CYTHON"] not in ["False", "0", 0]
+        if os.environ["WITH_CYTHON"] in ["False", "0", 0]:
+            os.environ["WITH_CYTHON"] = "False"
+            return False
 
     if "--no-cython" in sys.argv:
         sys.argv.remove("--no-cython")
--- a/setup.py
+++ b/setup.py
@@ -268,7 +268,7 @@ def check_cython():
     """
 
     if "WITH_CYTHON" in os.environ:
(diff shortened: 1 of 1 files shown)


vasole:
@kif

Is there some code missing? The code I was expecting was something in the lines of:

.. code:python

    # check if cython is not to be used despite being present
    def use_cython():
        """
        Check if cython is disabled from the command line or the environment.
        """
        if "WITH_CYTHON" in os.environ:
            if os.environ["WITH_CYTHON"] in ["False", "0", 0]:
                print("No Cython requested by environment")
                return False
    
        if ("--no-cython" in sys.argv):
            sys.argv.remove("--no-cython")
            os.environ["WITH_CYTHON"] = "False"
            print("No Cython requested by command line")
            return False
        return True

vasole:
In other words the line:

 return os.environ["WITH_CYTHON"] == "False"

should have been changed to:

return os.environ["WITH_CYTHON"] not in ["False", "0", 0]

t20100:
In this case, if WITH_CYTHON is "0" or 0, I would change the environment variable to "False" so that setup.py in sub-packages just have to check os.environ["WITH_CYTHON"] != "False".
If we leave multiple values meaning False, I'm pretty sure it's not gonna be tested everywhere...

PR #17, opened 2016-03-09
[CI] Use sdist on Linux and build without cython

### LINKED ISSUE #11: MANIFEST.in missing *.h and *.c
I know it is too early, but I put this issue to keep in mind that at each release we have to perform tests to check if we can generate the library from the sources obtained with python setup.py sdist.

### LINKED ISSUE #12: CI test sdist cython not installed

t20100:
- Closes #11
- Closes #12: use sdist on Linux
- Rename README.md -> README.rst
- Remove pip caching on Travis


kif:
Looks OK, tests are passing...

PR #24, opened 2016-03-21
Spec h5

PiRK:
New features: 

- SpecFile h5py-like API
- SpecFile to HDF5 conversion
- nested dictionary to HDF5 

Minor feature:

- ``io.utils.repr_hdf5_tree`` to build a string with the structure of a HDF5 file (I'm using this in tests and I think it can be a useful tool to list the content of a HDF5 file)

```
In [1]: from silx.io.utils import repr_hdf5_tree 

In [2]: print(repr_hdf5_tree("Downloads/sample.h5"))
+fields
	+fieldB
		-z=<HDF5 dataset "z": shape (256, 256), type "<f4">
	+fieldE
		-x=<HDF5 dataset "x": shape (256, 256), type "<f4">
		-y=<HDF5 dataset "y": shape (256, 256), type "<f4">
+particles
	+cellidx
		-x=<HDF5 dataset "x": shape (10000,), type "<i8">
		-y=<HDF5 dataset "y": shape (10000,), type "<i8">
	+pos
		-x=<HDF5 dataset "x": shape (10000,), type "<f8">
		-y=<HDF5 dataset "y": shape (10000,), type "<f8">
	+vel
		-x=<HDF5 dataset "x": shape (10000,), type "<f8">
		-y=<HDF5 dataset "y": shape (10000,), type "<f8">
		-z=<HDF5 dataset "z": shape (10000,), type "<f8">
```


vasole:
- It looks OK. Reading is as easy as using a dictionary that was the intended goal.

@kif @t20100 

- I just wonder if we should aim to a shorter file and class name. Something in the lines of spech5 for the file and SpecH5 for the class instead of specfileh5 and SpecFileH5

- The repr_hdf5_tree also looks too long. How does it compare to h5dump?

t20100:
Shorter names sounds fine

PiRK:
I've renamed all occurrences of ``SpecFileH5`` to ``SpecH5`` in module names and class names. 

Regarding ``repr_hdf5_tree``, it does not provide as many options as the ``h5dump`` command-line tool. It's actually a bit more similar to ``h5ls -r`` (https://www.hdfgroup.org/HDF5/Tutor/cmdtoolview.html#ex3), except that it returns a string and does not print it to the terminal. So I'm thinking ``h5repr`` or ``h5ls`` would be better. 

We could add a couple of command line arguments to the function to implement our own versions of  ``h5ls``. I'm thinking of something like: 

``h5ls(path_to_h5_file, pathtogroup="/", recursive=False)``

PR #27, opened 2016-03-29
Histolut

dnaudet:
**For review**

I will add the tests and complete the docs if accepted.
This code is not optimized (no fancy algorithm...)... It's faster than silx.math.histogramnd when more than one data set is used with the same LUT, but it can be improved a lot.
**It's written in cython, but I can easily rewrite it in C, no problem.**

I wrote too functions, the second one takes some input generated by the first. I could also write a class that hides all this, I don't know what's best.

Two functions : 
- histogramnd_get_lut : returns the LUT that can be used to build an histogram.
 - input : sample, bin ranges, number of bins, ...
 - returns the LUT and the shape of the histogram
- histogramnd_from_lut : builds the histogram.
  - input :
    - weights
    - LUT
    - (histogram shape and histogram dtype (*optional*)) OR histogram
    - min/max weights to be filtered out.
 - returns : the weighted histogram


vasole:
It is OK with me. Please add doc strings and tests.

vasole:
@dnaudet 

Are you already using this functionality?

dnaudet:
For some reason i got the hole upstream history in this pull request.

dnaudet:
Test.

dnaudet:
feea9ff fixes #58

Issue #2, opened 2016-01-26
About Plot1D2D


vasole:
Discussion Introduction
==================
[Plot1D2D.pdf](https://github.com/silx-kit/silx/files/104596/Plot1D2D.pdf)

Conclusion
=========

- Make clear we aim to have visualization widgets that implement common needs with an added value than just display. Example: there are many solutions to visualize an image, but if we want to offer the end user the possibility to change display options, extract profiles, calculate masks and more, it is better to have a common solution maintained by all of us than a set of individual solutions.

- To have an OpenGL backend is certainly a warranty for performance. High quality output can be assured by the matplotlib backend for instance.

- Current exposed methods are not considered fundamentally wrong. From a practical point of view they can be taken as starting point.

- New methods need to be added to accommodate developers’ need to access the items in the plot in order to efficiently update them.

- Current approach of choosing PlotWindow functionality at instantiation time needs to be completed by a simple enabling procedure.

- Missing (already foreseen) features: error bars, contour plots, iso-lines, histograms.

- Other missing features: maximum filter for image display, copy widget window (to clipboard, to file, ...)

- Logo contest: It was decided to have a logo based on the “vertical silex stone” with letters model. 

- There is need for early training and documentation.

t20100:
For the record, a few comments regarding the plot API (this does not change the need for refactoring and documentation of existing code):

* Need for a plot API?

  - Pros: Be able to switch rendering backend (e.g. matplotlib, OpenGL...).
  - Cons: Support a subset of functionnalities, yet another plot API, add complexity to the code.

* Some options regarding the API provided that there is no consensus on it:

  - Keep the API as it is in PyMca to support legacy.
  - Polish PyMca API but keep the general idea: limited changes in legacy.
  - Rename to use matplotlib like names:

    - Complete change of legacy code.
    - Expected to be easier for matplotlib users but it has some naming conflicts: matplotlib is PEP8 versus Qt CamelCase convention.

  - Write interaction directly over matplotlib (e.g. pan&zoom proof of concept: https://gist.github.com/t20100/e5a9ba1196101e618883):

    - Pros: Leverage all the features of matplotlib (with care in the interaction implementation)
    - Cons: Completely break PyMca's implementation, bound to matplotlib.

* What about notebook/web support?

I think the plane is the second option since the beginning, but the last option should be kept in mind to provide a more generic option if needed.

For longer term, a few links about a common plot API over multiple backends:

- http://www.almarklein.org/future_vis2.html
[...]

PR #31, opened 2016-04-01
Plot

t20100:
Related to #2 
This PR adds a gui packages with:
- qt.py to handle the different Qt bindings
- icons.py to provide access to icons (to be updated when data file handling is available)
- test package with a TestCaseQt (in silx/gui/test/utils.py) to support writting tests for Qt widgets.
- plot package that provides:
    - Plot.py: implementation of the PyMca Plot.py API at a single place. Its public API is mostly compatible with PyMca. The main API changes are:
       - an autoreplot mode (on by default),
       - the addition of get/setInteractiveMode methods to change the interaction
       - Separation of user-defined 'info' and curve/image 'style' parameters: Methods retrieving curves and images return an additional field with the item parameters.
    - PlotWidget.py: A Qt widget inheriting from Plot.py
    - PlotWindow.py: Adds a toolbar to PlotWidget. It is a subset of PyMca PlotWindow

    It is based on a matplotlib backend with a backend API simplified from PyMca plot backend in that is does not handle default values and does provide automatic behavior.

For now CI test of silx/gui are only run on appveyor (i.e. on Windows).
To disable the Qt tests, use the environment variable WITH_QT_TEST:
`WITH_QT_TEST=False python run_test.py`


vasole:
- Changes on qt.py sent to your project repository
- addImage inconsistency. We have xlabel, ylabel, xScale, yScale.
- In issue #2 you wanted to drop xScale and yScale in favor of origin and scale and I accepted it. Did you change your mind or you expect to change it in the future?
- The same applies to the insertMarker instead of addMarker
- Concerning the non-implemented getCurves method. Is it because you are hesitating about what to do (return an object or a dictionary) or because you consider it is simply not worth to implement it (that means you do not expect much evolution)? I was expecting to keep getAllCurves returning a list to simplify PyMca refactoring and to have your proposed getCurves returning an easier to maintain output than a list (ex. a dictionary). We can see the problem associated with the lists with your reimplementation of getAllCurves. If I have anything to say, I certainly prefer to return dictionaries rather than lists whenever possible.

t20100:
I did not made all the API modifications proposed in #2 .
But I think I'm still OK with what I proposed.

t20100:
Updated to add #2 API changes that were fully agreed.
The PyMca Plot API remains available (though with deprecation warnings).

For adding getCurves, I'd rather prefer to do it in another PR (same for adding zoom with e.g. +/- keys).

List of effective API changes from #2 :

Qt-like API typos
------

- Rename isActiveCurveHandlingEnabled, enableActiveCurveHandling to isActiveCurveHandling, setActiveCurveHandling
- Rename invertYAxis to setYAxisInverted
- Rename keepDataAspectRatio to setKeepDataAspectRatio
- Rename showGrid to setGraphGrid and add getGraphGrid
- Add missing getters: isDefaultPlotPoints, isDefaultPlotLines

add\* Methods
-------

- Rename insert\*Marker to add\*Marker
- addImage: change xScale and yScale arguements to origin and scale

clear and remove methods
------

- keep clear* and remove* methods but add a remove(legend=None, kind='all') that allows to do all clear* and remove* operations

replot
------

- Remove replot argument from all methods.
- autoreplot feature, with the possibility to disable autoreplot and explicitly handle it.

interaction
------

- Replace isDrawModeEnabled, setDrawModeEnabled, getDrawMode, isZoomModeEnabled, setZoomModeEnabled by getInteractiveMode, setInteractiveMode
- Add a pan and zoom mode
- Panning should be able while drawing (done with arrow keys).

Misc.
-----

- Add different signals for each kind of plot events (and keep the general one).
  This is partially done (done on a need basis) for events that can be directly connected to a slot (e.g. changing an axis log scale triggers an signal that can be directly connected to the corresponding toolbar button).
- Deprecate **kw from all methods

vasole:
- Very nice and fluid!

PR #42, opened 2016-04-11
Update readme and copyright

### LINKED ISSUE #5: specfile C library license is LGPL v2
Can we change it to MIT?
Otherwise, add an entry for it in copyright file to remember it.

### LINKED ISSUE #20: Readme is empty
The chosen abstract (single liner) is:
Toolbox for X-Ray data analysis 

May be replaced by smth better

t20100:
Closes #20 
Closes #5

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,12 @@
 silx toolkit
 ============
 
+The silx project aims at providing a collection of Python packages to support the development of data assessment, reduction and analysis applications at synchrotron radiation facilities.
+It aims at providing reading/writing different file formats, data reduction routines and a set of Qt widgets to browse and visualize data.
+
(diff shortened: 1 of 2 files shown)


vasole:
Shouldn't we say that the GUI depends on one of PyQt4, PySide or PyQt5?

t20100:
It SHOULD work on PyQt5 and PySide, but so far it's only been truly tested on PyQt4.
To date, automatic tests with PySide do not work because a method is missing in QTest, and it's not really been tested on PyQt5.
I'd keep that for later.

vasole:
Well, I have been using it with PyQt5.

By other hand, I do not consider an error to put in the readme where it should work. If it does not, it is a bug.

PR #47, opened 2016-04-13
Update Qt test to support Qt5

t20100:
- Close and delete PlotWidget in tests.
- Init a list of exisiting widgets that are not under silx control before running the tests.
  Then use this list when checking if tests are leaking widgets.


vasole:
That widget created by Qt (or PyQt) itself is really weird. Anyways, we found a clean way to deal with it.

Issue #48, opened 2016-04-13
Test PyQt5 wheels


vasole:
[]- Test supplied PyQt5 wheels.

[]- Figure out under which platforms are useful

Please **do not** install them on reference Mac. I need universal builds in order to be able to generate frozen packages.

Message from PyQt5 mailing list follows:

I've created a new set of wheels for the latest snapshot of PyQt5 v5.6.

Wheels for OS X and Windows can now be installed with pip3 from PyPi. Just do...

    pip3 install PyQt5

...which will also automatically install the approprate sip wheel as well.

PyPi doesn't host Linux wheels so that can be downloaded from...

    https://sourceforge.net/projects/pyqt/files/PyQt5/PyQt-5.6/

There is a link to the Linux wheel for sip on the sip download page.

The OS X and Linux wheels only have internal changes. If you installed the previous versions then there's not much point in installing these new ones.

The Windows wheels should now include the missing MSVC DLL and the OpenSSL DLLs.

Phil

t20100:
PyQt5 wheels are available on Windows and Mac (though only tagged for 10.6) ... and 'manylinux'...
For Python 3.5 only.
Integrated in appveyor CI, see #51

PR #51, opened 2016-04-14
[CI] Run tests with PySide and PyQt5 on Windows

t20100:
Make appveyor runs 3 tests:

- python 3.5, PyQt5 (wheel from pypi)
- python 2.7, PySide (wheel from pypi)
- python 2.7, PyQt4 (wheel from www.edna-site.org)

Relates to #48


vasole:
Great!

PR #52, opened 2016-04-21
Dictionary to HDF5, JSON and INI (dump and load)

PiRK:
This pull request renames module `io.dicttoh5` to `io.dictdump` and adds support of JSON and INI (handled by new module `io.configdict`, copied from `PyMca5.PyMcaIO.ConfigDict` with some changes). 

Main functions in `io.dictdump`:

- `dump`: basic dump of dictionary to HDF5, JSON and INI files
- `load`: basic loading of dictionary from HDF5, JSON and INI files

Files are passed to these function as file names or file handles (IO streams).

`dump` and `load` rely on:
- `dicttoh5`: slightly more customisable dictionary to HDF5 dump (`h5path`, `overwrite_data`, passing `h5py.Group.create_dataset` arguments)
- `h5todict`: basic loading of dictionary from HDF5, all dict values will be numpy arrays.
- `dicttojson`: call to `json.dump`, with parameter `indent` for pretty-printing
-  `dicttoini`: Call to `io.configdict.ConfigDict(ddict).write()`, to save dictionary to INI. Support of int, float, string, boolean, list, list of lists,  numpy arrays (1D and 2D).
- `io.configdict.ConfigDict().read` for loading back INI to dict.

In the INI format, all values are saved as strings, so some guess work is needed to load values back in their proper type. There are a few improvements compared to `PyMca5.PyMcaIO.ConfigDict`, to improve type detection:

- boolean interpretation now follows the same rules as ConfigParser: support of 'yes', 'true', 'on', 'no', 'false' and 'off' (case insensitive)
- strings can be preceded by the escape character (`\`)  in the INI file to ensure they are read back as strings and not interpreted as another type: `"\false"`, `"\1.0"`
- commas inside strings are always preceded by an escape character to prevent interpretation as a list 

The HDF5 format does not really preserve types, as all values are converted to numpy arrays before being written to file. For example, int and floats become numpy scalars (0D array), heterogeneous lists become arrays of strings...


vasole:
- Make sure scalars remain as scalars and not converted to arrays when writing HDF5.

Issue #40, opened 2016-04-08
FBFS: As python-h5py-dbg is missing in deb8, the automatic test during packaging fails.


kif:
This cannot go through in Debian as it would be considered "FBFS" (failed to build from scratch)

WARNING:run_tests:Patched sys.path, added: '/home/kieffer/workspace/silx/package/silx-0.0.1a3/.pybuild/pythonX.Y-dbg_2.7/build'
WARNING:run_tests:Test silx 0.0.1-dev3 from /home/kieffer/workspace/silx/package/silx-0.0.1a3/.pybuild/pythonX.Y-dbg_2.7/build/silx
INFO:run_tests:Import silx.test
WARNING:silx.io.test.test_specfile:de_DE.utf8 locale not installed on your system. An important i18n test will be skipped.
Traceback (most recent call last):
  File "run_tests.py", line 296, in <module>
    test_module = importlib.import_module(test_module_name)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/kieffer/workspace/silx/package/silx-0.0.1a3/.pybuild/pythonX.Y-dbg_2.7/build/silx/test/__init__.py", line 46, in <module>
    from ..io.test import suite as test_io_suite
  File "/home/kieffer/workspace/silx/package/silx-0.0.1a3/.pybuild/pythonX.Y-dbg_2.7/build/silx/io/test/__init__.py", line 32, in <module>
    from .test_dicttoh5 import suite as test_dicttoh5_suite
  File "/home/kieffer/workspace/silx/package/silx-0.0.1a3/.pybuild/pythonX.Y-dbg_2.7/build/silx/io/test/test_dicttoh5.py", line 30, in <module>
    import h5py
  File "/usr/lib/python2.7/dist-packages/h5py/__init__.py", line 13, in <module>
    from . import _errors
ImportError: /usr/lib/python2.7/dist-packages/h5py/_errors.x86_64-linux-gnu.so: undefined symbol: Py_InitModule4_64
[92317 refs]
E: pybuild pybuild:256: test: plugin custom failed with: exit code=1: env PYTHONPATH=/home/kieffer/workspace/silx/package/silx-0.0.1a3/.pybuild/pythonX.Y-dbg_2.7/build python2.7-dbg run_tests.py -i
[...]

PR #57, opened 2016-04-29
Make h5py an optional dependency

PiRK:
- removed `h5py` from install requirements in `setup.py`
- added `@skipIf(…)` for tests requiring `h5py`
- added error message logging and `ImportError` raising when trying to use a module or function requiring `h5py`
- added a line in Travis and AppVeyor config files to manually install `h5py` and keep running the tests, as it is no considered a mandatory dependency when running `pip install silx`


PiRK:
Does this provide a workaround for #40?

PiRK:
- added a note in function or module docstring, to mention dependency on h5py

PR #63, opened 2016-05-04
Standardization of documentation and installation instructions

PiRK:
Review of the installation instructions is necessary:
https://github.com/PiRK/silx/blob/doc/doc/source/install.rst

Specifically:

- Do we agree that we **officially** support only Python 2.7 and 3.5+? 
- I haven't been able to find a PyQt5 project on pypi. This seems to contradict a recent e-mail about availability of PyQt5 wheels. What is the easy way to install a Qt wrapper on Windows and Mac OS X?


vasole:
We support Python 2.7, 3.4 and 3.5.

On windows full functionality might only be available under Python 3.5 due to the use of obsolete compilers (VS 2008) under python 2.7

PR #64, opened 2016-05-04
[Plot] Add snap to curve with crosshair cursor + fix a few issues

### LINKED ISSUE #60: PlotWindow: Toggle color on mouse coordinates
@t20100 

- When active curve handling is enabled, there is an active curve, and points are shown, make a difference between mouse coordinates and point coordinates. That behavior was requested by beamlines (and implemented) for PyMca ScanWindow.

### LINKED ISSUE #61: [Plot] Update crosshair cursor and drawn primitives when panning with arrow keys
When panning with arrow keys, overlays like crosshair cursor and area displayed while drawing are not update.

Try to send a mouse move event when panning this way to sync the overlays.

### LINKED ISSUE #62: [Plot] Legend pane does not display active curve
The Legend pane does not take into account the active curve and so the color of the active curve is different between the legend pane and the plot area.

t20100:
Closes #60 
Closes #61 
Closes #62

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,7 @@ Dependencies
 
 The GUI widgets of the silx package depends on the following extra packages:
 
-* A Qt binding: `PyQt5, PyQt4 <https://riverbankcomputing.com/software/pyqt/intro>`_ or `PySide <https://pypi.python.org/pypi/PySide/>`_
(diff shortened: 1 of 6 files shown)


vasole:
- There is a runtime error at AppVeyor but it says tests are passing.

t20100:
There was an issue saving the wheel, I restarted the tests.

Issue #44, opened 2016-04-11
Add Levenberg Marquardt algorithm


vasole:
Refactor existing PyMca module.

vasole:
Closed by #72

PR #72, opened 2016-05-09
Levenberg Marquardt implementation

vasole:
Addresses #44


PiRK:
Regarding https://github.com/silx-kit/silx/pull/72/commits/4e307899ea5c8f4e74b3165f4d6db067f0a20dea#diff-3ce44ff7792abd000958391cf1b58b70L2, the [Sphinx doc](http://www.sphinx-doc.org/en/stable/domains.html#directive-py:currentmodule)<http://www.sphinx-doc.org/en/stable/domains.html#directive-py:currentmodule> says::

> in situations where documentation for things in a module is spread over multiple files or sections – one location has the py:module directive, the others only py:currentmodule.

My `.. pymodule::`directive was wrong, I meant `.. py:module::`.

payno:
if I understand correctly the line 26 :

"_logger = logging.getLogger(__name__)"

this is ok if we use this file as a module. But if we use it as a "single file" / demo we will get a warning.

payno:
Thanks to Thomas it s look like a call to "logging.basicConfig()" before _logger = logging.getLogger(__name__) is needed when we call it as a "single file"

silx/math/fit.py
```diff
@@ -0,0 +1,868 @@
...
+    :param full_output: bool, optional
+        non-zero to return all optional outputs. The default is 0
+
+    :param check_finite: bool, optional
+            If True, check that the input arrays do not contain nans of infs,
+            and raise a ValueError if they do. Setting this parameter to
+            False will ignore input arrays values containing nans.
+            Default is True.
+
+    :param left_derivative: bool, optional
```
dnaudet: left_derivative is described twice

doc/source/modules/math/fit.rst
```diff
@@ -0,0 +1,8 @@
+
+.. currentmodule:: silx.math
+
+:mod:`silx.math.fit`: Levenberberg Marquardt with constraints
```
dnaudet: Leven~~ber~~berg

Issue #28, opened 2016-03-30
Default data location


PiRK:
I'm creating this issue to discuss data packaging.

Initially we could gather all the data in a ``silx/data`` directory and make ``setup.py`` include it with the build. Having all the data (for the moment only GUI icons, but possibly also calibration data in the future, and maybe test related data) in a a single directory will make it easier to separate data and code for Debian packaging. Accessing data in our code could be done using ``from silx.resources import data_dir``. 

This is a draft for ``resources.py``:  https://github.com/PiRK/silx/blob/master/silx/resources.py

One alternative way of doing, instead of gathering all data in a sub-directory of the package, would be to store data together with the modules using it (e.g. putting icons in a subfolder of  ``silx/gui``) and setting  ``data_dir`` to the root package directory by default. And when we need to separate data and code, we could replicate the project structure in the data directory (e.g. ``/usr/share/silx/gui/icons``). It would make separating data and code slightly more complicated.

vasole:
I am more in favor of silx/data containing the different data.

However, I would like to foresee a similar mechanism to the one used to install the different subpackages where each subpackage can be told the prefix to be used when trying to find its own data. Example module A that stores its data in directory A/data whould be told that its data are to be found into data_dir/A/data

PR #75, opened 2016-05-12
Project data files + GUI icons

t20100:
Related to #28

Proposal to deal with project data files (GUI icons as *.png for now).

It follows setuptools [including data files doc](https://pythonhosted.org/setuptools/setuptools.html#including-data-files) and does not really care about Debian packaging issues except that data files are stored in a single directory and accessed relative to a `data_dir` variable.
It does not move data files between the source and the build tree.
It works for sdist, wheels, setup.py build and install.
It will not work if the packages are stored in a zip file (See [accessing-data-files-at-runtime](https://pythonhosted.org/setuptools/setuptools.html#accessing-data-files-at-runtime)).

It provides a `resource_filename` function to get the absolute path of a resource file.


vasole:
Anything makes sense to me if there is a module, named xxxx.py from here on, that can deal with the problem of finding the data files or the documentation files in the default case (no zip, standard installation) or following a predefined strategy (environment variable, registry variable or looking for those files in one directory within a set of paths relative to the xxx.py module. Those are solution proven to work.

t20100:
Updated way to handle data files.
This adds support zipped package and forzen application

PR #76, opened 2016-05-17
Update line profile of bilinear module

t20100:
- Add support for line profile of even width
- Add bilinear module to sphinx doc


kif:
Looks good to me.

PR #78, opened 2016-05-18
Spectoh5

### LINKED ISSUE #77: spectoh5: attributes not created when creating groups or datasets
This is a bug in my code. 

When reading a SpecFile and interpreting it as a HDF5 with `silx.io.spech5`, HDF5-like attributes are attached to some groups and datasets as a dictionary. For instance, the root group has the following `.attrs`: `{"NX_class": "NXroot"}`.

These attributes are currently not being written to the output HDF5 file when converting a SpecFile using functions from  `silx.io.spectoh5`.

PiRK:
Closes #77 (bug  attributes not created when creating groups or datasets when converting SpecFile to HDF5)

Minor documentation improvements.

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -75,7 +75,7 @@ see `Source package`_.
 Installing a Debian package
 ---------------------------
 
-Debian 8 packages are available at http://www.edna-site.org/pub/debian/.
+Debian 8 packages are available at http://www.silx.org/pub/debian/binary/.
 
 Download the following ``.deb`` files:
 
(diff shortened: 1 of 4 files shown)


PiRK:
Regarding our discussion about enforcing ascii when writing attribute values, I believe  ASCII encoding is implicitly done by my usage of `numpy.string_`, as all my attributes use only basic alphabetical characters without accents.

FYI, attributes are defined here: https://github.com/PiRK/silx/blob/master/silx/io/spech5.py#L294

PR #83, opened 2016-05-23
Add a profile selection/display toolbar for PlotWindow + ImageView widget with side histograms

### LINKED ISSUE #69: Master branch tests passing or not passing under windows with PyQt5?
@t20100 

After installation from master branch I get an error running the run_tests.py script. 

ERROR:silx.gui.plot.PlotTools: Error while converting coordinates (46.87500, 52.246696)with converter 'Exception'

silx_issue_17

t20100:
- A Toolbar with horizontal, vertical and free profile line with settable width using image.bilinear module that can be attached to a PlotWindow.
- A Toolbar to display and control the limits of a Plot.
- ImageView widget: plot an image with side histograms (taken from PyMca) with a def main(argv) and if __name__ == '__main__': to start the widget and display an image file

- Skip GUI tests when DISPLAY is not defined on Linux (e.g., running them on a server)
- Closes #69
- Fix run_tests.py coverage output


vasole:
- 1D ROI usage should not be there.
- Profile Menu. It should say "Free" and not "Fee"
- Can the radar view have a wider pen? It is difficult to be seen on my screen.

- Radar view: It would be convenient to switch to a hand cursor when one can move the viewing zone or (even simpler) to move the rectangle when moving (a left-clicked) mouse in the radar view even if not on the rectangle.

t20100:
Addressed the 4 comments.
For rectangle width, it is pixels on the screen.

PR #86, opened 2016-05-24
Make ConfigDict ordered

PiRK:
- Make `ConfigDict` an OrderedDict. This results in internal changes only, it does not affect the API. Files written with previous versions of  `ConfigDict` can still be read back. The relative order of values in a section is preserved, and the relative order of sub-sections in a section is preserved, but due to the output file format we cannot preserve the relative order between a value and a sub-section in a section (subsections are written after all top level values) 

 - Update docstrings.


t20100:
In the load function, you can do the same when loading json with
json.load(f, object_pairs_hook=OrderedDict)

PR #89, opened 2016-05-27
Histogramnd continued...

dnaudet:
Histogramnd, continued...

* histogramnd : function now returns the bin edges along with the histogram and weighted histogram. This breaks compatibility with the previous version if someone was directly unpacking the result (e.g : ```histo, w_histo, edges = histogramnd(...)```).

* new class HistogramndLut
  * lut is computed at instantiation time
  * ```HistogramndLut::accumulate(weights, ...)``` : applies the lut. This call's result is added to the previous call's result (stored in this instance).
  * ```HistogramndLut::apply_lut(weights, ...)``` : applies the lut and returns the result (i.e : does not accumulate it to this instance's histograms). User can provide a previously computed histo, in that case the behaviour is the same as histogramnd.
  * other minor stuff, see the doc.

* unit tests (around 50) for HistogramndLut : **nominal only**. I can add some **error** cases tests like the ones for histogramnd... but later.

I did not update the benchmark script to add HistogramndLut tho.


dnaudet:
**WIP** : return dictionnary instead of a tuple.
Dot not merge.

dnaudet:
How about a class instead of a dictionary?
Something like :

```python
class HistogramndResult():
  histo = ...
  weighted_histo = ...
  n_bins = ...
  bins_rng = ...
  edges = ...
  def accumulate(self, sample, weights):
    [call histogramnd and handles the parameters]
```
And if the user wants to accumulate he would do :
```python
result = histogramnd(sample, n_bins, bins_rng, weights)
result.accumulate(sample_2, weights_2)
histo = result.histo
weighted_histo = result.weighted_histo
```
Instead of (currently) :
```python
histo, weighted_histo = histogramnd(sample, n_bins, bins_rng, weights)
histo, weighted_histo = histogramnd(sample_2, n_bins, bins_rng, weights_2, histo=histo, weighted_histo=weighted_histo)
```

vasole:
The idea is to keep the "entrance" ticket as low as possible. Because of that we decided not to use a named tuple that could have been a nice compromise between a tuple and a dictionary.

If we do not care about compatibility with previous version (you are the only user), a dictionary is the simplest solution.

If we do care about compatibility with previous version, just adding a keyword "full_output" that by default is False solves things. If we are absolutely sure the number of return arguments will never exceed three, we can remain with a tuple. However, if we take a look at curve_fit, from scipy, we see that by default they return two items and that they added a "full_output" keyword returning the same first two items plus a dictionnary that could grow at will.

I do not consider a class a good idea.

t20100:
In this case, I like the return value being a class instance because it allows both direct use of the function + the result provides accumulation without having the user to take care of storing the intermediate results. It also looks the same as the histogram LUT API. (An alternative could also be to have a function for histogramnd and a separate class for acculumation).

Returning dicts is good for documentation and extension, but it does not allow to call the function as 
a single liner as when returning tuples. On the other hand, tuples prevent extension (without breaking the API) and are not well suited when returning more than 2 or 3 values.

```
result = myfunc(...)
histo = result['histo']
weights = result['weights']
```

versus

`histo, weights = myfunc(...)`

namedtuple (or something similar or derived from it) can allow both syntax. It is also possible to have the simple result being unpacked as a tuple, while extra information and methods being available.
For users, the use of namedtuple should not be visible, they will see 2 code samples: a single-liner and a bit more complex example. Simple syntax for the basic things and more complex syntax only for extra info.
For developers, that's different.

vasole:
If finally we go the class way, I would really like to have the full_output option set to False by default and returning a tuple. Why? Because with classes or dictionaries we risk to consume a lot more memory. This histogram business is going to handle 256x256x256 or 512x512x512 histograms. In the case of doubles that makes 1 GByte for the later. Returning a tuple makes sure that if we normalize one of the returned values by the other we are not keeping both, the normalized and the not normalized data and that we have no object around keeping duplicates.

The longer I think about it, the more in favor of the original tuple approach I am. Considering that numpy has not considered "extensibility" of their histogram functions via full_output keywords, I would remain with a fixed tuple for the time being.

dnaudet:
I added big red warnings in the doc about sample and weights not being contiguous : if so an internal copy has to be made.
I also added checks for the user provided histo and weighted_histo : it raises an exception if they're not C contiguous.
So I'd like to draw your attention to the fact that it's maybe another potential **pitfall** (that and the fact that a user has to make sure that he provides the exact same parameters when accumulating histograms).

t20100:
Following discussion on returned values, returning a class instance would make accumulation self-contained and avoid this kind of pitfall while still allowing to unpack results with a single liner call when no accumulation is necessary.

dnaudet:
@t20100 : my point exactly! :dart:

vasole:
The contiguity pitfalls are easily prevented in the wrapping code prior to call the C code. So they are certainly not an issue.

The user is supposed to provide as input what he got on output. I do not think it is so difficult to understand. If the user does not supply the output buffers, we generate them. If they are supplied, we use them. It is really straightforward use and good, old fashion, C programming way of doing things.

dnaudet:
Ok, so should I remove the exceptions and add another warning/info in the doc about the potential slow down?
If histo (and/or weighted_histo) is not contiguous then we have to perform 2 (and/or 4) more copies :
- one to create a contiguous array from histo (weighted_histo)
- one to put back the new values into histo (weighted_histo)

Old fashioned all right, good I don't know hehe :stuck_out_tongue_winking_eye:

vasole:
I do not mind having the exceptions. If the user takes what we generate, everything will work.

t20100:
A few comments:
- In histogram.py you could do `from .chistogramnd import chistogramnd as histogramnd` directly.
- For sample code in docstring, if you write them with >>> before the code, you can then test that the sample code is actually working with doctest.

Regarding returned value, I do prefer something like:
```
histo = histogramnd(n_bins=n_bins, bins_rng=ranges)
for sample, weights in ...:
    histo.accumulate(sample, weights)
```
Then, use: `histo.counts` and `histo.weights` or whatever names

over:
```
histo, w_histo, edges = histogramnd(sample, n_bins=n_bins,
                                    bins_rng=ranges, weights=weights)
for sample, weights in ...:
    histo, w_histo, edges = histogramnd(sample, n_bins=n_bins,
                                        bins_rng=ranges, weights=weights,
                                        histo=histo, cumul=w_histo)
```
Then use `histo` and `w_histo`

It's half the lines of code, avoid to explain what should be reentered to the function and looks more readable to me.

vasole:
Thomas, that is just a question of taste. I can also argue that in once case there is only one function to explain and to learn about.

t20100:
As we discussed about returned value, the same function (e.g., histogramnd) can return something that is for the user both a tuple (for single call use) and an object (for extended results and accumulation).

dnaudet:
@t20100 thanks Thomas, I'm adding the **>>>**.

WIP!

vasole:
Please note that there are merge conflicts to be solved. I sent them to your branch, but it added all the history of the master...

dnaudet:
I think I solved the conflicts (and pushed them into #f2ef5d2) .
I just gotta finish the return tuple...

dnaudet:
Done.

PR #90, opened 2016-05-31
Add Plot1D and Plot2D: 2 specific subclasses of Plotwindow

t20100:
- Add Plot1D a plot widget with tools for curves enabled.
- Add Plot2D a plot widget with tools for image enabled.
- Add Qt designer plugins for Plot1D and Plot2D


vasole:
The constructor should not call show. That can trigger unnecessary drawing.

It should follow Qt's way: one instantiates and calls show explicitly when desired.

t20100:
The idea is to have high level widgets with a "simple" API to be used from ipython (thus the show in the constructor).
I'd like to keep a single call when used from ipython, so I can either call show only when I am in ipython or provide wrapper functions (i.e., plot1D and plot2D) that does it... but it makes 2 entries in the API for the same thing...

Any API review/comments?

PiRK:
I like the idea of having a simple one-liner to load a widget, add the curve/image and show it. I don't think it is a good idea to restrict this behavior to IPython sessions, it could be used in a regular interactive python session as well.

t20100:
widget show done only when python run from the console

t20100:
Finally splitted Plot1D and Plot2D widgets from functions plot1D and plot2D to use from console

PR #93, opened 2016-06-01
Fix #91 (spec_date_to_iso8601)

### LINKED ISSUE #91: [io.spech5] spec_date_to_iso : "May" missing
#in spech5.py file:

Change :
```python
months = ['Jan', 'Feb', 'Mar', 'Apr', 'Jun', 'Jul',
              'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
```

To :
```python
months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul',
              'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
```

dnaudet:
Fix for #91
 Added unit tests for spec_date_to_iso8601.
**Nominal** tests only.
The **zone** keyword is **not** tested. I haven't added tests for the **error** cases either.
I checked that the new tests fail with the previous version of the function.

BTW the function only works if the provided spec dates are in english (Mon, Apr, Aug, Sun, ....). Are we sure this will always be the case?

### DIFF
--- a/silx/io/spech5.py
+++ b/silx/io/spech5.py
@@ -514,7 +514,7 @@ def spec_date_to_iso8601(date, zone=None):
         >>> spec_date_to_iso8601("Sat 2015/03/14 03:53:50")
         '2015-03-14T03:53:50'
     """
-    months = ['Jan', 'Feb', 'Mar', 'Apr', 'Jun', 'Jul',
+    months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul',
               'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
     days = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun']
 
--- a/silx/io/test/test_spech5.py
+++ b/silx/io/test/test_spech5.py
@@ -28,8 +28,11 @@ import os
 import sys
 import tempfile
 import unittest
+import datetime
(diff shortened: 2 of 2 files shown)


PiRK:
I have investigated Jérôme's suggestion of using `datetime`. It could simplify the function, especially if we expect to add more date formats in the future. Initially I expected SPEC to write a single format.

```
>>> from datetime import datetime
>>> d = datetime.strptime('Thu Feb 11 09:54:35+0100 2016', '%a %b %d %H:%M:%S%z %Y')
>>> d.isoformat()
'2016-02-11T09:54:35+01:00'
>>> 
>>> d = datetime.strptime('Sat 2015/03/14 03:53:50+0200', '%a %Y/%m/%d %H:%M:%S%z')
>>> d.isoformat()
'2015-03-14T03:53:50+02:00'
```

Des anyone have any objection to rewriting the entire function?

dnaudet:
I think this works only if you set the locale.LC_TIME right.

e.g :
```python
d = datetime.strptime('Mo Feb 11 09:54:35 2016', '%a %b %d %H:%M:%S %Y')
```
doesn't work if the computers locale is set to german (```locale.setlocale(locale.LC_TIME, 'de_DE.UTF-8')```)
It expects this :
```python
d = datetime.strptime('Do Feb 11 09:54:35 2016', '%a %b %d %H:%M:%S %Y')
```

Issue #105, opened 2016-06-21
[Plot] Mask tool possible enhancements


t20100:
List of ideas for improving the mask tool:
- Draw mask operations:
  - Additional drawing tools: circle (pencil can already do it), ring.
  - shrink/expand mask.
  - Edition of shapes as vector graphics.
  - Preview mask while drawing: draw rectangle and polygon while drawing, draw pencil shape while hovering the image
- Multiple mask levels support:
  - support overlapping mask levels.
  - Display mask levels as a list with + and - buttons to add/remove layers, and select and show/hide toggles for each mask.
- Optimize undo/redo, now store the whole mask, can store list of points when more compact in memory, also look at numpy.packbits/unpackbits for single mask level.
- Threshold:
  - Threshold range as a range slider with an histogram of the data values.
  - A tool to set the min or max value from a position in the plot (like a color picked tool)
- When direction is vertical, add a vertical scroll bar if needed.

t20100:
For the record, copied comment from @vasole in PR #106 :

The main root of the problem is that the mask is totally dissociated of the image used to define it. If the mask is totally dissociated from the image, as it seems to be your choice, you can design a mask at a much higher resolution than the image below it. So, you may think a pixel is selected when it is not.

To me, an image can potentially have data, representation and masks. Of course any of those on its own lead to a visual representation and the actual visualization does not care because it only needs to draw pixmaps. Keeping the simple concept of "legend", one can pack together items and choose to make them disappear if the "image legend" does so.

The analogue in 1D is a curve that has its data (positive and negative values) and its representation (without negative values with log scales). If one totally decouples both, bugs like #100 happen.

payno:
I think this is too much regarding current manpower. So let's move forward. I will close the issue and please reopen with more focus topics so we can have priorities.

Issue #104, opened 2016-06-21
[Plot] PlotWindow mask tool does not work on Debian 7 (matplotlib < 1.2.0)


t20100:
Matplotlib < 1.2.0 does not support transparent colormaps which are used by the mask tools to draw the mask layer.

vasole:
Several options:

- Use a non-transparent mask
- Force the use of the OpenGL backend
- Draw the mask using Qt
- Not caring. Is there an appropriate backport for Debian 7?
....

kif:
On Tue, 21 Jun 2016 06:58:44 -0700
V. Armando Solé <notifications@github.com> wrote:

> Several options:
> 
> - Use a non-transparent mask
> - Force the use of the OpenGL backend
> - Draw the mask using Qt
> - Not caring. Is there an appropriate backport for Debian 7?

Not caring: Debian 7 is not a target platform

t20100:
A workaround solution is to create a virtualenv with appropriate dependencies.

kif:
solution used in pyFAI
```
    def massif_contour(self, data):
        """
        Overlays a mask over a diffraction image

        @param data: mask to be overlaid
        """

        if self.fig is None:
            logging.error("No diffraction image available => not showing the contour")
        else:
            tmp = 100 * numpy.logical_not(data)
            mask = numpy.zeros((data.shape[0], data.shape[1], 4), dtype="uint8")

            mask[:, :, 0] = tmp
            mask[:, :, 1] = tmp
            mask[:, :, 2] = tmp
            mask[:, :, 3] = tmp
            while len(self.msp.images) > 1:
                self.msp.images.pop()
            try:
                xlim, ylim = self.ax.get_xlim(), self.ax.get_ylim()
                self.msp.imshow(mask, cmap="gray", origin="lower", interpolation="nearest")
                self.ax.set_xlim(xlim)
                self.ax.set_ylim(ylim)
            except MemoryError:
                logging.error("Sorry but your computer does NOT have enough memory to display the massif plot")
            update_fig(self.fig)
```

Issue #100, opened 2016-06-13
[Plot] Negative values lost when toggling log axes


vasole:
When visualizing a curve with negative values, the values are not shown when setting the axis as being logarithmic, that is desired behavior. However, when coming back to linear axis, the negative values are lost/not show.

PR #106, opened 2016-06-21
[Plot] Add a masking tools panel

### LINKED ISSUE #98: [Plot] Image mask tools
Functions (list taken from PyMca MaskImageWidget):
- Operation on the whole mask:
    - Clear
    - Inversion
- Drawing tools:
    - Brush (with settable width in image pixels)
    - Rectangle
    - Polygon
    - Above threshold
    - Below threshold
    - Within range
- Each draw tool can either paint or erase
- Multiple level of exclusive masks

Proposal: Regarding the number of tools and the fact that more tools will probably be needed, I'm thinking of making a DockWidget as for ROIs, e.g.:

maskdockwidget

Ideas for future:
- History/undo/redo
- Drawing tools: circle, ring
- Edition of shapes as vector graphics
- shrink/expand mask
- show/hide different level of mask
- Load (replace of add) and Save masks

t20100:
Closes #98 

Implemented functions :
- Multiple level of exclusive masks or a single level (as an option)
- Operation on the whole mask:
    - Clear
    - Inversion
    - Reset all mask levels
- Drawing tools:
    - Brush (with settable width in image pixels), which provides a way to draw circle
    - Rectangle
    - Polygon
    - Each draw tool can either paint or erase
- Thresholding:
    - Above threshold
    - Below threshold
    - Within range
    - Range can be loaded from current colormap range
- 10 levels of undo/redo
- load/save mask

Related issues: #104, #105


vasole:
It looks very nice. Some feedback:

- The transparency of the mask is too low (at least on my screen).
- Colormap dialog is modal. Is it intended?
- In my opinion, the mask should not disappear. You have masked the image, the image has a mask and it is drawn.
- The polygonal mask tooltip icon should explain the usage (left button to add points, right button to finish)

t20100:
Colormap dialog is modal because it does not sync with image changes (colormap dialog needs rework).
Addressed 3 comments.
Leaving the mask displayed when closing the mask tool can lead to inconsistencies if one change the image or remove it, the mask will stay...

vasole:
The main root of the problem is that the mask is totally dissociated of the image used to define it. If the mask is totally dissociated from the image, as it seems to be your choice, you can design a mask at a much higher resolution than the image below it. So, you may think a pixel is selected when it is not.

To me, an image can potentially have data, representation and masks. Of course any of those on its own lead to a visual representation and the actual visualization does not care because it only needs to draw pixmaps. Keeping the simple concept of "legend", one can pack together items and choose to make them disappear if the "image legend" does so. 

The analogue in 1D is a curve that has its data (positive and negative values) and its representation (without negative values with log scales). If one totally decouples both, bugs like #100 happen.

vasole:
@srio

Manolo's proposal: Provide the possibility to show all masks, a particular mask or none.

t20100:
Mask visibility toggling would be nice, but I'd keep that for later.
It's in the list of mask tools enhancements: See #105.

PR #109, opened 2016-06-23
[Plot] Update polygon interaction

t20100:
When drawing a polygon, this adds the possibility to end drawing the polygon by clicking with the left button near the first point of the polygon.
It is still possible to end drawing with a right click.

### DIFF
--- a/silx/gui/plot/Interaction.py
+++ b/silx/gui/plot/Interaction.py
@@ -235,9 +235,9 @@ class ClickOrDrag(StateMachine):
             self.initPos = x, y
 
         def onMove(self, x, y):
-            dx = (x - self.initPos[0]) ** 2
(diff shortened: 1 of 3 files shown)


vasole:
It is nice, however when combined with zooming, it can make the initial square too large.

Would it be too difficult to express the square in pixel coordinates in order not to get it influenced by the zoom level? If that is the case, we let it as it is now.

t20100:
Indeed, I missed that!
I fixed it with zoom on wheel.
I did NOT fixed it when zooming is done programmatically or with upcoming Ctrl+/Ctrl-), if OK I'd leave that for another issue.

vasole:
I think the problem when zooming with the combination of keys can be left for another occasion.

PR #110, opened 2016-06-23
Add a requirements.txt

### LINKED ISSUE #101: Add a requirements.txt file
As there is a growing bunch of optional dependencies, what about having a requirements.txt file with all those dependencies at the root of the project to document this + allow to install silx & optional dependencies with: `pip install -r requirements.txt`...

t20100:
List all project's dependencies.
Closes #101 

Some points to discuss:
- If a wheel is available for PyQt/PySide it installs it, otherwise it does not install PyQt/PySide. Do we want that?
  Alternative 1: Do not put pyqt/pyside in requirements.txt and add a requirements-qt.txt for this.
  Alternative 2: Choose one e.g., PyQt5 and enforces it (in requirements.txt only)
- if h5py is not installed and no wheel is available, it will not be installed

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,7 @@
 numpy
 h5py                      # For silx.io
 ipython                   # For silx.gui.console
+qt_console                # For silx.gui.console
 matplotlib >= 1.2.0       # For silx.gui.plot
 # PyQt4, PyQt5 or PySide  # For silx.gui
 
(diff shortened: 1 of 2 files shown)


vasole:
I think it needs discussion. I see advantages and disadvantages in all cases. I guess the problem behind is that pip is not a real packaging tool. Basically the options are:

1. If a wheel is available for PyQt/PySide it installs it, otherwise it does not install PyQt/PySide.
2. Do not put pyqt/pyside in requirements.txt 
3. Install PySide because it does not impose anything to the user

Concerning h5py, I guess what will not be installed is h5py (not silx)

t20100:
Updated:
- Add qt_console
- Ask for PyQt5 except on Windows with Python 2.7 where it asks for PyQt4
- Qt binding and h5py are only installed from wheels not from source.

Issue #84, opened 2016-05-23
[Plot] Anchors on profiles


t20100:
Enhancements for PlotWindow profile toolbar:
- [x] Disable profile actions when no image is displayed
- [x] Update profile when image is updated
- [ ] Add draggable anchors to free profile to allow changing an already defined profile.

vasole:
First two points seem feasible for 0.2.0 release...

t20100:
First 2 points implemented in #116

vallsv:
Closed by  #3011

PR #116, opened 2016-06-27
[Plot] Profile tool update

t20100:
Related to #84

- Disable profile actions when no image is displayed
- Update profile when image is updated
- Fix support of image with origin and scale


vasole:
Funny, the profile continues to be calculated even if the new image has nothing to do with the initial one.

PR #145, opened 2016-07-04
Explain data indices in Scan.data

PiRK:
Explain how to retrieve a data column (data corresponding to 1 detector) using transpose or slicing.


PiRK:
TODO: regenerate and push cython file

Issue #153, opened 2016-07-05
[Doc]Tuto plot


payno:
In the tutorial Getting started with plot widgets, in the chapter "Axes" (Actually at the bottom of the page) you might change : 
   "Different methods allows to change the plot axes:"

For something like "Different methods allow plot axes modifications:"

Issue #151, opened 2016-07-04
[Plot] Plot2D should diplay intensity along with X and Y in the bottom toolbar.  


WatierYves:
As in ImageView, in my humble opinion, the Plot2D  should display pixel intensity (value) when fly over with mouse.

kif:
On Mon, 04 Jul 2016 07:01:43 -0700
WatierYves <notifications@github.com> wrote:

> As in ImageView, in my humble opinion, the Plot2D  should display pixel intensity (value) when fly over with mouse. 
+1

Issue #139, opened 2016-07-04
[Plot] Active curve handling of plot labels


vasole:
If the active curve does not have associated labels, it should not change already existing plot labels.

Issue #138, opened 2016-07-04
[Plot] Plot2D profile should be activated as default.


vasole:
Extracting a profile is a very common operation on images. Therefore it should be available on Plot2D.

PR #162, opened 2016-07-05
[Plot] Fixes

t20100:
Addresses the following issues :
- Doc typo: #153,
- Plot2D: add profile toolbar #138, add image value in position information bar #151
- Plot addCurve/addImage xlabel/ylabel default behavior change: #139


vasole:
- I do not know up to what point one has to be strict concerning #139. Your current implementation does not change the axes labels the first time a curve with unsigned labels is activated. However, if there is other change of labels, the second time it become active it changes the labels. I guess that behavior is acceptable.

t20100:
If a curve has no associated labels, then it uses the 'default' labels provided by setGraph[X|Y]Label

Issue #143, opened 2016-07-04
[specfile] add backward compatibility layer with previous specfile


maurov:
Hello,

It would be nice to have a compatibility layer between specfile in SILX and specfile in PyMca. This would permit to upgrade all scripts reading SPEC files via the PyMca API replacing a single line:

 `from PyMca5.PyMcaIO import specfilewrapper as specfile`

with

`from silx.io.specfile import SpecFile as specfile`

Thanks,
Mauro

vasole:
Closed by #163

PR #163, opened 2016-07-05
specfilewrapper module

PiRK:
This PR addresses #143 

It also fixes a mistake in the data shape. The SpecFile C library stores the scan data in the same shape as in the original file (first index = line/row, second index = column = data for one detector). This is how `silx.io.specfile.Scan` stored the data until now.
On the other hand, the old python specfile wrapper transposed the data. This results in a more intuitive data access, as we noticed during the code camp, because specifying the first index allows to retrieve the entire dataset for one detector.

So from now on, the data should be transposed in `silx.io.specfile` and in `silx.io.specfilewrapper`.


vasole:
Incorrect handling of @CHANN and @CALIB. You can test it setting

#@CHANN 151 29 29 1
#@CALIB 0 2 0

They are totally ignored.

vasole:
You can rename myscandata class. If you want to be fully compatible, it should be "scandata" all in lower case. Just comment that PEP8 violation is to respect old API, therefore not viloating PEP8 :)

vasole:
Closes #163

Issue #156, opened 2016-07-05
[Plot] Add default title and icon to plot widgets


t20100:
When used from he interpreter, would be nice to see which windows are plots

t20100:
Setting icons on windows prevents them to use the application default.. so better not to set window icon by widget

Issue #155, opened 2016-07-05
[Gui] Rework qt.py


t20100:
- Add import uic?
- Doc: Explain what this module imports + links to PyQt* PySide documentation + an example to illustrate the fact that this is a flat namespace, e.g., widget = qt.QWidget() ...
- Add qt.QtBinding which is the Qt binding module used, so one can do from qt.QtBinding import [...]
  instead of using the string qt.BINDING to know which binding is used and have if/else construct to do extra imports.

t20100:
- No uic module in PySide
- Mistake: No way to do from qt.QtBinding import [...]

vasole:
Lines 32 and 34 should say Qt binding and not Qt bindings. The rest is Ok.

Issue #137, opened 2016-07-04
[Plot] Colormap widget does not update min and max values first time.


vasole:
When switching autoscale on/off the min and max data values are not updated **the first time**

Issue #158, opened 2016-07-05
Add a Qt designer plugin for the console widget


t20100:

t20100:
Reopen as not solved by #164

t20100:
Closing as not being contemplated in 5 years.
Please, reopen if needed.

PR #164, opened 2016-07-05
[Plot] More Fixes

t20100:
Addresses the following issues:

- rework qt.py: #155
- Add window titles for each plot widget when they have no parent: #156 
- ColormapDialog, set min/max with data range when in autoscale: #137


t20100:
Remove add IPythonWidget Qt designer plugin from this PR (reopen #158) as there is some issues under Windows (to investiguate later)

PR #170, opened 2016-07-07
[Plot] Update mask tool API

t20100:
- Rename getMask and resetMask to getSelectionMask and resetSelectionMask
- Add setSelectionMask
- Add get/setSelectionMask to PlotWindow (and so Plot2D)


vasole:
The implementation of setSelectionMask(self, mask, copy=True) does not match the documentation.

I guess `mask = numpy.array(mask, copy=False, dtype=numpy.uint8)` should be changed to `mask = numpy.array(mask, copy=copy, dtype=numpy.uint8)`

vasole:
My mistake. the copy keyword is used later on.

Issue #131, opened 2016-07-04
Auto-scale graph does not work as expected


andygotz:
When un-selecting a plot using the Legends control box I expect that when I then click on Auto-scale the plot will resize using only the values in the remaining (selected) plots. This does not happen and the plot appears with the same axes as if the un-selected plot was still there. Auto-scale does not seem to re-calculate the min and max of the graph. 

It would be nice if the plot re-calcuted the min+max taking into account only the active plots.

Andy

t20100:
Also check if resize is working as expected when curves are moved to right axis

t20100:
I could not reproduce this...

In IPython:
```
In [1]: from silx.gui.plot import plot1D
In [2]: %pylab qt
In [3]: plt = plot1D((1, 2))
In [4]: plt.addCurve((0, 1), (10, 11), legend='2')
```
Then unselecting a curve in the Legend widget and doing a reset zoom works.

t20100:
Tested with same data, it works.
2 options:
- Problem specific to a machine or
- Missunderstanding of autoscale toggling icons functions.

For second issue, we can provide more information about the function in the toolbar:
- Add more info to X and Y autoscale tooltips.
- Disable the reset zoom icon when no axis is in autoscale mode, change at least the tooltip and possibly the icon depending on X and Y autoscale mode to reflect this.
- When enabling autoscale on an axis also do a zoom reset.

vasole:
Should we open a specific issue asking for an enhancement? In fact what we are now missing is to apply the autoscale on the seen region if the X axis is fixed.

t20100:
Closing now as not reproduced and PR #171 adds info to button.
Open issue #179 for changing reset zoom behavior.

PR #171, opened 2016-07-07
[Plot] Add more info to reset zoom/autoscale tools + change behavior of autoscale buttons

t20100:
Related to #131:

- enable/disable reset zoom action depending on autoscale + change tooltip
- behavior change: reset zoom when checking an autoscale action + update tooltip


PiRK:
It seems to be working fine. Waiting for last CI tests to complete.

PR #172, opened 2016-07-07
[Plot] Fix ImageView bug + add profile tools

t20100:
- Add profile toolbar to ImageView
- Fix a bug in valueChanged signal with non default scale or origin


vasole:
Executing `python -m silx.gui.plot.ImageView` gives twice the profile toolbar.

PR #195, opened 2016-07-20
Fit interface

PiRK:
This is an early PR to discuss about API and possibly missing functionalities.

Main changes included in this PR:

- `silx.math.leastsq` -> `silx.math.fit.leastsq`
- new `silx.math.fit.fitmanager`, based on PyMca's `Specfit` module. This is a fit manager that loads data, handles constant and linear background substraction before fitting a model function, defining model functions, defining estimation functions to find initial parameters as input for iterative fitting, defining a custom numerical derivative function, managing fit constraints
- new `silx.gui.fit.FitWidget`, based on PyMca's  `SpecfitGui` module. This is a GUI to interface with `fitmanager`, modify constraints and fit parameters interactively, select fit functions, display fit results. 

`FitWidget` relies on other new submodules of silx.gui.fit (they should probably remain private so I haven't included them in the documentation for now)

- `Parameters`: based on PyMca5.PyMcaGui.math.fitting.Parameters
- `FitWidgets`: collection of widgets based on PyMca's FitConfigGui, MultiParameters, FitActionsGui, FitStatusGui
- `QScriptOption`: based on PyMca's QScriptOption

`fitmanager` relies on:

- `silx.math.fit.functions`: collection of functions (e.g gaussians...etc), based on PyMca's `SpecfitFuns`
- `silx.math.fit.filters`: strip and snip filters for background extraction;  Savitsky Golay smoothing filter; these are from  also extracted PyMca's `SpecfitFuns`
- `silx.math.fit.peaks`: `peak_search` function from  PyMca's `SpecfitFuns.seek`, and `guess_fwhm` from PyMca's `SpecfitFunctions`
- `silx.math.fit.fittheories` from PyMca's `SpecfitFunctions`

The main difference with PyMca's SpecfitGui and Specfit is that I removed a specific MCA fitting method, which consisted in isolating and fitting each peak separately. 
  
fitwidget1

The code of `fitmanager` requires more polishing.


vasole:
@t20100 

The PySide branch of our qt.py should define pyqtSignal = Signal in the same way the PyQt4 and PyQt5 define Signal=pyqtSignal. Given the predominance of the usage of PyQt4 and PyQt5 over PySide it is more than justified.

t20100:
If you want.
You can do it in this PR, it won't break anything.

PiRK:
Done (https://github.com/silx-kit/silx/pull/195/commits/dc56ede3ebcfdbca98b4b8f88db392833fd01ffa)

vasole:
- The name of the functions in the combo box does not need to match the actual names of the functions. It is advisable that you change the keys in fittheories..THEORY to the names used in PyMca.
- An improvement respect to PyMca would be to make use of a field (your current description field) to provide ToolTip information).
- FitManager. If I am not mistaken, it is advisable that every python class inherits from object (to be confirmed)

vasole:
- fittheories.test Non-gui silx modules should not rely on silx-gui modules.

vasole:
- When setting a constraint for a parameter, the free parameter is updated after the fit, but the constrained parameter value is not updated.
- In PyMca, when using the fit interface to set constraints, the interface is much smarter. It sets a default parameter name based on current parameter name and the current free parameters.

PiRK:
Regarding CamelCase, should this apply to all methods or do we want to maintain some compatibility with PyMca's SpecfitGui for the main methods? I'm mainly thinking about `setdata`, `estimate` and `startfit`.

vasole:
Convert them to CamelCase but keep the old ones calling the new ones but printing a deprecation warning.

PiRK:
I managed to fix the problem with the default parameter name not being updated when selecting a constraint.

I'm still investigating the other problem, but I can't reproduce it all the time. When I set the constraint to FACTOR, the constrained parameter seems to be properly updated after a fit most of the time, unless I set a factor that is significantly lower than the "real" one. When I set the factor to a significantly higher value, I don't seem to have the problem.

Is it possible that the leastsq function could sometimes give up on trying to fit a single parameter, and revert back to the initial value without failing with a `LinAlgError`?

vasole:
The only case not to variate the initial parameters without an error is when the initial parameters cannot be improved. You can check the number of iterations performed in order to give an informational message in that case.

The problem is not that one because the free parameter is changed.

vasole:
That lower limit is not at all arbitrary. The minimum you can ask to have a peak, is to be above noise. To allow to set it to above 1 sigma is already generous. In the past I was even limiting to more than 2.

 I agree that a warning informing the entered value has been ignored if smaller than the limit is convenient, but to remove the limit has no sense (already less than 2 for the sensitivity is questionable).

PiRK:
I called it arbitrary because the noise evaluation depends on another user defined parameter (fwhm). 

My main problem with this limit was the absence of warning/information about it, but you are right that we can solve this by finding a way to implement a warning.

vasole:
This can be already used.

PR #197, opened 2016-07-25
Fix index issue in leastsq()

PiRK:
The uncertainties calculation for `infodict['uncertainty']` had a loop on the wrong range, the length of the second dimension of `constraints` instead of the first dimension.   `constraints` has been transposed when it was included in _silx_.


vasole:
It looks good. Please, could you add a no regression test?

PiRK:
Done.

PR #199, opened 2016-07-26
Christian h5 octave

payno:
A simple class to make communication between Octave and Python (using hdf5 structures).
Mainly developed by CaptainNemoz.
For Octave version < 3.8 we need to add a small hack to make sure we don t loose information (adding an extra character for sq_string values )


payno:
It looks like I failed one merge, closing this pull request

PR #202, opened 2016-08-03
Document hdf5 string encoding

PiRK:
This PR improves documentation of `silx.io.spectoh5` regarding string encoding in generated HDF5 files.


dnaudet:
Looks good to me.

PR #203, opened 2016-08-04
Small modification of spech5 file structure

PiRK:
While I was working on a widget that can display a tree view of an HDF5 file or a specfile (using SpecH5), I found that the link `measurement/mca_0/info/data` was not working correctly due to a bug in SpecH5.
This caused the following error: `KeyError: u'Name /94.1/measurement/mca_0/info/data does not match any known dataset.'`

I could have just fixed it, but I don't think this link makes sense anyway. It was implemented because, at the time, I considered it was easier to make a simple link between groups (`measurement/mca_0/info -> instrument/mca_0/`) rather than implement a link for each dataset in `instrument/mca_0/` except `data`. 

This PR fixes the bug and changes the implementation by removing the group link and adding individual links for each dataset. The main API change that this causes is that the link `measurement/mca_0/info/data` no longer exists, but the data still exists as  `instrument/mca_0/data` and there is still the other link one level-up as `measurement/mca_0/data`.

@dnaudet : is this going to cause a problem for you? If it does, I can close this PR and just fix the problem without API changes.

### DIFF
--- a/silx/io/spech5.py
+++ b/silx/io/spech5.py
@@ -57,7 +57,12 @@ Specfile data structure exposed by this API:
               …
               mca_0/
                    data -> /1.1/instrument/mca_0/data
-                   info -> /1.1/instrument/mca_0/
+                   info/
+                       calibration -> /1.1/instrument/mca_0/calibration
+                       channels -> /1.1/instrument/mca_0/channels
+                       preset_time -> /1.1/instrument/mca_0/preset_time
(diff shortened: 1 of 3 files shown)


vasole:
- The link effectively does not make sense but it is like that.
- The link has to be made from measurement/mca_0/info -> 
instrument/mca_0/ as originally made.

PiRK:
In that case, I will close this PR and open PR #204

PR #204, opened 2016-08-05
Fix bugs in spech5

PiRK:
Alternative bug fix for broken links in `silx.io.spech5`, with no change to exposed hdf5-like structure. See (#203) for more explanations.


PiRK:
I just fixed a conflict that prevented merging. This can be merged, preferably before merging #220

PR #207, opened 2016-08-19
implement SpecH5Group.get()

PiRK:
Implement get method for SpecH5Group, to better mimic h5py.Group

The *getlink* parameter is ignored for the time being.


vallsv:
Missclick

PR #212, opened 2016-08-24
Feature weakref module

vallsv:
- Create silx.utils module
- Create a WeakMethod (class with callback)
- Create a WeakList

### DIFF
--- a/silx/utils/weakref.py
+++ b/silx/utils/weakref.py
@@ -115,3 +115,142 @@ class WeakMethod(object):
     def __hash__(self):
         """Returns the hash for the object."""
         return self.__obj.__hash__() ^ self.__method.__hash__()
+
+
+class WeakList(list):
+    """Manage a list of weaked references.
+    When an object is dead, the list is flaged as invalid.
+    If expected the list is cleaned up to remove dead objects.
+    """
+
+    def __init__(self, enumerator=()):
+        """Create a WeakList
+
+        :param iterator enumerator: A list of object to initialize the
+            list
+        """
+        list.__init__(self)
+        self.__list = []
+        self.__is_valid = True
+        for obj in enumerator:
+            self.append(obj)
+
+    def __invalidate(self, ref):
+        """Flag the list as invalidated. The list contains dead references."""
+        self.__is_valid = False
+
+    def _create_ref(self, obj):
+        """Create a weakref from an object
+        Can be overwrited.
+        """
+        if inspect.ismethod(obj):
+            return WeakMethod(obj, self.__invalidate)
+        else:
+            return weakref.ref(obj, self.__invalidate)
+
+    def __clean(self):
+        """Clean the list from dead references"""
+        if self.__is_valid:
+            return
+        self.__list = [ref for ref in self.__list if ref() is not None]
+        self.__is_valid = True
+
+    def __getitem__(self, index):
+        """Returns the object at the requested index"""
+        self.__clean()
+        return self.__list[index]()
+
+    def __iter__(self):
+        """Iterate over objects of the list"""
+        for ref in self.__list:
+            obj = ref()
+            if obj is not None:
+                yield obj
+
+    def __len__(self):
+        """Count item on the list"""
+        self.__clean()
+        return len(self.__list)
+
+    def __setitem__(self, index, obj):
+        """Set an item at an index"""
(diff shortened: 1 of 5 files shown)


silx/utils/weakref.py
```diff
@@ -0,0 +1,255 @@
...
+            list
+        """
+        list.__init__(self)
+        self.__list = []
+        self.__is_valid = True
+        for obj in enumerator:
+            self.append(obj)
+
+    def __invalidate(self, ref):
+        """Flag the list as invalidated. The contains dead references."""
```
dnaudet: the \***list**

silx/utils/weakref.py
```diff
@@ -0,0 +1,255 @@
...
+
+__authors__ = ["V. Valls"]
+__license__ = "MIT"
+__date__ = "24/08/2016"
+
+
+import weakref
+import types
+import inspect
+from ..resources import resource_filename
```
dnaudet: ?

dnaudet:
How about a WeakMethodProxy? Inherit from WeakMethod and override __call__.

vallsv:
- The absolute import is needed to force the module to load the lib saferef, else it tries to load itself :-D
- As you want for the proxy, but my feeling with proxy is it is more difficult to manage
- I dont understand your comment about list. I see a typo on the comment is it that?

Let me clean up a little more this request. I will try to have a full coverage of the code and to implement the sort function in a simple way.

dnaudet:
* no i meant *from ..resources import resource_filename* :wink: 
* see below
* yes i supposed the doc meant to say : The **list** contains dead references.

For the proxy, something along the lines of (there's probably a pitfall i didn't think of tho...) :
```python
class WeakMethodProxy(WeakMethod):
    def __call__(self, *args, **kwargs):
        fn = super(WeakMethodProxy, self).__call__()
        if fn is None:
            raise something
        return fn(*args, **kwargs)
```

vallsv:
Here it is ready to merge.

silx/utils/weakref.py
```diff
@@ -0,0 +1,312 @@
...
+class WeakList(list):
+    """Manage a list of weaked references.
+    When an object is dead, the list is flaged as invalid.
+    If expected the list is cleaned up to remove dead objects.
+    """
+
+    def __init__(self, enumerator=()):
+        """Create a WeakList
+
+        :param enumerator list or tuple: A list of object to initialize the
```
t20100: syntax: :param [type] variable_name:
-> :param iterable enumerator:

silx/utils/weakref.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def reverse(self):
+        """Reverse the list inplace"""
+        self.__list.reverse()
+
+    def sort(self, cmp=None, key=None, reverse=False):
+        """Sort the list inplace.
+        Not very efficient."""
+        sorted_list = list(self).sort(cmp, key, reverse)
+        return WeakList(sorted_list)
```
t20100: I think it should be: self.__list.sort() for inplace sort (as for reverse)

silx/utils/weakref.py
```diff
@@ -0,0 +1,312 @@
...
+    If expected the list is cleaned up to remove dead objects.
+    """
+
+    def __init__(self, enumerator=()):
+        """Create a WeakList
+
+        :param enumerator list or tuple: A list of object to initialize the
+            list
+        """
+        list.__init__(self)
```
t20100: It seems list inheritance is not used as all methods uses self.__list.
Is it needed?

silx/utils/weakref.py
vallsv: Hooo shit, i do the reverse thing from a long time...

silx/utils/weakref.py
vallsv: It will test cmp on references objects, i dont know if it is part of the ref API

silx/utils/weakref.py
vallsv: You mean if `__list` is needed? Yes you are right, maybe no.
But it is better to inherite from list for that: `isisntance(weaklist, list)`.
I can try to remove `__list`.

vallsv:
Here is a fix of the sort method, and a fix on the param docstring.
Ready to merge.

PR #213, opened 2016-08-25
Fix. Add the Octaveh5 from the test_octaveh5 in the try section to av…

payno:
Fix bug of h5py import from unittests.


vallsv:
Looks fine

PR #215, opened 2016-08-29
Fix weaklist

vallsv:
- Fix sort
- Add reversed
- Add get/set/del slice
- Add str/repr

### DIFF
--- a/silx/utils/weakref.py
+++ b/silx/utils/weakref.py
@@ -29,7 +29,7 @@ from __future__ import absolute_import
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "29/08/2016"
+__date__ = "30/08/2016"
 
 
 import weakref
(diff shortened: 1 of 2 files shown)


silx/utils/weakref.py
```diff
@@ -305,9 +343,19 @@ def reverse(self):
...
+        Not very efficient.
+        """
+        sorted_list = list(self)
+        sorted_list.sort(key=key, reverse=reverse)
         self.__list = []
         self.extend(sorted_list)
+
+    def __str__(self):
+        unref_list = list(self)
+        return unref_list.__str__()
```
t20100: Maybe do`return 'WeakList(%s) % str(unref_list)`

silx/utils/weakref.py
```diff
@@ -305,9 +343,19 @@ def reverse(self):
...
+
+    def __repr__(self):
+        unref_list = list(self)
+        return unref_list.__repr__()
```
t20100: Same, maybe do`return 'WeakList(%s) % repr(unref_list)`

vallsv:
Yes, its a good idea

vallsv:
Here it is

PR #229, opened 2016-09-08
[octaveh5] correct section title and underlining in rst doc

PiRK:
Sphinx was raising a warning due to the underlining being shorter than the title


payno:
Bad copy-paste, thanks

Issue #231, opened 2016-09-09
[Plot] BackendMatplotlib::addItem modifies input arrays


dnaudet:
BackendMatplotlib.addItem modifies the *x* and *y* input arrays when those arrays are numpy arrays (and the selected shape is polygon or polylines) :
```python
xView.shape = 1, -1
yView.shape = 1, -1
```

PR #232, opened 2016-09-09
Using reshape to get a reshaped view without modifying the input data.

dnaudet:
Proposed fix for issue #231


kif:
looks good to me

PR #233, opened 2016-09-12
[Gui] Add loadUi to qt.py

### LINKED ISSUE #230: [Gui] Add uic to the list of components
In PyQt4 and PyQt5 there is a uic module which can be used to load .ui file created with Qt Designer (there is something similar in PySide). Would it be possible to have this under silx.gui?

t20100:
This PR adds uic.loadUi to qt.py.
For PySide, it provides an alternative implementation which only has the 2 first arguments of loadUi implemented (and warnings when the 2 last are set).

Closes #230

### DIFF
--- a/copyright
+++ b/copyright
@@ -19,6 +19,11 @@ Files: silx/gui/plot/MPLColormap.py
 Copyright: Nathaniel J. Smith, Stefan van der Walt, Eric Firing
 License: CC0
 
+Files: silx/gui/_pyside_dynamic.py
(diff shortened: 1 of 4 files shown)


vallsv:
Looks nice.
Here is a ploblem on pyside
```
  File "C:\projects\silx\venv_test\lib\site-packages\silx\gui\_pyside_dynamic.py", line 41, in <module>
    import logger
ImportError: No module named logger
```

vallsv:
Let's do it...

PR #234, opened 2016-09-13
Add isosurface marching cubes implementation

t20100:
This PR proposes a C++ implementation of the marching cubes algorithm (http://paulbourke.net/geometry/polygonise/).
It supports processing data slice by slice and sub-sampling.
It outputs 3 arrays: unique vertices, associated normals (computed from the data, not from the generated triangles) and vertex indices to build the isosurface.

It lies in the math package.


payno:
1. "<" vs "<="
In the original algorithm (defined in http://paulbourke.net/geometry/polygonise/) there is a "<" instead of "<=" in your get_cell_code function (mc.hpp file). 
I am not sure but this is probably a strict inferior you want because we might want to have all values inferior to the isolevel be "grouped" in the isolevel ?

2. data structure
It would be nice to have more information about the data parameter in some function like the "process"  of the .hpp and the constructor in the marchingcubes.pyx or a reference to your http://paulbourke.net/geometry/polygonise/
Because a definition of the data must be complicated (z up system + defining vertices xy slice by xy slice  )

Otherwise good to be merged for me.

payno:
After a discussion with @t20100 about "<" vs "<=" the vertices truly delimiting the isocurve are defined later in the algorithm (process_edge) and this doesn't affect the algorithm.

PR #235, opened 2016-09-15
Fix a bunch of PEP8 warnings

vallsv:
- trailing space
- unused var
- malformed comments
- indentation

That's my new install of Eclipse...

I also removed few lines of dead code on  `silx/gui/plot/LegendSelector.py `. `rect` was created but not used. I think it is equivalent.


t20100:
Looks good to me.

PR #240, opened 2016-09-19
Provide a stable html-escape API

vallsv:
See notes:
- there is no stable API with Qt binding
- there is no stable API with Python
- It is maybe not a good idea to use cgi mudule, or xml module dependencies


t20100:
Looks good to me

PR #243, opened 2016-09-20
Update rgba conversion function + add icons

t20100:
- Add support of 3-arrays as input to rgba function
- Add basic tests of rgba function
- Add cube icons in svg and png format


t20100:
Updated icons documentation page, ready for review.

vallsv:
That's fine

PR #245, opened 2016-09-20
[Plot] Method to retrieve the displayed data range.

dnaudet:
Issue : #252
I couldn't find any method to get the actual data range (not the "view" or axis range).
Tests not written yet (nor the issue report), waiting for comments (or for someone to tell me that this feature is already available).

### DIFF
--- a/silx/gui/plot/Plot.py
+++ b/silx/gui/plot/Plot.py
@@ -198,7 +198,7 @@ __license__ = "MIT"
 __date__ = "23/02/2016"
 
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 import logging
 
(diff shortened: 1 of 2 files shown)


t20100:
No method in Plot provides this, this is only available as a private method in the backend, and it makes sense to have it in Plot.py.

- Right axes should also be handled, so either return both left and right Y range or allow to select the axis (e.g.,  getDataRange(axis='left')).
- What about returning a namedtuple?
- It's probably worth to cache the min/max of each curves (e.g., when adding a curve), not just the data limits.
- Using nanmin/nanmax is safer than min/max.

dnaudet:
Right!

PR #251, opened 2016-09-21
Feature h5 load API

vallsv:
Provide a generic loading API for HDF5-like files. It supports h5 files and spec files.

- Create `load` API
- Create `h5py_class` method to flag mimicked objects with the referenced class

### DIFF
--- a/silx/io/spech5.py
+++ b/silx/io/spech5.py
@@ -164,10 +164,16 @@ from .specfile import SpecFile
 
 __authors__ = ["P. Knobel", "D. Naudet"]
 __license__ = "MIT"
-__date__ = "15/09/2016"
+__date__ = "20/09/2016"
 
 logging.basicConfig()
 logger1 = logging.getLogger(__name__)
+
+try:
+    import h5py
+except ImportError:
+    logger1.debug("Module h5py optional.", exc_info=True)
+
 
 string_types = (basestring,) if sys.version_info[0] == 2 else (str,)  # noqa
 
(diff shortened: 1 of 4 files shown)


vallsv:
That's a rework of the input h5 without the fabio part.

That is the minimal API needed for the Hdf5TreeView.

Fabio will be part of another PR.

vasole:
I can see the top level "keys", but I cannot go into them. I would expect that using the "dictionary-like" API one could easily achieve it.

vallsv:
Can you clarify what do you mean? The API provides h5py objects, then i guess you complain about SpecH5 API?

Can you show us something which should work but dont?

In this case i think it is better to create a side issue to be able to merge this PR. Or i can remove Spec from this API while it is not compliant. Cause it is a needed API.

vasole:
What I mean is that one should be able to use the same widget to perform the same actions on SPEC files, HDF5 files or Image files. That means to perform a selection of a dataset or a group. If one opens a SPEC file on that widget, one should get the same thing as if the file would have been converted to HDF5 and opened.

You never seem to have shared that point because you always wanted to provide the object instead of the selection that can be used to retrieve the object but the goal is to be able to use the same widget of HDF5, SPEC and FabIO opened files and not to end up with a different widget for each type of source.

vallsv:
Sure SpecH5 have to provide the same API as h5py, that's the goal. If there is an issue here we have to fix SpecH5. That is why i would like to know where it is not the case.

vasole:
Just for the record. Verified with a test file. The tree was not extended beyond first level nodes.

PR #254, opened 2016-09-21
Hdf5 TreeView

vallsv:
Current state
--------------------

Provides an Hdf5 TreeView

- Create the module `silx.gui.hdf5` which expose few classes
- Provide the example `examples/hdf5widget.py`
- Can be launched using `./bootstrap.py python examples/hdf5widget.py`

Missing features
-------------------------
As we said in the silx meeting, we can postpone all this missing features, if we want to have the widget next week for the code camp. The main widget is working. Let me know if something is really needed.

- [ ] Update model to support weakref
- [ ] Create signals for `fileLoaded`, `beforeFileClosed` (can abort close), `fileClosed`, `fileSynchronized`
- [ ] Standard menu to close/sync (the example provide it by itself right now)
- [ ] Improve sorting speed / sort initial items of the model
- [ ] Improve the `synchronized` to take care of the state of the view (expanded/collapsed/selection should stay the same)
- [ ] Clean up the code after merge of https://github.com/silx-kit/silx/pull/251
- [ ] Minimal unittest coverage


vasole:
- Under python 3.5, PyQt5, I get the error  in Hdf5HeaderView.py, line 53 in `__init__`  setClickable(True) Attribute error: Hdf5HeaderView object has no attribute setClickable. Under Python 2.7 PyQt4 the problem is not there.

- Something to think about, links, particularly external ones should provide the actual file name and actual selection path used to retrieve the object.

vasole:
Just to make it clearer. In fact every selection should report the file and path used to perform the selection and not just selections concerning links.

vasole:
Good "starting" poing

PR #255, opened 2016-09-22
Optimize test image

vallsv:


PiRK:
Appveyor build does not start, but the commit is simple enough to merge anyway.

Issue #249, opened 2016-09-21
[fit widget] Improvements


PiRK:
- [x] move configuration parameters into the customizable configuration dialog (`AutoFwhm` checkbox), because they are specific to the selected fit theory (duplicate with #241)
- [x] address warning generated by `leastsq` when running a gaussian estimation
- [x] make the fit curve use the same `xlabel` and `ylabel` as the original data
- [x] try to replace `QScriptOption` with a simpler configuration dialog widget

PiRK:
[ ] handle background functions the same way as regular fit functions / theories (handle adding custom functions, custom configuration dialog…)

vasole:
That last feature I would put it as a separate enhancement issue for a later milestone.

PR #256, opened 2016-09-22
Simplification of fit configuration module.

PiRK:
This PR replaces QScriptOption with a new simpler widget.


vasole:
Looks good. Some minor things:

[] - Initialize "internal background" to 2 and 5000 instead of 1 and 10000.
[] - When the user has configured the "internal background" via the interface, the estimation should reflect the user parameters and not the default.
[] - Somehow there is a CIRITICAL error calculating the covariance matrix when using the internal background as background.

vasole:
- Where the interface says curvature it should say width.
- The curvature factor should always be one.
- An additional parameter for a smoothing width is needed for a complete characterization

vasole:
- Unit test needs to be updated (testFitConfig.py line 77). The assert is now to be made with a value of 2 and not 1.

PiRK:
This PR closes #249

vasole:
The estimation of the step is wrong. You should not remove the background. Just calculate the derivative and pass that to the estimate function for the Gaussian functions.

PR #258, opened 2016-09-23
Missing doc for mask keyword.

dnaudet:
Missing doc for mask keyword.


vallsv:
Hi. There was a conflict. I just fix it and merge it here https://github.com/silx-kit/silx/pull/260

PR #259, opened 2016-09-23
Tutorial on spech5 and spectoh5

PiRK:
So far this tutorial contains explanations about SPEC data files and about how they are exposed in `spech5`, and minimalistic code examples.


vasole:
Doc about SPEC file is wrong in what concerns MCAs.

PR #293, opened 2016-09-27
[Plot] Fix polygon interaction

t20100:
Fix some issues with drawing polygons:
- Some click were not creating new points (bug in testing if close to previous point)
- There were issues when dragging the mouse while closing the polygon


vallsv:
In `onRelease` if `btn` is not left or right, it will return None, while others returns a boolean. Is that safe? Or this code can't be reached with something else than left or right click?

t20100:
That was safe, but I added return False to make it clearer.
The returned value is used to know if the current interaction keeps the focus after mouse release.

vallsv:
Looks fine. Thanks a lot.

PR #294, opened 2016-09-27
Feature h5 fabio input (do not merge)

vallsv:
Must be merge after https://github.com/silx-kit/silx/pull/251

Current state
--------------------

- Feature Fabioh5 module providing h5py mimicked objects
- Currently fabio structure is exposed like that:
    - file
        - metadata (header of the first frame)
            - meta1: dataset
            - meta2: dataset
        - frames
            - frame1: 2d image
            - frame2: 2d image

Todo
--------

- [ ] Implement the structure discussed at silx meeting (2016-09-21)
    - file
        - data: 3d stack
        - info
            - frame1
                - meta1: dataset
                - meta2: dataset
            - frame2
                - meta1: dataset
                - meta2: dataset
- [ ] maybe use nexus structure for motors (positionner) and counters (measurment)


vallsv:
- The tree is now provided as expected
- I update the documentation
- I use a numpy feature to use the best type fitting the data, then it will use float16, float32, int, uint... I think it is the best.
- I also add the interpretation=image to the dataset, but it is not yet used in the widgets

PR #296, opened 2016-09-27
[console] Improve ImportError handling

### LINKED ISSUE #263: RichIPythonWidget missing can cause tests to fail
Detection and import of dependencies for the console widget needs to be improved.

This part of the conditional  import seems fragile : https://github.com/silx-kit/silx/blob/master/silx/gui/console.py#L116 

It is probably a good idea to drop support for older versions of IPython for this console feature  and make the presence of the `qtconsole` package mandatory to be able to use the widget. We already specify in the install documentation that the console widget depends on the  `qtconsole` library.

PiRK:
Remove logger error inside `except` clause just before raising the error, because the logger error cannot be quieted when the error is caught and handled downstream.
This closes #263.

Change handling of import, try ipython.qtconsole when qtconsole is present but has no RichIPythonWidget.

### DIFF
--- a/silx/gui/console.py
+++ b/silx/gui/console.py
@@ -111,6 +111,9 @@ if qtconsole is not None:
             qtconsole = None
         else:
             from qtconsole.inprocess import QtInProcessKernelManager
+    else:
+        from qtconsole.inprocess import QtInProcessKernelManager
+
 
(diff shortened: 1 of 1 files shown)


payno:
Looks good to me

PR #299, opened 2016-09-28
Fixed typo : rezetZoom -> resetzoom.

dnaudet:
Fix for CR #300

### DIFF
--- a/silx/gui/plot/Plot.py
+++ b/silx/gui/plot/Plot.py
@@ -2387,7 +2387,7 @@ class Plot(object):
             for legend in list(self._images):  # Copy has images is changed
                 image = self._images[legend]
                 self.addImage(image['data'], legend,
-                              replace=False, rezetZoom=False,
+                              replace=False, resetzoom=False,
                               pixmap=image['pixmap'], **image['params'])
 
     # Coord conversion


vasole:
@t20100 

If we follow camel case, resetZoom should be the correct option and resetzoom get the deprecation warning. I do not care so much and whenever possible I try to avoid more-than-one-word keywords and if it is a keyword already existing in matplotlib I just use it. If it were my choice, I would just silence the deprecation warning and accept both forms silently.

Qt would write resetZoom, xScale, yScale and so on because it is self-consistent on the whole API (including keywords) while matplotlib writes xscale and yscale. So, please clarify. In fact, matplotlib does whatever it wants at will in what concern keywords:

Line2D(xdata, ydata, linewidth=None, linestyle=None, color=None, marker=None, markersize=None, markeredgewidth=None, markeredgecolor=None, markerfacecolor=None, markerfacecoloralt='none', fillstyle=None, antialiased=None, dash_capstyle=None, solid_capstyle=None, dash_joinstyle=None, solid_joinstyle=None, pickradius=5, drawstyle=None, markevery=None, **kwargs)

t20100:
I also try to avoid more than one word arguments and when it happens my taste is to use lower case (without _ ).

xScale and yScale aguments come from PyMca where there is also xlabel and ylabel arguments.
I deprecated xScale and yScale in favor of origin and scale arguments in silx.

For silx plot API, I've used lower case for arguments, so resetZoom is a bug to me, I would only keep resetzoom.
I would keep lower case arguments in the plot unless we really want to enforce camelCase arguments through out silx.gui which would break compatibility with PyMca plot API (e.g.,  linewidth, xlabel, yaxis...).

vallsv:
Then, i merge that for the code quality.
But you still can continue to talk about the ideal API.

vasole:
@t20100 

OK. Let's say we assume that plot API will try to be lowercasewithoutunderscores.

t20100:
Ok for me.

PR #302, opened 2016-09-29
[Plot] Improve right axis support

### LINKED ISSUE #253: [Plot] Right axis range not updated.
The behaviour of the right Y axis (and all plots associated with it) is wrong.

```python
import numpy as np
from silx.gui import qt

from silx.gui.plot import PlotWindow

app = qt.QApplication([])

plotWin = PlotWindow()

plotWin.addCurve(np.arange(100), np.arange(100), legend='p1', yaxis='left')
plotWin.addCurve(2*np.arange(100), np.arange(100), legend='p2', yaxis='right')

plotWin.setKeepDataAspectRatio(True)
plotWin.show()
app.exec_()
```

t20100:
This PR improve right axis support:

- Closes #253
- Fix right axis support in pixelToData, dataToPixel methods
- Add support of aspect ratio for right axis

It also make mpl backend use getDataRange (instead of duplicate internal method) and it fixes getDataRange under certain conditions: image scale < 0, log axes.

It still misses non-regression tests.


t20100:
Add tests, thanks @dnaudet! 

Ready for review

payno:
Ok for me

PR #304, opened 2016-09-29
[Plot] Fix pan with arrow keys disabled at start-up.

### LINKED ISSUE #291: [Plot] Pan with arrow keys not enabled the first time
FocusPolicy is not set at the creation of the widget

t20100:
Closes #291 
setFocus in `__init__` and in `main()` of ImageView.


vallsv:
Looks fine

PR #305, opened 2016-09-29
OpenCL convolution: more readable kernels

pierrepaleo:
OpenCL 2D convolution kernels were re-written to be more readable.
They are also compatible with scipy.ndimage for even-sized convolution filters.


kif:
LGTM

PR #307, opened 2016-09-30
Fix errors in SIFT tests

pierrepaleo:
This PR fixes two errors with the sift tests:
- Fix error with array.sum() for new versions of pyopencl
- Do not perform test_matching if there is no "feature" module

Also, another (equivalent but more numerically stable) least squares routine is used for the linear alignment.

### DIFF
--- a/silx/image/sift/test/test_gaussian.py
+++ b/silx/image/sift/test/test_gaussian.py
@@ -121,7 +121,7 @@ class TestGaussian(unittest.TestCase):
         """
         Calculate a 1D gaussian using pyopencl.
         This is the same as scipy.signal.gaussian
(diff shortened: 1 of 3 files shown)


kif:
Hi Pierre,
About the commit 61f057b:
The test of matching of matching should be run even if the result cannot be checked to ensure no bugs is hidden there. Moreover the "L1 norm" is cheap enough so that it could be done in pure Python at little additional cost.

pierrepaleo:
OK, we revert to this behavior with 8e18d66886e15b36ba8ce883d4d5c0f7b8f7abca

kif:
LGTM

Issue #280, opened 2016-09-26
[gui] First parameter of widget must be parent widget


vallsv:
- When using the QtDesigner and promote widget, Qt expect to have parent widget as the first param

t20100:
Following discussions, when the widget's __init__ takes additional parameters we propose to handle this by having parent=None as the first argument of widget's `def __init__(self, parent=None, ....)`.
This is in contradiction with Qt convention (where parent is the last argument), but allows to promote widgets in the QtDesigner.
Anyway, it would be good to have a consistent approach for all widgets in silx.

PiRK:
For the widgets I have added to silx, the ones not following that convention are:

 - silx.gui.fit.FitConfig
    
    - ConstraintsPage
    -  BackgroundPage

- silx.gui.console

    - IPythonWidget
    - IPythonDockWidget

t20100:
I'm doing it for the widgets of the plot package

PR #313, opened 2016-10-03
[Plot] plot widgets have parent=None first + addImage origin and scale support float

### LINKED ISSUE #290: [Plot] addImage origin and scale parameters should accept a single float
addImage currently accept a 2-tuple of float, it would be convenient to accept a single float to set both values at once.

t20100:
This PR:
- Changes parameter order of plot widgets arguments to have parent first: Related to #280 
- Adds support of single float value for addImage origin and scale arguments: Closes #290


payno:
Looks good to me.

PR #323, opened 2016-10-05
Improve mask tool

vallsv:
- Change behaviour of aspect ratio and y-axis origin icons
- Display mask shape using the mask color
- Display shape for the pencil tool before drawing anything
- Update mask icon


vallsv:
Ready to be merged.

The preview of the mask is part of another PR: https://github.com/silx-kit/silx/pull/332

mretegan:
It seems that following this, the aspect ratio and y-axis origin icons are always present in the PlotWindow toolbar regardless of the value set by the user at instantiation.

vallsv:
Intresting. It would be good to fix it before the release.

t20100:
Opened issue #344 for this

PR #327, opened 2016-10-05
mainly pep8

### LINKED ISSUE #326: Many small things in Octaveh5
I'll submit a PR

kif:
close #326

### DIFF
--- a/silx/io/octaveh5.py
+++ b/silx/io/octaveh5.py
@@ -22,7 +22,7 @@
 #
 # ############################################################################*/
 """
-Python h5 module and octave h5 module have differente ways to deal with
+Python h5 module and octave h5 module have different ways to deal with
 h5 files.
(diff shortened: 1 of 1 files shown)


vallsv:
I think it is not a good idea to remove the `return None` of the `open`. At least it is clear that a `None` is returned. But i don't see why this function to not raise the exception.

Else it looks fine

kif:
On Wed, 05 Oct 2016 06:37:44 -0700
Valentin Valls <notifications@github.com> wrote:

> I think it is not a good idea to remove the `return None` of the `open`. At least it is clear that a `None` is returned. But i don't see why this function to not raise the exception.

Me neither ... @payno ?

-- 
Jérôme Kieffer
tel +33 476 882 445

vallsv:
Thanks for the fix. Let see it.

PR #335, opened 2016-10-06
[spectoh5] Fix critical bug when writing SpecH5Dataset to HDF5

PiRK:
Now that SpecH5Dataset is no longer a numpy array, we need to write the `.value` attribute to HDF5, and not the dataset itself.


vallsv:
Looks fine

PR #346, opened 2016-10-10
[Plot] Fix not possible to hide toolbar tool buttons

### LINKED ISSUE #344: [Plot] Keep ratio and invert Y axis push buttons always visible
Keep ratio and invert Y axis push buttons are always displayed in the toolbar.
See #323
This is a regression.

### LINKED ISSUE #345: [gui] Test hdf5 widget is using none standard decorator module
Use 
`from contextlib import contextmanager`
rather than
`from decorator import contextmanager`

t20100:
This PR:

- Closes #344 : aspect ratio and y axis orientation can be hidden as before
- Closes #345 : Replace non standard decorator module by contextlib
- Change PlotToolButton constructor parameters to have parent as first argument

### DIFF
--- a/silx/gui/hdf5/test/test_hdf5.py
+++ b/silx/gui/hdf5/test/test_hdf5.py
@@ -33,7 +33,7 @@ import time
 import unittest
 import tempfile
 import numpy
-from decorator import contextmanager
+from contextlib import contextmanager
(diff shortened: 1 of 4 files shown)


vallsv:
Nice. Thanks a lot.

PR #348, opened 2016-10-11
[Plot] Fix aspectratio

### LINKED ISSUE #347: [plot] KeepAspectRatio=True => Images/Plots not centered
Images and plot are not centered when keep aspect ration is set.

```python
import numpy as np
from silx.gui import qt as Qt
from silx.gui import plot

data = np.ones((100, 200))
data[0::5, 0::5] = 0

app = Qt.QApplication([])
plotWin = plot.PlotWindow()
plotWin.addImage(data, origin=(10, 30))
# plotWin.setActiveCurveHandling(False)
plotWin.setKeepDataAspectRatio(True)
plotWin.show()
app.exec_()
```

t20100:
Closes #347 : Bug with aspect ratio + add some tests for it


dnaudet:
Works fine, to be merged when CI is done. Thanks Thomas!

PR #349, opened 2016-10-11
Test before fixing issue on out Mac OS laptop

vallsv:
There is an issue on our laptop relative to the behaviour or QThreadPool.

It still must be fixed, based on this tests.


payno:
Looks good

PR #358, opened 2016-10-13
Code quality

### LINKED ISSUE #352: ThreadPoolPushButton test fails
I get this traceback:

```
ERROR:silx.gui.widgets.ThreadPoolPushButton:Error while executing callable silx.gui.widgets.test.test_threadpoolpushbutton._computeFail.
  File "/Users/marius/Software/silx/build/lib.macosx-10.11-x86_64-3.5/silx/gui/widgets/ThreadPoolPushButton.py", line 56, in run
    result = self.__callable(*self.__args, **self.__kwargs)
  File "/Users/marius/Software/silx/build/lib.macosx-10.11-x86_64-3.5/silx/gui/widgets/test/test_threadpoolpushbutton.py", line 54, in _computeFail
    raise Exception("exception")
Exception: exception
```

Is this error supposed to be logged at the end of the test? I get this:

```
Ran 691 tests in 62.155s

OK (skipped=1)
```

vallsv:
- Filter and test logging message (close #352)
- Ignore unreleased error widgets if the test fail
- Fix bunch of warnings (unused var, lib, space, trailing space...)
- Refactoring main function from ImageView as an example

### DIFF
--- a/examples/fftPlotAction.py
+++ b/examples/fftPlotAction.py
@@ -39,7 +39,7 @@ See shiftPlotAction.py for a simpler example with more basic comments.
 """
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
-__date__ = "20/09/2016"
+__date__ = "13/10/2016"
 
 import numpy
 import os
@@ -105,8 +105,8 @@ class FftAction(PlotAction):
 
         self.plot.clearCurves()
 
-        for i, curve in enumerate(allCurves):
-            x, y, legend, info, params = curve
+        for curve in allCurves:
+            x, y, legend, info, _params = curve
 
             if checked:
                 # FAST FOURIER TRANSFORM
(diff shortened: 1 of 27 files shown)


examples/fftPlotAction.py
```diff
@@ -105,8 +105,8 @@ def fftAllCurves(self, checked=False):
 
         self.plot.clearCurves()
 
-        for i, curve in enumerate(allCurves):
-            x, y, legend, info, params = curve
+        for _i, curve in enumerate(allCurves):
```
t20100: `_i` seems unused, why not just` for curve in allCurves:`

vallsv:
Fixed

examples/fftPlotAction.py
PiRK: Probably a remnant from a past version. I agree that `i` and `enumerate` are no longer necessary.

vallsv:
After talking to Pierre, i remove changes on indentation (our editors never agreed with each others).

PR #359, opened 2016-10-14
BackgroundDialog

PiRK:
bgwidget

This PR adds a `BackgroundWidget` widget and a `BackgroundDialog` widget, to help with picking parameters for *strip* and *snip* background computation algorithm prior to a fit. As the parameters are modified, the results are dynamically updated and displayed in a PlotWidget.

Other important features introduced by this PR:

- FitManager and FitWidget now have a *snip* background option by default

- the *strip* and *snip* background algorithms can now use anchors and be preceded by an adjustable smoothing filter (Savitsky-Golay)

- adding new custom background theories has been implemented in FitManager and FitWidget
       - background theories are no longer hardcoded inside FitManager, the default ones are now defined in a new `silx.math.fit.bgtheories` module (the same way default fit theories are defined in `fittheories` module)
  
- the custom configuration widgets must now be handled by `FitWidget` (using a new method `associateConfigDialog(theory_name, custom_dialog)`), instead of being defined as an attribute of `FitTheory`, to avoid mixing math and gui.

- `FitWidget` now has two *Configure* buttons, because both the fit theory and the background theory can have a custom configuration dialog. The *snip* and *strip* background theories are now associated with the new  `BackgroundDialog`.


PiRK:
bgwidget

PiRK:
(the typo in the widget title you can see in the screenshot has been corrected in the code)

vasole:
Just a minor comment. The SNIP background should have a default parameter much larger than 2. At the very least, twice the expected FWHM in points of the peaks. So, if the default parameter for peak search is 8 points, a reasonable parameter would be 16.

vasole:
There is a bug but it is independent of this PR.

PR #365, opened 2016-10-17
Improve masking tool

### LINKED ISSUE #364: [mask] Changing mask/unmask while creating a polygon is not taken into account

vallsv:
- Use QLabel for status bar
- Update default mack color according to colormap
- Rework doMask to take care of realtime changes (fix #364)
- Avoid to draw 2 times updateDisk at the same place

### DIFF
--- a/silx/gui/plot/Colors.py
+++ b/silx/gui/plot/Colors.py
@@ -26,7 +26,7 @@
 
 __authors__ = ["V.A. Sole", "T. VINCENT"]
 __license__ = "MIT"
-__date__ = "15/09/2016"
+__date__ = "17/10/2016"
 
 
 import numpy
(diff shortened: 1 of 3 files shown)


vallsv:
As side note, it would be nice to improve the draw_line to use a draw_polygon. draw_line to not respect the requested width.

t20100:
This looks good to me.

I found the 2 following issues:
- There seems to be a conflict between ctrl-c on the label and to one from the plot.
- The color of the preview (for pencil, rectangle and polygon) is not updated when the colormap is changed. It gets updated when a new tool is selected.
Those issues were already existing before this PR, so they don't necessarily need to be fixed now.
Do you want to tackle them now or leave it for another PR?

PR #368, opened 2016-10-18
[gui] ArrayTableWidget

PiRK:
This PR adds a widget  to view multi-dimensional numpy arrays as stacks of 2D arrays. The 2D array corresponding to the last two dimensions of the array is displayed in a QTableView. The indices of the other dimensions can be set by moving sliders, by typing text in a line edit or by using buttons/icons to browse the data.

arrayview
_4D array of shape (4, 5, 1000, 1000)_

Currently the widget is almost identical to the one in PyMca5, with only minor refactoring.

To be discussed: 
- module name: `silx.gui.array`? directly in `silx.gui` to reduce the number of folders?
- widget name: `ArrayWidget`? `ArrayTableWidget`? `NumpyArrayWidget`? `NumpyArrayTableWidget`?
- should we provide a way to display frames with any _perspective_ (not only frames whose axes are the last two dimensions of the array)? This is already implemented for 3D array, and could be generalized to higher dimensions.
- possibility to add a color to the cell's background depending on its value, to make a gigantic pixelized image? 

Todo list:
- [x] generalize perspective to n-dimensional case, to be able to see any  2D slice in any array
- [x] tests
- [x] documentation
- [x] tutorial
- optional colormap (difficult and potentially not very readable)


PiRK:
It seems that "array" as a package name will cause issues on some configurations:

```
$ python setup.py $BUILD_COMMAND

  File "setup.py", line 322, in <module>

    config = configuration()

  File "setup.py", line 318, in configuration

    config.add_subpackage(PROJECT)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 1000, in add_subpackage

    caller_level = 2)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 969, in get_subpackage

    caller_level = caller_level + 1)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 906, in _get_configuration_from_setup_py

    config = setup_module.configuration(*args)

  File "silx/setup.py", line 35, in configuration

    config.add_subpackage('gui')

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 1000, in add_subpackage

    caller_level = 2)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 969, in get_subpackage

    caller_level = caller_level + 1)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 906, in _get_configuration_from_setup_py

    config = setup_module.configuration(*args)

  File "silx/gui/setup.py", line 35, in configuration

    config.add_subpackage('plot')

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 1000, in add_subpackage

    caller_level = 2)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 969, in get_subpackage

    caller_level = caller_level + 1)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 906, in _get_configuration_from_setup_py

    config = setup_module.configuration(*args)

  File "silx/gui/plot/setup.py", line 35, in configuration

    config.add_subpackage('test')

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 1013, in add_subpackage

    dist = self.get_distribution()

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/misc_util.py", line 848, in get_distribution

    from numpy.distutils.core import get_distribution

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/numpy/distutils/core.py", line 10, in <module>

    from setuptools.command import easy_install

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 52, in <module>

    from setuptools.package_index import (

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/setuptools/package_index.py", line 15, in <module>

    from urllib2 import splituser

  File "/opt/python/2.7.9/lib/python2.7/urllib2.py", line 94, in <module>

    import httplib

  File "/opt/python/2.7.9/lib/python2.7/httplib.py", line 69, in <module>

    from array import array

ImportError: cannot import name array
```

t20100:
@vallsv started to place 'general purpose' widgets in silx.gui.widgets.
I would go for a silx.gui.widgets.ArrayWidget...
It is would be nice to optionally use a colormap to color the array.

As for the setup issue, try to add import array at the beginning of the top level setup.py... This worked for a similar issue with the io package.

PiRK:
I moved the widget to silx.gui.widgets. I'm thinking of using this PR to also move `console.py` to `silx.gui.widgets`, unless anyone disagrees with this move.

vasole:
- Personally I consider better to have as much as possible a single class on a single file. Therefore the orginal split. Qt itself separates them too.  Our HDF5 widgets too. Also, the FramwBroser could be used, and it was was used, for other things than the table.

- Again, unless you have clear reasons to change the name of a widget (i.e. widget name: ArrayWidget? ArrayTableWidget? NumpyArrayWidget? NumpyArrayTableWidget?) do not change it. It only asks for unnecessary work. 

In my opinion the existing widgets needed very little refactoring and I think you are just asking for unnecessary work.

vasole:
Concerning the colormap, I have seen it somewhere. Could it be on Oxidis?

vasole:
I hope we do not keep silx.gui.array or any module "array"anywhere (https://docs.python.org/3/library/array.html)

t20100:
I would keep console separated from other widgets as it has an extra dependancy on ipython...

vasole:
Please, do not misunderstand me. If you have any valid reason to change the name of a module, do so. What I do not want is names changed without any reason. For instance you could argue "have a shorter name", "support of something else than numpy arrays", ... but not changing them without any justification.

PiRK:
The only reason was having a shorter name, but I don't really have a strong opinion on this subject, so I'm fine with changing it back.

OK to put one class per file, but than I think I should add a subpackage to `widgets`, and expose the main widget (`NumpyArrayTableWidget`) in `widgets.__init__.py`, for the sake of maintaining a readable structure in `widgets`. If `FrameBrowser` starts being used elsewhere, we can then expose it later.

vasole:
If it supports other things than numpy arrays (for instance HDF5 datasets) then dropping "Numpy" from the name is fully justified. So, one thing to be checked.

vallsv:
Exposing all the widgets form package `silx.gui.widgets` into `silx.gui.widgets.__init__` is, i think, not a good idea, while this package is supposed to contain a lot of widgets.

PiRK:
@vallsv suggested that we could move the model to a new package `silx.gui.models`. It is to be discussed, it could be a good idea in the future if we consider having multiple models that each serve multiple views.

PiRK:
What about making the cells editable and providing a `getData` method to return the updated array? 
Would that be of any use to someone? It could be a low-level way to edit the data.

vasole:
To expose all the widgets at a single place does not shock me (PyQt exposes most of them at a single place) but I do not have a clear opinion.

t20100:
My opinion was to put 'simple' widgets (i.e., single file) into silx.gui.widgets and more complicated ones (e.g. plot, hdf5) or the one with extra dependencies (e.g., console) into their own package.
But this seems to need some discussion.

PiRK:
I created a separate issue for the colormap, this will not be implemented in this PR.

Regarding the icons of FrameBrowser, they should be changed with something more in line with the silx graphic charter, but that's not really my area of expertise.

PiRK:
Back to WIP, I need to sort the problem of 1D datasets being copied and document the API of FrameBrowser. Should be fixed before the end of the day.

vasole:
ArrayTableWidget setArrayData method is doing a copy systematically and it should not. (Reminder numpy.array defaults the copy keyword to True)

PiRK:
Fixed. I don't really need the numpy.array, I just need to guess the shape and dtype.

I also changed the behavior of the model, to raise a TypeError if the user requests copy=False but the data must be copied (instead of just logging a warning and forcing the copy).

t20100:
Late comment: with numpy.array copy=False makes a copy only if needed...
I personnally expect APIs taking an array-like object and a copy argument to behave the same way (so they might make a copy with copy=False).

PiRK:
I find the numpy behavior slightly dangerous. If someone use copy=False to get a memory view, because he wants to ensure that any change to the data will really modify the original data in memory, the hidden copy will cause a problem.

If we want this widget to be usable as an array editor, working directly on the original data without having to `setData`/`getData` at every modification, I suggest that we keep the new behavior (raise an exception if `copy=False` is requested but impossible). 

However, if we decide that direct editing of numpy arrays and h5py datasets is unsafe and must be disabled anyway, then I agree that we could make copy=False work the same way as numpy.array.

PR #371, opened 2016-10-18
[Plot] Fix leave event not generate for fast mouse movement

t20100:
This PR uses Qt mouse leave and move events rather than matplotlib enter/leave events to trigger enter/leave events in state machine.
This avoids missing leave events when there is no mouse move event in the outer part of the plot.


vallsv:
Looks to work fine. Thanks a lot

PR #380, opened 2016-10-24
[Plot] Fix display of negative values with log scales

### LINKED ISSUE #330: [plot1d] issue when switching to log scale:
```python
import numpy
from silx import sx
sx.plot(numpy.sin(numpy.linspace(-10, 10, 1000)))
#then switch to log-scale

%pylab
plot(numpy.sin(numpy.linspace(-10, 10, 1000))
yscale("log")
```
matplotlib gives a sensible graph (with discontinuities) while silx connects the last positive values.

t20100:
This PR fix display of negative values with log scale.
It replace values <= 0 with nan to show discontinuities.
It also filter xerror and yerror values if any of those is spanning error bars below 0 but changing those to nan.

Values stored with curves remains unchanged.

Closes #330

### DIFF
--- a/silx/gui/plot/BackendMatplotlib.py
+++ b/silx/gui/plot/BackendMatplotlib.py
@@ -159,6 +159,11 @@ class BackendMatplotlib(BackendBase.BackendBase):
             else:
                 errorbarColor = color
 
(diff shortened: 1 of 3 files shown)


kif:
LGTM

Issue #378, opened 2016-10-21
[doc] document the installation of a venv/virtualenv for working with silx


kif:
1) install vitrualenv
2) create a vitualenv
3) upgrade pip
4) install setuptools & wheel
5) install numpy, cython, scipy & matplotlib
6) install h5py, fabio, and silx
7) install pyopencl, mako

Comes after a discussion with VincentF-N

PiRK:
The PyQt dependency for silx.gui is a non-trivial issue, if we want to completely avoid using `sudo apt-get install`.

Even by adding Qt4 or Qt5 to the requirements to avoid having to explain how to install the Qt libraries, I'm stuck when trying to install PyQt, which requires SIP, which requires flex.

The most recent tutorial I found for installing PyQt in a virtualenv is https://www.metachris.com/2016/03/how-to-install-qt56-pyqt5-virtualenv-python3/ . It does not cover cases where ``apt-get install`` is not an option.

PiRK:
After installing *silx* in my venv (without pyqt), 3 sift tests end with an error:

`======================================================================
ERROR: test_gradient (silx.image.sift.test.test_image.TestImage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/users/knobel/venvs/silx_venv/local/lib/python2.7/site-packages/silx/image/sift/test/test_image.py", line 94, in test_gradient
    border_dist, peakthresh, EdgeThresh, EdgeThresh0, octsize, scale, nb_keypoints, width, height, DOGS, g = local_maxmin_setup()
TypeError: 'NoneType' object is not iterable

======================================================================
ERROR: test_local_maxmin (silx.image.sift.test.test_image.TestImage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/users/knobel/venvs/silx_venv/local/lib/python2.7/site-packages/silx/image/sift/test/test_image.py", line 133, in test_local_maxmin
    border_dist, peakthresh, EdgeThresh, EdgeThresh0, octsize, s, nb_keypoints, width, height, DOGS, g = local_maxmin_setup()
TypeError: 'NoneType' object is not iterable

======================================================================
ERROR: test_interpolation (silx.image.sift.test.test_image.TestImage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/users/knobel/venvs/silx_venv/local/lib/python2.7/site-packages/silx/image/sift/test/test_image.py", line 194, in test_interpolation
[...]

Issue #176, opened 2016-07-08
[CI] Add running GUI tests on Travis-CI


t20100:
For now, GUI tests only run with AppVeyor.
Add them to Travis for:
- [ ] Linux
- [ ] Mac OS X

vasole:
Can this issue can be closed? GUI tests are certainly run on linux. It is not the case for the Mac but I do not know if it is just our choice.

vallsv:
Thomas is working on, as i understand

t20100:
Yes coming soon

PR #381, opened 2016-10-25
[WIP][doc] installing silx in a virtualenv

PiRK:
This is a PR to add a  documentation page for installing silx in a virtualenv, for users with no sudo privileges.

It addresses #378  

The section about PyQt is still work in progress. We probably need to create our own PyQt manylinux wheels for python 2.7 and python 3.4.

Regarding the rest of the paragraphs, do we need more details?


t20100:
manylinux PyQt4 wheels would also be useful for travis.... Issue #176

kif:
LGTM

PR #369, opened 2016-10-18
[Plot] Add a Colorbar widget

t20100:
This PR proposes to add a widget displaying a colorbar synchronized with a PlotWidget displayed image or default colormap.
It wraps matplotlib colorbar in a QWidget to automate synchronisation.

TODO:
- [ ] More tests
- [ ] Add a button to edit the colormap
- [ ] Integrate with Plot2D and ImageView
- [ ] If possible rework layout to avoid having to much white space...

Closes #261

PR #383, opened 2016-10-26
Feature matplotlib init

vallsv:
Cherry-picked from https://github.com/silx-kit/silx/pull/369
- 389bb0b
- 4ad741c
- 744c196

Plus typo

### DIFF
--- a/silx/gui/plot/BackendMatplotlib.py
+++ b/silx/gui/plot/BackendMatplotlib.py
@@ -32,35 +32,17 @@ __date__ = "15/09/2016"
 
 
 import logging
-import sys
 
 import numpy
 
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


t20100:
Looks good to me ;)
Anyone else willing to do the review?

vallsv:
BTW isn't there a problem with PySide+Qt5?

t20100:
PySide is for Qt4.
There is a PySide2 project for Qt5 support: http://wiki.qt.io/PySide2

vallsv:
Oh, then it is fine. Thanks a lot for the feed back.
It would be very nice if they call it PySide4... :-D

vasole:
Are you sure you want to import matplotlib systematically when importing plot? At a certain point we should be able to choose a different backend.

PR #389, opened 2016-11-03
[sift] handle dtypes uint64 and float64

PiRK:
It would be preferable to handle image with these 2 dtypes, as numpy uses them as default in some situations. For example, when attempting naive conversion from RGB to black and white:

- numpy.sum(..., axis=-1) on RGB images (shape=(h, w, 3), dtype=uint8) returns uint64
- numpy.mean(..., axis=-1) on uint8 returns float64


pierrepaleo:
Looks good, thanks !

PR #403, opened 2016-11-10
[dictdump] Uniformize default file mode for various formats

PiRK:
Use "w" by default. "a" (append) causes loading issues with ini files, when the same dictionary is appended to a file: duplicated section headers.


kif:
LGTM

Issue #409, opened 2016-11-15
spectoh5.write_spect_to_h5 unclosed file when providing a filename when other exception


kif:
If there is an exception (for example an IO error or another hdf5 erro) the hdf5 file is not closed when leaving.
One should use try/finally.

kif:
This set of nested functions smells a bit

PR #411, opened 2016-11-15
attempt to address issue #409

kif:
This is a attempt for a fix, it still smells. I believe a class would be better on.


PiRK:
I'll accept the PR now to have the bug fixed, and keep the issue open to look into the refactoring later (after the trainings).

PR #420, opened 2016-11-15
[Plot] Fix data range not correct when some curves are hidden

### LINKED ISSUE #414: [plot] Hidden curve seems to be taken into account for reset zoom
Hiding a curve from the legend panel does not change the plot bounds when reseting the zoom

t20100:
Hidden curves were taken into account by Plot.getDataRange.
This PR does not take them into account for the data range and marks the data range as dirty when a curve is hidden or shown.

Closes #414


PiRK:
Looks good to me.

PR #431, opened 2016-11-23
Copy/paste table

PiRK:
QTableWidget with 4 actions for copy selection, copy all, cut selection & paste, and a context menu displaying the actions (right-click to open the menu).
Cut and paste is disabled by default.


PiRK:
I'm still working on trying to do the same for a QTableView (a QTableWidget without  data model).
The API documentation still needs to be added. For this I must wait until #368 is merged.

PiRK:
Ready for review, only API documentation are still missing (rst files).

PiRK:
Ready for review.

Note that this PR modifies the ``FitWidget`` and ``ArrayTableWidget`` to use these new table widgets.

I used this opportunity to set `editable=False` by default in the ArrayTableWidget. Now if someone *really really* wants to use this widget as a HDF5 file editor, he must prove that he knows what he is doing by explicitly setting both `editable=True` **and** `copy=False` (with `copy=True` he would end up editing a copy of the data).

PR #434, opened 2016-11-24
Refactor spectoh5 to avoid using nested functions & semi-global vars

### LINKED ISSUE #409: spectoh5.write_spect_to_h5 unclosed file when providing a filename when other exception

PiRK:
Create a writer class initialized with all parameters needed for the conversion. 
These parameters are now instance attributes rather than global variables accessed in sub-functions.
The actual conversion method then needs only a SpecH5 and a h5py.File. This simplifies the ``with ...:`` block that ensures that the HDF5 file is always closed.

Closes #409.
See also #411

### DIFF
--- a/doc/source/modules/io/spectoh5.rst
+++ b/doc/source/modules/io/spectoh5.rst
@@ -5,4 +5,4 @@
 -----------------------------------------------------
 
 .. automodule:: silx.io.spectoh5
-    :members:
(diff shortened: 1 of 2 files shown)


vallsv:
Looks good to me

Issue #412, opened 2016-11-15
[Plot1D] fit in zoomed area and in selected ROI


t20100:
The current fitting in plot1D seems to fit the whole data set.
It will be nice to be able to fit in the zoomed area (like PyMCA) or better yet, to fit the active ROI

vasole:
If it is not able to fit just the zoomed area, I would consider it a bug.

To fit the active ROI it would make the fit to be an action of the ROI and not of the plot. I would not go that way.

PR #435, opened 2016-11-25
Set data range in FitAction

PiRK:
When users click the fit icon in a PlotWindow, fit only data in the visible X limits.

Currently, the limits are only set when the Fit icon is clicked. 

For future reference, it is technically possible to set the limits automatically, with a callback, when the zoom is changed (see commented code in https://github.com/silx-kit/silx/commit/7c84dced3e9bc26ebf84ea391b52d167f1856151).


PiRK:
closes #412

PR #436, opened 2016-11-28
[specfile] prevent segfault when file is not a spec file

### LINKED ISSUE #424: [io] SpecH5 seg fault when the file is not spec file
If possible test the file before opening it with the C code to check that it is spec and raise an IOError instead of set fault.

PiRK:
Test for presence of #S or #F at the beginning of the file before opening the file.

Closes #424


vallsv:
Fine :-)

PR #442, opened 2016-11-29
[ArrayTableWidget] Add dimensions selector widget

### LINKED ISSUE #432: [arrayview] add widgets to select the table dimension
Currently we can only change the dimensions displayed in the table view programmaticaly. It would be convenient to add  an interactive way of doing this, through the array widget itself.

One solution would be to add a checkbox in front of each FrameBrowser, and configure them so that only to boxes can be ticked at the same time. Wen a check box is ticked, disable the frame browser for the corresponding dimension, and change the perspective.

Another solution would be to add two combo boxes, to select the row dimension and the column dimension. This is probably more intuitive, better for self documenting the interface, but would take some space in the widget (the first solution would just cause the sliders to be a bit shorter).

PiRK:
Add two comboxes to select the array dimensions corresponding to the table rows and the table columns. This allows changing the perspective directly from the gui.

screenshot from 2016-11-29 11 03 37

Closes #432

### DIFF
--- a/doc/source/Tutorials/array_widget.rst
+++ b/doc/source/Tutorials/array_widget.rst
@@ -10,6 +10,7 @@ ArrayTableWidget
 ----------
 
 Let's look at a simple usage example::
+
     >>> from silx.gui import qt
     >>> from silx.gui.widgets.ArrayTableWidget import ArrayTableWidget
     >>> import numpy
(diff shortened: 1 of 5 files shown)


PiRK:
Only the last 5 commits are relevant to this PR.

PR #445, opened 2016-11-29
Improve tests

vallsv:
- Display test name before testing (with -v)
- Update CI to use -v instead of -m
- Try to fix random issue on tests of hdf5tree by using mkstemp instead of NamedTemporaryFile + closing h5file from the model


PiRK:
Looks good.

PR #449, opened 2016-11-30
Create silx.io.open

vallsv:


PiRK:
This seems to conform with what we discussed during the silx meeting. I suggest we merge the PR and let history judge the consequences.

dnaudet:
Hi!
What about file access mode? And how about making a context manager?
(I don't know the if this is supposed to be an internal function or not so ignore my questions if they're out of scope)

t20100:
+1 for with statement support.

I guess there is no access mode because for the previous load function naming, you'd expect read mode only..

vallsv:
Access mode and context manager are not part of this patch. We can open a bug for the context manager, i will take care of it for fabioh5.the mode IS supposed to be read-only (but no idea if there is different read mode on hdf5)

PR #453, opened 2016-12-01
[Plot] pixel intensities histogram action

payno:


t20100:
For PlotWindow.__init__, it would not add an extra argument as there is already a lot of args. It can be enabled after initialization. It should be disabled by default for backward compatibility.
An alternative is to add the histogram action only to Plot2D where it is meaningful.

The current API of PlotWindow/Plot2D for accessing actions as attributes is not Qt-style, I would like to use getters in the future, so you can wrap the access to your action in a getIntensityHistogramAction()

Also, I would put the PixelIntensitiesHistoAction in PlotActions.

payno:
For this plot I need to get the information when the window is closing. So actually I simple define a new Plot1D and redefine the closeEvent. It do the work but it's not very elegant.
Do you have other ideas ? 
Can we imagine to have the PlotWindow always emitting a signal when closing ?

payno:
The icon should probably be redone. If you have any ideas let me know

Issue #391, opened 2016-11-04
[plot] color line


payno:
When executing the following code the curve is displayed in dark instead of pink or whatever any color.
But the color setting will work for any other curve after the first one.
 
`p=Plot1D()`
`p.addCurve(x, y, color='pink', legend='sin', linewidth=6.0)`
`#p.addCurve(x, y, color='red', legend='sin')`
`p.show()`

payno:
Not a bug, due to line selection

t20100:
You can disable curve selection on the widget to avoid this.

vasole:
As alternative, you can also decide which is your active curve color.

t20100:
This is not the first time someone gets disturb by this highlight and think that the provided color is not taken into account.

A few propositions:
- Disable active curve handling by default. This makes the behavior backward compatible by just enabling the active curve handling when creating the widget and is very simple to implement.
- Do not highlight any curve until one is selected by the user or through the API.

dnaudet:
Just for info, I was about to open the same issue yesterday! If we keep things this way maybe there should be at least an entry about it in the FAQ.

dnaudet:
Argh. I just spent the last couple hours trying to figure out why my plot was all black even though I used a colormap. :rage1: :volcano:
An entry in the FAQ won't be enough. :laughing:

vasole:
The problem is that the specialized widgets for 1D and 2D need to have "active" handling enabled by default because if not they do not know where to apply any of the functionalities: fit, ROI, profile, mask, ...

To disable it it by default can give more problems than just a "esthetic" problem. Probably @t20100 suggestion (Do not highlight any curve until one is selected by the user or through the API) is an appropriate compromise.

dnaudet:
I agree!
Thanks.

t20100:
Yes current behavior is miss-leading, the highlight only on user interaction would make it clear it is the highlight.

vasole:
@PiRK 

- Fit action will have to prompt a message asking the user to select a curve if there are more than one curve in the plot and none is selected as active. If there is only one curve, it can take it even if it is not active.

- The 1D ROI widget should not complain (just fill the Raw and Net fields with question marks or something similar if there is no curve active)

maurov:
Hello, my two cents comments on this:

- I also would like that the highlight would happen only when the user interact with the curve or if setted via the API, not by default.
[...]

PR #455, opened 2016-12-02
Minor refactoring of FitAction

PiRK:
Handling of multiple curves, none activated: if there is only one, take it.
Else, prompt the user to select a curve.

See #391 (in the future, no curve will be activated until one is explicitely clicked or programmaticaly activated)

### DIFF
--- a/silx/gui/plot/CurvesROIWidget.py
+++ b/silx/gui/plot/CurvesROIWidget.py
@@ -49,7 +49,7 @@ class CurvesROIWidget(qt.QWidget):
     """Widget displaying a table of ROI information.
 
     :param parent: See :class:`QWidget`
-    :paran str name: The title of this widget
+    :param str name: The title of this widget
     """
 
     sigROIWidgetSignal = qt.Signal(object)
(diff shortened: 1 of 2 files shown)


PiRK:
I'll put the status back to work in progress, because with @t20100  we think the `_getOneCurve(mode=...)` function could be implemented directly in the `getActiveCurve` method in the plot.

The default behavior would remain unchanged, but users would be able to pass a parameter `mode=…` to `getActiveCurve` to modify the return value when no curve is active.. Possible modes would be:

 - "none" (default): default behavior, return None when no curve is active
 - "unique": when no curve is active but there is only one curve in the plot, return the unique curve. If multilple curves are present and none is active, return "None".
 - "first": when no curve is active, return `getAllCurves()[0]`
 - "last": when no curve is active, return `getAllCurves()[-1]`

To be discussed in a meeting.

t20100:
If you want you can make a separate issue/PR to combine the `_getOneCurve` function with `Plot.getActiveCurve`, so possible discussions on this does not stop this PR.

payno:
Looks good to me

PR #458, opened 2016-12-05
Use spacer instead of widget

vallsv:
Hi Pierre,

Let me know if it is working as you expect. I only try the Array, not the Frame

### DIFF
--- a/silx/gui/widgets/ArrayTableWidget.py
+++ b/silx/gui/widgets/ArrayTableWidget.py
@@ -40,15 +40,7 @@ from .FrameBrowser import HorizontalSliderWithBrowser
 
 __authors__ = ["V.A. Sole", "P. Knobel"]
 __license__ = "MIT"
-__date__ = "25/11/2016"
-
-
-class HorizontalSpacer(qt.QWidget):
-    """Blank widget expanding horizontally"""
-    def __init__(self, parent=None):
-        qt.QWidget.__init__(self, parent)
-        self.setSizePolicy(qt.QSizePolicy(qt.QSizePolicy.Expanding,
-                                          qt.QSizePolicy.Fixed))
+__date__ = "05/12/2016"
 
 
 class AxesSelector(qt.QWidget):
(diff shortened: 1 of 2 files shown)


vasole:
Behavior to be tested. I recall to use a QSpacerItem not being exactly the same as the widget with the expanding size policy. 

If I remember well, the use of the spacer item does not prevent other widgets in the layout from growing while the used approach it does it.

vasole:
You can check the behavior running the following script (move the widget prior to resize it to see the two of them):

```
import PyQt4.Qt as qt

class HorizontalSpacer(qt.QWidget):
    def __init__(self, *args):
        qt.QWidget.__init__(self, *args)
        self.setSizePolicy(qt.QSizePolicy(qt.QSizePolicy.Expanding,
                                          qt.QSizePolicy.Fixed))

class Container(qt.QWidget):
    def __init__(self, parent=None, flag=None):
        qt.QWidget.__init__(self, parent)
        self.mainLayout = qt.QHBoxLayout(self)
        label = qt.QLabel(self)
        label.setText("This is a label")
        entry = qt.QLineEdit(self)
        entry.setText("This is a label")
        button = qt.QPushButton(self)
        button.setText("OK")
        self.mainLayout.addWidget(label)
        self.mainLayout.addWidget(entry)
        self.mainLayout.addWidget(button)
        if flag:
            self.mainLayout.addStretch(1)
        else:
            self.mainLayout.addWidget(HorizontalSpacer(self))

app = qt.QApplication([])
w1 = Container(None, 0)
w1.setWindowTitle("Widget")
w2 = Container(None, 1)
w1.setWindowTitle("Spacer")
w1.show()
w2.show()
app.lastWindowClosed.connect(app.quit)
app.exec_()
```

PiRK:
Hi Valentin,

Your suggestion seems to work. I can't find a difference of behavior between the stretch solution or the spacer solution. 

I made a small change in a PR to your branch, because I think it is better to have the slider take all the stretch and all other widgets to be fixed in size.

Issue #460, opened 2016-12-06
Warning in ArrayTableModel


vallsv:
```WARNING:silx.gui.widgets.ArrayTableModel:Frame axes cannot be changed for 1D and 2D arrays```

It is displayed when i use 2 time ``setArrayData`` with the same 2d array.

It is be generated by the model
```
        if n_dimensions < 3:
            _logger.warning(
                    "Frame axes cannot be changed for 1D and 2D arrays")
            return
```
When ``ArrayTableWidget.setNDimensions`` is called.

PR #461, opened 2016-12-06
[ArrayTableView] background and text colors

### LINKED ISSUE #382: [ArrayTableWidget] add colormap for cell background
It would be nice to have an option to add a color to the cells' background in the table view of ArrayWidget, based on the data value. (see #368 )

PiRK:
This PR adds a method to set the background and foreground colors of each cell by passing an array of RGB colors.

Closes #382


PiRK:
WIP: #460

PR #462, opened 2016-12-07
Data Viewer

vallsv:
This PR feature a set of widgets which allow to display any kind of data, but it is mostly design to display `h5py.Dataset` and `numpy.array`.

It contains 4 new widgets:
- `DataViewer` which allow to display any kind of data (the user dont have the choice on the view)
- `DataViewerFrame` which allow to display any kind of data (the user have the choice on the view)
- `NumpyAxisSelector` which allow a user to select axis from a `numpy.array`. A sliced/transposed array is provided in output.
- `DataViewerSelector` which allow to change the view of a `DataViewer` (that's an internal widget of for the `DataViewerFrame`)

It was designed to be able to add another views without too much problem (for 3d and image stack for example, BTW it already contains icons `view-3d.svg` and `view-2d-stack.svg`).

You can test it with the hdf5 tree like that:
```
./bootstrap.py python examples/hdf5widget.py ../data/hdf5/alltypes.h5
```
For Spec files, i send a small pull request to fix spech5 when a dataset do not contains a numpy.array.

dataviewerframe

I think the widget is enough good to be merged. If you have issues with features or the look and feel, i would prefer you to create a new issue. Cause I want to use it to help me for my coming work.


t20100:
A first few comments from a quick look at it:

- There is already a cube icon which is almost the same as the view-3d icon. I think we should avoid duplicated icons. You can update replace the cube icon (and its derivatives) with the view-3d icon.
- We've discussed that, but do we want to move widgets displaying data in a different package from widgets? (Not sure)
- Also to discuss, do we want to 'enforce' prefixing all getters with get for widgets for any new API?

vallsv:
> There is already a cube icon which is almost the same as the view-3d icon. I think we should avoid duplicated icons. You can update replace the cube icon (and its derivatives) with the view-3d icon.

I reuse your cube and your orange style, but optimized for a 16px rendering. Will the aim of the icon is not the same, i think it is better to have 2 different icon name.

> We've discussed that, but do we want to move widgets displaying data in a different package from widgets? (Not sure)

We can talk about it at a meeting. I create the issue https://github.com/silx-kit/silx/issues/475. It is not needed to aswer now, but it would be nice to decide it before 0.4 release.


> Also to discuss, do we want to 'enforce' prefixing all getters with get for widgets for any new API?

We said that it is better if you want to provide a Qt-property, but we do not enforce anything as i remember.

PR #465, opened 2016-12-09
Image stack view widget

PiRK:
StackView widget: browse through stack of images

A stack is a data volume, a 3D numpy array. The slicing plane can be changed. The colormap is applied globally to all the images, and the colormap autoscaling uses the global min and max values of the data volume.

The profile toolbar has an additional button to toggle between 2D and 3D mode (profile on active image displays a 1D curve, profile on volume displays an image).


vasole:
It is a work in progress but it is just to have a reminder. Please keep in mind to allow working with an h5py dataset or a sequence of images. In short, similar to what it has been made for the Table visualization.

vasole:
Given the names, I would prefer to talk about an ImageStack or a Stack rather than a Volume. To me this widget should be able to browse stacks of n dimensional data assuming the last two dimensions are the rows and columns unless specified in other way.

PiRK:
Question: should we enable the mask tool in this widget?

vasole:
Since we can always disable it, I would enable it by default because even if the widget is not connected to anywhere, the user may want to visualize the stack and save a mask to a file.

PiRK:
The widget is mostly ready for review. I still need to investigate a couple of issues:

- [ ] is there a way to cache the 3D profile, to avoid recalculating it when it is not necessary (e.g when browsing through the stack of images while the profile line remains constant and the perspective is not changed)?
- [ ] can we  (should we?) work with a list of 2D numpy arrays without casting it into a 3D numpy array?

vallsv:
Few proposal for array_like module:

- It would be nice to use `six` instead of an own definition of `string_types`.
- It would be nice to provide the `view()`
- I think it is better to remove the commented code. If you dont want to lose it, maybe you can create a branch containing that feature enabled.
- It can be nice to rename `TranposedDatasetView` into `DatasetView`.

PiRK:
Done, except for adding the `view()` method. As discussed, we don't foresee an immediate need for this, and it would be very limited, as the classes currently don't support changing the dtype without first explicitly casting the data into a numpy array. 
This could be implemented in the future, when needed. But currently the main usage for these `array_like`  classes is to access data as if it was transposed.

vallsv:
stack-proposal

Here is some proposal for the widget look and feel
- It is needed to enable the slice in the full cube by default
- It would be better to have an icon with both state: 1d and 2d
- I think it is better to remove the dock for the axes selection
- It would be better to display icons to help to select the axes of the slicing (it should help to understand to use) (it can be done radio buttons, but maybe checked buttons would be better (like buttons to select vline, hline... for the slicing))

Here is the zip for the svg [stack-proposal.svg.zip](https://github.com/silx-kit/silx/files/684682/stack-proposal.svg.zip)

vallsv:
Nice. Thanks for the fixes.

PR #466, opened 2016-12-12
Fix data contiguity check in SIFT

pierrepaleo:
This PR implements an extra check when passing non-contiguous numpy.ndarray to a SIFT plan.


vallsv:
Looks good

PR #472, opened 2016-12-15
Support ellipsis to reach scalar from spech5 datasets

vallsv:
This PR allow to use `[...]` or `[()]` to reach a scalar value, while `.value` is deprecated.

I am using `[...]` for scalar in `h5py.Dataset`, and the h5py documentation contains:

> To retrieve the contents of a scalar dataset, you can use the same syntax as in NumPy: result = dset[()]. In other words, index into the dataset using an empty tuple.`

Here is the result:
```
In [1]: import silx.io

In [2]: a = silx.io.open("../data/SpecFile/nofhdr.dat")

In [3]: a["1.1/title"][()]
Out[3]: '1  ascan  ss1vo -4.55687 -0.556875  40 0.2'

In [4]: a["1.1/title"][...]
Out[4]: '1  ascan  ss1vo -4.55687 -0.556875  40 0.2'
```

### DIFF
--- a/silx/io/spech5.py
+++ b/silx/io/spech5.py
@@ -181,7 +181,7 @@ from .specfile import SpecFile
 
 __authors__ = ["P. Knobel", "D. Naudet"]
 __license__ = "MIT"
-__date__ = "15/12/2016"
+__date__ = "19/12/2016"
 
 logging.basicConfig()
 logger1 = logging.getLogger(__name__)
@@ -731,7 +731,9 @@ class SpecH5Dataset(object):
 
     def __getitem__(self, item):
         if not isinstance(self.value, numpy.ndarray):
-            if item == Ellipsis or item == tuple():
+            if item == Ellipsis:
+                return numpy.array(self.value)
+            elif item == tuple():
                 return self.value
             else:
                 raise ValueError("Scalar can only be reached with an ellipsis or an empty tuple")
--- a/silx/io/spech5.py
+++ b/silx/io/spech5.py
@@ -181,7 +181,7 @@ from .specfile import SpecFile
 
 __authors__ = ["P. Knobel", "D. Naudet"]
 __license__ = "MIT"
-__date__ = "03/10/2016"
+__date__ = "15/12/2016"
 
 logging.basicConfig()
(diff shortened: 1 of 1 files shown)


vallsv:
HI. Do you try it using h5py or spech5?
Cause we have to mimick the h5py API. Then i do some tests and on my side it is a little different than what i expect (maybe the version of h5py? mine is 2.6.0):

```
In [3]: import h5py
In [4]: import numpy

In [5]: a = h5py.File("aaa", "w")
In [7]: a["data"] = numpy.string_("abcdef")

In [8]: a["data"][1]

ValueError: Illegal slicing argument for scalar dataspace

In [9]: a["data"][0]

ValueError: Illegal slicing argument for scalar dataspace

In [10]: a["data"][:]

ValueError: Illegal slicing argument for scalar dataspace

In [13]: r = a["data"][()]
In [14]: r, type(r)
Out[14]: ('abcdef', numpy.string_)

In [15]: r = a["data"][...]
In [16]: r, type(r)
Out[16]: 
(array('abcdef', 
       dtype='|S6'), numpy.ndarray)
```

Then only `...` and `()` are allowed, but `...` returns an array while `()` returns the numpy data. The result with a float is the same.

vallsv:
I apply your patch.

If i take take to refactor a common dataset from spech5 and fabioh5, i will add some tests to check the H5py dataset API for this cases, in order to check that the h5py API is stable.

vallsv:
> There is already a cube icon which is almost the same as the view-3d icon. I think we should avoid duplicated icons. You can update replace the cube icon (and its derivatives) with the view-3d icon.

I reuse your cube and your orange style, but optimized for a 16px rendering. Will the aim of the icon is not the same, i think it is better to have 2 different icon name.

> We've discussed that, but do we want to move widgets displaying data in a different package from widgets? (Not sure)

We can talk about it at a meeting. I create the issue https://github.com/silx-kit/silx/issues/475. It is not needed to aswer now, but it would be nice to decide it before 0.4 release.


> Also to discuss, do we want to 'enforce' prefixing all getters with get for widgets for any new API?

We said that it is better if you want to provide a Qt-property, but we do not enforce anything as i remember.

PR #474, opened 2016-12-16
[Doc] Documentation for ROI

### LINKED ISSUE #423: [plot] ROI widget is missing user documentation
At least a tooltip, better some inline help... And a page in the sphinx doc, for how to use it.

payno:
closes #423


vallsv:
Hi. I fix some typo.

But i think it would be good to:
- upload `netCounts.svg` and `rawCounts.svg`
- remove the line in the bottom of the curve of this images

I start to fix it by myself, but the SVG is not there, then it is not very easy to edit.

vallsv:
Thanks a lot. That's fine for me.

vasole:
- I prefer "from a .ini file" instead of "from an ini file" (I agree, it is just a matter of taste).
- It would be convenient to provide figures based on our plot widget. For instance:
```
import numpy
from silx.sx import plot
x = numpy.arange(1000.)
peak = 5000. * numpy.exp(-0.5 * (x-500.) * (x - 500)/300)
background = 10 * x + 4. 
y = peak + background
plt = plot(x, y)
handle = plt.getWidgetHandle()
handle.ax.fill_between(x, y, background, where=(450 < x)  & (x  < 550), facecolor="red") 
plt.show()
```

Issue #146, opened 2016-07-04
[Plot] Add colormap object


vasole:
We could create a colormap object:
- It could still be accessed as a dictionary 
- It would be simpler to define colormaps
- Easy to manipulate

vallsv:
After a talk with Thomas, and following the refactoring he did on plot items, it would be nice to:
- Refactoring the plot3d colormap as a non mutable plot object (should be easy)
- Then
    - provide setters and events
    - connecting it to the item objects
    - updating the GUI manipulating colormap

payno:
After a discussion with @t20100 and @vallsv, the Colormap object should be included in the silx.gui.plot and used by plot and plot3d modules. It should also :

- not contained the 'autoscale' variable should be replaced by None values for vmin and vmax.
- None is a valid value for vmin and vmax
- Colormap object should have a changed signal
- Should have a copy function
- should have a __getitem__ function
- should have a fromDict and getDict function
- should have a getColorMapRange function (returning the vmin, vmax tuple )
- should have a getColorMapLut fonction returning the list of colors for this colormap
- should have an applyColormap(data, vmin, vmax) function
- should have a copy function

t20100:
Closed by #898

PR #480, opened 2017-01-05
[plot3d] Add silx.gui.plot3d package

t20100:
This PR imports plot3d project as silx.gui.plot3d.
It provides a ScalarFieldView widget based on a Plot3DWindow main window (which central widget is  a Plot3DWidget). This widgets allows to display isosurfaces and a cutting plane on a 3D volume of scalars. The bounding box of the volume has some ticks and labels.
It also provides a Tree widget: SFViewParamTree which allows to control the display of isos

It is based on OpenGL2.1 (and extensions that are core in OpenGL 3).
So it adds a dependency to PyOpenGL (and an OpenGL driver...).

Internal packages are organised this way:
- glutils: Loads PyOpenGL and provide classes to handle OpenGL resources
- scene: Provides a hierarchical scene structure handling rendering and interaction. It includes its own event system (for rendering dirty flags for now), transformation matrix classes with caching, and a node structure to organise the OpenGL rendering.
- utils: Miscellaneous: tick axes layout and video saving in MNG format encoding
Those packages are not directly exposed from the widget level API, and the internal API are most probably subject to change with future developments.

The widgets level API (ScalarFieldView) should hopefully be relatively stable.
For the cutting plane, the silx.gui.plot3d provides a colormap class... This class should be the same as for the plot (see #146), so it should be adapted.

Finally, compared to silx standard, it currently has barely no tests.

TODOs:
- For this PR:
  - [x] Documentation
  - [x] Sample code
  - [x] Manual tests: on Mac, Windows, python 3, PySide, PyQt5
- What I would like to leave for other PRs:
  - True test suite
  - Continuous integration


t20100:
Added some documentation and sample code.

By the way, I forgot credits and thanks to @dnaudet for his contribution to this package and his feedbacks!

t20100:
A few updates:

- ScalarFieldView added to dataviewer thanks to @vallsv 
- Added methods resetZoom + clearIsosurfaces to ScalarFieldView
- Fix video save in MNG

payno:
No issue with code or user API for me. Just one remark which might be important (that we saw with @vallsv) :

    It's looks like the PlaneOrientationItem is not taking into account the name of the axes setted by the user
    in PlaneOrientationItem you have : 
    "
        _PLANE_ACTIONS = (
            ('3d-plane-normal-x', 'YZ-Plane',
             'Set plane perpendicular to X axis', (1., 0., 0.)),
            ('3d-plane-normal-y', 'XZ-Plane',
             'Set plane perpendicular to Y axis', (0., 1., 0.)),
            ('3d-plane-normal-z', 'XY-Plane',
             'Set plane perpendicular to Z axis', (0., 0., 1.)),
        )
    "
    
    And in the example (examples/viewer3DVolume) if you set different axis names like: "window.setAxesLabels('Y', 'X', 'Z')" you can end up with some misleading information.

Some cosmetic optimisation which would be nice from my point of view but no urgency: 

   - update camera view position icon : it would be nice to get something more "main stream" with an eye or a camera and the field of view
   - viewer3d example (also from @vallsv ) : 
        - adding and deleting an isosurface (+ - buttons) would be better slightly move to the right (under level and color items) from @vallsv
        - double click on the isosurface level : 
                + reset it to 0
                + launch probably a subroutine and don't wait for validation
                + don't reset the QSlider value of the iso level
        - transparency : It would be nice to an information about the setted value (and min and max)

To end some more general question about the structure :

    - Isosurface might be in an independent file ? 
    - would  it be interesting to create a geometry module (under plot3d) (including for example transform, scane.utils, primitives...)
    - some functions might also be needed somewhere else (like angleBetweenVectors, segmentPlaneIntersect, Plane ...) and generate an higher module (under silx )

t20100:
Plane names and parameter tree remarks were addressed.

Manually tested, it worked on:
- Debian (7 and 8) with python2.7/PyQt4 and PySide and python3.4/PyQt4, PyQt5.
- Windows with python2.7/PyQt4.8 and python3.5/PyQt5.6

There is some issues on Mac OS X python3.5/PyQt5.7.
Not tested (yet) on Mac OS X python2.7/PyQt4.
I can fix Mac OS X support in another PR if you like.

t20100:
Fixed MacOSX issue
Added support for Qt5 high resolution display

payno:
Nice feature.

PR #486, opened 2017-01-06
Hdf5widget example

vasole:
With #485 one can see the images but due to the scaling one does not see differences when seeing the images browsing dimension 0 or dimension 1 (the default for the hypercube).

This pull request makes sure there are changes browsing on any dimension by taking a pattern that is not a multiple or divider of any of the individual dimensions.


vallsv:
Looks good to me

PR #488, opened 2017-01-06
Use RecordTableView as view of DataViewer

vallsv:
RecordTableView allow to display
- Structured numpy arrays (https://docs.scipy.org/doc/numpy/user/basics.rec.html) which are also provided by HDF5
- And 1D numpy array as a vertical table

The file `compound_array.h5` from my `tmp 14 days` provides some examples of records


PiRK:
I have a problem accessing the data (filesystem not correctly mounted), at the moment. The code looks good.

Issue #495, opened 2017-01-09
[DatasetView] transpose method does not work as expected on already transposed DatasetView instance


PiRK:
In the current implementation, `DatasetView.transpose(t)` does not take into account whether or not the data was already transposed. Instead of transposing the already transposed view, it transposes the underlying original data with a new transposition tuple `t`.

This examples illustrates the difference in behavior between a regular numpy array and a DatasetView.
```
>>> from silx.utils.array_like import DatasetView
>>> import numpy
>>> 
>>> a = numpy.arange(4).reshape((2, 2))
>>> a
array([[0, 1],
       [2, 3]])
>>> a.T
array([[0, 2],
       [1, 3]])
>>> a.T.T
array([[0, 1],
       [2, 3]])
>>> numpy.array_equal(a.T.T, a)
True
>>> 
>>> 
>>> b = DatasetView(a)
>>> numpy.array_equal(b, a)
True
>>> numpy.array(b.transpose((1, 0)))
array([[0, 2],
       [1, 3]])
>>> numpy.array(b.transpose((1, 0)).transpose((1, 0)))   # bad result
array([[0, 2],
       [1, 3]])
>>> numpy.array_equal(b.transpose((1, 0)).transpose((1, 0)), 
                      a.transpose((1, 0)).transpose((1, 0)))    # bad result
False
```

PR #496, opened 2017-01-09
[array_like] fix double transposition issue

PiRK:
Fix a problem when transposing a view that was already transposed (#495). Add tests for this issue.
Implement `.T` as an alias for `.transpose()`.


vallsv:
Looks fine

PR #498, opened 2017-01-10
[StackView] Make getStack return same object set in setStack

PiRK:
Make getStack return the original data, no matter what the current view is.

Add getCurrentView() method to: return the view currently displayed in the widget.


vallsv:
Looks fine

Issue #417, opened 2016-11-15
[plot2d] Add a method to get profile data from the API


t20100:

t20100:
Proposal: make profileWindow accessible from the Plot2D so user can retrieve the curve from there

PiRK:
closed by #505

PR #505, opened 2017-01-11
[Plot] PlotWindow API

### LINKED ISSUE #501: [plot] Make PlotWindow API more Qt-like
PlotWindow actions and dock widgets are accessed through attributes and properties (when lazy-loading is necessary) which is not Qt-like.
To make it Qt-like this should be accessed through getter methods.

To maintain backward compatibility, we just need to keep the attributes and properties (and remove their documentation).

PiRK:
Closes #501 (make plot API more Qt-like with getters for actions and dock widgets)


PiRK:
Addresses #417 as well: make profileWindow accessible from the Plot2D so user can retrieve the curve from there

PR #509, opened 2017-01-12
[spech5] implement .compression and .chunks

PiRK:
This improves the compatibility of SpecH5Dataset with application designed to handle h5py datasets.


PiRK:
Are there any other obvious h5py dataset attributes to be implemented, while we are at it?

Issue #394, opened 2016-11-07
[imshow] provide a histogram


kif:
like we have profiles ... print the histogram of pixels intensities

payno:
done

PR #522, opened 2017-01-16
Intensity histogram action enhancement

payno:


payno:
I still have an issue when guessing the number of bins. If it is a B&W image, he will try to fit with a number of bins of 516 and many interval will be empty. 
This gave a bad looking output

outputhisto

payno:
To solve the number of bins in a B&W image: adding a (bad) hack. Checking if type is integer and if minVal==0 and maxVal==255

payno:
For now it is missing : 
   - should we lock the possibility to change the line type from the button ''change curve line and marker style"
   - error is not displayed
   - some optimization by removing a fiew loops

vallsv:
Looks fine.

payno:
closes #394

PR #524, opened 2017-01-16
[fitmanager] fit actual function, not sum of base functions

### LINKED ISSUE #523: [fitmanager] fitmanager assumes that fitfunction can be decomposed into sum of base functions
A `FitTheory` defines a number of parameter names. When the estimation function returns a number of estimated parameters that is larger than the number of defined parameter names, `FitManager` assumes that that the function to be fitted is a sum of N times the fit function (N being the number of estimated parameters divided by the number of base parameters for this theory).

This worked for all theories defined up to now, because the functions in `silx.math.fit.functions` that accept a variable number of parameters are defined as a sum of a base functions:

 `sum_gauss(x, h0, c0, fwhm0,  h1, c1, fwhm1) = sum_gauss(x, h0, c0, fwhm0) + sum_gauss(x, h1, c1, fwhm1)`

A good example for which this assumption breaks is a fit theory that defines a polynomial whose order is defined by the number of parameters.
- `poly(x, a, b, c) = a*x**2 + b*x + c` ->order=2 
- `poly(x, a, b, c, d) = a*x**3 + b*x**2 + c*x + d` -> order=3 

`poly(x, a, b, c) ≠  poly(x, a) +  poly(x, b) + poly(x, c)`

PiRK:
Close #523 

I still need to add a unittest for this.

### DIFF
--- a/silx/math/fit/fitmanager.py
+++ b/silx/math/fit/fitmanager.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*#########################################################################
 #
-# Copyright (c) 2004-2016 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2017 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


PiRK:
I will do the test in a separate PR when I add the polynomial theory.

PiRK:
commits included in PR #528

PR #526, opened 2017-01-16
[plot] Add more formats to plot SaveAction

### LINKED ISSUE #437: RGB image export from Plot2D with original resolution
On the one hand the ability of silx/Plot2D to directly export greyscale tiffs from the Plot2D window is great for further data processing and on the other the export of png and jpg snapshots from the current graph is handy for electronic logbooks etc. 

It would be great to also have the possibility to export a false color image in the resolution of the original data without any axes or labels (just using the colormap as used in the plot window applied to the data as exported in the tiff/edf export). This would be great to have for images that one wants to use in publications and would give the option to later on add separate overlay layers to the exported image as the experimental resolution of the raster-data is usually known.

### LINKED ISSUE #503: [plot] Add save as svg to plot

t20100:
In the SaveAction:

- Add save of plot as SVG, closes #503
- Add save of active image with colormap applied as RGB in PNG and TIFF, closes #437

This PR also contains some refactoring:
- New module: silx.gui._utils with QImage<->numpy.array
- Moved colormap functions from matplotlib backend to silx.gui.Colors

### DIFF
--- a/silx/gui/_utils.py
+++ b/silx/gui/_utils.py
@@ -35,6 +35,7 @@ __license__ = "MIT"
 __date__ = "16/01/2017"
 
 
+import sys
 import numpy
 
 from . import qt
--- a/silx/gui/_utils.py
+++ b/silx/gui/_utils.py
@@ -1,0 +1,99 @@
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2017 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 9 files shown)


t20100:
One remark: When saving active image as RGB, the data array is converted to an RGB image, but neither negative image scale nor Y axis orientation is taken into account.
So the saved image can have axes flipped from what is seen is the plot (as it is already the case when saving image data to file from the plot).

t20100:
Ready for review

PiRK:
LGTM

PR #527, opened 2017-01-16
Update sx package

### LINKED ISSUE #426: [sx] sx does not load gui when in a notebook
When sx is imported in a notebook it does not load gui stuff.
Do we want to change that?

In this case, ideally it would need to check if it is run locally or remotely...

### LINKED ISSUE #482: Test sx module under different condition before next release
PR #481 has changed the sx module.
It needs to be checked again with different ipython/pyqt configurations..

### LINKED ISSUE #511: [sx plot] color keyword ignored
Matplotlib plot command supports color keyword but silx.sx.plot ignores it.

t20100:
- Add color, linestyle, linewidth and marker keywords to sx.plot, closes #511 
- Add a warning when sx is imported in a notebook: silx.gui not loaded, closes #426 
- Test sx with different (I)python with and without notebook, closes #482

### DIFF
--- a/silx/sx/__init__.py
+++ b/silx/sx/__init__.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2017 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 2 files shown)


t20100:
Add a fix of silx.io imports: sx was importing deprecated load function

t20100:
Ready for review

PiRK:
LGTM

Issue #339, opened 2016-10-07
[fit] Add higher degree polynomials 


vasole:
Higher degree polynomials should be added to the list of functions and/or backgrounds. Ideally one should be able to just configure the degree of the polynomial. If that is not possible, a very simple solution would be to offer up to 5th order polynomials.

PiRK:
closed by #528

PR #528, opened 2017-01-16
[FitWidget] Add polynomial function (and background)

PiRK:
This PR implements polynomial fit theories for FitWidget. It also improves display of fitted curve : delete curve when fit fails, instead of leaving the old curve, same for estimated curve.


PiRK:
Closes #339

payno:
under review

vasole:
The configuration of the background takes the whole data range and not just the portion selected prior to the fit.

PR #530, opened 2017-01-16
[data] New package silx.gui.data

### LINKED ISSUE #475: Create a silx.gui.data
Maybe we have enough widget about "data" in `silx.gui.widgets`. And i guess we will have more widget about that.

In order to have a stable API, it would be nice to decide to create `silx.gui.data` before the release of silx-0.4. silx-0.3 does not provides this kind of widget anyway.

PiRK:
This PR moves ArrayWidget, DataViewer (and all associated modules) and NumpyAxesSelector from silx.gui.widgets into silx.gui.data.

Closes #475

I'm keeping it as work in progress because I have not been able to run any tests (issue with OpenGl).


PiRK:
The CI tests seem to stall randomly. `run_tests` ends succesfuly, then nothing happens for an hour.
I can eventually get all green lights by re-running the tests repeatedly until they all pass, but it's probably not very wise to do so without finding the problem. 

But I have no clue what can cause this, because I have mostly just been moving modules from one package to another, and added documentation, without touching the actual code nor changing the tests.

vallsv:
It looks like to be the same issue that we had few weeks ago. Then it is still here...

vallsv:
It is maybe import-dependant (order of the imports), but i see nothing strange on the logs. It always fail, anyway the number of time you restart it?

PiRK:
Yesterday I was able to make it succeed after restarting the build a few times. 

You can see the build for commit https://github.com/silx-kit/silx/pull/530/commits/67f6a1e7e1996908a76906eafab071f62fd21669 succeeded, then the build for the next commit https://github.com/silx-kit/silx/pull/530/commits/0c1277dce10c3ccdf5cb9f50d3f37331e3e3e823 failed. There was no change in import order and no change in source code in this commit.

vallsv:
Looks fine

PR #535, opened 2017-01-17
[DatasetView] Fix python 3 incompatibility, add test

PiRK:
Don't use `sys.maxint`, as it is not available in python 3.


vallsv:
Looks fine.

PR #542, opened 2017-01-18
[plot3d] Handle missing PyOpenGL or QtOpenGL

### LINKED ISSUE #532: [plot3d] Add message when QtOpenGL is not installed
On debian QtOpenGL is installed separately. Add an error message in plot3d package to warn if it is missing

### LINKED ISSUE #540: DataViewer: Fix availability of Plot3d
OpenGL can be available and QGLWidget can be missing. t20100 will fix plot3d module import. Maybe there is fix to do in the DataViewer, and DataViewer tests.

t20100:
This PR adds a function to check if plot3d can run (i.e., PyOpenGL and QtOpenGL are available).
This function is used by plot3d tests and DataViewer.

Closes #532, closes #540

### DIFF
--- a/doc/source/modules/gui/plot3d/index.rst
+++ b/doc/source/modules/gui/plot3d/index.rst
@@ -51,6 +51,8 @@ Widgets gallery
 Public modules
 --------------
 
+The following sub-modules are available:
+
 .. toctree::
    :maxdepth: 2
 
@@ -60,6 +62,10 @@ Public modules
    sfviewparamtree.rst
    toolbars.rst
    actions.rst
+
+The :mod:`silx.gui.plot3d` package provides the following function:
+
+.. autofunction:: isAvailable
 
 Sample code
 -----------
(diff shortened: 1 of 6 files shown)


t20100:
Changed: plot3d raise ImportError if PyOpenGL or QtOpenGL is not available.

Ready for review

vallsv:
I dont try it, but for tests without opengl

Instead of
```
test_plot3d_suite = SkipPlot3DTest
```

it have to be something like that no?
```
def test_plot3d_suite():
    suite = unittest.TestSuite()
    suite.addTest(SkipPlot3DTest())
    return suite
```
Or it is anyway working?

t20100:
It's working like that: suite.addTest take either a TestSuite or a TestCase:
https://docs.python.org/2/library/unittest.html#unittest.TestSuite.addTest

vallsv:
Then it's fine

PR #570, opened 2017-01-23
Support h5node to dataviewer

vallsv:
It is useful to display extra informations from HDF5Tree

Basically local name (and not the name using path from external h5 file)


payno:
lgtm

Issue #562, opened 2017-01-23
should six.io.uilts.h5ls take as constructor argument silx.io.fabioh5.File


CaptainNemoz:
this will insure compatibility with the silx toolkit.
Might also take an edf file has input

vasole:
Closed by #576

PR #576, opened 2017-01-23
[h5ls] Support spech5 and fabioh5

PiRK:
Addresses #562. The basic feature is easily achieved by using silx.io.open.

The display format of datasets is still inconsistent between HDF5 datasets and SpecH5Dataset objects:   `str(h5f[key])` returns a string representation of the complete data array for SpecH5Dataset. FabioH5 objects are still  to be tested.


vallsv:
That's fine

Issue #564, opened 2017-01-23
StackViewMainWindow in silx.gui.plot.StackView: unselect profile


srio:
after working with profile tool, it remains active even when it is unselected, so the profile window appears when passing from one slice to another

payno:
on it

payno:
If the profile window is closed then the profile lines are cleared but the selection mode is not changing

vasole:
@PiRK Check if solved by #577

PiRK:
Closed by #577

PR #577, opened 2017-01-23
Stackview behavior improvement

### LINKED ISSUE #560: [StackView] delete profile when perspective is changed
When the perspective changes, the previous profile line is not deleted (even when it is out of the range of the new X&Y axes).

payno:
fixes #560 and #564


PiRK:
LGTM

Issue #554, opened 2017-01-23
[gui] Bug with python 3 on Mac OS X


t20100:
In qt/__init__.py:  import _macosx -> from . import _macosx

PR #579, opened 2017-01-23
[plot] Fix ImageView image scale bug

### LINKED ISSUE #557: [plot] ImageView histogram histogram does not always have the right bounds
With some data sets, the horizontal histogram is not displayed with the right bounds…. E.g. not aligned with the image or not fully visible.

t20100:
Fix a bug in the scaling of ImageView horizontal histogram when scales are set.
closes #557


t20100:
Added another single liner fix to close #554.

Ready for review.

vallsv:
Looks fine

PR #581, opened 2017-01-24
Support msk as input mask

vallsv:
Here is the implementation of 

- I rework the load function to use more exception instead of returning messages
   - Let me know what you think, i am not very satisfy
- And i implement the msk loading using fabio as an optional dependency


vallsv:
Save as Fit2D msk file is added as suggested by t20100.

PR #583, opened 2017-01-24
Improve text formatter

### LINKED ISSUE #565: Data format in ArrayTableWidget
Hi guys,

It would be nice that the data format could be changed in ArrayTableWidget -- for example, interactively choosing the number of displayed digits of float numbers.

vallsv:
- Clean up get/set format to use only TextFormatter
- TextFormatter provides signal on format changed
- Update models when the format changed
- Plus few unittest for TextFormatter

It should help to fix #565. There is way way to provide a common widget to custom different widgets at a same time.

### DIFF
--- a/doc/source/modules/gui/data/index.rst
+++ b/doc/source/modules/gui/data/index.rst
@@ -2,7 +2,7 @@
 .. currentmodule:: silx.gui
 
 :mod:`data`: Widgets for data visualization
---------------------------------------------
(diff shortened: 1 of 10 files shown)


PiRK:
Looks good, except that the  TextFormatter module does not have a RST doc page.

vallsv:
Rst pages was updated as requested.
Thanks for the feed back.

PR #587, opened 2017-01-24
[plot] Fix zoom interaction

### LINKED ISSUE #430: [plot] zoom area not available when starting on a curve
When one starts a zoom area on a selectable curve, the zoom interaction is not available because the curve selection takes the priority.
This is disturbing and not necessary as the curve selection only needs the click and the zoom can coexists with curve (or image) selection.
Images can also be dragged. In that case this interaction cannot coexist with the zoom area selection.
Furthermore, curves are selectable by default.

t20100:
This PR adds the possibility to start drawing the zoom area on a selectable element of the plot (like a curve) closes #430

### DIFF
--- a/silx/gui/plot/PlotInteraction.py
+++ b/silx/gui/plot/PlotInteraction.py
@@ -1,7 +1,7 @@
 #  coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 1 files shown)


t20100:
Ready for review

payno:
Works for me

PR #591, opened 2017-01-26
[plot] resetZoom with data margins and keep aspect ratio on

### LINKED ISSUE #452: [plot] resetZoom with keep aspect ration does not keep data margins
It seems data margins are not enforced when aspect ratio is respected.
Margins are applied first and then aspect ratio is kept which can lead to margins smaller than set...
The margins should be made larger than requested to keep aspect ratio, not smaller.

t20100:
This PR fixes an issue with Plot.resetZoom method with data margins and keep aspect ratio on: Closes #452

### DIFF
--- a/silx/gui/plot/BackendBase.py
+++ b/silx/gui/plot/BackendBase.py
@@ -473,7 +473,7 @@ class BackendBase(object):
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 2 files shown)


vallsv:
LGTM

PR #593, opened 2017-01-26
[plot3d] Fix color dialog issue on MacOS

### LINKED ISSUE #555: [plot3d] Issue with color picker in parameter tree on MacOS
On MacOS El Captain with python 3.4, PyQt4.8.7, there is a issue with the color picker: changing one color in the param tree changes other colors… Cancel and OK buttons are added to the dialog each time  the dialog is opened.

t20100:
This PR solves issues that might occur on MacOS with the plot3d parameter tree widget:
- Size of icons too large on retina display
- Buggy color dialogs

closes #555


vallsv:
LGTM

PR #595, opened 2017-01-27
[PeriodicTable] mostly documentation

### LINKED ISSUE #558: [PeriodicTable] improve documentation
- currently there is no documentation on how to construct a custom list of elements and pass them to the widgets' constructor
- doc of getSelection is inconsistent (return type: list of symbols or list of items?)
- inconsistency: setSelection() takes a list of symbols, getSelection() return a list of objects (PeriodicTableItem). Maybe make getSelection() return list of symbols and add getSelectedItems() for items. Not sure we can change this, check legacy compatibility issues with PyMca widget

PiRK:
Add example in documentation to explain adding of custom data to elements.

Make `bgcolor` attribute a simple RGB string rather than a `QColor`

Closes #558


vallsv:
LGTM

PR #596, opened 2017-01-27
Improve DataViewer

vallsv:
Few other fixes
- Support CTRL-C where it is useful
- Fix and improve display with dataset using external compression plugins
- Do not propose Plot3D when 3d data have a dimension =1


PiRK:
LGTM

Issue #575, opened 2017-01-23
[ci] Update CI systems setup to install dependencies


t20100:
Check that dependencies (like fabio) are installed during continuous integration so that as much test as possible are run.

t20100:
Status after PR #607 (which adds fabio) gets merged:
- On appveyor (Windows): pyopencl, Mako and PyOpenGL (as well as drivers/mesa impementation) are not installed and related tests are not run.
- On travis (Linux and MacOS): All dependencies are installed and all tests are run (except GUI tests on Linux with python 2.7 and 3.4 which are missing a PyQt wheel to be enabled).

vasole:
As far as I know PyOpenGL is a source package not requiring compilation.

What is the problem of having it for AppVeyor?

t20100:
No problem... but it will also need OpenGL drivers or mesa lib.
I'll add it when I actually have some tests using OpenGL.

PR #602, opened 2017-02-01
[CI] Add python3.6 to appveyor

t20100:
This adds python3.6 as an environment to test on Windows.
Still need to check the availability of wheels for project dependencies

### DIFF
--- a/ci/appveyor.yml
+++ b/ci/appveyor.yml
@@ -25,6 +25,10 @@ environment:
         VENV_TEST_DIR: "venv_test"
 
     matrix:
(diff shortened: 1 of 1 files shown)


vasole:
In terms of wheels I was able to get (I did not try more):

- numpy 1.12
- matplotlib 2.0
- PyQt4
- PyOpenGL
- h5py

That already tests the most commonly used parts of silx.

t20100:
Now should work. This doesn't change anything but adding python3.6.
In particular, it does not update installation of dependencies like fabio, see #575.

Issue #601, opened 2017-02-01
Install error in virtualenv/python3.4 on debian8 using pip


vincefn:
Installing silx in a brand-new virtualenv on debian8 with pip3.4 fails:
############
**favre@gpu2-1304:~$ export BASEDIR=~favre**
**favre@gpu2-1304:~$ cd $BASEDIR**
**favre@gpu2-1304:~$ mkdir -p test-env**
**favre@gpu2-1304:~$ virtualenv -p python3.4 test-env/**
Running virtualenv with interpreter /usr/bin/python3.4
Using base prefix '/usr'
New python executable in test-env/bin/python3.4
Also creating executable in test-env/bin/python
Installing setuptools, pip...done.

**favre@gpu2-1304:~$ source test-env/bin/activate**

**(test-env)favre@gpu2-1304:~$ pip3.4 install --upgrade pip**

Downloading/unpacking pip from https://pypi.python.org/packages/b6/ac/7015eb97dc749283ffdec1c3a88ddb8ae03b8fad0f0e611408f196358da3/pip-9.0.1-py2.py3-none-any.whl#md5=297dbd16ef53bcef0447d245815f5144
  Downloading pip-9.0.1-py2.py3-none-any.whl (1.3MB): 1.3MB downloaded
Installing collected packages: pip
  Found existing installation: pip 1.5.6
    Uninstalling pip:
      Successfully uninstalled pip
Successfully installed pip
Cleaning up...

**(test-env)favre@gpu2-1304:~$ pip3.4 install numpy**

Collecting numpy
  Using cached numpy-1.12.0-cp34-cp34m-manylinux1_x86_64.whl
Installing collected packages: numpy
Successfully installed numpy-1.12.0

**(test-env)favre@gpu2-1304:~$ pip3.4 install silx**

Collecting silx
  Using cached silx-0.4.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-gw4145o3/silx/setup.py", line 505, in <module>
        long_description=get_readme(),
      File "/tmp/pip-build-gw4145o3/silx/setup.py", line 86, in get_readme
        long_description = fp.read()
      File "/mntdirect/_users/favre/test-env/lib/python3.4/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
[...]

PR #603, opened 2017-02-01
Double fix README encoding issue

PiRK:
Remove non-breakable space from README.rst, specify encoding=utf-8 in setup.py when reading README.RST
Addresses #601


vallsv:
Looks fine

PR #604, opened 2017-02-02
Median filter

payno:
This is a first version of the median filter plot action.

Basically the source code is from PymCA and we linked this with a PlotAction.
I added this on the 'image' module under a filter module. This module is needed only if in a close future other filters will be integrated. Is it the case ?
Median filter and median tools came from an (old) version of scipy.
Q: It might be interesting to change this ?
@kif told me he has some opencl code to manage at least the sorting (from PyFAI).

For now performances when having a large 'filter width' are slowing down the GUI so
it might be interesting to create a thread to launch the filtering. But it might be overkilled if we have a faster implementation.
This is needed by @CaptainNemoz. Do you have an idea about the kind of image size /filter width will be needed ?
Maybe someone else need this ? @PiRK  and @vasole for next pymca ? What are your constrain ?


vasole:
If you do not need the conditional version, probably you can manage with just numpy.median.

For plotting purposes one usually does not need a wide filter and the one from PyMca or from numpy are appropriate. If it is for something more serious, then an optimized version with OpenCL would be desirable.

If we go that way, it would be nice to be able to choose one or the other just by a keyword/flag.

If we have both implementations available (in PyMca, in numpy and in pyFAI) then to add it to silx it should not be a big effort.

In terms of needs of PyMca, the currently available version seems to be good enough.

CaptainNemoz:
From what I tested during the code camp with some ID19 data, the 
needed filter size was 15. I think that in the case of imaging with 
detectors we have at ESRF, this might be a maximum.

  Christian

kif:
On Thu, 02 Feb 2017 01:42:04 -0800
CaptainNemoz <notifications@github.com> wrote:

> 
>   From what I tested during the code camp with some ID19 data, the 
> needed filter size was 15. I think that in the case of imaging with 
> detectors we have at ESRF, this might be a maximum.

For info a 4kx4k image with a median filter of 15x15=225, if taken in
float64 takes ~30GB of RAM (which my computer does not have).

A more realistic case: 2048x2048 filter over a 15x15 allocates 2 GB in uint16 
the median filter, as implemented on numpy takes then ~10s:

In [11]: import numpy
In [12]: r=numpy.random.randint(0,65000,size=2048*2048*225).astype("uint16")
In [13]: r.shape=2048,2048,225
In [14]: r.nbytes
Out[14]: 1887436800
In [16]: %time s=numpy.median(r,axis=-1)
CPU times: user 8.92 s, sys: 392 ms, total: 9.32 s
Wall time: 9.31 s

vasole:
@kif 

Are we talking about the same type of use of the median filter?
The implementation we have needs the size of the input array plus that of the kernel (negligible) instead of the the size of the input array multiplied by the size of the kernel.

CaptainNemoz:
Exact, I didn't understood the Jerome's statement. For me the median 
filter size ks is a ksxks array convoluted with the image.
Christian

kif:
On Thu, 02 Feb 2017 06:06:17 -0800
V. Armando Solé <notifications@github.com> wrote:

> Are we talking about the same type of use of the median filter?
> The implementation we have needs the size of the input array plus that of the kernel (negligible) instead of the the size of the input array multiplied by the size of the kernel.
> 

The numpy implementation probably needs to "densify" the array before
the calculating the sort... this is one issue, the second is the actual
calculation time using this method.

Cheers,

kif:
On Thu, 02 Feb 2017 06:16:29 -0800
CaptainNemoz <notifications@github.com> wrote:

> Exact, I didn't understood the Jerome's statement. For me the median 
> filter size ks is a ksxks array convoluted with the image.

Actually the issue comes from the fact the median filter is non linear filter, not a
convolution.

dnaudet:
Just FYI (for comparison purposes) I wrote a cython medfilt2D in xsocs because the one in scipy was randomly crashing on me.

payno:
Behavior : Scipy is having a different management of the pixel close to boundaries. It is not computing the pixel median or creating a mirroring of the pixels values at boundaries. Actually I don't see the point of such a process and I simply reduce the size of the window. 
Is it an issue for some of you ? 

Question : I creating a new file QMedianFilterDialog under silx.gui.widget only containing this dialog. I was considering calling it DialogXXX to include other stuff, it might be done in the future except if you already have stuff to be included in.

Gain of performances vs scipy are probably due to there mirroring behavior or a bug because performance gain is quit big from my point of view, 
It might due to a bug.
I will add some unit tests.

perfmedianfilter

vasole:
Reducing the size or doing nothing seem to me the most appropriate behaviors. There is yet other alternative that is to repeat the value at the boundary to fill the median window. I suppose in most cases this later approach will be almost equivalent to doing nothing.

vasole:
Just for the record. The C++ standard library provides std::nth_element. Probably the original algorithm could have been speed up by using std::nth_element instead of quickselect in the case of the non-conditional median filter implementation.

kif:
the scipy.ndimage.filters' median_filter offers all find of mode for traiting the boundary conditions.

payno:
Update on perf vs scipy with the few modifications
perfmedianfilter

PR #615, opened 2017-02-07
[plot3d] Add nearest/linear interpolation option to cut plane + fixes

### LINKED ISSUE #561: [plot3d] cutting plane nearest/interpolation option
Add an option to toggle interpolation on/off for the cutting plane

### LINKED ISSUE #572: [plot3d] plot dragging scaling and centering in 3d
This _may_ be specific to MacOS or a trackpad.
When zoomed in on a data view, left-click and dragging is not properly scaled - the data (or the plane) moves too slowly.
Also, when rotating with the mouse, the center of rotation is always in the center of the data. It would be nice if this was also where the click was made, as for the zoom-with-the-mouse.

### LINKED ISSUE #612: [plot3d] ParamTree : setLevel with null data range.
The exception raised when calling SFViewParamTree::setLevel when the data min and max are equal (i.e : range == 0) is misleading. It's actually the exception raise by Qt : it complains about QSlider::setValue getting a double instead of an int (it might just be just because that particular double is NaN in that case, but we're probably better off casting it to int anyway).

t20100:
This PR updates ScalarFieldView and associated param tree widgets:
- Fix setLevel when min == max, closes #612 
- Fix pan of object/cut plane when zoom is closed + use picked position as center for rotation interaction when over an object, closes #572 
- Add an option to change cut plane interpolation between nearest and linear, closes #561

### DIFF
--- a/examples/viewer3DVolume.py
+++ b/examples/viewer3DVolume.py
@@ -86,7 +86,7 @@ def load(filename):
 
             data = numpy.array(data, order='C', dtype='float32')
 
-    elif filename.lower().endswith('.npy'):
+    elif magic[:6] == b'\x93NUMPY':
         data = numpy.load(filename)
 
     else:
--- a/examples/viewer3DVolume.py
+++ b/examples/viewer3DVolume.py
@@ -65,7 +65,9 @@ def load(filename):
     if not os.path.isfile(filename.split('::')[0]):
         raise IOError('No input file: %s' % filename)
 
-    if '.h5' in filename.lower():
+    with open(filename.split('::')[0], 'rb') as f:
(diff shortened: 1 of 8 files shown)


dnaudet:
Tested fixes for #612, #572 and #561 on Linux. All is fine.

t20100:
Still need to avoid texture glitches with nearest interpolation.

t20100:
Fixed glitches due to numerical instability with nearest interpolation.

Also added a few updates:
- viewer3DVolume.py checks file type from magic header rather than file extension.
- Fixed parameter tree isolevel slider range not updated when loading new data with a different range.
- Add display of visible back-faces of surfaces.

PR #617, opened 2017-02-07
Add png still image as a fallback for process-working icon 

### LINKED ISSUE #616: [icons] process-working icon is not available as png
I encountered this issue with an installation where gif and mng image format plugins are not available:
```
silx/gui/icons.py, line 217, in getQFile
    raise ValueError('Not an icon name: %s' % name)
<type 'exceptions.ValueError'> Not an icon name: process-working
```
We should add a png (frozen image) just in case.

t20100:
closes #616


vallsv:
Looks fine

Issue #247, opened 2016-09-21
[Plot] Return objects instead lists for Plot getCurve|Image methods


t20100:
Methods to get curves and images returns a list with data and dicts.
The returned data is the 'internal copy' used by the plot.

Proposition: Return an object with accessors (e.g., getX(copy=True) for curve) that returns by default a copy of the data and a **getitem** for backward compatibility with list.
This would also document the available information.

vasole:
Agreed in today's meeting we can go ahead.

t20100:
Closed by PR #621.

PR #621, opened 2017-02-14
[Plot] Object-oriented API

t20100:
Proposition for a major refactoring of Plot to provide object-oriented API (See #247)... yet backward compatible.

Idea: For this PR, make a first step to introduce objects for plot primitives (curves, etc..) with only getters as public API, no setters and no signal of modifications (but prepare for that).
I'd leave colormap as objects and event as objects for another time.

Plot primitive classes are defined in PlotItems.py

This is a proposition, do not hesitate to discuss the API....

Turn plot primitives into objects:
- [x] curves as object
- [x] images as object
- [x] markers as object
- [x] items as object
- [x] use object-oriented API all over silx
- [x] update documentation


vasole:
I would aim for something like a generic object that simplifies handling of some of the foreseen features (support of NXdata, calibrations, ...). To keep compatibility with the existing API it does not seem to be difficult since basically all what one needs is to implement ``__getitem__`` for the object and current addWhatever methods to become shortcuts to create the object in-the-fly.

t20100:
As discussed we propose to keep deprecation warnings for now and remove them for next release so we can identify usage of __getitem__ in silx.
We'll have to open an issue for that when merging this PR.

t20100:
This PR turns Plot get* results into objects with getters with the same name as arguments of Plot.add* methods (and same 'objects').

So all items (hline, polygon etc) are represented by a single class (where we could have a class for each primitive), and histograms are curve objects (where a special histogram object would allow to bin edges and avoid changing histogram style when changing curve style). Those possible changes are left for other PR.

It aims at being backward compatible with previous API (though I would expect some regressions as there's quite a bit of refactoring there...).

Plot primitives object only have public getters. This is a first step towards using objects with also setters and events.
The update of primitives is still based on recalling the corresponding add* method which looks weird when using objects.

vallsv:
BTW fix for profile issue is now merge. Commits 735fe42 and e610845 can be removed.

vallsv:
Few comments

- I would remove the "get" from base objects ```data()``` instead of ```getData()``` while it is not widgets, we can't have conflict with properties. But it is a matter of taste.
- Your ```Base``` object should not have link to plot.
- If you want to have Qt signals on ```Base``` items, it should inherit from QObject, and provide a parent owner in the constructor. On PyQtGraph that's the case for most of objects (everything inherite from ```QGraphicsObject```). Then i guess it would be nice.
- ```QGraphicsItem``` have a ```ZValue```. It would be better to use ```setZValue``` instead of ```setZLayer```.
- In StackView, do you really need to create params from image? I guess you did it to avoid warnings? But it would be better to generate it from the Image class it-self, that's already done there.

vasole:
I agree with most of @vallsv's remarks.

Perhaps I am old fashioned, but I think setters and getters simplify making derived classes. If I remember well, the original idea was to implement getters, getters and, where considered relevant, add the property (for instance, one could have setData, getData with the property data if relevant).

Concerning zlayer or zvalue, my only comment is that we should avoid inventing a new keyword. Zvalue looks appealing to me if we can use floats and there is no conflict with visualization of 3D data. If it cannot be used as a value, then matplotlib uses zorder:
 
http://matplotlib.org/examples/pylab_examples/zorder_demo.html

PiRK:
> I would remove the "get" from base objects data() instead of getData() while it is not widgets, we can't have conflict with properties. But it is a matter of taste.

I have a small preference for keeping the "get" prefix. I'm always surprised when I need to add parenthesis after something that looks like a property or an attribute. But as you say, it's just a matter of taste and convention, so if I need to get used to `data()`' so be it.

t20100:
- I would prefer to keep the get prefix: the whole Plot API is done this way, so it remains consistent with it (even if it is not consistent with Qt).
- For StackView returned values, I opened issue #627. This PR is already large, and that can be done afterwards. To me the current implementation is only temporary.

t20100:
- Renamed ZLayer to ZValue
- Renamed base class from PlotItems.Base to PlotItems.Item and change `__init__` signature (no arguments)

As there is no signals yet, there is no need to inherit from QObject (and we can mimic signals without QObject inheritance).
I would keep the `getPlot()` method rather than `parent()`, because this method is expected to return a `Plot` instance. This also keeps the `parent` name for future use if we want to add `Group` objects if we need.

vallsv:
There is still problem with the merge of the histogram change. Sorry for that...
```
  File "C:\projects\silx\venv_test\lib\site-packages\silx\gui\plot\Plot.py", line 621, in addCurve
    if params['xerror'] is not None or params['yerror'] is not None:
NameError: name 'params' is not defined
```

vallsv:
LGTM, let's merge it tomorrow if nobody complains.

PR #622, opened 2017-02-15
[FitAction] remove fitted curve prior to fitting the same data again

### LINKED ISSUE #619: [FitAction] remove previous fit result from plot before starting new fit
Comment by @CaptainNemoz : while calculating a new fit, the old fit result is still displayed on the plot, so the user might believe the fit is finished and didn't produce any visible change. 

The fitted curve should be removed from the plot prior to starting the new fitting. If possible, a wait cursor and/or a wait button should tell the user that the fit is being computed.

PiRK:
Close #619


payno:
LGTM

PR #629, opened 2017-02-16
Improve plot histogram

vallsv:
- Avoid to display warning message if it is not expected
- Fix axes labels
- Fix window histogram parent
- Fix histogram x-normalization according to colormap of the source image


payno:
LGTM

PR #638, opened 2017-02-24
[DataViewer] Add NXdata viewer

PiRK:
This PR adds NXdata views for the  `silx.gui.data.DataViewer` widget. The *@interpretation* is taken into account to determine the dimensionality of the plot.

Generic NXdata parsing functions are added to a new `silx.io.nxdata` module.


vasole:
Please take into account that the interpretation attribute is not always present. If it is present, it is of course better to consider it.

vasole:
- hdf5widget.py option  create NXdata does not work under Python 3.5 or Python 3.6

vallsv:
I just check the overview and it looks fine.

Here is few comments:

- `view-nxdata` images can be removed while it is not used?
- Should we merge it with `dataviews.py`
- It would be better to import widgets only when it is needed (as it is done on `DataViews.py`). Then only when the expected dataview needs `ArrayCurvePlot`, `ArrayImagePlot`, `ArrayStackPlot`, `ArrayTableView`. I would like to avoid to load too much things at the startup of an application. If we don't need OpenGL and Matplotlib to display something, it's better to not load it.

I ask myself for this other comments:

- Mostly typos:
    - `NxDataView` instead of `NXdataview`
    - `isNxData` instead of `isNXData`
    - `signal_is_2D` instead of `signal_is_2d`
    - `nx_` instead of `NX_` in functional API

PiRK:
OK for most comments. 

Not sure about CamelCase conventions when it comes to concepts such as *NXdata* which use a mix of upper case and lower case in their names. Maybe I should use pure lower case *nxdata* whenever possible, and stick to *NXdata* in CamelCase contexts such as class names or QWidget methods.

I also agree that the functional API in `silx.io.nxdata` should disappear, as it is 100% duplicating the class methods. I started with only functions, then it became obvious that I needed a class to avoid the clumsy `@validate_NXdata` decorator, and it was a mistake to then keep using the functions in NXdataViews.

More issues to be fixed:
 - until we have an addScatter method in plot, I need to compute my own RGBA array for addCurve, but I should at least avoid importing matplotlib
 - calibration objects are to be discussed in a silx meeting

vasole:
Concerning CamelCase conventions. Two capitals are kept as capitals. More than that follow CamelCase.

Examples:

HtmlIO
PyMca

Therefore the correct terms if you consider NX and data as separated things would be NXData and NXDataViewer

If you consider NXdata as the item on its own and not a combination of two things, then NXdataViewer is the correct thing since it is a viewer of NXdata and not viewer of Data for NeXus (NXDataViewer).

I would propose to follow NXdataViewer approach (a Viewer of NXdata)

vallsv:
It is usually not the point of capitals, but the point of acronyms. And in this case `NXdata` is not an acronym. I would like to avoid this kind of stylistic class name: `FReLoN`, and you can see that there is no more 2 capitals together. But it's fine for me while we are consistent with ourself.

I still ask for the functional API with underscore. Any capital hurting me: `is_valid_NDdata`.

PiRK:
Ready for review (again). For the moment I refrain from documenting `silx.math.calibration`. There is an open issue #695 with milestone 0.5 for discussing the API of the calibration objects.

vallsv:
Looks fine.

PR #641, opened 2017-02-28
Fix bug in Hdf5TableModel callback

PiRK:
key in lambda function was fixed to its value in the last iteration of the "for" loop, when getting HDF5 attrs. This caused the same attribute value to be displayed for all attributes in the HDF5  DataViewer.


vallsv:
LGTM

PR #642, opened 2017-03-01
Improve deprecated decorator

### LINKED ISSUE #636: @deprecated: add optional arguments (reason, replacement, deprecation date)
It would be convenient if we could add more information when deprecating a method.

There are ways to create decorators that take optional arguments and use introspection to decide if they should behave as regular `@decorator` or as a `@decorator_generator()`:
http://yati-sagade.github.io/2012/10/02/decorators-optional-args/

PiRK:
Close #636


vallsv:
LGTM

Issue #650, opened 2017-03-06
[mask] Mask does not remain when changing image


t20100:
The mask is discarded if the image is replaced by one of the same size while it can stay there.
As a consequence, the mask does not work well with the StackView as changing the image reset the mask.

PR #651, opened 2017-03-07
replace=False preserve mask in StackView when updating frame number

alemirone:
useful for exploring 3D stacks


t20100:
This will close issue #650

PiRK:
Looks good to me.

PR #652, opened 2017-03-07
[plot] OOplot add setters to Plot items objects

t20100:
This PR follows PR #621 and adds setters to most of the attribute of Plot item objects.
It provides:
- Splitting of `PlotItems` module into an `items` package.
- Setters to Plot items classes.
- Moving rendering of primitives from `Plot` module to item's classes.
- Automatic synchronization of the Plot (with asynchronous replot) when item objects are updated. 
- Update of the doc accordingly.
- Update Plot backend to remove active color stored in backend (which is not necessary).
- Fix a typo in median filter doc.

It does **NOT** provide:
- Signals upon modifications
- Update of silx code using addCurve|Image... where it can directly modify the Plot item objects.

As PR #621, it aims at being backward compatible with previous API (though I would expect some regressions as there's also quite a bit of refactoring there...) and this is a proposition, so do not hesitate to discuss the API....


t20100:
Update add* methods to avoid unnecessary copies (see #663)

t20100:
Added a copy argument to addImage and addCurve so caller can choose to copy or not provided data.
Ready for review.

This follows Plot 
Some improvements (not for this PR):
- add Histogram and a Scatter classes
- split the Shape class into Polygon, Rectangle, Polyline etc...

vallsv:
Let's merge it.
Ill open an issue for the getPixmap API.

PR #662, opened 2017-03-11
[Plot] OO API. Missing compatibility __len__ internal method.

vasole:
Missing method for true compatibility.

### DIFF
--- a/silx/gui/plot/PlotItems.py
+++ b/silx/gui/plot/PlotItems.py
@@ -438,6 +438,10 @@ class Curve(Item, LabelsMixIn, SymbolMixIn, ColorMixIn, YAxisMixIn, FillMixIn):
         else:
             raise IndexError("Index out of range: %s", str(item))
 
+    def __len__(self):
+        """Compatibility with PyMca and silx <= 0.4.0"""
+        return 5
+
     @staticmethod
     def _logFilterError(value, error):
         """Filter/convert error values if they go <= 0.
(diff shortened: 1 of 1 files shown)


vasole:
@t20100
@PiRK 

Triggered during tests with PyMca. 
pymcasilx

t20100:
It sounds weird to have a curve having a length of 5, if it has a length, I would expect it to be the number of points.

In the code raising the issue, `len(oldStuff)` is useless, the comparison to `None` is enough:
```
oldStuff = self.getCurve(legend)
if oldStuff is not None and len(oldStuff):
    oldX, oldY, oldLegend, oldInfo, oldParams = oldStuff
```
https://github.com/PiRK/pymca/blob/c605f6c52f4ccebec0250604bd56070d99e0e569/PyMca5/PyMcaGui/silxplot/PyMcaScanWindow.py#L319

vasole:
If one aims for full compatibility with the old API, where a tuple was expected, the __len__ should be there.

vallsv:
Should we also inherite graphics object from tuple in case people was using `instanceof` on the result?
It will make use really crazy if we can't break anything from a 0.x version of the lib.

t20100:
It is not possible to return objects with *full* compatibility with previously returned `list` type.
`__getitem__` should provide compatibility in most cases.
If you want *full* compatibility, the best is to inherit from the `Plot` and override `getCurve` and similar methods so that they really return a list as before.

vasole:
The simplest way up to now (0.4.0) to check if one was under silx or under PyMca was to check for the length of the returned list. That is not asking for full compatibility but minimal one. If for something we were supposed to retrieve as a list, where the only difference with the previous API was the length, we cannot get its length ....

vasole:
On 27/03/2017 09:43, Valentin Valls wrote:
>
> Should we also inherite graphics object from tuple in case people was
> using |instanceof| on the result?
> It will make use really crazy if we can't break anything from a 0.x
> version of the lib.
>

It is not that we break something from version 0.x of the lib. It is
that we already broke something since the beginning (the length of the
returned list) and now simply there is not length at all when it cost
peanuts to have it.

Issue #261, opened 2016-09-23
[Plot] Plot 2D: is it possible to add a colorbar (optional)?


lucarebuffi:

t20100:
Yes, and that's definitely missing!
I started to wrap matplotlib colorbar so it synchronizes with a PlotWindow (and so it follows the 'active' image's colormap).

vasole:
Closed by #664

PR #664, opened 2017-03-13
New colorbar

payno:
- [x] First draw of the widget. Only getter

- [x] deal with log 

- [x] deal with signals of the parent 

- [x] Unit tests

A lot of stuff is missing actually but the designed should be close to this:

v2
v1

If something looks wrong for you please let me know.


dnaudet:
Nice!
Would it be possible to add something to hide the controls at the bottom (constructor keyword or something) please?
Thanks!

payno:
No problem

vallsv:
I think it would be better to not provide any of the autoscale or log/linear.

We already have a dialog to edit colormap. Then i really think it is better to provide a readonly widget. Or at worst with a minimalistic button with a drop menu.

Do you have a usecase?

payno:
This is the first version with no 'interaction' with the colormap. It will be added when we will have a colormap object. Will simplify a lot.

There is some question remaining :

1.  I created two configuration, one which display the ticks values and one which doesn't but add min and max labels on top and bottom of the Gradation. For this there a Configuration list. But I think this is probably overkilled. Just one parameter in the constructor 'show min-max' label might be enough (but it will implicitly hide TickBar labels, something I am not a enjoyed about).

2. I had some problems with the display type. So I am using a QFontMetrics to now if I should switch to scientific writing. But this is not very good as well as overkilled. Maybe It can stay like this until we find something better ? Or if you have ideas ?

3. I let the user to select the type of format he want to have for the TickBar (standard, scientific, let the TickBar decide). Maybe this option is not so needed for now ?

vallsv:
Looks nice since there is no border between ticks and color bar!

BTW, do you think it is difficult to move the ticks inside the colorbar and not outside? That's anyway fine like that.

I only have a problem with the `ColorbarWidget` constructor.

- I think it would be better to remove most of the option from the constructors. Choose a default behaviour and that's it.
- You have to provide all this customization as properties. We don't really need to have real Qt properties (it can be useful), but at least getter and setter for:
     - `showNorm(True/False)` (if it is only useful for the editable widget it can be removed)
     - `showAutoscale(True/False)` (same a above)
- `Configuration` is really not a good idea. Remove that from the constructor and create getter/setter/property on your widget:
    - I dont know, you can have a bunch of features like `showTicks(True/False)`, `showTicksValues(True/False)`, `showMinMaxValues(True/False)` if it is not mutual exclusive.
    - If it is mutual exclusive use a mode concert `setDisplayMode(MIN_MAX_VALUE / TICK_VALUES...)`.
- I don't think it is good to have a dependency with a `plot`. But i guess we don't have choice right now.

vallsv:
I try to fix unittests by removing `TestTickBar` but there is still problem with `ValueError: Integers to negative integer powers are not allowed`.

t20100:
This PR will close #261

vasole:
When replacing the power of ten, I think you should make sure your 10 is a float to prevent integer overflow or underflow. For instance, on my machine numpy.power(10, 10) can give me a negative number.

vallsv:
Hi. Iv got again a bunch of comments:
- The `ColormapWidget` have to work when there is no colormap. I send you a PR with a new failing tests (https://github.com/payno/silx/pull/4)
- The same PR also show that some constructed widget are not well released (maybe tests have to be tuned)
- I send you a PR to clean up the legend (https://github.com/payno/silx/pull/3)
- The method `setColormap` must take a single object as parameter `setColormap(colormap)`, else it will be difficult for further uses. Right now it will be the dict you construct inside the function.
- I would rename `setNTicks` to `setTicksNumber`/`setTicksCount` (as you have `setTicksDensity`)
- I would say `ColormapWidget` should provides getter and setter to hide/show min/max ticks and so on, but right now it is fine. We can see if it is useful later.

We can work together tomorrow, if you like.

payno:
thank for the review> I will modify it. 

But for the last point in fact you can but you have to access to the 'sub widgets' to show and hide those kind of thinks. Maybe we can expose the sub widgets function ...

PR #668, opened 2017-03-15
Ocl 2 d median

kif:
Direct port from pyFAI.


t20100:
A question of code organization: `sift` is in the `image` package and `medfilt` is in `opencl`.
Do we want to have all functions depending on opencl in the `opencl` package (i.e., `sift` should be there too) or to put `medfilt` in `image` or `math` package?

Do we want to have a single entry point for median filter either in C++ or opencl?
A related question of API consistency: see #696

A few minor comments:
- In medfilt.py you could use a global variable and a function instead of the `_MedFilt2d` class 
- include for_eclipse.h is missing in the repo, but this does not seems to cause issues.
- silx.opencl.medfilt docstring says its 1d, 2d, 3d median filter, but unless I missed something it is only 2d.

payno:
I think that having a single entry point would be better. And depending if we have opencl or not we can used opencl or C++ code.

vasole:
To me a median filter is a mathematical operation and it should be available from math with a keyword qualifier specifying if the C++ or the OpenCL versions are to be used.

Concerning sift, it is the same. Registration is not restricted to images. It should be available from math with keywords specifying if using FFT or sift. If math wants to perform registration of a 2D array using sift or image wants to perform registration using sift, they just have to provide the means to access the sift implementation under opencl.

I am also in favor of having all OpenCL dependent code under opencl, but that does not exclude to access it from other places. For instance, I find logical that the median filter OpenCL implementation is located under opencl while having it accessible from math on request.

payno:
I think you have an issue in silx/opencl/common on allocate_cl_buffers function (~l500) if your device is None then you can call ocl.device_from_context(context) with context==None when it should be a list

kif:
replaced by #799

PR #675, opened 2017-03-17
Improve setup.py to be able to use pip install while numpy is not here

vallsv:
I also would like to:
- Move cmdclass declaration to the bottom
- And remove all those "###########" separator,  cause now the file is much more declarative and executive, then it looks already very readable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ import io
 import sys
 import os
 import platform
+import shutil
 import logging
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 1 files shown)


t20100:
Looks good to me.

Sounds good to move cmdclass declaration to a single place and to remove "###########", there is just the doc generation and test part which spreads across multiple functions/classes.

I just added a few missing docstrings.

vallsv:
I reduce uses of global var, clean up a little the sphinx import. But i did not remove the `######`, maybe that is not needed.

PR #685, opened 2017-03-24
Transparency slider for image on a plot

PiRK:
This PR depends on PR #652 (for using  `image.setColormap()`)

It provides slider widgets that can be linked to a specific image on a plot and adjust the alpha channel of its colormap. 
i believe that this could be of general use for *silx*. In PyMca it could be used for plotting detector images on top of a sample picture. 

TODO: 
- discuss API and implementation
- tests
- documentation

Specific questions to be discussed: 
- is there a better way to change the transparency of any colormap than the way I use in `BaseImageAlphaSlider._valueChanged`? I try to get or to compute a RGBA colormap vector from the actual colormap used for the image, I modify it and then use the modified cmap for the image.
- If there is no easier way, should I move the colormap vector computation code to `silx.gui.plot.Colors`?


vasole:
In principle the silx API is able to provide the pixmap associated to the image. Therefore you just need to update its alpha channel and there is no need to calculate any new pixmap.

PiRK:
@t20100 : is this a correct way of doing the alpha channel modification on the pixmap?

```
img = self.getImage()
imgdata = img.getData(copy=False)
pixmap = img.getPixmap(copy=False)
pixmap[:, :, 3] = alpha_value
img.setData(imgdata, pixmap)
```

Is it mandatory to pass again the unmodified data to `img.setData()`, or could I just pass the new pixmap as a single parameter?

vasole:
@t20100 

The problem is that the functional API (as well as the OO API) accept as data a pixmap. However, the method setData of the OO API, when pixmap is not provided, sets self._pixmap to None instead of setting it to the input data if the input data is a pixmap. That behavior prevents the simpler use getPixmap, setPixmap and forces the multiple calls exposed above.

t20100:
As it is `getPixmap` returns the provided pixmap when adding the image, so it can be None if it was not provided.

To me, we should add a separate method to get the image as it is displayed no matter what it is (data, pixmap set or RGB data). I think this is what you need.

The `setData` method requires both data and pixmap to be provided to ensure that data and pixmap are consistent (i.e., they have the same shape). Yet, we can add ways to update just one of them at a time.

To support transparency, I think it is best to add a `get|setAlpha` method to image.
This needs update of the backend, but matplotlib seems to support that.

vasole:
@t20100 

Given the slowness of matplotlib when calculating pixmaps, would not be simpler to have the signatures:

setData(data, pixmap=None)
setPixmap(pixmap, data=None)

That assures you both have to be consistent (although for me it is not a requirement). If one does not have a pixmap, it is to be calculated from the data but if it is provided it is used.

t20100:
`setData(data, pixmap=None)` is already there, and `setPixmap(pixmap)` would allow to update the pixmap.

I would keep `getPixmap` returning the provided pixmap (which can be None) so one can retrieve what he provided and add a new method like `toRgba` or `getRgbaImage` which returns either the provided pixmap or the pixmap calculated by applying the colormap if pixmap is None.

For support of RGBA images only (without data), maybe we can have a separate RgbaImage class... Not sure though.

PiRK:
I've added tests, in my opinion it can be merged (if they pass CI).

I haven't added documentation, to keep some freedom in case we want a potential future refactoring of the plot package (grouping actions, toolbuttons and this kind of specific action buttons together, for example?).

PR #687, opened 2017-03-27
Add min/max combo function

t20100:
This PR provides a function `min_max` that computes min, max, argmin, argmax and optionally strictly positive min and its index in a single pass.

With positive min computation the benchmark I ran gave a 2x to 9x speed-up compared to numpy.
Without positive min computation, the speed-up is no as good. From the benchmark I ran, it is around 2x for 1 million integers and for 1 million floats, the speed-up is between no speed-up (for random data) to about 3x (for ordered data) compared to numpy nanmin/nanmax.

This combo operation is useful for the plot management.


kif:
one could pass the option to cython not to check for array initialized.
One comment opened to discussion:
instead of the numpy.asanyarray, use numpy.ascontiguousarray and then define the cython function's signature: number[::1]. It would make the code faster most of the time  and  lose of performance when the array is in fortan layout.

PR #692, opened 2017-03-28
[plot] Reorganizing plot and plot3d packages

t20100:
This PR is reorganizing plot and plot3d in order to prepare for an OpenGL backend for the plot:
- It moves OpenGL shared modules in silx.gui.plot._glutils.
- It moves Plot backends to silx.gui.plot.backends.
- It makes a silx.gui.plot._utils package and add ticklayout there.


vallsv:
I just have a comment about `from . import _matplotlib` in the `__init__` of backend package.

I think it would be better to avoid to import all the available backends when it is not needed.
Maybe at one point matplotlib will not be needed at all. Then the backend `__init__` should be empty.

What do you think?

t20100:
The issue with matplotlib is that we want to change its settings to use the Qt backend corresponding to what we are using... and as far as I know that can only be done for sure basically when importing matplotlib the first time.

This is why `import _matplotlib` is there to be imported as early as possible.

vallsv:
I was not aware of that. But I think it would be anyway better to load it explicitly on the plot module, and not on the `init` of the `backend` package. I mean, it should not be the responsibility of the `backend` package to load it. But it is up to you.

t20100:
Done

vallsv:
Thanks a lot.
Let's merge it.

PR #693, opened 2017-03-28
[plot] Function to convert data to colors with a colormap

t20100:
This PR adds a function `applyColormapToData` which convert a numpy array to an array of RGBA colors by using a colormap. Data can be of any dimension.


PiRK:
LGTM

PR #697, opened 2017-03-29
[plot] Scatter object

PiRK:
Add a Scatter item


PiRK:
Regarding the Plot interface, I have done the bare minimum: just an `addScatter` method.

I agree with @t20100 that we should avoid adding several methods (getActiveX(), setActiveX(), getAllXs(), getX(), removeX()...) to Plot whenever we define a new item. 

The easy way to avoid this would be to make `_setActiveItem, _getItem, _getItems` public. This methods hade a keyword `kind='image/curve/scatter'`.

PR #698, opened 2017-03-30
Changing API of median filter

### LINKED ISSUE #696: silx.math.medianfilter API using (width, height) rather than (height, width)
`medianfilter` docstring says  2D `kernel_dim` is (kernel_width, kernel_height), while numpy convention is usually (height, width).
This makes it inconsistent with median filter in scipy and opencl median filter in PR #668.

Also if a single value is provided for kernel_dim, it is expand to 2D with the second dimension being 1, while in this case, other median filters are using the given value as the size in each dimension.

To me this is worth an API change.

payno:
- Kernel dims move from (width, height) to (height, width).
- if a single value is given, default behavior is now to run a 2D median filter instead of a 1D, => move (dim, 1) to (dim, dim)

close #696


t20100:
Why not using the signature of scipy `medfilt2d(input, kernel_size=3)` as in PR #668 ?
This would make all API consistent and easier to switch from one implementation to the other.

payno:
True, I will change  it

PR #703, opened 2017-04-05
fabioh5: Fix float precision from headers

### LINKED ISSUE #680: [fabioh5] time_of_day
From EDF file we have to convert `time_of_day`. It looks like a big float: "1469116891.704490".

Right now it uses a float32, but the result looks to be not accurate, while i guess this curve should be more linear.

screenshot from 2017-03-23 09 31 16

vallsv:
Relative to https://github.com/silx-kit/silx/issues/680


PiRK:
LGTM

PR #707, opened 2017-04-06
Make sx.plot style match the longest color name possible

### LINKED ISSUE #706: [sx] plot function not matching the right color name
To handle style, `sx.plot` matches the first color name it founds rather than the longest.
Depending on color dictionnary internal order, passing style `'black'` can match with `'b'` (i.e., blue).

t20100:
Closes #706


vallsv:
Looks fine

PR #708, opened 2017-04-06
[plot] Split Image class

### LINKED ISSUE #686: [plot] Image API must provide a way to access to the displayed pixmap
The plot `Image` item provides a custom `set/getPixmap`.
On the previous API `getPixmap` was providing the displayed image, anyway a pixmap was defined.
It is not the case anymore.

The image item must provide an API to reach the displayed image (the pixmap or the data transformed with the colormap).

t20100:
This PR splits Image class into ImageBase which is inherited by ImageData (2d data + colormap) and ImageRgba (RGB(A) image).

It follows API proposed in issue #686 except for `getGrayscaleImageData` which is not available (there are different ways to compute a grayscale image + this can be added later if needed).
So `Image.getPixmap` is no more available.

Closes #686

### DIFF
--- a/doc/source/modules/gui/plot/items.rst
+++ b/doc/source/modules/gui/plot/items.rst
@@ -15,10 +15,14 @@ Curve
    :members:
    :inherited-members:
 
-Image
------
+Images
+------
 
-.. autoclass:: Image
+.. autoclass:: ImageData
+   :members:
+   :inherited-members:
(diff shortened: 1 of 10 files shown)


t20100:
Ready for review

t20100:
Intergate PR #714 and rebase to tip.

PR #711, opened 2017-04-06
Fix support of RGB image for the mask tool

### LINKED ISSUE #710: [plot] Mask does not support RGB image
Mask fails for RGB images because:

- it expects active image (saved in self._data) to be a 2d array:

```
@@ -876,7 +876,7 @@ class MaskToolsWidget(qt.QWidget):
             self._scale = activeImage.getScale()
             self._z = activeImage.getZValue() + 1
             self._data = activeImage.getData(copy=False)
-            if self._data.shape != self.getSelectionMask(copy=False).shape:
+            if self._data.shape[:2] != self.getSelectionMask(copy=False).shape:
                 # Image has not the same size, remove mask and stop listening
                 if self.plot.getImage(self._maskName):
                     self.plot.remove(self._maskName, kind='image')
@@ -911,8 +911,8 @@ class MaskToolsWidget(qt.QWidget):
             self._scale = activeImage.getScale()
             self._z = activeImage.getZValue() + 1
             self._data = activeImage.getData(copy=False)
-            if self._data.shape != self.getSelectionMask(copy=False).shape:
-                self._mask.reset(self._data.shape)
+            if self._data.shape[:2] != self.getSelectionMask(copy=False).shape:
+                self._mask.reset(self._data.shape[:2])
                 self._mask.commit()
```

- Then it fails when trying to use threshold (which should be disabled in that case as it is meaning less)

t20100:
Closes #710 + fix a typo for RGB image support in OOPlot

### DIFF
--- a/silx/gui/plot/MaskToolsWidget.py
+++ b/silx/gui/plot/MaskToolsWidget.py
@@ -876,7 +876,7 @@ class MaskToolsWidget(qt.QWidget):
             self._scale = activeImage.getScale()
             self._z = activeImage.getZValue() + 1
             self._data = activeImage.getData(copy=False)
-            if self._data.shape != self.getSelectionMask(copy=False).shape:
+            if self._data.shape[:2] != self.getSelectionMask(copy=False).shape:
                 # Image has not the same size, remove mask and stop listening
                 if self.plot.getImage(self._maskName):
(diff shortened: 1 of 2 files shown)


t20100:
Closing as integrated in PR #716

PR #712, opened 2017-04-06
[W.I.P]Start refactoring ProfileToolBar

payno:
Start refactoring of ProfileToolBar. Now one ToolBar for 2D and one for 3D

### DIFF
--- a/silx/gui/plot/ImageView.py
+++ b/silx/gui/plot/ImageView.py
@@ -59,7 +59,7 @@ from . import PlotWindow, PlotWidget
 from . import PlotActions
 from .Colors import cursorColorForColormap
 from .PlotTools import LimitsToolBar
-from .Profile import ProfileToolBar
+from .Profile import Profile2DToolBar
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


payno:
Question shared with @PiRK : This refactoring bring two classes :  Profile2DToolBar and Profile3DToolBar inheriting from ProfileToolBar which became an abstract class. So previous code outside silx should deal with it. Is it bad enough to 
- do not do this refactoring
- Let implicitly ProfileToolBar be 2D abd create a AAbstractToolBar
- something else...

But from what I recall other stuff needed to be refactored due to 3D feature.

payno:
Some detected issues:
    a bottleneck is that we have one plot per dimension. This bring to increase code complexity.
    We actually have a ProfileMainWindow which is used but only for 2D profile.
    A way to simplify would be to always use the same PlotWidget and calling a setDefault1DInterface and a setDefault2DInterface to switch from 1D and 2D. But user wont be able to tune is own PlotWidget for profile.
    We can also always define ProfileMainWindow and a boolean to know if this widget is only used to managed the plot or also for the window show/hide.

payno:
not useful for now

Issue #691, opened 2017-03-28
[mask] Split Mask and MaskToolsWidget in two classes


PiRK:
It would be useful to have a BaseMask object that does not enforce a 2D mask array. We could inherit this to create an Mask2D (or ImageMask or just keep the current generic name Mask) and a Mask1D (or ScatterMask). 

Currently more than half of the Mask object is already generic and would just need a few `assert len(mask.shape) == 2` to be removed to create a BaseMask. 
The methods that need a specific implementation are `save, updateRectangle, updatePolygon, updatePoints, updateStencil, updateDisk, updateLine`.

The only challenge is that a 2D mask only needs to know about the data shape, whereas a ScatterMask will need to know about the actual data coordinates. 

The same applies to `MaskToolsWidget`.

t20100:
Maybe have a look at PR #239 and xsocs: https://gitlab.esrf.fr/kmap/xsocs.

vasole:
Can we not have a mask that adapts to the shape of the underlying data? The mask can be reshaped at will.

t20100:
For images, the mask is reshaped if the image changed.

PiRK:
The issue I see is that for an image the `updateRectangle/Circle`... methods can just return a mask array of the proper shape without much knowledge of the data, because the data "position" is given by its indices in the array. 

For a scatter, we need to look at the actual data (x, y) for each point to figure out if a point is in the rectangle/circle... before we can update the mask.

vasole:
Why can you not use the data "position" as given by its index? 

Mask size equal to number of points

1D data ->  data[i] is associated to mask[i]
2D data -> data[i, j] is associated to mask[i *j + j]

vasole:
Is this to be considered closed by #716?

PiRK:
Yes, I will close it.

PR #716, opened 2017-04-07
Refactoring Mask

PiRK:
The Mask tools need some refactoring to be able to do masks on other type of data (not just images).

Older history is available here: https://github.com/PiRK/scripts/commits/master/ScatterMaskToolsWidget.py

Addresses #691


PiRK:
Closes #710  (https://github.com/silx-kit/silx/pull/716/commits/f2a9636a6052acd48fb51aa521c81635a4517721)

vasole:
An example or the possibility to execute the new modules would simplify code review.

t20100:
I encountered a few issues while playing with it (on macOS, python3.5, PyQt5.8):

With both the image mask and the scatter mask example, there is an issue with the 'between' threshold. Here are the log for the image mask:
```
/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/BaseMask(
   1035                 max_ = float(self.maxLineEdit.text())
   1036                 self._mask.updateBetweenThresholds(self.levelSpinBox.val
-> 1037                                                    min_, max_)
   1038                 self._mask.commit()
   1039 

/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/BaseMask(
    263         :param bool mask: True to mask (default), False to unmask.
    264         """
--> 265         stencil = numpy.logical_and(min_ <= self._getValues(),
    266                                     self.getDataValues() <= max_)
    267         self.updateStencil(level, stencil, mask)

AttributeError: 'ImageMask' object has no attribute '_getValues'
```

With the scatter mask example:
- The button synchronise min/max of the threshold tool with the colormap does not update the min/max values.
- When setting log scale on an axis:
```
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/PlotActions.py", line 265, in _actionTriggered
    self.plot.setXAxisLogarithmic(checked)
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/Plot.py", line 1964, in setXAxisLogarithmic
    self.resetZoom()
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/Plot.py", line 2315, in resetZoom
    self._backend.resetZoom(dataMargins)
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/backends/BackendMatplotlib.py", line 541, in resetZoom
    ranges = self._plot.getDataRange()
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/Plot.py", line 430, in getDataRange
    self._updateDataRange()
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/Plot.py", line 397, in _updateDataRange
    bounds = item.getBounds()
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/items/core.py", line 96, in getBounds
    return self._getBounds()
  File "/Users/tvincent/src/silx/build/lib.macosx-10.6-intel-3.5/silx/gui/plot/items/core.py", line 621, in _getBounds
    displayed=True)
ValueError: too many values to unpack (expected 4)
```
This seems an issue with plot.items Scatter and Point classes conflict of cache attribute.

Otherwise, looks good to me.

PiRK:
I fixed all the problems you mentioned, but I found another one for which I don't see an obvious solution: 
the "active" scatter handling is different from the active image handling.

The first image added is the active one, but there is no active scatter unless we specifically activate one. 
Currently the method `_activeScatterChanged` uses `activeScatter = self.plot.getScatter()` which returns the last scatter added to the plot. This is the scatter used for displaying the mask, not the real scatter data.

PiRK:
We cannot use `_setActiveItem(...)` because we don't want the scatter data to be all black (selection color).

PiRK:
Actually the scatter does not change color when it is selected, so the solution was simple (https://github.com/silx-kit/silx/pull/716/commits/22760fac91e7994f8ec99ba5933e84ba3381d42a)

The PR is ready for review.

There is still an issue of lack of public API to get/set the active scatter, but this can be discussed and implemented in a separate PR, because it is also a problem for histograms.

vallsv:
Tested with pyFAI-drawmask without any problem.

PR #719, opened 2017-04-10
[fabioh5] Support UB matrix

vallsv:
Here is the first implementation.

- A `sample` group is created only if expected and contains few informations from `UB_pos/mne` and `sample_pos/mne`.
- It only reach data from the first header (i do not check hole the frames for efficiency)
- Nexus requests float for this datasets. `numpy.array` used by it-self `float64`. I think it's fine.
- I add `interpretation=scalar` for this 3 datasets. Like this we display it as raw data.
- I use `medipix.edf` to create a unittest. I think it is a very old data, then it would be nice to test the result with an up-to-date file.

screenshot from 2017-04-10 17 49 36

### DIFF
--- a/silx/io/fabioh5.py
+++ b/silx/io/fabioh5.py
@@ -597,8 +597,13 @@ class SampleGroup(LazyLoadableGroup):
             data = self.__fabio_reader.get_unit_cell_alphabetagamma()
             data = Dataset("unit_cell_alphabetagamma", data, attrs=scalar)
             self.add_node(data)
-            data = self.__fabio_reader.get_orientation_matrix()
-            data = Dataset("orientation_matrix", data, attrs=scalar)
+            # According to this issue the UB matrix is not yet available
+            # in the Nexus specification (it only can describe the U matrix)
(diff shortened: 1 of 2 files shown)


vasole:
Please, change the name orientation_matrix to ub. 

People from NeXus have realized they made a mistake and the orientation_matrix is just U and not the product UB.

Sorry for the inconvenience.

vallsv:
There is not problem, it is a small fix.

vallsv:
BTW what about `ub_matrix` instead of just `ub`?

vasole:
I am fine with ub_matrix too. I have given the NeXus people until our release to decide.

vallsv:
This PR can way before merging.

vasole:
It looks OK to me. If the PR does not interfere with any other one, we can validate it when NeXus people have taken a decision on the name (ub or ub_matrix).

vasole:
This needs to be merged in order to simplify testing that spech5 and fabioh5 provide the same information and in the same form. At this moment it does not seem to be the case.

PR #722, opened 2017-04-11
[spech5] Support UB matrix

### LINKED ISSUE #721: [spech5] Expose UB matrix as nexus sample
The SPEC format store UB matrix.

It would be nice to expose it in an NXsample.

BTW, the specification for the UB matrix is not yet defined in Nexus. It should be done soon.

References:

* https://github.com/silx-kit/silx/issues/682 (where Armando talk about this feature form SPEC)
* https://github.com/nexusformat/definitions/issues/559 (issue relative to the missing UB matrix in Nexus)
* http://download.nexusformat.org/doc/html/classes/base_classes/NXsample.html#nxsample (NX specification)

PiRK:
See #719. 
Closes #721

### DIFF
--- a/silx/io/spech5.py
+++ b/silx/io/spech5.py
@@ -1020,14 +1020,14 @@ def _dataset_builder(name, specfileh5, parent_group):
         if not "G3" in scan.scan_header_dict:
             raise KeyError("No UB matrix in a scan without a #G3 header line")
         array_like = numpy.array(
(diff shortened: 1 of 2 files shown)


PiRK:
Btw, this PR could serve as an example to show the (tedious) way of adding a new group and new datasets to spech5, if someone else needs to do this in the future.

vasole:
It looks fine. The pending issue is to know if it will be called ub_matrix or ub but that is other point.

Issue #720, opened 2017-04-11
[plot] Marker (symbol) size


PiRK:
It would be good to have an option to adjust the marker size for points (curves and scatters). This would improve the scatter mask display by allowing to mark the masked points with bigger points.

PR #723, opened 2017-04-11
[plot] symbolsize for scatter and curve

PiRK:
This PR adds the possibility to change the symbol size for scatter and curve items. 

i'm not sure about `_DEFAULT_SYMBOL_SIZE = 1.0`. I wasn't able to find the matplotlib default value for markersize (confusing documentation).


t20100:
Closes #720

PR #726, opened 2017-04-12
[plot] Objecterize events

### LINKED ISSUE #246: [Plot] Let's plot events be objects
Plot events are currently dict.

Proposition: Make them instance of some Event class with attributes or methods to access their content and a **getitem** for backward compatibility with dict events

vallsv:
Convert dictionary-like events from plot to object.

Closes #246

Comment:
- I try to use `qt.Qt.LeftButton` instead of `left`
    - It is half done, cause i am not sure everybody is fine with it. We can talk about that.
- I used `scenePos` (like pyqtgraph), but the code uses `dataPos`, we can talk about that.
- Is `MouseMoved` needed button (Plot.py line 2514). I add a FIXME there.
- `ItemClickedEvent`
    - I merged `curveClicked` and `imageClicked` into a single `ItemClickedEvent` (containing item and data index: `col, row` for images).
    - `ItemClickedEvent` for curves was not providing index (it looks like the full xdata and ydata was sent). I add a FIXME there.
    - I think it would be better to remove any data or data index from events. It's too much specific. Items should provide a getter to return data from coord. We still can provide this info for compatibility, but no more.
- `PlotWidget.notify` emit a `sigPlotSignal`. I dont know if PlotEvent classes have to be emitted. There is a FIXME.
- Draw events
    - I decide to manage it by creating custom draw item + generic set of item event a-la PyQtGraph `RegionChangeStarted`, `RegionChangeFinished`, `RegionChanged`. We can talk about that.
    - There is a bunch on new items inheriting from `Points` (`DrawItem`, `RectangleDrawItem`, `LineDrawItem`).
    - [x] `DrawItem` have to be discussed and reworked.
    - I think `VLineDrawItem` and `HLineDrawItem` are not needed (a generic line is enough)
    - Inheriting from `Points` for `DrawItem` is maybe not a good idea (it have to be managed as `[x1 x2 x3], [y1 y2 y3]` instead of `[x1 y1] [x2 y2] [x3 y3]`). It creates a bigger boilerplate
    - `RegionChangeStarted` was created but it is not used (it's better if it's part of another PR)
    - i remove `parameters` key. It looks to be unused. More data can be anyway stored in the events or in the items if needed.
- Hover events (for markers)
    - Implemented as `ItemHoveredEvent`
    - However i think it would be better to have both `Enter` and `Leave` events
- Marker events
    - `ItemClickedEvent` was used instead of `markerClicked`
        - A strange behaviour here is `x` and `y` (from the dictionary-like event) do not refer to the mouse but to the marker position (maybe it is normal, it would be nice to confirm)
    - `RegionChangeStarted` and `RegionChangeFinished` was used instead of `markerMoving`/`markerMoved`.
        - But i am not very happy while it was needed to provide mouse position to `RegionChange` and this event was not designed for that.
        - I think the use of this event here is fine.
        - Then would it be possible to remove the mouse position? If it is not, i can create a small hack for compatibility.
- `ActiveItemChangedEvent`
    - There is not only one activeItem, then this code have to be reworked (i added a FIXME)
    - On the plot the emit of the activeItemChanged is filtered when previousItem is none. Maybe it is a bug  (i added a FIXME)
- `interactiveModeChanged`
    - Implemented as `InteractiveModeChangedEvent`
- `contentChanged`
    - Implemented as `ChildAddedEvent` / `ChildRemovedEvent`
- `setGraphCursor`
    - Implemented as `CursorChangedEvent`
-  `activeItemChanged`
    - It is implemented as `ActiveItemChangedEvent` but we have to talk about it cause in fact there is many active items. I think the way it is implemented by the PR do not respect the thing
    - The plot function emitting `ActiveItemChangedEvent` have to be cleaned up i think.
    - I add a FIXME, the emit is inhibited when previousItem is None, i dont understand way
    - I create a `_kindFromItems`, to convert `Image` class to `image` string for example. The same function is used in the plot. It have to be merged somewhere.
- `setYAxisInverted`
    - Implemented as `AxisOrientationChangedEvent`
    - An enum provide the states DEFAULT/INVERTED
- `setXAxisLogarithmic`
    - Implemented as `AxisNormalizationChangedEvent`
    - An enum provide the states LINEAR/LOGARITHMIC
- `setXAxisAutoScale`
    - Implemented using `AxisAutoScaleChangedEvent`
- `setKeepDataAspectRatio`
    - Implemented using `AxisKeepDataAspectRatioChangedEvent`
- `setGraphGrid`
    - Implemented using `GridChangedEvent`
- [x] Fix uses of dictionary getter from the library


vallsv:
We can talk about this first API.
I do not test manually a lot of things.

vasole:
Does the ``__getitem__`` method of those objects return the ancient dictionary?

vallsv:
Yes, that's the idea. I create a munch of unittests before refactoring to take care of it.

vallsv:
I improve few things as discussed yesterday
- `LimitsChangedEvent`: takes the plot on the construction + merge API for Y-range `getYRange(axis='left')`
- Use already existing `Shape` objects and delete `DrawItem` (which was introduced by this PR)
- `ItemRegionChange`: Remove mouse position from the constructor

We still have to talk again about
- `ActiveItemChangedEvent`
- `_kindFromItems`
- Remove compatibility code used for event which was not existing in PyMca (basically when a signal exists)
- Remove `Plot.graphCallback` + `Plot._callback`?
- Package for `silx.gui.plot.events`
- Axis events
- qt.Qt.LeftButton instead of left
- scenePos (like pyqtgraph) or dataPos

t20100:
`enum.Enum` which is used in `PlotEvent` is not available in python2.7..
There is a backport which we can have in third_party: https://pypi.python.org/pypi/enum34

vallsv:
I was not aware it was not avaialbe.
But if we provide it in silx, i will use it for other part of the lib.

t20100:
For enum, see PR #839

t20100:
For compatibility of events with PyMCA see calls of `_callback` in https://github.com/vasole/pymca/blob/master/PyMca5/PyMcaGraph/backends/MatplotlibBackend.py

PR #739, opened 2017-04-19
735 download

### LINKED ISSUE #735: Utilities to dowload test data
Ported from pyFAI

kif:
close #735


kif:
A couple of comments:
* Damien suggests to handle checksums ... do we want/need this ?
* shall it be available to be used from outside silx ? (possibility to set alternative URLs, ...)

dnaudet:
IMO if only used for silx tests, checksums are probably overkill and adds unnecessary complexity. I just added them to the xsocs download script because it didn't take me long.

t20100:
`download_all` is not tested and uses a `getimage` method that was rename as `getfile`.

I would not provide default values for `project`, `url_base` and `env_key` in `TestResources.__init__`.
I'd rather provide them when instanciating it.

What about renaming `utilstest` to `testresources`, I find utilstest not meaningfull.

dnaudet:
It looks like you update the json file before actually downloading the file, shouldnt it be the other way around?

PR #746, opened 2017-04-19
Improve viewer script

vallsv:
- create menus: open, exit, about
- update viewer on selection changed
- always use raw mode if it is not requested (it is an update of the DataViewer behaviour)


PiRK:
LGTM

PR #747, opened 2017-04-20
[FitAction] keep a constant color for fitted curve

### LINKED ISSUE #742: [fitAction] Cycling colors
The fact of showing the estimation introduces changes on the shown curves.

I see the interest, but I would not show the estimated curve unless explicitly requested. 

You should also keep in mind that if you do so, in addition you would have to handle the plot update if the user changes the estimation parameters by hand.

At this point (v 0.5.0) I would just get rid of the plot of the estimated curve.

PiRK:
Ensure the fit curve keeps the same color by reusing a curve object

Also, remove the display of the estimation curve. To do it properly, one would need to update the plot also when the estimated parameters are manually modified.

Closes #742

### DIFF
--- a/silx/gui/plot/PlotActions.py
+++ b/silx/gui/plot/PlotActions.py
@@ -51,7 +51,7 @@ from __future__ import division
 
 __authors__ = ["V.A. Sole", "T. Vincent", "P. Knobel"]
 __license__ = "MIT"
-__date__ = "16/02/2017"
+__date__ = "20/04/2017"
 
 
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


vasole:
I think there is a problem. The fit curve is not deleted either on a new estimate (click on estimate) or after a click on the fitAction button. For instance, the user can re-adjust the fitting region from the plot and then click again the fitAction button.

vasole:
Well, I have seen in PyMca the behavior is the same we have now. Let's see if people complain.

PiRK:
I thought the non-deletion of the fit when the fitting region changes was a feature, to have several fits visible on different data ranges. Of course, it means the previous fits must be explicitly deleted after the zoom has changed.

vasole:
Do not worry. Leave it as it is and wait for feedback.

Issue #669, opened 2017-03-15
[plot3d] MacOS multiple display issue


t20100:
With viewer3DscalarField example, there is some issues with dual display:
- with mirror display: second screen turn off.
- with dual display: starting on HDPI screen and moving window to second screen makes mouse interaction only working in the corner of the window

t20100:
Closed by PR #749

PR #749, opened 2017-04-20
[plot3d] Some fixes

### LINKED ISSUE #694: [plot3d] Change data info in scalarfield vidw param tree
In the parameter tree of the 3d view, offset and scale are in x, y, z order while shape is given as z, y, x.
Replace shape by size or dimension and order it the same as offset and scale.

t20100:
This PR:

- Change shape to size in parameter tree to match order of offset and scale (closes #694)
- Fix starting on HighDPI screen and moving window to second screen makes mouse interaction only working in the corner of the window (see #669) + display text with a consistent size

### DIFF
--- a/silx/gui/_glutils/font.py
+++ b/silx/gui/_glutils/font.py
@@ -67,7 +67,11 @@ ULTRA_BLACK = 99
 """Thickest characters: Maximum font weight"""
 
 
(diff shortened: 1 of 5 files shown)


t20100:
Closes #669, with this change, I could not reproduce mirror display issue on macos

PR #753, opened 2017-04-21
Add documentation for the API to get the ROI information

### LINKED ISSUE #413: [plot] ROI widget: add and document an API to get the ROI information

PiRK:
Closes #413


vasole:
That implements/documents a getter. Do we have a setter?

PiRK:
We have a setter, `fillFromROIDict`.  Its usage is somewhat complicated to understand, the dictionary parameter is not documented.

vasole:
@t20100 

If we want to 100% follow CamelCase conventions, one should write Roi and not ROI (because of three capitals). I would propose to have self.fillFromROIDict = self.fillFromRoiDict and equivalents for compatibility.

PR #758, opened 2017-04-25
[spech5] aborted scans

### LINKED ISSUE #756: [spech5] error with scans aborted after 0 points
For spec files containing scans that are interrupted before the first point
(`aborted after 0 points.`) the `measurement` group contains all counters and
scalers, but on accessing them, an error is thrown.

PiRK:
closes #756

Catch the error when using `data_column_by_name`  and return an empty array.

Use custom errors for handling errors thrown by the C library, instead of trying to interpret them.


vasole:
We need to discuss the meaning of scan_aborted in your code. Many scans are aborted and not because of that are empty.

PiRK:
I can rename it to self.scan_aborted_0pts. Or empty scan. This name is only used in the tests.

vasole:
Any of those (scan_aborted_0pts or empty_scan) would make it.

PR #761, opened 2017-04-25
Animated PNG icons

vallsv:
Provide animated icons using many PNGs images

- It will be used if MNG is not supported (GIF was providing a very bad rendering)
- It will be always used in Qt5, instead of MNG (in Qt5 MNG rendering using alpha transparency have a bad rendering)


PiRK:
LGTM

Issue #757, opened 2017-04-25
[Plot] Histograms


dnaudet:
Hi all, I have a couple issues with the display of histograms :

Context : displaying histograms :
``` python
Plot.addCurve(x, y, histogram='left')
# with len(x) = len(y) + 1
```

* the first one was, AFAIK, already there in 0.4 : the x/y log actions throw an exception (because x and y lengths aren't the same probably).
* the second one is a regression and probably appeared with the introduction of objects : addCurve throws an exception if x and y don't have the same length.

payno:
I made a few modification, now this should be ok. But we should wait for @t20100 to review it before merging.
There is quit a lot of new classes and I don't know the structure of the code. My patch is maybe not ''adapted" to the OO architecture.'

Issue #688, opened 2017-03-28
[plot] Split Curve class item into separated classes


t20100:
As is, histograms, scatters and curves are all defined in a single class.
The API would probably be clearer if they where separated objects in the OO API even if they share a common method to add them to the plot.

vasole:
While I see the interest for the histogram, I really fail to see it for the scatter. To me a scatter is just a curve without line connecting the points.

vallsv:
In this single case, if you want to create generic widget to custom styles of your items it will be already useful to split concepts.
And I guess if you want other display like voronoi or what, this class can be very useful.
BTW `PyQtGraph` provides `PlotCurveItem` and `ScatterPlotItem` (http://www.pyqtgraph.org/documentation/graphicsItems/scatterplotitem.html)

vasole:
t20100 has clarified a point that had been escaped to me. The scatter object handles a third coordinate and it is indeed a representation of 3-dimensional data.

t20100:
closed by #762

PR #762, opened 2017-04-25
Create an Histogram item

payno:


t20100:
This will close #688 and close #757.

payno:
LGTM to.
I made part of it to .

PR #763, opened 2017-04-26
Refactoring third party

vallsv:
- Use `third_party/_local` mechanism
    - Only `six` is concerned right now
- Fix debian 8 and 9 packaging to avoid to package `third_party/_local` (debian 7 still use it, but who cares)


PiRK:
LGTM

PR #771, opened 2017-04-28
Correct bug affecting calculation of ROIs

### LINKED ISSUE #770: [ROI] Bug calculating ROIs on curves with negative or unsorted X values
The calculation is made on the unsorted values what gives wrong results.

vasole:
This PR

[x] closes #770
[x] implements a non-regression test
[x] provides a simpler access to the ROIs (setRois, getRois, calculateRois).

The later can be removed if considered convenient.


vallsv:
LGTM

PR #778, opened 2017-05-02
773 median

kif:
some work remains


vasole:
Perhaps we could profit this PR to  rename the first argument of the medfilt2d definition (input is a built_in function)

payno:
Your modifications have been included into the fix 773 pull request.
So I will close this one if it is ok for you ?

vasole:
closed because of being merged into 773.

PR #779, opened 2017-05-02
Support complex for curve, image, cube and stack

### LINKED ISSUE #775: [viewer] Display of complex data
The viewer needs to handle complex data.
Trying to display complex data raises an error depending on the selected view.
This is somewhat related to issue #277.

One proposition is to display the modulus of complex (bumpy.absolute), or to add a selection 'filter' to choose what to display.

vallsv:
This implementation only filter data with numpy.absolute in case of complex.

The toolbar do not display the raw data, but the filtered result

close #775

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -66,6 +66,21 @@ def _normalizeData(data):
     return data
 
 
+def _normalizeComplex(data):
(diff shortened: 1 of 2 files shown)


vasole:
Functionality is there. Waiting for the CI to finish.

vasole:
Tested with silx view.

PR #782, opened 2017-05-03
Fix 773

payno:


vasole:
- Please check the names are medfilt (not medianfilt), medfilt1d and medfilt2d.
- The situation is much worse than before. Running your benchmark on my machine:

medianfilter780

payno:
Is your compiler dealing with openMP ?
Because I test it on an other computer and tendency is the same ?

payno:
Implementation using queue didn't seems to be a good idea. I move back to vector with fix size.
Seems to make a difference on @vasole computer.

vasole:
Yeah, with your modifications now silx is twice faster than scipy or pymca.

PR #783, opened 2017-05-03
Provide matplotlib as package

### LINKED ISSUE #777: Provide a way to import matplotlib
Some glue is needed to import `matplotlib` and Qt at the same time.

While silx provide Qt, it is needed to also provide `matplotlib` externally.

- create `silx.gui.matplotlib` package
- move `Modestimag` inside
- fix imports

vallsv:
- isolate matplotlib dependency into a package
- move colormaps as resources
- there is no changes on the code, it is only a move of functionality from a module to another one

Close #777


vasole:
Is there no danger of having a module under silx.gui trying to import matplotlib to get the wrong thing?

vallsv:
At the end, the matplotlib backend is the only thing which should import matplotlib, then i guess no.

t20100:
@vasole in this case (which is unlikely), ``from future import absolute_import`` should be used.

The idea to have it outside the plot is for application willing to use both silx and matplotlib directly (e.g., pyFAI) to share matplotlib initialisation with silx.

t20100:
A few comments:
- I would prefer to have all colormaps in a single file by using either structured array (https://docs.scipy.org/doc/numpy/user/basics.rec.html) or ``numpy.savez``.
- We need to update the copyright file regarding files containing viridis and cie colormaps (CC0 license + different copyright)
- I think functions ``get(MPL)Colormap`` and ``get(MPL)ScalarMappable`` functions in ``silx.gui.plot.Colors`` should go into ``silx.gui.matplotlib`` package.
- The way ``get(MPL)Colormap`` is done could be improved, but this can be done in another PR.

Finally, do we want to merge this before the release?

vasole:
Do as you want. I still do not see the need because all what an application has to do is to is to choose which of PyQt4, PySide or PyQt5 is going to use prior to access either silx or matplotlib for the rest to be properly initialized. I never needed to touch the inners of the qt wrapper nor the matplotlib backend.

t20100:
We also need to set the backend in matplotlib 'rcParams' and to choose the figure widget from qt4 or qt5 backend.

I'm also wondering if this package would be better suited in ``silx.gui.plot.matplotlib`` as matplotlib is only used in the plot.

vasole:
"""" 
We also need to set the backend in matplotlib 'rcParams' and to choose the figure widget from qt4 or qt5 backend.
"""

But that is to be made automatically by the matplotlib backend based on the Qt binding already used. I never, ever, needed to set rcParams externally. The backend has to use the Qt binding used. That's all.

vallsv:
* I said i really think it is better to see a single colormap as a single resource. By experience it is more easy to manage.
* I don't reach deeply `Colors.py`, but at one point `getColormap` will return something else than MPL color map, then it makes sense to provide it here. But for an external use of matplotlib, you are right, maybe it can be useful to provide MPL colormaps.
* For an external use, i would prefer the short version `silx.gui.matplotlib` than `silx.gui.plot.matplotlib`. Same level as `silx.gui.qt`. But it's up to you, i am fine with both.

t20100:
There seems to be a small difference in the `getSupportedColormap` behavior, I'll make you a PR for that.

vallsv:
Thanks a lot. It's now merged.

PR #785, opened 2017-05-04
Add unit test on ascent or lena for the median filter

payno:


kif:
LGTM

kif:
Nota: the "lena" method still exists but it raises a runtime error.
I would suggest to test for ascent and fall-back on "lena" in the other case.

Nota: lena is available as external-resource as well.

PR #788, opened 2017-05-04
Fix 787

payno:
Need PR #668 to merged, before this one be valid.
Not sure this will need any unit test ?
Maybe just making sure this is launching ? Is this useful ?


vasole:
The unit test could be as simple as taking one of the existing unit tests for the median filter and make it accessible from image.

payno:
this PR has been modified by the PR  #803

PR #794, opened 2017-05-04
Rework setup.py for cython and openmp

vallsv:
- The `Build` class allow to catch options from the build command
- The `BuildExt` class take care of the cytonization and filter flags for OpenMP
- No more global
- No more command-line parsing and hack of sys.args
- No more process outside the setup() function
- Command-line argument have the priority on env variables
- Env variables are parsed (0/1/true/false... else are ignored)

```
python setup.py build --no-cython --no-openmp

python setup.py build --force-cython --openmp

export FORCE_CYTHON=yes
export WITH_CYTHON=True
export WITH_OPENMP=1
python setup.py build

export FORCE_CYTHON=no
export WITH_CYTHON=False
export WITH_OPENMP=0
python setup.py build
```
Limitation: Build options must be after `build` argument.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@
 # ###########################################################################*/
 
 __authors__ = ["Jérôme Kieffer", "Thomas Vincent"]
-__date__ = "26/04/2017"
+__date__ = "04/05/2017"
 __license__ = "MIT"
 
 
@@ -47,6 +47,7 @@ logger = logging.getLogger("silx.setup")
 
 
 from distutils.command.clean import clean as Clean
+from distutils.command.build import build as _build
 try:
     from setuptools import Command
     from setuptools.command.build_py import build_py as _build_py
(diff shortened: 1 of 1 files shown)


vallsv:
Let me know if this version is acceptable for you

vasole:
Just a question (I am lazy to look at the code).

Are the options 0, False, no equivalent and the options 1, True, yes equivalent when dealing with the environment variables? It would be very cumbersome to have to remember that for WITH_CYTHON one has to use True/False and for WITH_OPENMP 1/0

vasole:
In the code we test for false and true. Should the capitalized forms False and True be added? I say it just because it is what it have been naturally used in the opening post.

Following @vallsv decision, the PR can be approved as it is or with the added cases.

vallsv:
@vasole The env variable content is converted to lower case before the test. And `true` is same as `1` or `yes`. If the env variable contains an unexpected value, a warning is displayed and the variable is ignored. I think that's the safer way for env variable.
```
+        value = content.lower()
+        if value in ["1", "true", "yes", "y"]:
+            return True
+        if value in ["0", "false", "no", "n"]:
+            return False
```

PR #798, opened 2017-05-05
Merge opencl fix conflict

payno:


payno:
No working as it should.

PR #799, opened 2017-05-05
OpenCL Median filter on images

kif:
Big commit with all modifications over the 2 last month

replace #668


vasole:
I have tried to review it but I am not sure about my machine being an appropriate tool:

- I do not get the same output as with scipy.ndimage.filters median_filter with mode set to nearest but I get a warning  saying max actual groupsize: 1024, expected: 512

- It takes an enormous amount of time (50 seconds for ascent image and 55 kernel)  but I do not have a dedicated GPU

payno:
LGTM

PR #801, opened 2017-05-09
Fix setup

### LINKED ISSUE #800: setup.py fails when not using cython
This is due to 'fake cythonize' not patching list of source files.

t20100:
Closes #800 + minor typos


vallsv:
Nice catch. LGTM.

PR #803, opened 2017-05-09
Fix 787 v2

payno:


t20100:
conditional argument is not available in opencl implementation, I would propose to remove this argument from the unified function.

PR #804, opened 2017-05-09
Fix scatter plot with colormap for OpenGL backend

### LINKED ISSUE #802: [plot gl] scatter plot with colormap are not properly displayed
To reproduce:
from silx import sx
w = sx.Plot2D(backend='gl')
w.show()
x, y, v = numpy.random.random(1000*3).reshape(3, 1000)
w.addScatter(x, y, v)
w.resetZoom()

This seems to come from an issue with color type handling

t20100:
Closes #802 
This fixes support of RGB(A) uint8 color array internally to support color produced by applying colormap to scatter plot.

### DIFF
--- a/silx/gui/_glutils/VertexBuffer.py
+++ b/silx/gui/_glutils/VertexBuffer.py
@@ -179,7 +179,8 @@ class VertexBufferAttrib(object):
                  size,
                  dimension=1,
                  offset=0,
-                 stride=0):
+                 stride=0,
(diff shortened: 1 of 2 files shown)


payno:
LGTM

t20100:
@vallsv sorry, I'm a little too british!

I'll change that along with a future PR, no need to do it now for the release, `normalisation` is not in public API,

PR #806, opened 2017-05-09
Update a Travis test to use silx compiled with asserts

### LINKED ISSUE #796: [CI] Make a test in DEBUG mode
To avoid problems with nightly builds, at least one of the CI tests should be run in DEBUG mode.

It looks to me that it would be preferable to do it for one of the linux tests.

### LINKED ISSUE #810: [package] Fix -DNDEBUG on debug mode

vallsv:
- Fix flag compiler to have expected behaviour with `--debug`
- Fix flag compiler according to the use of `python-dbg`
- Configure Travis Linux Python 3.4 to be compiled and tested with asserts

It closes #796
It closes #810

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
           env:
               - BUILD_COMMAND=sdist
               - WITH_QT_TEST=False
+              - PIP_INSTALL_EXTRA_ARGS="--global-option build --global-option --debug"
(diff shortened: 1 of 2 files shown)


t20100:
Why not leaving the management of `-g` to `numpy.distutils` and co, even if they put multiple ones?

Issue #813, opened 2017-05-10
[packaging] seg fault during nightly build


t20100:
Reproduced by importing twice pyopencl with python2.7-dbg:
```
$ python2.7-dbg
>>> import pyopencl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/pyopencl/__init__.py", line 28, in <module>
    import pyopencl._cl as _cl
ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.55.0: undefined symbol: Py_InitModule4_64
[44627 refs]
>>> import pyopencl
Segmentation fault
```

The `silx.opencl.common` module seems to import twice pyopencl leading to seg fault in nightly build.
This can be cleaned up, but the issue will remain when importing twice pyopencl... This needs more investigation.

t20100:
I'm closing this issue, nightly build now OK.

PR #815, opened 2017-05-10
Clean pyopencl import

t20100:
This PR removes duplicated init code in opencl.common.
It is workaround for issue #813 as it avoids pyopencl's seg fault when imported twice with python2.7-dbg on Debian 8...
But the root of issue #813 which is not in silx remains unsolved.


t20100:
With this, debian 8 packaging terminates successfully (though opencl does not get tested in debug)

PR #819, opened 2017-05-11
Remove median filters hard scipy dependency

vasole:
Two of the tests were relying on scipy.

### DIFF
--- a/silx/image/test/test_medianfilter.py
+++ b/silx/image/test/test_medianfilter.py
@@ -28,30 +28,21 @@
 
 __authors__ = ["H. Payno"]
 __license__ = "MIT"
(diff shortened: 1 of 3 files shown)


kif:
It is a non-sense to check the median filter on ordered data.

vasole:
1 - That test checks if the two median filters implementations are accessible from image. Nothing else.
2 - A median filter with a (1,1) kernel is non-sense on **ANY** data.

payno:
Those test are in image module. Image module is calling opencl and math median filters. The goal is just to check that the link between those is done and not to test directly tests (which are already in math and opencl modules).
But maybe those test are just overkilled.

PR #836, opened 2017-05-22
[mask] minor mask improvements

### LINKED ISSUE #834: [mask] setSelectionMask can't process return value of getSelectionMask()
When no mask is set, e.g. when the mask tools widget has just been initialized, the following code prints an error message:

`self._getMaskToolsDockWidget().setSelectionMask(self._getMaskToolsDockWidget().getSelectionMask())`

The error log is:
`ERROR:silx.gui.plot.MaskToolsWidget:Not an image, shape: 1`

We could avoid this by making `setSelectionMask` accept an empty array (and possibly None?) as a synonym for `resetSelectionMask()`

PiRK:
Adds `resetSelectionMask` to dock widget API.

Fixes also a few  minor sphinx warnings.

EDIT: Initially this PR was meant to close #834, but this will need a bit more work than expected.

### DIFF
--- a/silx/gui/plot/CurvesROIWidget.py
+++ b/silx/gui/plot/CurvesROIWidget.py
@@ -197,6 +197,7 @@ class CurvesROIWidget(qt.QWidget):
            - ``"from"``: x coordinate of the left limit, as a float
            - ``"to"``: x coordinate of the right limit, as a float
            - ``"type"``: type of ROI, as a string (e.g "channels", "energy")
+
         :param order: Field used for ordering the ROIs.
              One of "from", "to", "type", "netcounts", "rawcounts".
              None (default) to get the same order as displayed in the widget.
--- a/silx/gui/plot/MaskToolsWidget.py
+++ b/silx/gui/plot/MaskToolsWidget.py
@@ -610,6 +610,5 @@ class MaskToolsDockWidget(BaseMaskToolsDockWidget):
     :paran str name: The title of this widget
     """
     def __init__(self, parent=None, plot=None, name='Mask'):
-        super(MaskToolsDockWidget, self).__init__(parent, name)
-        self.setWidget(MaskToolsWidget(plot=plot))
-        self.widget().sigMaskChanged.connect(self._emitSigMaskChanged)
+        widget = MaskToolsWidget(plot=plot)
+        super(MaskToolsDockWidget, self).__init__(parent, name, widget)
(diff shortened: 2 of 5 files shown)


PiRK:
Fixing #834 seems to cause unexpected side effects, so I reverted the initial commit and I'll fix the issue directly in PyMca by avoiding calling `setSelectionMask` with an empty array.

This PR is now just a minor docstring improvement and exposes `resetSelectionStack` in the dock widget.

PiRK:
Would it be OK to set `self._z = self.plot.getActiveImage().getZLayer() + 1` in `MaskToolsWidget`?

I have a case for which my active image is on z=1, and when it is changed while the mask widget is hidden, the image is redrawn on top of the mask, which is then invisible (https://github.com/vasole/pymca/pull/58#issuecomment-303717716)

vasole:
@t20100 

It would be better than to have something hard  coded to 1. I would say proposed solution should be the default for the case the application has not explicitly set it.

In any case, an application should be able to specify at what level/layer wants the mask.

t20100:
+1 for what @PiRK proposed and optionally to set the Z value of the mask.

PiRK:
After looking at the code, tt seems `self._z = activeImage.getZValue() + 1` is already implemented in the `_activeImageChanged` and `_activeImageChangedAfterCare` slots.

The issue only exists when the mask is set while the mask widget has never been set visible, so neither one of these slots is connected, and the Z layer is set to its init value (1). I will assess whether it is possible to connect `_activeImageChangedAfterCare` already in `__init__` .

PiRK:
Actually it would not help to have the slot connected in __init__, because _activeImageChangedAfterCare would automatically disconnect itself as soon as the active image is set with a different shape as the current mask (initialized as a (0, 0) shape array). 

The broader question is whether we want to change the mask widget to have the same behavior whether is is hidden or shown, or keep it as it is. As it is, it's designed to fully work and accept image changes only when visible. When hidden, it only shows the mask as long as the shape of the active image does not change, else it removes the mask and disconnects itself.

t20100:
The mask widget was only designed to work with an active image.
It's already be extended to scatter plots and is not ready for synchronization with other mask widgets...
It's probably time to consider a refactoring.

vasole:
The synchronization can be achieved by the application as being shown with the work on PyMca.

Can this PR be merged?

PR #841, opened 2017-05-23
Clean-up logging

### LINKED ISSUE #825: [logging] Clean up library from logging.basicConfig
`logging.basicConfig` should never be used by a library. The applications have to take care of the logging handlers.

You can read for example https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library

Maybe there is special cases, like notebooks or ipython consoles. But i hope it already provide configurer logging system.

t20100:
This PR closes #825 by:
- Removing `logging.basicConfig()` from silx (replaced by do nothing logging handler).
- Calling `logging.basicConfig()` in `sx` module when used from the console. 
- Creating all module's logger using `logging.getLogger(__name__)`
- Renaming deprecation logger to `"silx.DEPREACTION`"

### DIFF
--- a/silx/__init__.py
+++ b/silx/__init__.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2015-2016 European Synchrotron Radiation Facility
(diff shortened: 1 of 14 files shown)


vallsv:
LGTM

PR #851, opened 2017-05-31
Add support of Qt5.4's QOpenGLWidget to OpenGL plot backend and plot3d

### LINKED ISSUE #793: [plot gl][plot3d] Use Qt>=5.4 new OpenGL widget API

t20100:
This PR adds support of Qt>=5.4 QOpenGLWidget (which deprecates QGLWidget), closes #793.
It does so by providing a common OpenGLWigdet used by OpenGL plot backend and plot3d which inherits either from QGLWidget or QOpenGLWidget and support High-DPI with Qt5.

It also improves the support of cases where OpenGL is not available (PyQt4.QtOpenGL not installed, OpenGL2.1 not available,...).
In those cases, it displays an error pop-up once and displays an empty widget instead of the OpenGL rendering.

### DIFF
--- a/silx/gui/_glutils/FramebufferTexture.py
+++ b/silx/gui/_glutils/FramebufferTexture.py
@@ -66,49 +66,48 @@ class FramebufferTexture(object):
         self._previousFramebuffer = 0  # Used by with statement
 
         self._name = gl.glGenFramebuffers(1)
-        gl.glBindFramebuffer(gl.GL_FRAMEBUFFER, self._name)
 
-        # Attachments
-        gl.glFramebufferTexture2D(gl.GL_FRAMEBUFFER,
-                                  gl.GL_COLOR_ATTACHMENT0,
-                                  gl.GL_TEXTURE_2D,
-                                  self._texture.name,
-                                  0)
+        with self:  # Bind FBO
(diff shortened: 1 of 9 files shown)


vallsv:
Looks interesting, but just a comment about `OpenGLWidget`.
- As a base component it would be much better to not display any popup. We can talk about that to find another way for that. If it anyway can't render anything inside, it is maybe more interesting to display the message embedded in the widget (it is like web browsers when flash is not installed or not authorized: it is rendered as an hashed box with a text message).
- And as base Qt component is is maybe interesting to move it inside `silx.gui.qt.OpenGLWidget`. But it is just a proposal.

I don't check other parts.

t20100:
Agreed for first point, and it should not be too complicated to do.

For the second point, the OpenGLWidget as is, is not meant to be public. It fits the needs for the OpenGL plot backend and plot3d widgets but not more for now. For instance, not all features of QGLWidget and QOpenGLWidget are exposed and not all methods are defined in the fallback.

t20100:
Removed the pop-up and display the error message inside the widget instead.

vallsv:
Looks fine as i remember :-)

PR #852, opened 2017-05-31
QDoubleValidator() cannot be used without passing parent object

mguijarr:
Hey guys,

I tried to use latest silx and got an error message about `QDoubleValidator()` signature ; it seems it is
invalid to call `QDoubleValidator()` constructor without at least a parent object or `None`, look:

```
   Python 2.6.6 (r266:84292, Dec 26 2010, 22:31:48)
   [GCC 4.4.5] on linux2
   Type "help", "copyright", "credits" or "license" for more information.
   >>> from PyQt4.QtCore import QT_VERSION_STR
   >>> from PyQt4.Qt import PYQT_VERSION_STR
   >>> from sip import SIP_VERSION_STR
   >>>
   >>> print("Qt version:", QT_VERSION_STR)
   ('Qt version:', '4.6.3')
   >>> print("SIP version:", SIP_VERSION_STR)
   ('SIP version:', '4.10.2')
   >>> print("PyQt version:", PYQT_VERSION_STR)
   ('PyQt version:', '4.7.3')
   >>> from PyQt4.QtGui import QDoubleValidator
   >>> v=QDoubleValidator()
      File "<stdin>", line 1, in <module>
      TypeError: arguments did not match any overloaded call:
      QDoubleValidator(QObject): not enough arguments
      QDoubleValidator(float, float, int, QObject): not enough arguments
    >>>
``` 
This PR fixes the problem I had.

### DIFF
--- a/doc/source/Tutorials/fitconfig.rst
+++ b/doc/source/Tutorials/fitconfig.rst
@@ -122,7 +122,7 @@ used by our fit function to scale the *y* values.
             self.scalingFactorEdit.setToolTip(
                 "Enter the scaling factor"
             )
-            self.scalingFactorEdit.setValidator(qt.QDoubleValidator())
+            self.scalingFactorEdit.setValidator(qt.QDoubleValidator(self))
 
             self.ok = qt.QPushButton("ok", self)
             self.ok.clicked.connect(self.accept)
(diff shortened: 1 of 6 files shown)


vasole:
Yep, I recall that problem appearing from a certain PyQt version on. I already felt on that one too.

PR #853, opened 2017-06-01
Refactoring actions

payno:
- [x] move PlotActions.Py into sub modules
- [x] fix import from PlotActions
- [x] add deprecation warning in PlotActions.py
- [x] Update documentation of plot actions


payno:
filter => medfilt
only import control and io in actions
shoudl decorator module be renamed as depracation (and include warning file)  ?

PR #857, opened 2017-06-01
Fix macos OpenGL widgets text display

### LINKED ISSUE #790: [plot gl] On macOS the text label are clipped
Sometimes the last character is not fully displayed.

This seems to come from the text rendering by Qt not providing the correct size to initialize the buffer where to render, but it needs more investigation.

t20100:
- closes #790
- Avoids useless calls to rasterText in OpenGL plot backend


vallsv:
LGTM

Issue #858, opened 2017-06-01
[spech5] Segmentation Fault in SpecH5


carichte:
silx.io.spech5.SpecH5 is supposed do be called with a file-path as argument.

As soon as any other kind of object is passed to SpecH5, this will result in a Segmentation Fault (tested on different machines)

This usage is not unusual:
```
with open('/data/visitor/ma3331/id01/knno-47-008-SSO.spec') as fh:
    s5 = silx.io.spech5.SpecH5(fh)
```

I mentioned it before in the mailing list: http://www.edna-site.org/lurker/message/20170424.143447.51ee80ca.en.html

Together with relative paths from `SpecH5Group.visit` (see https://github.com/silx-kit/silx/issues/614#issuecomment-305529363) fixing this would allow to use the `write_spec_to_h5` function to insert spec scans in existing h5 files.

PiRK:
I made a commit that implements the relative path behavior for `visit` and `visititems`. It appears that this change did not break the `spectoh5` module, which is a good sign. 
I'm now a bit worried about this change in behavior affecting other users who got used to dealing with full pathes.

vasole:
Can this be closed?

carichte:
Yes it seems to work. Thanks!

carichte:
Sorry I only checked before the .visit methods.

processing an opened spec file does not seem to work yet.

For python3, it fails because of `isinstance(filename, file)` which only works for python2.

Apart from that, it fails because in line:

https://github.com/silx-kit/silx/blob/master/silx/io/spech5.py#L1634

a `self.` is missing, leading to a `TypeError: coercing to Unicode: need string or buffer, file found`

PiRK:
I just proposed a quick patch to fix the 2 issues you pointed at. 

It seems that it would be useful to add a unittest before we can safely close this issue.  I can do that early  next week.

vallsv:
Thanks a lot Pierre. But take care of your holiday, we can manage the hot fix.

vallsv:
It is merged.
I created a test coverage for SpecH5 using file descriptor. It can be improved later.

Issue #614, opened 2017-02-07
[spech5 & fabioh5] Should visit() and visititems() follow links?


PiRK:
Currently `h5py.Group.visit` and `visititems` don't follow links when they recursively visit an HDF5 tree. So I also made `SpecH5Group.visit` ignore links.

But this `h5py` behavior could change in the future, see https://github.com/h5py/h5py/issues/671 (milestone 2.8). 

As the SPEC to h5 conversion relies on `visititems`, it would make our life easier if links were not ignored.  (see  #609). 

I don't expect that we will ever see a circular link in a SpecH5 or a FabioH5 object.

carichte:
Closely related but probably another issue:
There is another discrepancy between `SpecH5Group.visit` and `h5py.Group.visit`:

while the former always returns absolute paths of the file, the latter returns relative paths within the group.
I mentioned this here: http://www.edna-site.org/lurker/message/20170424.143447.51ee80ca.en.html

Would it not be better to adapt the h5py behavior?

PiRK:
Sorry, I believe I skipped your last message while processing all the mails when I came back from holidays.

I will investigate the discrepancy between `h5py.Group.visit` and `SpecH5Group.visit`. I suppose that when I first wrote SpecH5 my opinion was that we should include as much information as possible, and finding a relative path from an absolute path is easier than the opposite, but now I agree that we should mimic the `h5py` behavior as closely as possible.

Regarding the idea of accepting file handles as input for `SpecH5`, it will not be difficult to implement. But internally I will still need to find out the file name, in order to create a   `silx.io.specfile.SpecFile` object. So using an already open file handle will not optimize the processing in any way.

I definitely need to add a type test to prevent the segfault caused by passing a non-string to `silx.io.specfile.SpecFile`

carichte:
Maybe it is enough to avoid a segfault and throw an error instead. 
It would be interesting, though, to change a bit the behavior of `write_spec_to_h5`, to accept also SpecH5Group next to SpecH5 and a file path.
But if I'm right, this is already possible when using `silx.io.spectoh5.SpecToHdf5Writer` instead, together with visit giving relative paths.

PiRK:
I believe that you are right, `SpecToHdf5Writer.write` can already handle `SpecH5Group` objects.

I will modify `write_spec_to_h5` to accept groups. This should be trivial, it just needs to be added to the unit tests.

PiRK:
[...]

PR #859, opened 2017-06-02
[SpecH5] specfile segfault and SpecH5.visit behavior

PiRK:
Addresses #858
 
CHANGELOG: SpecH5Group.visit and .visititems now provide the relative name/path to the callback function (same behavior as h5py.Group.visit and .visititems)


PiRK:
See also #614

PR #863, opened 2017-06-06
[Mask] Make setSelectionMask update mask origin, scale...

PiRK:
If the plot has an active image when `setSelectionMask` is called, it is reasonable to use its origin, scale, Z...

This enables programmatic mask setting even if mask widget is not connected to active image signals.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -460,7 +460,7 @@ class BuildExt(build_ext):
 
     COMPILE_ARGS_CONVERTER = {'-fopenmp': '/openmp'}
 
-    LINK_ARGS_CONVERTER = {'-fopenmp': ' '}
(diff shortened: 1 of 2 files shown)


t20100:
If you already have an active image can't you call `_activeImageChanged`?
And don't you need to register a listener to handle active image changes?

PiRK:
Yes, calling  `_activeImageChanged` will work as well.

Regarding connecting a listener, I'm a bit hesitant because the existing 2 callback methods can potentially delete the mask if one of my widget ends up in a state with no active image. This is a problem, when synchronizing the mask between several widgets. But obviously this issue already exists when the mask widget is first activated and the connections are made.
I guess I will need to be very careful to never delete the image before replacing it with another one.

Issue #865, opened 2017-06-07
[plot] Zoom back on right click might not go back to default zoom level


t20100:
As zoom on wheel interaction is not stored in zoom history, zoom-back on right click might not zoom back up to starting zoom level (if wheel zoom has been used first).

At least, a right click when the zoom history is empty should do a reset zoom.

Maybe consider some way to store some zoom level in the history when zoom on wheel is used, e.g., when starting zoom on wheel after other zoom interaction has been performed.

vasole:
I find your _at least_ solution very elegant.

t20100:
PR #866 implements the _At least_ option.
Do we want more?

vasole:
I'm happy with it. In PyMca users had to go explicitly to the zoom reset icon. So, your solution is already better.

PR #866, opened 2017-06-07
Add resetZoom on right click to zoom interation mode

t20100:
This PR implements the _At least_ option of #865 by calling `resetZoom` when the history of zoom levels is empty.
To be compatible with previous event behaviour, when zoom history is empty, it only sends `mouseClicked` event when `resetZoom` didn't changed the limits.


vasole:
Are you sure the axis to be checked is right and not left?

PR #871, opened 2017-06-08
Add nx file extensions to the open dialog

### LINKED ISSUE #870: [silx view] nxs extension is not part of the open dialog
The main open dialog do not show `nxs` files by default. That's easy to fix.

vallsv:
closes #870

### DIFF
--- a/silx/app/view.py
+++ b/silx/app/view.py
@@ -158,6 +158,7 @@ class Viewer(qt.QMainWindow):
         extensions = collections.OrderedDict()
         # expect h5py
         extensions["HDF5 files"] = "*.h5"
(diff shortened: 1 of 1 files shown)


vasole:
While we should keep HDF5 Files as *.h5, I would add *.h5 also under NeXus files.

vasole:
Some additional comments:

- We generate NeXus files with h5 extension
- I have seen HDF5 and NeXus files with the hdf extension.

vallsv:
I see your point. I dont think it is needed but let's duplicate the extension.
By default all supported extension are displayed anyway.

PR #877, opened 2017-06-13
[Plot] Fix SaveAction with OpenGL backend

### LINKED ISSUE #850: [plot gl] Saving a plot snapshot fails with OpenGL backend + PyQt4
The SaveAction is using QPixmap.grabWidget which does not support QGLWidget.
It could use Plot.saveGraph as CopyAction and PrintAction.

t20100:
This PR closes #850.

It removes the option to save a snapshot of the plot as JPEG (anyway, I see little interest to save a snapshot in a lossy format...).
A snapshot of the plot can be saved as either PNG or SVG.


payno:
replace by #879 , fix conflict

PR #878, opened 2017-06-14
[Plot] Add colorbar to PlotWindow

### LINKED ISSUE #730: [plot] Add ColorBar to Plot2D
Now that we have a colorbar that synchronizes itself with the plot, I think we can: 
- Have it embedded in Plot2D, hidden by default to keep backward compatibility (e.g.,  a `Plot2D.setColorBarVisible` method)
- Have it enabled when using `sx.imshow`

t20100:
This PR adds a `ColorBarWidget` to the `PlotWindow` (closes #730) inside the central widget (having the colorbar in a DockWidget looks ugly).

The colorbar is hidden by default in the PlotWindow.
In the Plot2D it is displayed by default and a checkable Action which toggles the visibility of the colorbar is added to the toolbar after the colormap action.

Do we want to put a ToolButton with a menu instead of a checkable action as for aspect ratio and Y axis orientation?


PiRK:
If believe a checkable action with a tooltip, as you did, is adequate enough for toggling visibility of a widget.

t20100:
Do we want to want to display the colorbar by default in the `StackView` as well?

vasole:
I do not have a defined opinion on it. The advantage of showing it would be that it would be clear if one is applying a colormap to each image or the same colormap to all of them.

t20100:
I added the colorbar to StackView by default.
I found a few issues with the ColorBar while integrating it to PlotWindows.
In this PR I only fixed the one that broke the tests, I'll open an issue for the others and if it's OK fix them in another PR.

t20100:
We added an action to show/hide the colorbar to both Plot2D and StackView (where the colorbar is shown by default).

Issue #872, opened 2017-06-09
[mask] Load dialog do not display all supported files as default


vallsv:

vallsv:
It looks to be fixed.

PR #880, opened 2017-06-14
Load mask extensions

vallsv:
- Support TIFF images. It was not the case before, but the tiff extension was provided (maybe a regression)
- Display all available extensions by default

#872


payno:
LGTM

PR #882, opened 2017-06-14
FBP in silx

pierrepaleo:
This PR adds a Filtered Back-Projection (FBP) utility in silx.

For now, the only backend is OpenCL. The GPU version uses textures (image2d_t). There is a CPU version without textures.
 The Backprojection kernel is adapted from [PyHST2](https://forge.epn-campus.eu/projects/pyhst2) with Alessandro's permission.


silx/opencl/backprojection.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+    def filtered_backprojection(self, sino):
+        """
+        Compute the filtered backprojection (FBP) on a sinogram.
+
+        :param sino: sinogram (`numpy.ndarray`) in the format (projections, bins)
+        """
+
+        self.filter_projections(sino)
```
kif: as filter_projection alters the device memory, it should be locked ...
If this method is the "main", the best would be to lock here (and not in backprojection). 
Unless 2 locks are advices to prevent dead-locks.

kif:
I still have one issue:
```
ERROR: test_fbp (silx.opencl.test.test_backprojection.TestFBP)
  File "/users/kieffer/workspace-400/silx/build/lib.linux-x86_64-2.7/silx/opencl/test/test_backprojection.py", line 140, in test_fbp
    t, res = self.measure()
  File "/users/kieffer/workspace-400/silx/build/lib.linux-x86_64-2.7/silx/opencl/test/test_backprojection.py", line 117, in measure
    result = self.fbp.filtered_backprojection(self.sino)
  File "/users/kieffer/workspace-400/silx/build/lib.linux-x86_64-2.7/silx/opencl/backprojection.py", line 426, in filtered_backprojection
    self.filter_projections(sino)
  File "/users/kieffer/workspace-400/silx/build/lib.linux-x86_64-2.7/silx/opencl/backprojection.py", line 389, in filter_projections
    self.num_projs
  File "/usr/lib/python2.7/dist-packages/pyopencl/__init__.py", line 515, in kernel_call
    global_offset, wait_for, g_times_l=g_times_l)
TypeError: No registered converter was able to produce a C++ rvalue of type unsigned long from this Python object of type numpy.int32

Ran 1 test in 0.203s
```

kif:
LGTM
The support of gpyfft is left for later (dependency on clfft & debian9)

Issue #689, opened 2017-03-28
[plot] Add signals to plot items


t20100:
This allows one to listen to a dedicated object in the plot (e.g., a marker) without listening the whole plot.

t20100:
This issue becomes important if we want to ensure proper synchronization between object (e.g., between image, colorbar, colormap dialog).

PR #891, opened 2017-06-15
[Plot] Fix ColorBarWidget

### LINKED ISSUE #889: ColorBarWidget issues
PR #878 hightlighted a few issues in the `ColorBarWidget`

Issues:
- `ColorBarWidget` is not synchronized when `PlotWidget`'s default colormap is updated.
- Changing the colormap from `ColormapDialog` does not update the colorbar.
- From the console, doing an `addImage` with a different colormap does not update the colorbar immediately. It is only updated when `PlotWindow` gets the focus when clicking on it.
- ColorBar.py l809 `assert(type(self._vmin) == type(self._vmax))`: why is it needed? One can enter vmin=1, vmax=1.5 and it should work.
- If setting a colorbar with vmin == vmax, the colorbar gradient is not updated and displays the previous colormap.
- Using ImageData.setColormap on the active image does not update the colorbar.

Those issues were found using PR #878 branch.

Improvements:
- large values of vmin or vmax enlarges the width of the colorbar. The QLabels could span over/under the ticks to minimize the width of the colorbar + we could display those values with a limited number of significant digits (and have the full value displayed in a tooltip). Anyway, I would not enlarge the colorbar gradient above a certain width.
- It is probably worth listening for active image changes only when the ColorBarWidget is visible.

Refactoring:
- Refactor set|getColormap to only store it once in _ColorScale (and do the checks once) and have proxy to access it in other classes.
- Use a common implementation of checks and erroneous cases (e.g., negative bound with log scale) for the plot and the colorbar.
- In _ColorScale, it looks possible to create the QLinearGradient during setColormap rather than for each paintEvent.

Fixing some of those issues would probably be easier with item's signals (#689) and colormap object ( #146).

t20100:
This PR fixes issues in the ColorBarWidget. It fixes most of #889 issues. 

Issues:
- [x] `ColorBarWidget` is not synchronized when `PlotWidget`'s default colormap is updated.
- [x] From the console, doing an `addImage` with a different colormap does not update the colorbar immediately. It is only updated when `PlotWindow` gets the focus when clicking on it.
- [x] ColorBar.py l809 `assert(type(self._vmin) == type(self._vmax))`: why is it needed? One can enter vmin=1, vmax=1.5 and it should work.
- [x] If setting a colorbar with vmin == vmax, the colorbar gradient is not updated and displays the previous colormap.
- [x] Missalignment of tooltip and color gradient (due to widget margins?).

Improvements:
- [x] large values of vmin or vmax enlarges the width of the colorbar. The QLabels could span over/under the ticks to minimize the width of the colorbar + we could display those values with a limited number of significant digits (and have the full value displayed in a tooltip). Anyway, I would not enlarge the colorbar gradient above a certain width.
- [x] It is probably worth listening for active image changes only when the ColorBarWidget is visible.

Refactoring:
- [x] Refactor set|getColormap to only store it once in _ColorScale (and do the checks once) and have proxy to access it in other classes.
- [x] In _ColorScale, it looks possible to create the QLinearGradient during setColormap rather than for each paintEvent.

### DIFF
--- a/silx/gui/plot/ColorBar.py
+++ b/silx/gui/plot/ColorBar.py
@@ -539,7 +539,7 @@ class _ColorScale(qt.QWidget):
         """yPixel : pixel position into _ColorScale widget reference
         """
         # widgets are bottom-top referencial but we display in top-bottom referential
-        return 1 - float(yPixel)/float(self.height() - 2*self.margin)
+        return 1. - (yPixel - self.margin) / float(self.height() - 2 * self.margin)
 
     def getValueFromRelativePosition(self, value):
         """Return the value in the colorMap from a relative position in the
--- a/silx/gui/plot/ColorBar.py
+++ b/silx/gui/plot/ColorBar.py
@@ -92,7 +92,6 @@ class ColorBarWidget(qt.QWidget):
         self.layout().setSizeConstraint(qt.QLayout.SetMinAndMaxSize)
         self.setSizePolicy(qt.QSizePolicy.Minimum, qt.QSizePolicy.Expanding)
 
-
     def getPlot(self):
         """Returns the :class:`Plot` associated to this widget or None"""
         return self._plot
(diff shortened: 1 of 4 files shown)


payno:
Use a common implementation of checks and erroneous cases (e.g., negative bound with log scale) for the plot and the colorbar. => should be done by the colorbar object

t20100:
Initially planed but to be done in other PRs:
- Use a common implementation of checks and erroneous cases (e.g., negative bound with log scale) for the plot and the colorbar.
- Using `ImageData.setColormap` on the active image does not update the colorbar. This should fix changing the colormap from `ColormapDialog` does not update the colorbar. It requires item's signals and should be fixed by PR #689.

PR #892, opened 2017-06-16
[dictdump] Make sure files are closed even if an error occurs

### LINKED ISSUE #883: [io] dicttoh5 doesnt close the file if it fails
dicttoh5 doesn't close the hdf5 file it opened if there is an exception when writing to it.

PiRK:
Use context managers in `dicttoh5` and `h5todict`. Closes #883


vallsv:
LGTM

PR #897, opened 2017-06-19
Zenodo DOI in doc

### LINKED ISSUE #896: [doc] Provide DOI in documentation

kif:
close #896


vallsv:
About `On Linux, to install silx with pip, you must install numpy first`, would it be possible to change the sentence while it is not anymore the case?

vallsv:
Thanks for the fix.
LGTM too.

PR #898, opened 2017-06-19
[gui] Colormap object

payno:
PR relative to #146 

- [x] Move existing Colormap object from plot3d to plot
- [x] Add missing functionalities to the object + signal
- [x] Integrate the object in existing code
- [x] check back widgets using colormap
- [x] Update doc

### DIFF
--- a/doc/source/modules/gui/plot/colormap.rst
+++ b/doc/source/modules/gui/plot/colormap.rst
@@ -1,5 +1,5 @@
 
-.. currentmodule:: silx.gui.plot.colormap
+.. currentmodule:: silx.gui.plot.Colormap
 
 :mod:`Colormap`: Colormap API
 =============================
--- a/doc/source/modules/gui/plot/getting_started.rst
+++ b/doc/source/modules/gui/plot/getting_started.rst
@@ -343,7 +343,7 @@ When updating an image, if ``origin`` and ``scale`` are not provided, the previo
 Colormap
 ++++++++
 
-A ``colormap`` is described with a :class:`Colormap` as follows:
+A ``colormap`` is described with a :class:`.Colormap` class as follows:
 
 .. code-block:: python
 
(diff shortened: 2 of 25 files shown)


payno:
Some question reminder : 
    should the function converting the colormap to a dict should be called getDict or toDict ?

dnaudet:
or asDict ;)

t20100:
I'd avoid `getDict` to keep `get*` for 'attributes' of the class.

vallsv:
`_getDict`?

payno:
I think I will go for toDict

payno:
Q: 
- now that the Color is integrated in the plot is it still interesting to conserve the examples/colorbar ?

payno:
the management of the colrmap by the colorbar will be part of an other PR ( #913 ), this should prevend not regression, the next one should fix remainging issues of #889

payno:
From my point of view the class Colormap in plot3D/scene/function.py should inherit from the general Colormap and the should probably be changed to. What do you think @t20100

t20100:
Indeed, examples/colorbar.py is now pointless.
What would be interesting is to have an example of making a widget inherited from PlotWidget by adding actions, colorbar and tools (see #916).

vallsv:
The ColorBar widget is a widget which can be used without any dependency. Then an example showing the features is still useful.

t20100:
I'd prefer to keep `plot3d.scene` separated from `plot.Colormap` because the whole scene API was made using Python properties rather than get/set and it does not directly depend on Qt (so no QObject in there), but the top-level `plot3d` widget API should use it.

t20100:
@vallsv, agreed if it does not use `Plot2D` (which is what this PR provides).

vallsv:
Few comments:

```
from silx.gui.plot import Colormap as genColormap
```
- It would be good to remove the `as`. Just use Colormap.

```
_DEFAULT_COLORMAPS = {
    'gray': 0,
    'reversed gray': 1,
    'red': 2,
    'green': 3,
    'blue': 4,
    'temperature': 5
}
```
- It would be good to add a little comment about why this things have numbers.

```
NORMS = 'linear', 'log'
"""Tuple of supported normalizations."""

NORMALIZATIONS = ('linear', 'log')
"""Tuple of managed normalizations"""
```
- There is a duplication code. It would be good to use only one of this.
- I would prefer to have enum, but i am maybe the only one
- It would be nice to have const for LOGARITHM  and LINEAR
- If we want to expose the Colormap object as gui or plot module, which can be convenient, it would be nice to expose constants as static attributes from the object it-self.

Most of getter function does not have summary docstring. But i dont check the rendering of the doc anyway. I would prefer
```
Return the upper bounds of the colormap is exists, else None
:rtype: float, None
```
than
```
:return: the upper bounds of the colormap or None
```

vallsv:
You should implement `__eq__` instead of `compare`, and also `__hash__` can be very useful.

PR #903, opened 2017-06-20
[Plot] Merge Plot and PlotWidget classes

t20100:
This PR merges Plot.py and PlotWidget.py modules into a single PlotWidget module.


vallsv:
LGTM

PR #909, opened 2017-06-22
[plot] Create plot axes object

vallsv:


vallsv:
Maybe the documentation have to be fixed.

PR #913, opened 2017-06-23
Colorbar colormap object

payno:
This will need the merge of #898 before this one


payno:
I change a lot the examples/colorbar since the colormap refactoring and @t20100 modifications.
I moved it for a global view. It might also to check that there is no incoherence and to have an overview of the behaviour of colormaps.
 
More interesting there is inside a _ColormapEditor class that might be interesting if one day (or when) the Colorbar will allow modifications of the colormap. There will probably modification to do because it have been developed quiet straightforward.

colormapchecked

payno:
- [ x] link colormap signal with the plot
- [] check plot connection
- [x] add unit test to check that ColorBar is updated if the colormap of the plot change
- [] test viewer3d
- [] test imageview
- [] test statckviex

t20100:
LGTM, there is still some cases where the colorbar is not synchronized, but this is due to missing signals in plot items: #689

PR #920, opened 2017-06-27
Plot documentation

PiRK:
Fix documentation references to defunct class Plot.
Document item getters.
Fix a broken import in fit action.


vallsv:
LGTM

PR #921, opened 2017-06-28
Add finite argument to min_max

t20100:
This would be needed for PR #898 to improve colormap autoscale when there is inf in the image data.


payno:
LGTM

PR #931, opened 2017-07-05
[StackView] implement getPlot()

### LINKED ISSUE #927: StackView duplicates plot API
With the refactoring of the plot axes (#909) and colormap (#913) it has become apparent that duplicating a subset of the plot API in StackView was a bad idea and is causing extra work. 

I suggest that we deprecate the duplicated API (`setGraphXLabel, resetZoom, getActiveImage, getGraphTitle,setGraphTitle,  isKeepDataAspectRatio, setKeepDataAspectRatio`...) to give us the option of removing the methods in a few releases,  and just provide a `getPlot()` method to replace them all.

The only drawback to this is that it will indirectly expose the entire plot API instead of just the useful subset of methods.

PiRK:
Implement a `getPlot()` method, remove dependency on legacy `StackView` functional API in `Profile3DToolbar`.

Closes #927

### DIFF
--- a/doc/source/modules/gui/plot/stackview.rst
+++ b/doc/source/modules/gui/plot/stackview.rst
@@ -13,7 +13,9 @@
 
 .. autoclass:: StackView
    :members:
-   :exclude-members: remove, setInteractiveMode, addItem
+   :exclude-members: remove, setInteractiveMode, addItem, getActiveImage,
+       resetZoom, setYAxisInverted, isYAxisInverted, getSupportedColormaps,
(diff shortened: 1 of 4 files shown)


PiRK:
Should `getPlot()` return a weakref proxy, or is it OK to return the plot itself?
If we want to return a proxy, should we provide a callback argument to allow users to be informed when the plot is destroyed (`getPlot(callback=None)`)?

t20100:
Return the plot widget itself, the caller is responsible for releasing it.

PR #933, opened 2017-07-05
Bugfix in backprojection

pierrepaleo:
This PR fixes a bug in the Backprojection class when using it to reconstruct multiple slices.


PiRK:
Great.

PR #935, opened 2017-07-06
[debugging AppVeyor] Force PyQt5==5.8.2

PiRK:
Force PyQt5==5.8.2 to see if tests pass.


PiRK:
This hack is a temporary workaround for issue #936, which might take a while to fix.

PR #937, opened 2017-07-06
Fix tests failing with PyQt5.9 on Windows

### LINKED ISSUE #936: [GUI] tests and hdf5viewer example fail with PyQt5.9 wheel on Windows

t20100:
Change time.sleep to qWait in hdf5 tree model tests.
This should fix failing tests with appveyor.

Closes #936


vasole:
If it is a PyQt bug (as it seems to be the case), I would take a extreme programming approach and skip the affected tests in case the PyQt version is exactly 5.9.1.

t20100:
Closing as not solving the issue.

PR #938, opened 2017-07-07
fix bug introduced by PR #928. Setup.py not updated (PyTest) for the …

payno:
…default inplace option


PiRK:
This should fix the nightly build issue.

PR #941, opened 2017-07-07
FBP: Bugfixes and tests

pierrepaleo:
This PR adds another bug fix in the Backprojection class, and adds a unit test to ensure that multi-slice reconstruction goes smoothly.


vasole:
The title says WIP while the flag says ready to merge....

pierrepaleo:
It should be OK, I was waiting for the CI to succeed.

PR #942, opened 2017-07-07
medfilt2d: take kernel_size into account

pierrepaleo:
This PR adds a small bugfix in the OpenCL median filter which was not actually taking into account the kernel size.


PiRK:
LGTM. The two AppVeyor tests still failing are related to the known PyQt 5.9 issue.

PR #945, opened 2017-07-10
Change test image size.

### LINKED ISSUE #943: [opencl] Test crashing under windows with Python 3.5
The test_transform test crashes systematically on my machine under Python 3.5 but not under Python 2.7

After looking at it, it seems ocl.platform[0].device[0] is the Intel HD Graphics 5500 and ocl.platform[0].device[1] is the i5-5200 CPU.

The test passes if instead of using an image 3000x3000 one uses a 1024x1024 image.

vasole:
Closes #943


t20100:
Closed as already solved by #946

Issue #944, opened 2017-07-10
[opencl] Clean sift.test.test_transform


pierrepaleo:
The `test_transform` module should be cleaned, and involve images of moderate size (not 3k x 3k).

vasole:
Duplicate of #943 ?

vasole:
OK, I see, your issue goes beyond just the fix and advocates code cleaning.

I have just proposed PR #945 that just fixes the image size but it does not clean anything.

pierrepaleo:
Addressed by  #946.

PR #946, opened 2017-07-10
[sift] transform module and test

pierrepaleo:
This PR aims at addressing #944 and #943. 
It also makes the fix proposed in #945.


kif:
LGTM

PR #965, opened 2017-07-13
PrintPreview

PiRK:
This PR implements a print preview dialog and an associated plot action.

The source is PyMca. I also looked into Qub, but their print preview handles only raster images, without SVG capabilities.

TODO:

- [x] dialog
- [x] action or toolbar


PiRK:
TODO: fix centering of comments in `addSvgItem`

PiRK:
Question : should the print preview action remain in its own module (currently silx.gui.plot.actions.printpreview), or moved to silx.gui.plot.actions.io together with `PrintAction`, or should we have a common module for both print actions outside of io, or should io become a package including print preview?

If I add print preview to actions.io, we are getting close to 1000 lines. Moreover, the print preview action relies on dedicated widgets, whereas the other io actions are mostly widget less, so for me it makes sense to give it a separate module..

PiRK:
I transmogrified the actions into tool buttons, to separate the geometry setting from the print preview.

t20100:
`PrintPreviewDialog` has some attributes with dostring, Are they part of the API?
If so, it would be better to have some methods to access them to follow Qt-style API.

vasole:
In terms of functionality it seems to be fine. 

@t20100 I let you decide concerning the API to approve the merge.

t20100:
I'll see that with @PiRK when he's back.

t20100:
LGTM, to merge when CI is happy.

PR #975, opened 2017-07-19
[plot3d] Show/Hide bounding box and orientation indicator

t20100:
This PR adds API to ScalarFieldView and Plot3dWidget and checkbox in the ParamTree -> Style to:
- Show/Hide the bounding box, its axes and grid
- Show/Hide the top-right orientation indicator


PiRK:
LGTM

Issue #760, opened 2017-04-25
Allow context menus for charts


rozyczko:
Enhancement request.

It would be beneficial to have a possibility of adding a custom context (right mouse click) menu to the charts.
There is a lot of operations which can be added to a plot, which shouldn't necessarily go in the toolbar, so context menus are ideal.

vallsv:
That's something planed at some time. A refactoring was done for the coming 0.5 which allow us to implement this kind of stuffs. But yet it is not planed.

vasole:
Yes, it is something planned. At this point the right click is set to a zoom back but it should be possible to overwrite that. For instance, assuming that should be the behavior if no context menu is applied or if the context menu only action is just the zoom back.

t20100:
PR #976 removes right-click zoom interaction from the plot.
This allows to add context menus to the plot through QWidget API as illustrated in the sample code provided in PR #976 (https://github.com/t20100/silx/blob/bf45c30e497559dae77c923d46b27e33a2354223/examples/plotContextMenu.py).

@rozyczko, does that fit your needs?

rozyczko:
@t20100 
Tested the example script on plot-zoom branch.
Looks good and is exactly what we need.
Once merged, this issue can be closed.
Thanks

vasole:
Closed by #976

PR #976, opened 2017-07-19
[plot] Remove zoom-back from right click + add ZoomBackAction

t20100:
This PR:
- removes zoom-back control on right click from zoom interaction.
- adds a `ZoomBackAction` PlotAction.
- adds a plotContextMenu.py example adding a context menu to a `PlotWidget`.

To do so, the zoom history is now available from the `PlotWidget` as a separated `LimitsHistory` object accessible through `PlotWidget.getLimitsHistory`.

There is still a right click interaction for closing polygon with the draw polygon interaction mode.

Can we consider this PR is closing issue #760 ?


vasole:
Should not we add an example of Plot Context menu calling one of the already existing and predefined actions to consider #760 closed?

For instance, an action "Toggle Crosshair" or "Toglle Logy" on a PlotWidget (therefore without icons)

vasole:
I have just seen the example is already there.

t20100:
This PR is not enough for proper plot context menu implementation.
In the example, the context menu is attached to the PlotWidget which is a QMainWindow.... So the context menu is also defined over toolbars and dock widgets.

t20100:
Example updated to add context menu to plot area (i.e., plot backend widget) rather than PlotWidget.
This avoids to have the context menu over dock widgets included e.g., in a PlotWindow...
It also provides the position in plot area widget which can be converted to graph coords with `PlotWidget.pixelToData`.

vasole:
The example works but the functionality of zoom back is lost in all the other examples.

Can't we use setContextMenuPolicy to keep or not the legacy behavior? 

From the Qt doc:

customContextMenuRequested is emitted when the widget's contextMenuPolicy is Qt::CustomContextMenu, and the user has requested a context menu on the widget

t20100:
This would make a difference of behavior of the right click for the user between having a context menu or not.

I proposed to restore zoom back on Ctrl+left click (which changes the current behavior of right click and the way to do the zoom-back).
@vasole prefers a default context menu with zoom-back as first action to keep some backward-compatibility with the current behavior.

To use Qt API instead of providing our own for the context menu, we can set a default context menu with zoom-back (in the backend widget) that can be disabled with Qt.NoContextMenu or overridden with Qt.CustomContextMenu (but not extended).

vasole:
To be able to revert zoom with a right click has a long tradition at the ESRF dating from well before PyMca. Newplot already provided zoom control via the context menu. We already talking about 17 years. To go beyond that, I should remember what SpecPlot was doing and I do not remeber it. Considering that NewPlot was following SpecPlot I would not be surprised to be talking about even longer time.

I would certainly prefer to have a default context menu with zoom back as first (and perhaps only) that can be disabled with Qt.NoContextMenu or overridden with Qt.CustomContextMenu (but not extended).

t20100:
I implemented a default context menu containing only 'zoom back'.
This context menu can be overridden with a custom context menu as in the example or removed (NoContextMenu policy).
The not so obvious part for this for users is that it has to be done on the backend idget and not on the PlotWidget.

In the polygon drawing interaction mode, I removed the close polygon on right click which is "hidden" by the context menu anyway. This forces to close the polygon by clicking on the first point.
An alternative would be to disable context menu while drawing.

PR #997, opened 2017-07-27
Read from internal framebuffer rather than from widget framebuffer

### LINKED ISSUE #996: [plot gl] Save snapshot as PNG with OpenGL backend issue with PyQt4 on Debian 8
Get image from internal off-screen framebuffer rather than screen framebuffer as in plot3d.

t20100:
Still need to be tested on macOS and  windows.

closes #996


t20100:
Tested on macOS and Windows, ready to merge

vallsv:
LGTM

PR #999, opened 2017-07-31
Add None as a possible colormap. This wont draw the ticks, min, max o…

### LINKED ISSUE #989: [specfile] Change license to MIT

payno:
…r gradient but only the the gradient boundary (in order for the user to know the colormap is here.). This is used especially for RGB(A) image. This /close #989


vallsv:
LGTM

Issue #995, opened 2017-07-27
[silx view / spech5]  Very slow access to mesh scans with mca data


vasole:
Tried with a map obtained as a mesh scan containing mca data and it is very slow.

It seems something needs to be optimized. It looks as if the GUI would be trying to access the data multiple times because with the example file a single readout of the whole file is of the order of 10 seconds but the GUI is much slower than that.

Probably some cache at the spech5 would already speed up things.

PiRK:
There seems to be a large time difference when accessing a SPEC file directly in python compared to accessing it in the Qt loop. The former access is 5 times faster.

Test:

```
from silx.io import spech5
import time

# 199MB file
with spech5.SpecH5("../data/mesh_and_mca.dat") as h5f:
    start = time.time()
    # builds a (78026, 1024) array
    a = h5f["1.1/measurement/mca_0/data"]
    end = time.time()
    print("Simple access in python ", end - start)

from silx.gui import qt

class Dummy(qt.QPushButton):
    def __init__(self):
        qt.QPushButton.__init__(self)
        self.clicked.connect(self.timeit)

    def timeit(self):
        with spech5.SpecH5("../data/mesh_and_mca.dat") as h5f:
            start = time.time()
            a = f["1.1/measurement/mca_0/data"]
            end = time.time()
            print("Access in Qt loop ", end - start)

app = qt.QApplication([])

d = Dummy()
d.show()

app.exec_()
```

Result:
```
$ python timespech5.py 
('Simple access in python ', 10.502830028533936)
('Access in Qt loop ', 51.92558002471924)
```

PiRK:
This is a simpler script to replicate the issue::
```
from silx.io import spech5
import time
from PyQt4.QtGui import QApplication

with spech5.SpecH5("../data/mesh_and_mca.dat") as f:
    start = time.time()
    a = f["1.1/measurement/mca_0/data"]   # builds a (78026, 1024) array of float64
    end = time.time()
    print("Simple access in python ", end - start)

app = QApplication([])

with spech5.SpecH5("../data/mesh_and_mca.dat") as f:  
    start = time.time()
    a = f["1.1/measurement/mca_0/data"] 
    end = time.time()
    print("Access after initializing QApplication ", end - start)
```

All accesses after `app = QApplication([])`  take 5 times longer than the ones before.

These are the unsuccessful optimization attempts, so far:
 - releasing the GIL in `specfile.Specfile.get_mca` when calling the C library function `SfGetMca()`
[...]

PR #1004, opened 2017-08-01
Improve computation time of SpecH5

vallsv:
This PR will improve the computation time of SpecH5 in the context of the Hdf5Tree.

- SpecH5 have enough information to return request using `getlink` and `getclass` without loading the data at this level. This patch takes care of it.
- SpecH5 will also cache datasets per files
- Links to the same dataset will only be computed one time


PiRK:
I've added additional improvements implemented while troubleshooting #995. But the issue with the Qt thread slowing down `spech5` is still to be solved.

vallsv:
LGTM

vallsv:
Sorry. Remember me the behaviour of the links?
```
f[.../data]
f[.../info/data] # is it already cached?
```
If you already know the way to convert ".../info/data" -> ".../data" it would be easy to cache the links with a small patch on `_get_full_key`.

Issue #87, opened 2016-05-25
[Plot] Profile, allow to choose sum or average


vasole:
Allow an icon toggling (like with aspect ratio) associated to the width (for instance a "w" and a "w" with a horizontal bar on top) in order to allow the user to choose if the average profile or the sum is desired.

t20100:
closed by #1929

PR #1023, opened 2017-08-08
[Profile] add sum option

PiRK:
Addresses #87. Adds a toolbutton to select "sum" or "mean" method for computing profile.

In "sum" mode, when linewidth > 1, the parts of the profile close to the image edge that do not have a full coverage are currently ignored. So the summed profile can be slightly shorter than the averaged profile.

Maybe we should also clip the averaged profile to the same length, for a better consistency between the two modes. An average profile curve computed on a lower number of samples might look noisier than the middle of the curve.


PiRK:
After discussing with @t20100, the test for full coverage should not be done in bilinear.pyx, but rather when drawing the profile.

PiRK:
This will have to be done in a better way: the profile line needs to stop at the image edge, to indicate that we don't calculate the sum when the profile area gets out of the image's limit. Then the bilinear interpolation will receive `src` and `dest` coordinates inside the image limit and not need to check the coverage itself.

PR #1028, opened 2017-08-09
Widget screenshots in documentation

### LINKED ISSUE #1027: Screenshots visibility in doc
The screenshots of the widgets are currently hard to find in our documentation. There is currently no standardization regarding their location. 

Here are a few suggestions on how to improve this:

- We should have one "official" screenshot per public widget, with a file name identical to the widget's class name. Some widgets can have secondary screenshots to illustrate their various states..
- The official screenshot should be updated whenever the widget is updated in a visible manner.
- The screenshot should be displayed at the class level documentation, possibly included in the class docstring (like it's done in https://github.com/silx-kit/silx/blob/master/silx/gui/data/DataViewer.py#L47, for instance)
- We should have a single "Widgets gallery" page for the entire `silx.gui` package, with one subsection per subpackage (`plot`, `data`...), displaying the same official screenshot next to a link to the class documentation. Or else we could put one gallery per subpackage, like it's already done in the `silx.gui.plot` documentation page. Or both (but trying not to duplicate images or rst code is important).
.

PiRK:
Closes #1027 

This PR ensures each widget has an associated screenshot, displayed in a gallery (in `modules/gui/index.rst`) and in the class docstring. 

`plot` and `plot3d` widgets have additional galleries at the module documentation level, so for the moment I did not add the screenshots to each class docstring.

To be discussed:
- should we also add a gallery of screenshots of higher level applications using silx widgets?
- is there a way to avoid code duplication for the `plot` and `plot3d` galleries? I could not find a way to use `.. include::`  because the relative paths of the images are not the same in `modules/gui` and `modules/gui/plot`


vallsv:
I think it would be better to avoid to break the overview of the API page.

Then i think it would be better to
1. Remove 
```
+Sub-packages and modules
+------------------------
```
2. Put the gallery as a main entry of the doc
3. Remove the include of the gallery from `modules/gui/index`, optionally we can create a link to the gallery, there is no problem.

What do you think?

vallsv:
LGTM

PR #1029, opened 2017-08-09
Forward  projector in OpenCL

pierrepaleo:
This PR adds a forward tomographic projector ("Radon transform") in OpenCL. 
The projector adapted from the PyHST2 code (it uses the [Joseph interpolation scheme](http://ieeexplore.ieee.org/abstract/document/4307572)).

It works on both GPU and CPU, although Mac OS is not supported (due to the restriction of workgroup size to 1). Textures are used when available.


payno:
LGTM

PR #1036, opened 2017-08-16
[debian] Be able to remove sift from the library

### LINKED ISSUE #1014: [legal] could you make the build/test process work without the sift directory
Hello, as explained in a private email, it would be great if we could remove sift for the official Debian package.

So the setup.py and the test should be adapted in order to allow a simple

rm -rf silx/opencl/sift

thanks a lot

vallsv:
Patch from Frederic-Emanuel Picca from Soleil.

Closes #1014


vallsv:
I think, if sift is not packaged, it would be better to skip tests instead of removing it. I can fix that.

vallsv:
- Rework the test to skip it if the directory do not exists (to avoid catching useful exception)

vallsv:
Tests availability is now done on existence of the directory.

PR #1048, opened 2017-08-22
[spech5] Major refactoring

### LINKED ISSUE #608: [fabioh5] Implement visit and visititem
As discussed on 2017-02-01 at the Silx meeting

- It's part of the h5py API
- It is used by the converter of @PiRK

PiRK:
This PR cleans up the `spech5` module. It uses the same base classes as `fabioh5` now. Future maintenance will be much easier now.

As a result, there is a new module `commonh5`, where I moved the `fabioh5` classes that were also relevant to `spech5`.

A secondary goal for this PR was to implement generic  `visit` and `visititems` methods, working for both `spech5` and  `fabioh5` (closes #608).

CHANGELOG NOTE: *`Group.keys()`, `Group.values()`, `Group.items()` now return lists in PY2 and iterators in PY3*, exactly like `h5py.Group`. Previously, `SpecH5Group` always returned lists and `fabioh5.Group` always returned iterators.. 

Next step, after this is merged, will be to create a generic conversion to HDF5 tool based on `spectoh5`.


vallsv:
LGTM

Issue #1039, opened 2017-08-16
Problem with GUI test on appveyor and travis


vallsv:
I guess system pops can explain things...

![appveyor - ci 4](https://user-images.githubusercontent.com/7579321/29354793-d9e95792-826e-11e7-8899-3d896159dce8.png)
![appveyor - ci 3](https://user-images.githubusercontent.com/7579321/29354797-de009318-826e-11e7-8b43-ec6ecb90f033.png)

vallsv:
Well, it looks to be not the case on appveyor: 
![appveyor - ci 5](https://user-images.githubusercontent.com/7579321/29356797-b6861b76-8275-11e7-911a-836beac0a1d1.png)

t20100:
Did you tried the registry patch proposed in: https://github.com/appveyor/ci/issues/1715?

vallsv:
Yes, but it have to be applied before our scripts have hands on it, else it have no effects.

vallsv:
- I create a branch to play around that https://github.com/silx-kit/silx/pull/1024 (it should fix problems a have with the mouse, but not this one, it also introduce few other things we can cherry pick one of this days)
- And GUI tests was disabled by https://github.com/silx-kit/silx/pull/1050 while it is not fixed

vallsv:
And i have no feed back from Travis

vallsv:
Travis answer that they will check it is possible to disable the popup. But ask us if we can close it by our-self. I think it is not so easy while we never know when and when the dialog pop.

PR #1050, opened 2017-08-23
Do not test the GUI on Appveyor

vallsv:
- while there is a problem with appveyor


vallsv:
See https://github.com/silx-kit/silx/issues/1039 and https://github.com/appveyor/ci/issues/1715

PR #1051, opened 2017-08-23
[spech5] Lazy mca loading

PiRK:
These optimizations mostly solve #995.  

The previous PR #1048 already took care of caching the data, to ensure it is loaded only once. 
Now the MCA dataset is also lazy loaded, so the HDF5 tree will not become unresponsive when opening a group containing a heavy MCA dataset (unless we want to see the actual data).

The issue causing the loading to take 5 times longer than usual in a Qt application is still not fixed.


vallsv:
LGTM

PR #1053, opened 2017-08-24
[convert] New module and application

### LINKED ISSUE #609: [fabioh5] Test it with a generic "silx convert" to hdf5
As discussed on 2017-02-01 at the Silx meeting

- We have to take care of SoftLink
- @pierrepaleo can be interested by this feature (it's good to have a beta tester)
- We can extract from edf metadata few nexus-normalized detector metadata if expected

PiRK:
This PR creates a new module `silx.io.convert`, based on `silx.io.spectoh5`, to convert any h5py-like object to HDF5. 

`silx.io.spectoh5` is now deprecated, but should still work. It is a module that defines aliases to keep the `spectoh5.write_spec_to_h5` and the old `spectoh5.convert` functions working.

The PR also adds a new `silx convert` app (closes #609).


PiRK:
I will stop here with this PR, and I will open an issue to remind us that unittests are to be added for `silx.app.convert`. I quickly tested the application with a couple of data files (EDF and SPEC), and with a few combinations of all the possible options.

PiRK:
Before merging, maybe we should discuss whether or not we want to keep the advanced options (*chunks, compression, shuffle*). These options apply globally to all datasets.

I added them because I think they can be useful when converting simple and homogeneous data formats (a list of EDF images, for instance). But maybe they can cause issues for heterogeneous formats, like a SPEC file with a mixture of scalar positioners, 1D detector data and 2D mca data.

t20100:
After discussion with @vasole, it would be nice if compression is only for data.
Possible options:
- provide a minimum dataset size for which compression is applied;
- only apply compression in detectors group.

vallsv:
Just a non-blocking comment about `Hdf5Writer`. I don't see the point of choosing hard/soft links. Hard links are not links. Then is there a use case for using hard links? If no, it's always better to remove unused code.

vallsv:
My misunderstanding of hardlinks clarified by @vasole. Then it's fine like that.

vallsv:
Just a small comment. It would be good to write somewhere that the file was generated by silx-convert with the "silx version". This kind of information can be very useful at one point. Maybe Nexus files have a specific field for that?

PiRK:
I agree that the silx version would be a useful metadata for the output file.

Issue #379, opened 2016-10-24
[resources] Adding extra resource dirs.


dnaudet:
The resources (icons, ...) system functions (getIcon, ...) in silx are very practical, but they don't allow (AFAIK) for a user to add its own folder(s) to the lookup path, which means that a client project has to duplicate silx code into its own project to use these functionalities (well that's what I did anyway).

So maybe it would be interesting to allow the user to register a folder and a namespace (to avoid conflicts) to the list of paths into which silx looks for resources?
And then you could get files like resource_filename('myNs:foo/myFile').

vallsv:
Close by https://github.com/silx-kit/silx/pull/1054

PR #1054, opened 2017-08-24
Prefix for resources

vallsv:
- Fix timeout on test of external resources
- Add `is_dir` and `list_dir`
- Allow to specify a prefix to resource names: "silx:foo.png". It allow to use the same framework for other projects. It will reduce a lot of code duplication, especially on the GUI part.
- Move resources used in tests on a `silx.test.resources` package

```
# For example we can do that in PyFAI

import silx.resources
silx.resources.register_resource_directory("pyfai", pyFAI.resources)
# The package "pyFAI.resources" must have the same structure of the one from silx.

filename = silx.resources.resource_filename("pyfai:calibrant/my_calibrant.C")

import silx.opencl.utils
filename = silx.opencl.utils.get_cl_file("pyfai:my_program")

import silx.gui.icons
icon = silx.gui.icons.getIcons("pyfai:my_funky_icon")
```


t20100:
Closes #379

t20100:
Allow `register_resource_directory` to take either a package name or a path (in case resources are not stored in a python package, e.g., for Debian).

Question to discuss: For external resources, `getQIcon` is adding `gui/icons/` to the icon path. I'd prefer to use absolute path for external resources rather than forcing users to store them in a gui/icons/ directory. @vallsv has the opposite opinion. Any other opinions?

t20100:
For the test of resources, it would be better to avoid having data files outside `silx.resources` to ease extraction of data from package.
An option is to create a temporary folder with some files to test it, but it prevents testing pkg_resources...

vallsv:
- registration function was reworked to be able to force a directory
- resources for tests was removed (and temp file are not used)

kif:
TV: namespace + path VV suggest 'namespace:path', 
    TV agrees with a single parameter, while a 2-parameter field would look easier in a first instance 
    but more complicated in the long term as there will always be 2 parameters to forward in every function to be called. ('gui/icons/icon.png', project='silx')
    This field should be unique, so a single string makes more sense.
    examples: "gui/icons/icon.png" or "silx:gui/icons/icon.png" or "pyFAI:gui/icon/another.png"
    We agreed that if there is a prefix before colon, the path provided after should be complete related to the project's resource root.
    i.e. "silx:logo" will NOT resolve logo into gui/icon/logo.png.

PR #1060, opened 2017-08-28
[silx view] Fix physical/access path

### LINKED ISSUE #1059: [silx view] Displayed physical links are not accurate
I was based on the fact that it was not possible to reach the real path used to access to dataset from this dataset, but it looks like it is not possible to access to the real location of the data.

I am confused...

Maybe i was always used a wrong definition...

```
In [1]: import h5py

In [2]: f = h5py.File("../data/hdf5/buggy/pierre_bad_links.h5")

In [3]: a = "/1.1/measurement/mca_0/data"

In [4]: b = "/1.1/instrument/mca_0/data"

In [6]: f.get(a, getlink=True).path == b
Out[6]: True

In [7]: f.get(a).name != a
Out[7]: False

In [8]: f.get(a).name
Out[8]: u'/1.1/measurement/mca_0/data'

In [6]: f.get(a, getlink=True).path == b
Out[6]: True

In [7]: f.get(a).name != a
Out[7]: False

In [9]: h5py.__version__
Out[9]: '2.7.0'
```

`a` is a link to `b`. Reaching `a` should return the dataset pointed by `b`. But it is not the case here.

vallsv:
- Fix the way node paths are computed with our commonh5 objects (now it match h5py)
- Fix the physical path, which is the hdf5 path when every intermediate links are removed
- A lot of test cases to be sure it is h5py compliant, and that we display the right thing

The screenshot shows that the path `/soft_link_to_group/dataset` is a link to `/group/dataset` and it is not trivial. Previous computation was based on wrong assertion.

Closes #1059

screenshot from 2017-08-28 17 23 09

### DIFF
--- a/silx/gui/hdf5/_utils.py
+++ b/silx/gui/hdf5/_utils.py
@@ -144,15 +144,44 @@ class H5Node(object):
         return object.__getattribute__(self.__h5py_object, name)
 
     def __get_target(self, obj):
-        basename = obj.name.split("/")[-1]
-        link = obj.parent.get(basename, getlink=True)
-        if isinstance(link, h5py.ExternalLink):
-            external_obj = obj.parent.get(self.basename)
-            return self.__get_target(external_obj)
-        elif isinstance(link, (h5py.SoftLink)):
-            return self.__get_target(obj.parent[link.path])
-        else:
+        """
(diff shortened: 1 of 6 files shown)


t20100:
LGTM

PR #1072, opened 2017-09-01
[test] Robustness of GUI tests

vallsv:
Just a test without checking Plot events


vallsv:
No not waiting for Plot events looks to be fine.

PR #1081, opened 2017-09-04
Fix plot errors with latest version of matplotlib

t20100:
This PR should fix errors spotted by CI that happens with new version of matplotlib: v2.1.0rc1:

- silx.gui.plot.test.testPlotWidget.TestPlotCurveLog.testPlotCurveErrorLogXY
- silx.gui.plot.test.testPlotWidget.TestPlotCurveLog.testPlotCurveToggleLog

It does not change anything about PySide that also seems to be broken on CI...

### DIFF
--- a/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/silx/gui/plot/backends/BackendMatplotlib.py
@@ -143,7 +143,7 @@ class BackendMatplotlib(BackendBase.BackendBase):
                 errorbarColor = color
 
             # On Debian 7 at least, Nx1 array yerr does not seems supported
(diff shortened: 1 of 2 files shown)


vallsv:
Can be merged if no more test failure than before :-D

vasole:
Do the tests fail if we use the actual base clase init method insead of super?

vallsv:
The issue you mentioned (with PySide) is not relative to this one.
But about PySide it is also what we think about, but we are still trying to understand what can explain this problem.

vallsv:
LGTM
Tests succeeded, but the PySide issue, as expected.

Issue #1038, opened 2017-08-16
Could you provide a .desktop file for the silx application


picca:
It would be nice to have a silx-view.desktop file in order to access the viewer from the menues.

A nice icon should be great also ;)

vallsv:
Sounds interesting. Here is some entries https://stackoverflow.com/questions/25284879/install-desktop-file-with-setup-py

t20100:
I created a .desktop file + 48x48 icon in PR #1086 
@picca is that enough for you? Do you know projects that integrate the copy of those files to their setup.py?

Not sure how to have python installing it at the right place (handling --user and virtualenv).
Packages to check: https://github.com/welshjf/install_freedesktop, https://launchpad.net/python-distutils-extra (there's a Debian package).

t20100:
Closed by #1086

PR #1086, opened 2017-09-05
Add .desktop file and silx icon for silx viewer

t20100:
This PR adds a .desktop file for Linux desktop menu integration and an icon (Related to #1038).


t20100:
I've checked the files by manually copying them to ~/.local/share/applications/ and ~/.local/share/icons/

picca:
Hello,

look at here :)

https://stackoverflow.com/questions/25284879/install-desktop-file-with-setup-py

vallsv:
BTW i think it would be better to move the `desktop` directory inside our `package` directory. And they recommend to follow the reverse DNS convention for the file name, then something like `org.silx.SilxViewer.desktop`. What do you think?

picca:
obey the policy ;)

t20100:
Moved and renamed.
I will not do the installation of .desktop from the setup.py, the solution proposed in https://stackoverflow.com/questions/25284879/install-desktop-file-with-setup-py does not seem to work with `--user` (It should work if you install the package with pip as root but it is pointless...).
Instead I'm updating our unofficial debian packaging.

t20100:
installation OK on debian8: ready for review.

t20100:
closes #1038

vallsv:
As we have `org.silx` for the app, what about doing the same for the icon? I am always afraid of name collision.

t20100:
There seems to have no icon (at least in Debian gnome icons package) using reverse DNS naming...

vallsv:
Then it's fine for me.

vallsv:
LGTM,
No luck, people ony use rnice without desktop :-)

PR #1096, opened 2017-09-08
[plot3d] Rework interaction to only use left button

### LINKED ISSUE #957: [plot3d] Rework pan/rotate interactions
Now the zoom/pan/rotate interaction uses both mouse buttons and the wheel.
This is not convenient on macOS laptops where the right click is emulated with Ctrl+click and it prevents adding context menu on right click.

The idea is to provide 2 separate interaction modes for pan and rotate which only uses the mouse left button.

t20100:
This PR changes the implementation of Plot3DWidget and ScalarFieldView to free the right click.
It adds buttons for interaction modes (rotate, pan and pan plane for ScalarFieldView) in a toolbar.

To do so, it :
- Moves silx.gui.plot3d.Plot3DActions to silx.gui.plot3d.actions as in plot
- Adds Plot3DWidget.setInteractiveMode
- Adds QAction to set interactive mode
- Reworks ScalarFieldView to use the previous changes

To test: examples/viewer3DVolume.py (new icons on the top-left corner)

Closes #957


vallsv:
LGTM, waiting for tests on Travis

PR #1099, opened 2017-09-08
Fix setup.py for sift installation

### LINKED ISSUE #1098: sift module not installed
On debian 8 with python2, sift module is not installed due to evaluation of the wrong path in silx/opencl/setup.py.
Using the absolute path of `__file__` should solve the issue.

t20100:
sift was no more installed...
closes #1098


vallsv:
LGTM

PR #1100, opened 2017-09-08
[plot3d] Add a drop-down list tool button to reset viewpoint

### LINKED ISSUE #1097: [plot3d] Add a drop-down list to set viewpoint
This is more concise than the current toolbar

t20100:
Merge PR #1096 first !

This PR adds a tool button with a drop-down list of pre-defined viewpoint positions to reset the point of view and use it instead of a bunch of QAction in a toolbar.

closes #1097


vallsv:
Here is a commit to display icons on the menu, if you like https://github.com/t20100/silx/pull/10

vallsv:
LGTM

PR #1112, opened 2017-09-11
[plot3d] Add support of any colormap

### LINKED ISSUE #956: [plot3d] Add support of any colormap
Replace shader implementation of colormap by texture based colormap as it is done in plot OpenGL backend.

t20100:
This PR adds the support of virtually any colormap (rather than a subset).
It adds the new matplotlib colormap (viridis and co) to the colormap options in the plot3d parameters tree.
Closes #956


payno:
LGTM.
Just one simple question, the 'GL2' keyword is for compatibility until openGL 2 I think this is what I recall but just to be sure.

t20100:
I initially named the rendering methods `*GL2` as it is using OpenGL2, in case I need to add a separate set of rendering methods for other version of OpenGL.

PR #1115, opened 2017-09-12
Fix issues with object caching on PyOpenCL > 2015.2

kif:
Only for classes deriving from "OpenclProcessing"


kif:
I decided to leave the refactoring of  sift in another bug #1133

pierrepaleo:
Looks good, although the change of caching behavior (formerly enabled by default) in pyopencl seems weird to me... do we have a "public notice" from Andreas Klöckner on this matter (apart from the pyopencl mailing list) ?

PR #1122, opened 2017-09-12
[fix] add check on range_histo to make sure no numpy.nan and/or numpy.inf a…

payno:
…re in the range


t20100:
The PR includes modifications to `silx/math/fit/filters.c` and `silx/io/specfile.c` while the `.pyx` files were not changed.
Could you remove those files from the PR?

Otherwise LGTM.

payno:
my bad. Done

PR #1129, opened 2017-09-12
[FBP] PEP8

payno:


t20100:
Could you add the example to the example gallery in the doc (https://github.com/silx-kit/silx/blob/master/doc/source/sample_code/index.rst) ?

t20100:
And maybe use a longer name for the example filename.

payno:
waiting for the forward projector to be merged and we will add utils to generate a Shepp-Logan phantom

payno:
Now the backprojection example is free from external resources. 
It deal with it using the forward projection.
Question : 
should we also include an example for the projection ?
I think the current example is enough.

PR #1140, opened 2017-09-13
[io/gui] Support for h5pyd

vallsv:
Here is some tests to check the use of `h5pyd` in `silx.io`.

This PR still contains ugly hacks.

screenshot from 2017-09-13 16 49 44


vallsv:
Here is an exception occurred while accessing to the default HDF5 provided by h5serv.

```
<type 'exceptions.ValueError'> Big-endian buffer not supported on little-endian compiler   File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/ColorBar.py", line 213, in _activeImageChanged
    data=image)
  File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/ColorBar.py", line 159, in setColormap
    data=data)
  File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/ColorBar.py", line 389, in setColormap
    self.colorScale.setColormap(colormap, data)
  File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/ColorBar.py", line 518, in setColormap
    self.vmin, self.vmax = self._colormap.getColormapRange(data=data)
  File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/Colormap.py", line 247, in getColormapRange
    min_, max_ = min_max(data, min_positive=False, finite=True)
  File "silx/math/combo.pyx", line 316, in silx.math.combo.min_max (silx/math/combo.c:14106)
    return _finite_min_max(data, min_positive)
  File "silx/math/combo.pyx", line 208, in silx.math.combo._finite_min_max (silx/math/combo.c:12566)
    def _finite_min_max(cython.floating[::1] data, bint min_positive=False):
```

t20100:
A few quick comments:
- print leftover in `silx/gui/hdf5/Hdf5Item.py`
- Add h5pyd to requirements.txt and to optional dependencies in the optional dependencies in the installation doc

t20100:
If we have to deal with the endianness of datasets everywhere in the code, there's a lot of fun coming up!

vasole:
Reading HDF5 is supposed to handle endianness transparently. I guess a different history is to have a function written by us that receives an HDF5 dataset when in fact is designed to use an array.

vallsv:
I guess the problem with endienness is only for our cython code.

vasole:
Looks good for the time being. Just fix the conflicts and the PR can be merged.

Concerning the errors related to external links, we'll see if that is actually needed in the feature as it is not clear the problem is at our side.

vallsv:
It's now rebased.

PR #1142, opened 2017-09-13
OpenCL: LinAlg operators

pierrepaleo:
This PR adds a LinAlg class of OpenclProcessing. The aim is to perform operations on device while supporting src/dst being on host as well.
For example, the current operators (gradient, divergence) support images being numpy.ndarray, pyopencl.Array or pyopencl.Buffer; and the destination can also be one of these. 

The branch is derived from pierrepaleo/silx:projector because I am going to modify the tomography operators.


pierrepaleo:
Two iterative reconstruction algorithms are now available in silx: SIRT (Simultaneous Iterative Reconstruction Technique) and a Total Variation regularization algorithm.

The SIRT implementation is compatible with ASTRA results, although it might be good to have a unit test. Also, I wonder if the SIRT and TV examples be merged in one single file ?

pierrepaleo:
Working on an issue when using CPU

pierrepaleo:
The iterative reconstruction is working on GPU. 
On CPU, we have troubles with the Xeon E3-1245 v5 when using the AMD implementation (it works fine with others CPUs like E5-2643 v3, E5-1650 v4). The Intel implementation seems to be working regardless of the (Intel) CPU.

payno:
LGTM. Only doc link to the doc is missing and exposition in the image module. 
I will do it next week

Issue #1127, opened 2017-09-12
[convert] Multiple edf files vs. multi-frame stack


PiRK:
`silx convert *edf` will by default write a scan_0 group and merge all data into it. 

Maybe option `--create-root-group` should be set by default for multiple EDF files ?

Anyway, `silx convert *edf` will probably most often be used to create a multi-frame file. Should we try to detect cases when the edf files are a stack of images, and do a `cat *edf > multiframe.edf` prior to running convert? Or add an explicit `--edf-stack` option to do this?

vasole:
My experience trying that command under windows is an error message even after solving #1118 

I would say that unless directives are provided indicating a merge, each file should correspond to an individual entry in the final file. Probably not what the user wants but it is sure to work.

That should be already the current implementation so that at least it does not crash.
We could offer to detect the pattern ROOT_SCANNUMBER_DETECTORNUMBER_INDEXNUMBER.edf  that should be easy to couple with any SPEC file provided the user requests for it (or a SPEC file is provided)

vasole:
Let me pout it clear. **We have to** provide a tool able to take a set of EDF (or any other image acquisition format files) and associate them to a particular scan of a SPEC file. In other words, we have to be able to provide a merging tool. So, I am not asking for a lot of cases. Just two:

- The user has explicitely requested to take all input files as members of a stack. Therefore all the frames have to have the same shape.

- The user has followed the convention most widely accepted at the ESRF concerning the naming.

The simplest solution is of course to implement just the first one giving the **_possibility to specify the output path_** in the HDF5 file. That would allow to make the things simpler by taking a two steps approach. For instance, the SPEC file is converted and then the associated images are addded to a particular entry explicitely asking to be merged and under what path. A URI would do the job. So basically the option:

- The user has explicitely requested to take all input files as members of a stack. Therefore all the frames have to have the same shape.

Plus the possibility to specify -o outputfile.h5::/path/to/place would be enough.

The example below looks to me pretty obvious to understand:

silx convert --stack *.edf -o destination_file:::/entry_0/instrument/detector_x

PR #1135, opened 2017-09-12
[silx convert] unittests

PiRK:
Closes #1058

PR #1148, opened 2017-09-14
[convert] Improve optional arguments

### LINKED ISSUE #1104: [silx convert] Do not compress everything in the output hdf5
After discussion with @vasole, it would be nice if compression is only for data.
Possible options:

-  provide a minimum dataset size for which compression is applied;
-  only apply compression in detectors group.

(From discussion in PR #1053)

### LINKED ISSUE #1105: [silx convert] Add silx version information in output hdf5 file
From discussion in PR #1053:

> It would be good to write somewhere that the file was generated by silx-convert with the "silx version". This kind of information can be very useful at one point. Maybe Nexus files have a specific field for that?

PiRK:
This PR implements several changes for convert:
- set a minimum dataset size for applying compression and chunking (closes #1104)
- create root group by default, provide an option to disable it if the users really wants to merge files (temporary fix for #1127 until we deal with stacks of files)
- write a `silx_version` dataset (closes #1105)
- implement URI, remove `--hdf5-path` option (mentioned in #1127)


vasole:
I find the 1 MByte limit safe but too small as there are file systems where the minimum block that can be read or written is already 4 MBytes. 

The performance issues due to compression are when the chunk size is beyond the cache used for chunking, The default cache chunk size is 1 MByte by default. What we have to make sure/control is that the chunk size is not bigger than the cache chunk size.

vasole:
After discussion, it seems the simplest solution is:

- Base compression or not of the dataset on the number of elements of the dataset (i.e. medipix images are small and would not be compressed based on size). A default limit of 500 elements seems convenient and user should be able to overwrite the default.

- Perform a check about the size (in bytes) of the chunk is below the currently used cache size.  See slides 30-37 of: 

https://support.hdfgroup.org/HDF5/doc/Advanced/Chunking/Chunking_Tutorial_EOS13_2009.pdf

PiRK:
To be merged *after* #1135

vallsv:
LGTM

PR #1149, opened 2017-09-14
[plot] ComplexImageView widget

### LINKED ISSUE #277: [Plot] Display of complex data
Code for RGBA lin/log display using standard circular table:

```

def phase2rgb(s):
    """
    Crates RGB image with colour-coded phase.
    """
    ph = np.angle(s)
    t = pi / 3
    nx, ny = s.shape
    rgba = np.zeros((nx, ny, 4))
    rgba[:, :, 0] = (ph < t) \* (ph > -t) + (ph > t) \* (ph < 2 \* t) \* (2 \* t - ph) / t + (ph > -2 \* t) \* (ph < -t) \* (ph + 2 \* t) / t
    rgba[:, :, 1] = (ph > t) + (ph < -2 \* t) \* (-2 \* t - ph) / t + (ph > 0) \* (ph < t) \* ph / t
    rgba[:, :, 2] = (ph < -t) + (ph > -t) \* (ph < 0) \* (-ph) / t + (ph > 2 \* t) \* (ph - 2 \* t) / t
    return rgba

def complex2rgbalog(s, amin=0.5, dlogs=2, smax=None):
    """
    Returns a 2D RGBA image with colour-coded phases and log10(amplitude) in brightness.
    Args:
        s: the 2D complex data array to be converted to RGB
        amin: the minimum value for the alpha channel
        dlogs: amplitude range displayed, in log10 units
        smax: if specified, all values above max will be displayed with an alpha value of 1
    Returns:
    """
    rgba = phase2rgb(s)
    sabs = np.abs(s)
    if smax is not None:
        sabs = smax * (sabs>smax) + sabs * (sabs<=smax)
    a = np.log10(sabs + 1e-20) 
    a -= a.max() - dlogs  # display dlogs orders of magnitude
    rgba[:, :, 3] = amin + a / dlogs * (1 - amin) * (a > 0)
    return rgba


def complex2rgbalin(s, gamma=1.0, smax=None):
    """
    Returns RGB image with with colour-coded phase and linear amplitude in brightness. Optional exponent gamma is applied to the amplitude
    """
    rgba = phase2rgb(s)
    a = np.abs(s)
    if smax is not None:
        a = smax \* (a>smax) + a \* (a<=smax)
    a /= a.max()
    rgba[:, :, 3] = a**gamma
    return rgba

def colorwheel(text_col='black', fs=16, ax=None):
    """
    Color wheel for phases in hsv colormap.
    Args:
        text_col: colour of text
        fs: fontsize in points
        fig: an instance of matplotlib.figure.Figure (for offline rendering) or None

    Returns:
        Nothing. Displays a colorwheel in the current or supplied figure.
    """
    xwheel = np.linspace(-1, 1, 100)
    ywheel = np.linspace(-1, 1, 100)[:, np.newaxis]
    rwheel = np.sqrt(xwheel ** 2 + ywheel ** 2)
    phiwheel = -np.arctan2(ywheel, xwheel)  # Need the - sign because imshow starts at (top,left)
    #  rhowheel=rwheel*np.exp(1j*phiwheel)
    rhowheel = 1 * np.exp(1j * phiwheel)
    if ax is None:
       ax = plt.gca()
    ax.set_axis_off()
    rgba = complex2rgbalin(rhowheel * (rwheel < 1))
    ax.imshow(rgba, aspect='equal')
    ax.text(1.1, 0.5, '$0$', fontsize=fs, horizontalalignment='center', verticalalignment='center',transform=ax.transAxes, color=text_col)
    ax.text(-.1, 0.5, '$\pi$', fontsize=fs, horizontalalignment='center', verticalalignment='center', transform=ax.transAxes, color=text_col)
```

t20100:
This PR adds a `ComplexImageView` widget to the plot widget zoo.

This widget is meant to display a complex image and allows to switch between different modes through a drop-down tool button: 
- amplitude
- phase
- real part
- imaginary part
- combined amplitude (brightness) + phase (color-coded) with either linear or log scaled amplitude. Credits goes to @vincefn for the color conversion.

closes #277 

TODO:
- [x] Add a few proxy to the plot API.
- [x] Documentation, screenshots.
- [x] Check what to use for profiles when displaying combined view.
- [x] Add to data viewer


vincefn:
This looks good ! Few remarks:
- It would be good to have the 'phase' view default to the hsv colormap, which is periodic (and identical(?) to the map used for RGBA)
- Is it possible to add a choice of the range in amplitude for the log(RGBA) view ?

vallsv:
What about  `ℝ` and `iℝ` for your icons? That's common math notations. `Im` confused me, cause i thought the first char was a `L`.

vincefn:
Yes it would be better to be consistent with the colormap, I have to check the choice I made in pynx.
Regarding the profile I think it would be more expected to plot the amplitude, even if the phase often holds more information. It's straightforward to switch to the phase display, which updates the plot, so should be fine.

t20100:
This PR also contains a few fixes with features not working with empty or only not finite images:
- Profile (https://github.com/silx-kit/silx/pull/1149/commits/81e752a25bbac9a5c671dff9f02bed97f9ca83bf)
- Colormap (https://github.com/silx-kit/silx/pull/1149/commits/a5936c204d3effab4d60078240e4d4de719dbb6f, https://github.com/silx-kit/silx/pull/1149/commits/d6f9b8354f07673f6e5a23cbeb226b938dd8e037)
- Mask (https://github.com/silx-kit/silx/pull/1149/commits/129f8cc3c15df91b1797b9af16ef76063f38c83c)

Regarding colormap management:
- HSV is the default for the phase (but it can be changed),
- A+phase and log(A)+phase colormaps uses HSV + alpha channel (cannot be modified),
- Other modes (amplitude, real, imaginary) each have its own colormap.

Still missing: a way to provide the range in amplitude for log(A)+phase mode.

t20100:
I add a dialog to set the amplitude range accessible from the tool button when log(A)+phase mode is used.
The range is defined with max amplitude and displayed delta in log10 units.
@vincefn, is this OK, or a dialog to set min-max directly would be better?

Otherwise, I should have addressed all comments.

vincefn:
The dialog to set the delta in log10 units is perfect for me. Nice icons !
However I'm not seeing the expected change in  amplitude (alpha channel).

t20100:
Done. Alpha channel range was [0.5, 1], now using the full range.

vallsv:
LGTM

vallsv:
- It would be good to improve the status bar to display the complex value for the data if it is possible.
- I also think it would be better to avoid popup and to use a panel to configure things. And the way to access to this alpha popup is i think not the best.

But it is already fine as it is. Then it is merged.

vallsv:
We also have to talk to @PiRK cause now he is overriding the `_View2D` for PyMca, then maybe, there is things to talk around that.

vincefn:
It would indeed be great if this was automatically accessible (replace the normal Image view) in the silx view (DataViewer) when the data is complex.

vallsv:
That's already working like that, if you like to try. And it is now merged on the master.

vincefn:
I updated 1min early I guess - it's great !!
Now deployed in my /sware/exp/pynx/devel/bin/activate !

t20100:
@vallsv, yes the dialog is not very convenient and a `DockWidget` would certainly be better.
To show the complex value in the status bar would be good to, and to also show amplitude and phase curves in the profile... this means that all tools surrounding the plot supports complex data.

PR #1151, opened 2017-09-14
[plot] assert plot not none on the _BaseMasktoolsWidget

payno:
Warning : this bring to an API modification BaseMaskToolsDockWidget(self, parent=None, plot=None, name='Mask') -> BaseMaskToolsDockWidget(self, plot, parent=None, name='Mask')

### DIFF
--- a/silx/gui/plot/MaskToolsWidget.py
+++ b/silx/gui/plot/MaskToolsWidget.py
@@ -211,8 +211,9 @@ class MaskToolsWidget(BaseMaskToolsWidget):
 
     _maxLevelNumber = 255
 
-    def __init__(self, parent=None, plot=None):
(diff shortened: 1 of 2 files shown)


payno:
This modification is including a modification of the API BaseMaskToolsDockWidget(self, parent=None, plot=None, name='Mask') -> BaseMaskToolsDockWidget(self, plot, parent=None, name='Mask').
what is the procedure in this case: adding a warning in the documentation ? Something more ?

vasole:
As far as I know we had decided to have parent=None as the first argument of all the widgets for easy promotion in the designer. Why is the modification?

payno:
ok, my bad I though we wanted to have parent as the first parameter **with** a default value and not the first in any case.

In this case an assert with doc should do it.

PR #1164, opened 2017-09-18
Rework for h5py types

vallsv:
Folling the presentation of Gerd Heber on HDF5, it is more or less possible to display and store ASCII and UTF-8 properly using `h5py`.

This document http://docs.h5py.org/en/latest/special.html explains how to store and retrieve HDF5 datatypes which are not supported by numpy. Mostly enum, ref and UTF-8 variable-length.

The current master:
- Store UTF-8 as HDF5 ASCII type, which is very bad
- Tries to decode ASCII as UTF-8, which is also bad
- Do not support ref, enum, UTF-8 properly

This PR improve things:
- Fixes the formatter to display special types properly (enum, ref, strings, vlen)
- Fixes informations displayed in the tree (which can be tested using `/tmp_14_days/valls/all_types.h5`)
- Fixes informations displayed in the HDF5 info view (now the real HDF5 type is also displayed)
- Fixes views and intermediate structures to propagate h5py special types properly (h5py monkey-patch the numpy dtype, sometime it is lost)
- Provides an hexadecimal viewer for opaque types.

There is still work to do on fabioh5.

### DIFF
--- a/silx/gui/data/ArrayTableModel.py
+++ b/silx/gui/data/ArrayTableModel.py
@@ -34,7 +34,7 @@ from silx.gui.data.TextFormatter import TextFormatter
 
 __authors__ = ["V.A. Sole"]
 __license__ = "MIT"
-__date__ = "24/01/2017"
+__date__ = "26/09/2017"
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


vallsv:
I will postpone the last check to another PR. If it is possible to merge that, that will be easier.

PiRK:
LGTM. So many exotic types.

PR #1167, opened 2017-09-18
[treeview] Provide an API to select a single tree node item

vallsv:
This API allow to select an item of the Hdf5TreeView using an h5py node. This can be useful for dialog windows.

Feature requested by Alessandro Mirone.

- If the item is found, parent items are expended, and then the item is selected.
- If the item is not found, the selection do not change.
- A none argument allow to deselect everything


vallsv:
That's ready to merge

Issue #1158, opened 2017-09-17
[legal] do you have the licenses of all the resources icons ?


picca:
Everythings in the title :))

I try to use cme in order to generate and upgrade the copyright file. But I do not have the license of all theses svg, png files.

picca:
When I run

`cme upgrade dpkg-copyright`

I get this

```The following paths are missing information:
- doc/source/conf.py: missing copyright and license
- doc/source/img/silx.ico: missing copyright and license
- silx/gui/plot/MPLColormap.py: missing copyright and license
- silx/image/bilinear.c: missing copyright and license
- silx/image/shapes.c: missing copyright and license
- silx/image/sift.py: missing copyright and license
- silx/io/specfile/specfile.c: missing copyright and license
- silx/math/combo/combo.c: missing copyright and license
- silx/math/fit/filters/filters.c: missing copyright and license
- silx/math/fit/functions/functions.c: missing copyright and license
- silx/math/fit/peaks/peaks.c: missing copyright and license
- silx/math/histogramnd/chistogramnd.c: missing copyright and license
- silx/math/histogramnd/chistogramnd_lut.c: missing copyright and license
- silx/math/marchingcubes/marchingcubes.cpp: missing copyright and license
- silx/math/medianfilter/medianfilter.cpp: missing copyright and license
- silx/opencl/utils.py: missing copyright
You may want to add a line in debian/fill.copyright.blanks.yml
```

picca:
I think that some of the problem are due to licensecheck wwhcih does not regognize the license header.
Since it work with a bunch of files, it would be  nice to investigaet and understand why it fail sometimes.

vasole:
- The icons were generated at the ESRF so, they are under the same license as silx (MIT).

- Concerning the c files, some (all?) of them are generated by cython.

- The file silx/opencl/utils.py contains the license but not the explicit Copyright leading sentence.

PiRK:
- `silx/gui/plot/MPLColormap.py` has been recently moved to silx/gui/plot/matplotlib/Colormap.py`. The copyright header seems to not be standard (CC0). We could put our own copyrigth header, as far as I understand CC0 allows us to do whatever we want. But CC0 is more permissive than MIT, so I don't see the point of doing this.
 - `doc/source/conf.py` was auto-generated by sphinx, then slightly modified by us. Are we allowed to add our copyright header?
 - `silx/image/sift.py` is a one liner, but I will add our copyright anyway
[...]

PR #1173, opened 2017-09-20
Optimize all SVG icons and remove any possible licenses

t20100:
This PR makes sure all SVG icons are optimized (a few where not) and further reduce them by removing metadata (possibly containing copyright/license information).
This related to #1158 and probably closes it?


vallsv:
LGTM

PR #1174, opened 2017-09-20
[io] Support npz in silx.io.open

### LINKED ISSUE #1152: [silx view] Also open .npz files
npz files can be compressed and contain multiple arrays contrary to npy

vallsv:
This PR allow to load `npz` files using `silx.io.open`.

numpy supports more types than h5py can support, especially 'U' (UTF-32 strings) and 'O' (python object), then we can't be sure everything will be displayed properly. But for common data types it should be fine.

`npy` files still using FabioH5 and then are exposed using the NeXus layout.

Closes #1152

screenshot from 2017-09-20 14 42 19

### DIFF
--- a/silx/app/view.py
+++ b/silx/app/view.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "20/09/2017"
+__date__ = "21/09/2017"
 
 import sys
 import os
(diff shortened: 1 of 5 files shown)


vallsv:
Hi @vincefn, if you want to try before the release feedback is welcome. I only test with small and simple dataset. I guess you use a more complex data.

vincefn:
This works for me, I could open a bunch of npz files, including 2D/3D, nice !

Side issue: if a 2D array type is bool, it is not recognised as an image. Which is useful e.g. for masks. May need to move this to a new issue. I shared an example npz file where 2 arrays are bool through the esrf cloud.

t20100:
I opened issue #1177 for viewing boolean arrays.

What about opening `.npy` files the same way as `.npz` files? I see little interest in opening them with a HDF5 structure as compared to EDF there is no header and no metadata.

vincefn:
I agree it would be simpler to expose npy files in the same way.

vallsv:
Ok, let's do it.

vallsv:
Fixed. It's done as expected.

PR #1178, opened 2017-09-20
hdf5treemodel now closes files that have been opened

alemirone:


vallsv:
Thanks for the PR. There is definitely something to do here, but it is not enough. I will try to improve it.

Maybe another way would be to define this `__del__` in the object returned by our `silx.io.open` function.

vallsv:
Superseded by https://github.com/silx-kit/silx/compare/close-file-on-model-destruction?expand=1

Issue #1162, opened 2017-09-18
[plot] Split PlotWindow toolbar to smaller toolbars


t20100:
I think it would be better to have multiple toolbars for `PlotWindow`: one for the modes, one for the controls (log-scale, aspect ratio, etc...) and one for I/O (copy, save, print).
I makes it easier to add a toolbar and actions near the beginning (i.e., right after the modes) and it improves code reuse: one can use the toolbars he wants to make his own `PlotWidget`.
I did it this way in `plot3d`.

t20100:
Maybe it would be good to start a 'silx.gui.plot.tools' package for plot toolbars.

PR #1179, opened 2017-09-21
[plot] Add action show/hide axis

payno:
simple action to show and hide axis.
This only manage the case of the public API ('manage' the call to setAxesDisplayed)


t20100:
Few comments:

- I would definitely add the action to the available actions but I think it should not be added to existing widgets' toolbar.
  Extra actions can be added by the user in a subclass (and issue #1162 should simplify adding actions to toolbars by splitting it down).
- In `PlotWidget`, you still need to call `self._setDirtyPlot()` as this is handling the repaint of the plot, no?
- `_sigAxesVisibilityChanged` can be public (maybe name it `sigAxesDisplayChanged` to match `setAxesDisplayed` method)
- It seems `PlotWidget.isAxesDisplayed` is missing, it would be best to add it rather than adding it to each backend.
- For the icon, what about using the full screen icon? Otherwise, I send you how to "optimize" the SVG file.

payno:
> I would definitely add the action to the available actions but I think it should not be added to existing widgets' toolbar.
Extra actions can be added by the user in a subclass (and issue #1162 should simplify adding actions to toolbars by splitting it down).

Seems fair

> In PlotWidget, you still need to call self._setDirtyPlot() as this is handling the repaint of the plot, no?
I am not sure I see

yes bad line removal. Fixed

> _sigAxesVisibilityChanged can be public (maybe name it sigAxesDisplayChanged to match setAxesDisplayed method)

Complicated topic. Because for now only the case to call of the public API is handled. But in some cases only one axes can be modified (in the future ?). Maybe we can add an information about the axis concerned. But it will probably evolved and as we don't have a lot of time to think about it (if this is in the next release), have it private seems  'the safest'.

> It seems PlotWidget.isAxesDisplayed is missing, it would be best to add it rather than adding it to each backend.

Yes. But I thing a way or another it will be linked to the backend. For example in OpenGL it is link to  _plotFrame.displayed

> For the icon, what about using the full screen icon? Otherwise, I send you how to "optimize" the SVG file.

Maybe. This is true that it will in a way 'force' image to take all the place available. But it is also removing axis. I think this depend on users. But good for me.

t20100:
You're right, let's keep `_sigAxesVisibilityChanged` private.

For `PlotWidget.isAxesDisplayed`, we can keep a duplicate of the backend state in `PlotWidget`. This is done for example for axis states, it minimizes the code duplicated in each backend.

For the icon it's up to you, it's a matter of taste ;)

payno:
For PlotWidget.isAxesDisplayed I would also go to keep it private for now. For the same reason as _sigAxesVisibilityChanged

PR #1197, opened 2017-09-25
Sift upgrade

kif:
The idea is to use the OpenCL-Processing class which allows easy sharing of context (with the idea to re-use context and buffers between different modules)


kif:
Let's check if it actually works

t20100:
There is 2 cythonized files that got commited in the way : `silx/math/chistogramnd.c`,` silx/math/medianfilter/medianfilter.cpp`. It would be nice to remove those.... and re-consider storing cythonized files in github...

The` silx/opencl/alignment.py` is completely commented and looks to be a duplicate of `silx/opencl/sift/alignment.py`

Otherwise, a lot of refactoring and typo/style fixes. LGTM

PR #1206, opened 2017-09-28
[nxdata] Fix decoding attributes

PiRK:
Reading a byte string from an attribute and comparing it with a unicode string caused an error in python 3, while trying to find a dataset with a specific `@long_name` attribute.


vallsv:
LGTM

PR #1208, opened 2017-09-28
[silx view/io] Support filename::path

### LINKED ISSUE #1157: [silx view] directly open data from hdf5 path from the command line
It would be very practical if it was possible to supply an hdf5 path to directly display a given data array.

For example:
`silx view data.h5:/entry_1/instrument_1/detector_1/data`
or:
`silx view data.h5 --h5path=/entry_1/instrument_1/detector_1/data`

Often searching for the correct hdf5 path can be long, and it will only get worse once files are nexusified...

Maybe also add `--viewer=Cube` to choose the view ? 8-)

vallsv:
Merge first https://github.com/silx-kit/silx/pull/1223, i can then rebase this PR if you like.

- Add a generic proxy for python object (`silx.utils.proxy.Proxy`)
- Implement `filename::path` in `silx.io.open()`
   - The implementation returns a proxy to the target node which also owns the opened file
   - This proxy provides API of the target object, plus common API of opened files (`close` method and context management)
- Update `silx view` to avoid to check files before using `silx.io.open`
   - As result the viewer is always displayed anyway the files from the command line exist or not.

Closes  #1157 

```
valls@linkdick ~/workspace/silx.git (support-path-in-filename) $ ./bootstrap.py silx view 201606id01-FZP-S0013-result-new.npz::obj ../all_types.h5::float toto ../data/cant_be_read.h5
ERROR:silx.app.view:Filename 'toto' must be a file path
ERROR:silx.app.view:Unable open file (unable to open file: name = '/workspace/valls/data/cant_be_read.h5', errno = 13, error message = 'Permission denied', flags = 0, o_flags = 0)
```

screenshot from 2017-09-28 18 27 08

### DIFF
--- a/silx/app/test/test_view.py
+++ b/silx/app/test/test_view.py
@@ -26,7 +26,7 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "14/09/2017"
+__date__ = "29/09/2017"
 
 
 import unittest
@@ -110,7 +110,6 @@ class TestLauncher(unittest.TestCase):
     def testWrongOption(self):
         try:
             result = view.main(["view", "--foo"])
-            self.assertNotEqual(result, 0)
         except SystemExit as e:
             result = e.args[0]
         self.assertNotEqual(result, 0)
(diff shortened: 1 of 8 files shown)


t20100:
From a quick look at it, proxy.py license/copyright is not correct.

vallsv:
That's MIT license

vallsv:
The new proxy uses only original code. Then the license is fine.

PR #1213, opened 2017-09-29
[plot] Fix save image data as PNG 

t20100:
Save image data as PNG was broken (related to introduction of `Colormap` object)

This PR fixes it.


vallsv:
LGTM

PR #1225, opened 2017-10-02
[plot] Add tests for Imageview

t20100:
This PR adds some tests for `ImageView`.

It has some issues with matplotlib 2.1.0rc1 on macOS python2.7, PyQt4 where QWidget recursive repaint and Cocoa errors occurs.


t20100:
This PR also removes circular references that where causing tests to fails because of remaining widgets...

matplotlib 2.1.0rc1issue will be fixed in next release.

PiRK:
LGTM

PR #1226, opened 2017-10-02
Changelog v0.6

PiRK:
Please double-check, in case I forgot a new feature or an important bug fix..


vasole:
@t20100 @vallsv 

I find --use-opengl-plot do cumbersome to type that looks like discouraging the user to type it. Options like --gl or --opengl are much more intuitive to type.

kif:
OK we go like that

PR #1231, opened 2017-10-02
Fix float widget according to locale

### LINKED ISSUE #1200: Localized numeric values are not correctly interpreted
I've found an issue (latest master) when using the silx viewer, on a French-localized computer (mac):

- open a stack of images from the silx viewer
- open the colormap dialog. The min/max values are printed with a dot as separator ('5.2')
- it is not possible to enter a new value with a  dot separator ('4.7') because (I assume) a localised float-format validation is enforced in the entry box (I guess Qt auto-recognised the locale)
- if I enter a comma-separated value ('4,7'), I get a popup widget with a 'ValueError', could not convert string to float, because silx expects the C locale
- the colormap dialog cannot be opened any more because it continues to remember the comma-separated value entered, which triggers an immediate ValueError which can't be corrected.

vallsv:
- Force C locale from `silx view`
- Create a simple `FloatEdit` widget (it shoul be improved for a better feedback)
- Fix many tools using `FloatEdit`

Closes #1200

### DIFF
--- a/silx/app/view.py
+++ b/silx/app/view.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "21/09/2017"
+__date__ = "02/10/2017"
 
 import sys
 import os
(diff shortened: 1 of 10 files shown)


t20100:
LGTM, don't forget to re-cythonize all files for the release.

PR #1234, opened 2017-10-02
Filter fabio metadata

vallsv:
- Avoid to store metadata headers only used to decode the input file format
- Robustness on raw header


PiRK:
LGTM. AppVeyor tests fail probably d/t missing fix #1240

PR #1246, opened 2017-10-03
[plot3d] Improve axes indicator

### LINKED ISSUE #993: [plot3d] Improve overlay orientation indicator integration
For now, it uses a semi-transparent square background.
Either make the background fully transparent or being a semi-transparent circle would look better.

t20100:
This PR improves the look of the top-right axes orientation indicator by displaying the axes in a semi-transparent circle rather than in an opaque square.

Closes #993

### DIFF
--- a/silx/gui/_glutils/gl.py
+++ b/silx/gui/_glutils/gl.py
@@ -101,7 +101,10 @@ def enabled(capacity, enable=True):
     :param bool enable:
         True (default) to enable during context, False to disable
     """
-    if enable:
+    if bool(enable) == glGetBoolean(capacity):
+        # Already in the right state: noop
+        yield
+    elif enable:
         glEnable(capacity)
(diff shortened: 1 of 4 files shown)


vasole:
Not working under windows? The GL tests do not pass

https://ci.appveyor.com/project/ESRF/silx/build/0.6.0.2708

t20100:
Test fails for other reason (#1273)

vasole:
They may fail for other reason but, why do we only see them when WITH_GL_TEST=True and not then WITH_GL_TEST=False?

https://ci.appveyor.com/project/ESRF/silx/build/0.6.0.2708

t20100:
Because both tests with `WITH_GL_TEST=True` are also both using `PyQt5`

vallsv:
LGTM

Issue #1237, opened 2017-10-03
[Tomo] FBP test fails under windows


kif:
investigation under way

kif:
the computer has 3 drivers: intel, amd and nvidia. 
the one selected for running the tests is:
'GeForce GT 640' on 'NVIDIA CUDA'

I will test all 3 devices and check if any of them works

kif:
It works when using Intel and AMD CPU driver.

kif:
using the CPU code on the GPU does not help,
nor using volatile shared arrays.

kif:
@pierrepaleo  do you have an idea ?

kif:
Compiles neither on MacOS

vasole:
I confirm it runs smoothly on CPU under windows.
On GPU, the example does not generate a good sinogram.

vasole:
Closed by #1247

PR #1247, opened 2017-10-03
Disable filtering in fourier space for now, until gpyfft is useable.

kif:
Contains some clean-up but disable filtering in Fourier space. The results are too variable from one implementation to another.

Numpy does the job for now: the feature is available but not yet at full speed.


kif:
close #1237

t20100:
LGTM, let's see if CI is still happy.

PR #1248, opened 2017-10-03
[plot3d] Add rotate scene around vertical axis action

### LINKED ISSUE #1125: [plot3d] Silx 3D viewer continuous rotation
It would be nice to have a button to activate the live rotation of the 3D view (along one chosen axis ? Or just the vertical one).

Would be useful, with the possibility of changing the iso surface value / cut plane location.

t20100:
This PR adds an action to start/stop rotating the scene around the central vertical axis.
This action is added to Plot3DWindow toolbar.
While it is rotating, the scene remains interactive.

Closes #1125


PiRK:
LGTM

vincefn:
Nice !

PR #1258, opened 2017-10-05
[Profile] fix #1256

### LINKED ISSUE #1256: [profile] won't show the profile window
Under windows PC.
When I launch the silx view and display some dataset using the `Image` and `Image stack` views :
- if I am on full screen mode the profile window won't appear (and won't be launched, I can't see them)
But if I am not on the full screen mode then It work ??? !!!
- if not on full screen mode work perfectly.

Don't know where the issue is (profile window, viewer or somewhere in the middle).

payno:
This should fix the issue #1256 
But I wonder why the showProfileMainWindow of the profile has such a complicated behavior.

From my point of view having it displayed close to the mouse would be good too no ?

### DIFF
--- a/silx/gui/plot/Profile.py
+++ b/silx/gui/plot/Profile.py
@@ -653,19 +653,17 @@ class ProfileToolBar(qt.QToolBar):
             winGeom = self.window().frameGeometry()
             qapp = qt.QApplication.instance()
             screenGeom = qapp.desktop().availableGeometry(self)
-
             spaceOnLeftSide = winGeom.left()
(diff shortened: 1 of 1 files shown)


t20100:
Then you would have to move the profile window away from the plot each time you start a profile and the window is hidden.

To me in an application, it would be better to have the profile window and the plot both included in a single window.

vasole:
With floating windows, my experience is that sometimes show() is not enough and one needs to call the widget method raise_() after show because it can be "shown" but under other widget/window.

Issue #1261, opened 2017-10-06
[silx.math.combo] min_max


payno:
If I open a terminal and go for:

```
import numpy
from silx.math.combo import min_max

l = numpy.arange(10, numpy.int64)
min_max(l) 
```
This raises 
```

ERROR:silx.gui.qt._qt:<class 'TypeError'> No matching signature found   File "C:\Users\opdau\Documents\payno\silx\build\lib.win-amd64-3.5\silx\gui\plot\ColorBar.py", line 213, in _activeImageChanged
    data=image)
  File "C:\Users\opdau\Documents\payno\silx\build\lib.win-amd64-3.5\silx\gui\plot\ColorBar.py", line 159, in setColormap
    data=data)
  File "C:\Users\opdau\Documents\payno\silx\build\lib.win-amd64-3.5\silx\gui\plot\ColorBar.py", line 389, in setColormap
    self.colorScale.setColormap(colormap, data)
  File "C:\Users\opdau\Documents\payno\silx\build\lib.win-amd64-3.5\silx\gui\plot\ColorBar.py", line 518, in setColormap
    self.vmin, self.vmax = self._colormap.getColormapRange(data=data)
  File "C:\Users\opdau\Documents\payno\silx\build\lib.win-amd64-3.5\silx\gui\plot\Colormap.py", line 251, in getColormapRange
    min_, max_ = min_max(data, min_positive=False, finite=True)
  File "silx/math/combo.pyx", line 320, in silx.math.combo.min_max (silx\math\combo.c:14195)
    return _min_max(data, min_positive)
  File "silx/math/combo.pyx", line 117, in silx.math.combo.__pyx_fused_cpdef (silx\math\combo.c:3524)
    def _min_max(_number[::1] data, bint min_positive=False):

```

This append in our windows PC but work on other console. Seems like he is not handling numpy.int64 and in some examples replace it for int32.

t20100:
This comes from the fused type `_number` which does not use sized types.

t20100:
closed by #1262

PR #1262, opened 2017-10-06
[silx.math.combo] Fix 1261

payno:


vallsv:
I found that on our code, which should be more explict.

From chistogram
```
import numpy as np
ctypedef fused lut_t:
    np.int64_t
    np.int32_t
    np.int16_t
```
or from fabio
```
cimport numpy as cnp
ctypedef fused any_int_t:
    cnp.int8_t
    cnp.int16_t
    cnp.int32_t
    cnp.int64_t
```

payno:
should close #1261

t20100:
chistogram also depends on numpy at C level, there is a `cimport numpy as np`.

t20100:
Ideal solution would be to have signed types definition without relying on numpy.

PR #1285, opened 2017-10-11
[specfile tests] Close specfile objects

PiRK:
This PR closes files properly in tests, so that we don't need to call `gc.collect()` before deleting the test files.

There is still a problem with `test_spectoh5`, caused by `convert(self.spec_fname, self.h5_fname)`. This line still requires a `gc.collect` for a reason unknown.

### DIFF
--- a/silx/io/commonh5.py
+++ b/silx/io/commonh5.py
@@ -32,12 +32,13 @@ import collections
 import h5py
 import numpy
 from silx.third_party import six
+import weakref
 
 from .utils import is_dataset
 
 __authors__ = ["V. Valls", "P. Knobel"]
 __license__ = "MIT"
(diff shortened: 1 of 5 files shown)


PiRK:
The issue with `test_spectoh5` and `convert` is that it uses `silx.io.open`, which has a known bug: it tries to open the file with FabIO first, then with SpecH5, and the first attempt keeps a file handle.

This will be fixed with the next FabIO release. In the meantime, we need to continue calling `gc.collect`.

vasole:
Tests are failing.

PiRK:
I forgot to test Python 2. It seems Python 3 is more permissive than Python 2.
Now it is fixed.

vallsv:
LGTM. Nice clean up.

PR #1286, opened 2017-10-11
1199 ocl image

### LINKED ISSUE #1199: [OpenCL] interpolation & decimation
Provide interpolation (nearest/bilinear) and decimation (Lanczos2/3) functions and a couple of kernel using them.

kif:
In this PR we define a set on common tools to process images like:

- [x] Cast to float 
- [x] scale intensity in a given range (normalization)
- [x] intensity histogram
- [ ] convert from color to grayscale
- [ ] Bilinear interpolation (single or multiple color chanels)
- [ ] Lanczos interpolator
- [ ] image rotation
- [ ] homography
- [ ] gaussian blur 
- [ ] convolution
- [ ] sRGB - RGB - Lab - Luv color space tranformation
- [ ] desaturation

close #1199


kif:
All features are not yet implemented, but some useful ones are. I believe it is worth merging into the master branch

t20100:
LGTM  if CI is happy with it.

To allow it to scale to more image processing, maybe it is worth to split the `ImageProcessing` class into different processing classes (sharing a base class and with the possibility to share some buffers).... to be discussed if it is possible/better.

PR #1289, opened 2017-10-13
Byte offset in OpenCL

kif:
This is an implementation of the byte offset decompression in OpenCL. It may accelerate the reading of CBF files on high end GPUs (to be validated)


kif:
I may have a couple of improvements (for speed)

t20100:
I proposed a few minor modifications in a PR on this branch, otherwise LGTM!

PR #1291, opened 2017-10-16
[LegendSelector] keep a curve active after renaming it

PiRK:
If a curve is active, make sure it is still active after renaming it in the legend selector.


t20100:
LGTM

PR #1293, opened 2017-10-18
PySide2 support

### LINKED ISSUE #1283: [gui] Add support of PySide2
PySide2 is already available in conda (and not in pypi as of today) and it is supported in matplotlib.
We should support it.... and drop PySide...

t20100:
This PR is adding experimental support of PySide2.

It only adds PySide2 as a supported binding but it **does not** fix issues specific to PySide2 (i.e., tests does not pass with PySide2).

Related to #1283


t20100:
I tested PySide2 with miniconda3 on macOS:
- Plot1D, Plot2D and ScalarFieldView widgets are working, except for the print tool button (error message: `QPrintDialog: Cannot be used on non-native printers`)
- many tests fail until test `testDropExternalFile (silx.gui.hdf5.test.test_hdf5.TestHdf5TreeModel)` set fault.

A lot more testing/debugging is needed to truly support `PySide2`.

vasole:
Please, get rid of the conflict.

vallsv:
I try with the web interface, but it is not working

vallsv:
It finally looks to be fine. Let's merge it.

PR #1295, opened 2017-10-18
[plotWidget] reset color and style cycling

PiRK:
This PR ensures that calling `addCurve(replace=True)` has the same effect as `clear() + addCurve(replace=False)` on the next curves' color and style.


t20100:
LGTM

PR #1301, opened 2017-10-20
[image] typo "uin8" -> "uint8" in _convertImageToRgba32

PiRK:
The typo maybe causes an extra copy when data is already in uint8, but it still does the job.


t20100:
Nice catch!

PR #1302, opened 2017-10-20
[plot] Log-log scale bug

PiRK:
With mpl >= 2.0.0, axes are cleared prior to applying log scale. This PR ensures the log status of the other axis is preserved, as well as the X and Y limits (unless autoscale is active).


PiRK:
It seems to survive the test of switching to log scale while the axes limits are negative.
MPL prints a few `RunTimeWarning`, and when I switch back back to linear scale my graph is zoomed around *(x, y) = (1,1)*.

PiRK:
There is also a problem with the grid disappearing, in mpl 2.0.1

PiRK:
see #1307

PR #1303, opened 2017-10-20
[plot3d] ScalarFieldView support transforming data with a 3x3 matrix (for non orthogonal support)

### LINKED ISSUE #951: [plot3d] Support non-orhtogonal axes

t20100:
This PR adds support of a 3x3 matrix in the data transformation (`get|setTransformMatrix`).
This is meant to display a dataset defined in a non-orthogonal coordinate system.
The displayed axes with ticks are the orthogonal axes and by default the cutting plane can be set to the direction of the 3 orthogonal axes (you can define other cut plane orientation from the API).

This PR breaks the API because the cutting plane was previously defined in the "array" coordinates and it is now defined in "data" coordinates with transformations applied.
This PR updates the `CutPlane` class API.

closes #951

### DIFF
--- a/silx/gui/plot3d/SFViewParamTree.py
+++ b/silx/gui/plot3d/SFViewParamTree.py
@@ -665,7 +665,7 @@ class DataSetItem(qt.QStandardItem):
 
         self.setEditable(False)
 
-        klasses = [DataTypeItem, DataShapeItem, OffsetItem, ScaleItem]
+        klasses = [DataTypeItem, DataShapeItem, OffsetItem]
         for klass in klasses:
             titleItem = qt.QStandardItem(klass.itemName)
             titleItem.setEditable(False)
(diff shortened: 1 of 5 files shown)


t20100:
To test it, you can add the following code to `examples/viewer3DVolume.py`:

```
window.setTransformMatrix(((1., 1., 0.),
                           (0., 1., 0.5),
                           (0., 1., 1.)))
```

t20100:
TODOs:

- [ ] Check ROI display/returned value
- [ ] Add transform matrix to param tree (and eventually optionally allow to edit it from the GUI)
- [ ] Check if cut plane API break is an issue and can be avoided (by adding an argument to choose the coordinate system)

t20100:
`CutPlane` API is updated but now backward compatible.

t20100:
Finally took the time to complete this one.
It should be fully backward-compatible.

The cut plane can be specified either in the array or data coordinates (i.e., with transforms applied).
For now retrieving an image with the API from the cut plane is only available if the plane is perpendicular to an axis of the array.

The method `SelectedRegion.getDataRange` returns the bounding box of the selection (in case the provided 3x3 matrix is a rotation or a skew).

payno:
LGTM

PR #1306, opened 2017-10-23
[NXdataView] Add 1D data as curve, not scatter

PiRK:
Even if x values are not monotonic, it can still be a curve (mesh scan, parametric data...)

It makes more sense to add 1D data always as curves, and add a scatter only in the case of 2 axes and one signal dataset (x, y, values). Users can toggle the curve style to display points, if they want.

### DIFF
--- a/silx/gui/data/NXdataWidgets.py
+++ b/silx/gui/data/NXdataWidgets.py
@@ -26,7 +26,7 @@
 """
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
-__date__ = "27/06/2017"
+__date__ = "23/10/2017"
 
 import numpy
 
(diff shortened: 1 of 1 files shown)


PiRK:
Related to #1236

t20100:
Do you want to always add a curve in this case but change the default: lines for monotinic curves, points for everything else? Knowing that the user can change the curve style from the GUI.

vasole:
To show always as curves seems to be obvious.

For the default behavior, I guess we'll have answers for everybody.

My preference would be to use the matplotlib default for curves that seems to be line without symbol but with the possibility to set the default by applications desirable (default point, point-line, just line).

My point is that if the action to toggle lines is applied to the plot or to the active curves. If it is applied to the plot, then we should not switch line representation as function of the x ordering.

It is not something critical because user can choose interactively.

PiRK:
As far as I car tell, our action toggles the *default* style, probably not curves with an explicit non-default style. So if I add a curve with an explicit `linestyle=" "`, I don't think the user will be able to enable the lines.

PiRK:
I just realized that I'm always dealing with a single curve, so it's OK to not specify a style when adding a curve, and then change the plot default style to set the linestyle.

t20100:
@sole, yes in the end it is probably better (and simpler) to always show the data the same way (i.e., with lines) without checking whether it is monotonic or not.
The kind of plot to use (scatter or curve) should be provided by NXData somehow.

So LGTM as is.

Issue #1136, opened 2017-09-12
[plot] Issue with setting log scale for curves with negative data


t20100:
There is an issue with curves with negative data when setting to log scale and then moving the mouse over the plot raises an issue with python3.5, matplotlib 2.0.2 and PyQt5 (most probably) on linux.

Changing the minimum version of matplotlib for which to clear the plot before toggling log scale solves the issue.
The calls to `cla` where added in https://github.com/silx-kit/silx/commit/8a0b783b99a8172ef0edc002f02ef92ca2cafd04 for matplotlib 2.1.0

PR #1307, opened 2017-10-26
Matplotlib 2.1.0 bug with log scale

PiRK:
Change the way to fix #1081 because `cla()` caused lots of side effects.

I kept `>= "2.0.0"` as a condition because of #1136. We need to test if this issue is still solved.

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -2647,6 +2647,13 @@ class PlotWidget(qt.QMainWindow):
         xAuto = self._xAxis.isAutoScale()
         yAuto = self._yAxis.isAutoScale()
 
+        # With log axes, autoscale if limits are <= 0
+        # This avoids issues with toggling log scale with matplotlib 2.1.0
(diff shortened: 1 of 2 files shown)


vasole:
Is it possible to add debug level logging information instead of just "pass"?

PiRK:
Good idea.

t20100:
Even with this fix, I still get issues when the limits of the axes are fully negative before switching to log scale:
```
from silx import sx
w = sx.plot((-1, -2))
w.setGraphXLimits(-2, -1)
w.setGraphYLimits(-2, -1)
```
and toggle log scale.
Tested on debian8 (python 3.4, matplotlib 2.1.0, PyQt 5.3.2).

t20100:
What I tested that seemed to work in all case was to set limits > 0, calling matplotlib draw and then setting log scale.
The problem is that this introduce some flicker when switching to log scale.

t20100:
I've added an extra workaround in a PR on this branch (see https://github.com/PiRK/silx/pull/8)
It now seems to work for what I've tested (negative limits, no postive data) with mpl 1.4.1, 1.5.3, 2.0.2 and 2.1.0.

With this, to me this issue is now fixed on our side.

Issue #1317, opened 2017-10-31
[plot] Colormap sampling subject to rounding errors


t20100:
For the record, there is an issue when getting the LUT from a colormap by sampling 256 values of the colormap.
One cannot expect to get the exact internal LUT due to numerical imprecision when converting colors applied with a colormap from float in [0, 1] to uint8.

```
from silx.gui.plot.Colormap import Colormap
import numpy
# Create a gray scale colormap with linear normalization and autoscale and apply it to 256 values
colors = Colormap().applyToData(numpy.arange(256))
numpy.all(numpy.equal(colors[:, 0], numpy.arange(256)))
```
This returns `False`, while we should expect `True`.

For colors defined with 256 levels in [0, 1] this error does not occurs when using float32 to go from float to uint8:
`numpy.all(numpy.equal(numpy.linspace(0, 1, 256, dtype=numpy.float32) * 255, numpy.arange(256)))`
-> True
`numpy.all(numpy.equal(numpy.linspace(0, 1, 256, dtype=numpy.float64) * 255, numpy.arange(256)))`
-> False

PR #1318, opened 2017-10-31
[plot] Add a method to sample a Colormap into a color array

### LINKED ISSUE #1132: [plot] Add a method to Colormap to get the LUT
The ColorBarWidget, plot3d and plot OpenGL backend at least uses some similar code to get the color LUT from a Colormap.
This should be provided once by the Colormap object

t20100:
This PR adds a `Colormap.getNColors` method that returns an array of color sampled linearly from the colormap LUT (not taking into account normalization and range).
This allows to share code between `ColoBar`, plot OpenGL backend and plot3d.

Closes  #1132

There is an issue while getting the color LUT this way, see #1317


vallsv:
LGTM

Issue #1292, opened 2017-10-18
[PyPI] Provide link to silx home page in package metadata.


vasole:
Home page should point to www.silx.org and bug tracker to the issues in GitHub.

t20100:
I couldn't find how to set the bug tracker URL from setup.py.

PiRK:
It seems this is another feature that has been removed from the PyPi web interface. I could not find anything either.

Can we remove the 0.6.1 milestone?

vasole:
What it cannot be found is the bug tracker. The home page (www.silx.org) should be there.

PiRK:
This part of the issue can be closed (#1319).

t20100:
Closed by #1319

PR #1319, opened 2017-10-31
Update project homepage and zenodo URL

t20100:
This PR point project URL to www.silx.org and update zenodo URL to point to silx umbrella DOI (so it will always point to the latest release).

Related to #1292


kif:
LGTM

PR #1323, opened 2017-11-06
[CI] Remove PyQt5 version pinpointing on appveyor

t20100:
This PR makes use of the latest version of PyQt5 on appveyor.


t20100:
Related to #936

t20100:
Still seems to fail

t20100:
@vasole, tests fails the same way when OpenGL tests are disabled: https://ci.appveyor.com/project/ESRF/silx/build/0.6.0.2812/job/jj64869ou574f0er

t20100:
To make test pass, I disabled the following tests:

```
diff --git a/silx/gui/hdf5/test/test_hdf5.py b/silx/gui/hdf5/test/test_hdf5.py
index 8e375f2..455ce9b 100644
--- a/silx/gui/hdf5/test/test_hdf5.py
+++ b/silx/gui/hdf5/test/test_hdf5.py
@@ -833,9 +833,9 @@ class TestHdf5TreeView(TestCaseQt):
 def suite():
     test_suite = unittest.TestSuite()
     loadTests = unittest.defaultTestLoader.loadTestsFromTestCase
-    test_suite.addTest(loadTests(TestHdf5TreeModel))
+    #test_suite.addTest(loadTests(TestHdf5TreeModel))
     test_suite.addTest(loadTests(TestNexusSortFilterProxyModel))
-    test_suite.addTest(loadTests(TestHdf5TreeView))
+    #test_suite.addTest(loadTests(TestHdf5TreeView))
     test_suite.addTest(loadTests(TestH5Node))
     return test_suite
 
diff --git a/silx/gui/widgets/test/__init__.py b/silx/gui/widgets/test/__init__.py
index 7affc20..ef1c916 100644
--- a/silx/gui/widgets/test/__init__.py
+++ b/silx/gui/widgets/test/__init__.py
@@ -38,7 +38,7 @@ __date__ = "19/07/2017"
 def suite():
     test_suite = unittest.TestSuite()
     test_suite.addTests(
-        [test_threadpoolpushbutton.suite(),
+        [#test_threadpoolpushbutton.suite(),
          test_tablewidget.suite(),
          test_periodictable.suite(),
          test_printpreview.suite(),
```
Not sure commenting `TestHdf5TreeView` is needed...

vallsv:
No idea why the tests was not skipped
```
In [2]: from PyQt5 import Qt

In [3]: Qt.qVersion()
Out[3]: '5.9.2'

In [4]: Qt.qVersion() == "5.9.2"
Out[4]: True

In [5]: Qt.PYQT_VERSION
Out[5]: 329985

In [6]: Qt.PYQT_VERSION_STR
Out[6]: '5.9.1'

In [7]: Qt.PYQT_VERSION_STR == "5.9.1"
Out[7]: True

In [8]: import sys

In [9]: sys.platform == "win32"
Out[9]: True
```

vasole:
The tests were not passing on other Qt versions too (5.9, 5.9.1, 5.9.2). Perhaps it is safer to tests if qt.qVersion() starts with 5.9

vasole:
For instance, you test against Qt 5.9.2 but your message mention skipping under windows and Qt 5.9.1. If you want to do exactly that, you are not doing the proper test.

vallsv:
Yes, you are right. But the error is on the skip comment

vallsv:
But i would prefer to check if the situation change for the next release that is why i dont use the prefix "5.9"

vasole:
I will try to test if the problem is at the Qt or at the PyQt5 side. I have a former version of Qt compiled. So, I will compile latest sip and PyQt5 but using that version of Qt.

vallsv:
I spend time for that, then it is close to be done on our Windows computer

vallsv:
It looks like the segfault can be removed by an extra `prosessEvents` somewhere, but as result Qt deadlock.

It looks like this code create the deadlock on Python3 / Windows / PyQt 5.9.1 / Debug build.
- I did not try other PyQt versions
- I did not test on other systems
```
> python
from PyQt5 import Qt
qapp = Qt.QApplication([])

class Runnable(Qt.QRunnable):
    def __init__(self):
        super(Runnable, self).__init__()
    def run(self):
        print("AAAAAAAAAAAAA")

Qt.QThreadPool.globalInstance().start(Runnable())

# There is a deadlock on exit
exit()
```

vallsv:
In fact this code create a deadlock in the debug mode, but segfault 50% of time in release mode.

vallsv:
Easier way to segfault:
```
> python
from PyQt5 import Qt
qapp = Qt.QApplication([])
Qt.QThreadPool.globalInstance()
exit()
```

t20100:
Closing as integrated in #1328.
For further discussion, see #936.

Issue #1311, opened 2017-10-26
[plot3d] Add a general purpose widget for 3D rendering


t20100:
Target is a replacement candidate for PyMca Object3D widget that can display:
- 3D scalar field on regular grid (as ScalarFieldView)
- 3D scatter plot
- 2D image on regular grid
- 2D scatter plot

and the associated widget to set the parameters of the scene.

t20100:
Status:

Scene items:
- [x] 3D scalar field on regular grid with iso-surfaces and cut plane (one for now)
- [x] 3D scatter plot as 3D points
- [ ] 2D image on regular grid:
   - [x] as 2D image
   - [ ] as surface plot
- [x] 2D scatter plot as: points, lines or surface (through Delaunay triangulation)
- [x] Clip plane (one for now)
- [ ] 3D mesh

Features:
- [ ] Per item transform:
  - [x] translate, rotate, scale, 3x3 matrix
  - [ ] configurable center of rotation
- [ ] Per item bounding box 
- [ ] Per item lighting enable/disable
- [ ] Picking of item and position in items
- [ ] Planes interaction
- [ ] Parameter tree widget to update item properties/scene content

Extra:
- [ ] Multiple cut planes
- [ ] Multiple clip planes

Misc:
- [ ] Documentation
- [ ] Sample code
- [ ] Tests

t20100:
Closing as most features are available.
Using one issue per required new feature is now a better way to track this.

PR #1326, opened 2017-11-08
Add Delaunay implemenation to silx.third_party

t20100:
This PR adds `libqhull` wrapper and `Delaunay` class from `scipy` v1.0.0 to `silx.third_party`.
This wrapper has been patched to be standalone and remove dependency to lapack.

When removing `silx.third_party._local` as part of the packaging, `scipy.spatial.Delaunay` is used instead of the local copy.

Delaunay triangulation is need for upcoming plot3d feature (#1311).

I quickly benchmarked 4 `libqhull` wrappers with 10**6 2D points:
- PyMca (in C)
- PyMca  (in cython)
- matplotlib.tri.Triangulation
- scipy.spatial.Delaunay

matplotlib is the fastest, and scipy the slowest, but they all remains within 10%. 

Some alternatives are:
- Use matplotlib.tri.Triangulation
- Update PyMca wrapper to libqhull2015.2
- Implement another Delaunay.....


kif:
Nice baby ...

payno:
LGTM

PR #1328, opened 2017-11-09
Refactoring test options

vallsv:
Proposal to clean up the test process.

- Create a container `TestOptions` to configure the test process
- Remove environment variable from the tests
- `run_tests` configure `TestOptions` according to the environment variable and the command line
- [x] Talk about `SILX_OPENCL` which is part of the silx library, and not only part of the tests
- [x] Fix `silx test` app


silx/gui/hdf5/test/test_hdf5.py
```diff
@@ -26,11 +26,12 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "22/09/2017"
+__date__ = "09/11/2017"
 
 
 import time
 import os
+import sys
```
t20100: this seems not used

silx/gui/widgets/test/test_threadpoolpushbutton.py
```diff
@@ -26,21 +26,25 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "15/12/2016"
+__date__ = "09/11/2017"
 
 
 import unittest
 import time
+import sys
```
t20100: it seems not used

t20100:
This makes it cleaner!

t20100:
Can't you share the implementation of `configureTestOptions` and `configure_test_options` in `silx.test.utils`?

As `run_tests.py` is used in other projects I think the import and use of  `<project>.test.utils` should be optional.

vallsv:
I agree but it's difficult. `options` is from the command args and then part of the API of each programs. But we can talk about it.

vallsv:
Now it is reworked as discussed with @t20100

t20100:
Fixed branch conflict.

LGTM: This is cleaner and improves reuse in other project.

PR #1329, opened 2017-11-10
CHANGELOG for 0.6.1

PiRK:


PiRK:
I removed the ready to merge label for the time being, so we can discuss whether we should also fix #1330 for this minor release.

PR #1331, opened 2017-11-10
Refactoring test options [debug CI]

vallsv:


vallsv:
Problem found... a typo... as always with Python...

Issue #772, opened 2017-04-28
[Plot] DockWidgets should be just containers


vasole:
This is just a heads up.

The PR #771 illustrates a situation that can be common to other docked widgets.

It seems reasonable to have all the functionality available in the contained widget independently of being inside a dock widget or not. In other words, the dock container should not add other functionality than the docking capabilities.

This seems to have been discussed but I did not find an associated issue.

t20100:
+1
Yes we started discussing it, to me the DockWidget should only manage layout stuff, not functionalities, and ideally be created by who is using the widget (e.g., PlotWindow in our case). This is clearly not the case as it is now.

vasole:
Add to guidelines.

PR #1334, opened 2017-11-13
[CurvesROIDockWidget]add a setter to activate the middle ROI marker

PiRK:
The functionality was implemented, but not publicly exposed.

### DIFF
--- a/silx/gui/plot/CurvesROIWidget.py
+++ b/silx/gui/plot/CurvesROIWidget.py
@@ -46,7 +46,7 @@ ROI are defined by :
 
 __authors__ = ["V.A. Sole", "T. Vincent"]
 __license__ = "MIT"
-__date__ = "27/06/2017"
+__date__ = "13/11/2017"
 
 from collections import OrderedDict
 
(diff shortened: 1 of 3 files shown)


t20100:
Why do you add a new already deprecated `CurvesROIDockWidget.setMiddleROIMarkerFlag`method?

PiRK:
There was no proper way to do this without either breaking the internal consistency (not doing all plot operation in  the same widget), or breaking #772, or implementing the functionality in both widgets and deprecating it in the DockWidget.

So I decided to refactor the code to conform with #772, by moving everything from the `DockWidget` to `CurvesROIWidget`. I tried not to break the compatibility, but obviously it will depend on how much of the internal API our users are accessing.

TODO (?): give a direct getter `CurvesROIWidget` in `PlotWindow`. Currently one needs to use `PlotWindow.getCurvesRoiDockWidget().roiWidget`

t20100:
+1 for a direct access to CurvesROIWidget in PlotWindow, this is inline with #772.

PR #1337, opened 2017-11-14
Fix PATH when using a python in debug mode

vallsv:
Use a "-pydebug" suffix in case of Python debug interpreter, else the synchronization is not done at the right place


vallsv:
I checked the way the ABI `d` is computed by the `wheel` and `pip` module.

PR #1343, opened 2017-11-16
[plot3d] Add new primitives to OpenGL internal scene structure

t20100:
This PR adds new rendering primitives and updates existing ones.
It is a step towards #1311


vallsv:
LGTM

vallsv:
I think it would be better to move the shader as resource to have at least a syntax highlighting, and a smaller module, and maybe something more flexible, but it's also fine like that.

t20100:
The primitive file has grown a lot and need refactoring: it should at least be split into smaller modules.
I like to have the shader code close to the python code because they are tightly coupled, but I can move the shaders in the resource, that would be inline with the opencl part.

PR #1350, opened 2017-11-21
Check segfaults on PyQt 5

vallsv:
Do not merge it.

It's only a branch to check PyQt issues with our non-patched project.

```
python run_tests -v silx.gui.test.test_issue_936.suite_1  # which is now fixed on PyQt upstream
python run_tests -v silx.gui.test.test_issue_936.suite_2
python run_tests -v silx.gui.test.test_issue_936.suite_3
```


vallsv:
Outdated

PR #1351, opened 2017-11-21
[plot] Improve marker implementation in matplotlib backend

### LINKED ISSUE #1348: [PlotWidget] Marker text does not move with marker
When a marker is dragged, the text does not move. It moves after any other interaction happens (curve clicked, pan, zoom).

This issue seems to be present with all mpl versions.

t20100:
This PR improves matplotlib plot backend marker text support:

- Closes #1348: Marker text does not move with marker 
- Hides marker text when marker is not visible in the plot area
- Update line marker text position when plot limits change
- Refactor marker in the backend
- Remove unused overlay argument in backend addMarker


PiRK:
LGTM

PR #1352, opened 2017-11-23
[plot3d] Fix isosurface level editor in parameter tree widget

### LINKED ISSUE #1280: [plot3d] Input field of isosurface level is limited to 2 digits
This can prevent setting the level to the desired value.

t20100:
This PR allows to set any value for isosurface level from the parameter tree widget rather than limiting resolution to 10e-2.

closes #1280


vallsv:
LGTM

PR #1353, opened 2017-11-23
[plot3d] Update plot3d sample code documentation

t20100:
This PR removes the sample code available from silx.gui.plot3d module documentation page.
It was duplicating the sample code gallery.


vallsv:
LGTM

PR #1355, opened 2017-11-24
[plot3d] Add light direction control from param tree

t20100:
This PR adds 2 sliders to control the light direction in the param tree.
The sliders control azimuth and altitude angles.

To test, use examples/view3DVolume.py and go to  Style-> Light Direction in the param tree widget.


vallsv:
LGTM

PR #1372, opened 2017-11-30
[CI] Fix travis osx python 2.7

t20100:


t20100:
This PR makes use of the previous default osx worker as the new default osx worker that was just introduced does not seems to provide (yet?) `pip`.

Related to #1373

vallsv:
LGTM

PR #1375, opened 2017-11-30
Expose supported extensions

vallsv:
A convenient way to expose supported format.

We also can talk about refactoring this thing using a descriptor class per codec. It can be very useful to provides more informations.

### DIFF
--- a/silx/io/__init__.py
+++ b/silx/io/__init__.py
@@ -36,6 +36,7 @@ from .utils import is_dataset
 from .utils import is_file
 from .utils import is_group
 from .utils import is_softlink
+from .utils import supported_extensions
 
 # avoid to import open with "import *"
 __all = locals().keys()
--- a/silx/io/__init__.py
+++ b/silx/io/__init__.py
@@ -26,7 +26,7 @@
 
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
-__date__ = "15/05/2017"
+__date__ = "30/11/2017"
 
 
 from .utils import open  # pylint:disable=redefined-builtin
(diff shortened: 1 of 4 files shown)


silx/io/fabioh5.py
```diff
@@ -53,6 +53,36 @@
 _logger = logging.getLogger(__name__)
 
 
+_fabio_extensions = set([])
+
+
+def available_extensions():
```
t20100: Why not calling it `supported_extensions` as the `silx.io` one?

silx/io/utils.py
```diff
@@ -53,6 +53,38 @@
...
+        formats["NeXus files"] = set(["*.nx", "*.nxs", "*.h5", "*.hdf"])
+    formats["NeXus layout from spec files"] = set(["*.dat", "*.spec", "*.mca"])
+    if flat_formats:
+        try:
+            import fabioh5
+        except ImportError:
+            fabioh5 = None
+        if fabioh5 is not None:
+            formats["NeXus layout from fabio files"] = set(fabioh5.available_extensions())
+    if numpy is not None:
```
t20100: numpy should always be available

silx/io/__init__.py
```diff
@@ -36,6 +36,10 @@
...
 from .utils import is_group
 from .utils import is_softlink
+from .utils import supported_extensions
 
 # avoid to import open with "import *"
-__all__ = ["save1D", "is_dataset", "is_file", "is_group", "is_softlink"]
+__all = locals().keys()
+__all = filter(lambda x: not x.startswith("_"), __all)
+__all = filter(lambda x: x != "open", __all)
+__all__ = list(__all)
```
t20100: I personally preferred the previous `__all__` definition even if it needs to be maintained by hand.

vallsv: This do what it is supposed to do. Filtering out open.
The previous way is not maintainable, and will grow for every.

silx/io/fabioh5.py
vallsv: It is not the same output, but yes, it can be called the same

vallsv:
I updated the thing according to your comments. But it still use filters for the `__all__`.

Maybe it would be better to remove everything. `open` is not anymore a reserved word in Python3 if i am not wrong, and no body should use `import *`. Then this filters looks to be mainly useless and without use case.

PR #1379, opened 2017-12-01
[NXdataView] Fix issues with non unique axis labels

PiRK:
When two axes use the same @long_name attribute, we got  an error because we relied on these names to be unique.

This PR fixes this bug and adds an assertion in `NumpyAxesSelector.setAxisNames` to enforce unicity of names.

It also fixes the bug of the signal name being used by mistake as ylabel instead of the Y axis name , and adds support of a `@long_name` attribute on the signal dataset.


vallsv:
LGTM

PR #1384, opened 2017-12-06
remove dependency on qtconsole for debian8:

### LINKED ISSUE #1383: Nightly debian8 builds failed ....
ipython-qtconsole and ipython3-qtconsole are build/runtime dependencies.

Apparently they are not needed with ipython >2.x which is the case both for debian

kif:
Apparently this is not needed with recent ipython (>2)

close #1383


kif:
packaging builds (at least on edna-site)

PiRK:
From what I understand, in` IPython==2`  there was no need for qtconsole. For `Ipython>2`, `qtconsole` is a an additional dependency to get   `RichIJupyterWidget`. 

it's probably not a hard dependency to build the debian package, but it means the console tests will just be skipped.

The nightly silx log suggests that the `ipython-qtconsole` packages have been uninstalled on silx.org.
`
dpkg-checkbuilddeps: Unmet build dependencies: ipython-qtconsole ipython3-qtconsole`

kif:
On Wed, 06 Dec 2017 11:23:32 +0000 (UTC)
Pierre Knobel <notifications@github.com> wrote:

> The nightly silx log suggests that the `ipython-qtconsole` packages have been uninstalled on silx.org.

Indeed.

I am upgrading the documentation build to work (also) on notebook
(ipynb) files in addition to RST. This prevents the manual conversion
ipynb -> rst and uses a package available as part of jessie-backports,
using the recent version of ipython but incompatible with the old one
needed for qtconsole.

This way, there are fewer generated files in the git tree and the doc
looks better.

Cheers,

Jerome

PR #1387, opened 2017-12-07
Support fabio file series on FabioH5

vallsv:
Support file series from FabIO.

About the implementation:
- It only support 1 frame per images (else an assertion will be raised). We can change that but it have a pre-computation cost.
- It was not done to be efficient (but to be implemented fast), files are read 3 times. We can improve that, but let see what's happen on real examples.

I would like to re-implement file_series in fabio as a container of frames. It can create something much more convenient to use. In the mid time i hope the current implementation is enough.


vallsv:
Ready for a review

PR #1388, opened 2017-12-07
[convert] Handle file series as a multiframe file

### LINKED ISSUE #1127: [convert] Multiple edf files vs. multi-frame stack

PiRK:
Closes #1127 
Depends on #1387

### DIFF
--- a/silx/app/convert.py
+++ b/silx/app/convert.py
@@ -32,6 +32,8 @@ import numpy
 import re
 import time
 
+import silx.io.fabioh5
+
 
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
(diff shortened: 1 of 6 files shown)


PiRK:
Ready for review.

vasole:
Very useful. There are minor points but they will be addressed in separated issues.

PR #1397, opened 2017-12-08
[gui]Colormap dialog modal

### LINKED ISSUE #1192: [ColorMapDialog] modal to non modal ?
I would personally prefer to have a NonModal ColormapDialog.
This way users can 'play with the different colormap' and other tool more easily from my point of view and just keep the dialog open somewhere.

payno:
should fix #1192 

- [x] have a gui valid for modal and non-modal 
- [x] take an object in set colormap
- [x] listen to the colormap state
- [x] edit the colormap
- [x] check sync between two colormapdialog editing the same colormap
- [x] replace autoscale button and use one button 'auto' per boundary
- [x] unit tests
- [x] add colormap name to the combobox if not existing yet
- [x] update the colormapAction

### DIFF
--- a/silx/gui/plot/Colormap.py
+++ b/silx/gui/plot/Colormap.py
@@ -307,8 +307,8 @@ class Colormap(qt.QObject):
         """
         if vmin is not None and vmax is not None:
             if vmin >= vmax:
-                err = "Can't set vmin and vmax because vmin >= vmax"
-                err += "vmin = %s, vmax = %s" % (vmin, self._vmax)
+                err = "Can't set vmin and vmax because vmin >= vmax " \
+                      "vmin = %s, vmax = %s" % (vmin, vmax)
                 raise ValueError(err)
 
         self._vmin = vmin
--- a/silx/gui/plot/Colormap.py
+++ b/silx/gui/plot/Colormap.py
@@ -214,9 +214,9 @@ class Colormap(qt.QObject):
             value)
         """
         if vmin is not None:
-            if self._vmax is not None and vmin >= self._vmax:
-                err = "Can't set vmin because vmin >= vmax."
-                err += "vmin = %s, vmax = %s" % (vmin, self._vmax)
+            if self._vmax is not None and vmin > self._vmax:
+                err = "Can't set vmin because vmin >= vmax. " \
+                      "vmin = %s, vmax = %s" % (vmin, self._vmax)
                 raise ValueError(err)
 
         self._vmin = vmin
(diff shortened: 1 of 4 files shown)


payno:
I started the fix of #1192 
Two questions related:
   - What is your opinion about removing the `autoscale` and replace by two buttons (one for vmin and vmax). This would add a button (which is bad).  But give all option accessible for user. I think it would be better.
   - can we get rid of the old API in the sense replacing the setColormap function with the signature (name, normalization, autoscale, vmin, vmax) to simply give a colormap ?

t20100:
My answers (point by point):
- You could have a ComboBox with 'autoscale', 'manual', 'autoscale min.', 'autoscale max.'. Is autoscale on only one side actually needed?
- The ColormapDialog API is documented in the 'Internals' part of the plot, and there is little chance someone is directly using it. To me you can change it without taking care of the deprecation.
- We need to discuss what exactly the dialog is editing and how to sync it with the plot content.

vallsv:
I try with 2 dialog boxes, it looks like linear/log and min/max are not well synchronized. I haven't tried to check why.

Here is my test:
```
%gui qt
from PyQt4 import Qt
app = Qt.QApplication([])

from silx.gui.plot.Colormap import Colormap
colormap = Colormap(name='red', normalization='log', vmin=1., vmax=2.)

def createDialog(colormap):
    from silx.gui.plot.ColormapDialog import ColormapDialog
    dialog = ColormapDialog()
    dialog.setModal(False)
    dialog.setDataRange(1., 100.)
    dialog.setColormap(colormap)
    dialog.show()
    return dialog

dialog = createDialog(colormap)
dialog2 = createDialog(colormap)
```

vallsv:
It is also important to change the buttons label:
- For non-modal: Just `Close`. And if we want something to go back to the previous value (the value when the dialog was opened it should be `Reset`, and it should not close the dialog but only change the colormap)
- For modal: `Ok` is fine, but `Cancel` should reset the colormap to the previous value. I dont know if it is the case.

payno:
> I try with 2 dialog boxes, it looks like linear/log and min/max are not well synchronized. I haven't tried to check why.

thanks for the bug. Should be fixed now

> For non-modal: Just `Close` ... and `reset`

yes

> `Cancel` should reset the colormap to the previous value

It is the current behavior

payno:
Colormap names are not 'normalized' in the sense they can start by an upper case or a lower one.
Historically the colormapDialog was moving all first character in an upper case but was only dealing with names in lower case.
Now that we want to be able to deal with all possible colormap (not by default) this trigger an issue of normalization or not.

Anyway I can either:
- remove the colormap rename (by changing in some case the first character as an upper case)
- have a dictionary to make the association between the presented name (to the user ) and the real colormap name.

I would rather go for the first option since it will simplify (a few) the code and keep the original names. But I guess there was a good reason for changing those ?

t20100:
I don't know the initial reason, but to me capitalized names looks better in ComboBox and lower case names better in API..
`ColormapDialog` code can be simplified by using Python `str.title()` method which does exactly what is needed.

vallsv:
If the "titled" is only a matter of presentation in can be done by the combobox only for the presentation. Storing the lower case (or the real one) and using a non case sensitive to check and compare this names can be a good idea.

t20100:
You need to keep the real name which is not necessarily lower case.
I'm looking at this for #1411.

vallsv:
LGTM

PR #1400, opened 2017-12-08
[doc] Documentation for silx applications

PiRK:
Documentation pages for `silx view` and `silx convert`

 - [x] silx view
 - [x] silx convert (after #1388)


vallsv:
LGTM

Issue #1398, opened 2017-12-08
Random test fail with travis macOS python3 environment


t20100:
```
======================================================================

ERROR: testLoadSaveFit2D (silx.gui.plot.test.testMaskToolsWidget.TestMaskToolsWidget)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/testMaskToolsWidget.py", line 254, in testLoadSaveFit2D

    self.__loadSave("msk")

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/testMaskToolsWidget.py", line 231, in __loadSave

    self._drawPolygon()

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/testMaskToolsWidget.py", line 105, in _drawPolygon

    self.mouseMove(plot, pos=(0, 0))

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/utils.py", line 164, in mouseMove

    TestCaseQt.mouseMove(self, widget, pos=pos, delay=delay)

  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/contextlib.py", line 88, in __exit__

    next(self.gen)

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/utils.py", line 115, in _waitForMplEvent

    raise RuntimeError("MPL event %s expected but nothing received" % mplEventType)

RuntimeError: MPL event motion_notify_event expected but nothing received

======================================================================

ERROR: testLoadSaveNpy (silx.gui.plot.test.testMaskToolsWidget.TestMaskToolsWidget)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/testMaskToolsWidget.py", line 249, in testLoadSaveNpy

    self.__loadSave("npy")

  File "/Users/travis/build/silx-kit/silx/venv/lib/python3.6/site-packages/silx/gui/plot/test/testMaskToolsWidget.py", line 231, in __loadSave

    self._drawPolygon()

[...]

PR #1402, opened 2017-12-08
[CI] Always use Travis macOS xcode7.3 image

t20100:
Trying to workaround #1398


t20100:
Dismissed as not working

PR #1408, opened 2017-12-12
[test] Fix a poor test and provide a new context manager

PiRK:
This test was never executed, except in a virtual environment without h5py. 
This PR adds a context manager for testing missing optional dependencies, even if they are actually installed.

### DIFF
--- a/silx/test/utils.py
+++ b/silx/test/utils.py
@@ -222,10 +222,6 @@ def test_logging(logger=None, critical=None, error=None,
     return decorator
 
 
-
-
-
-
 # Temporary directory context #################################################
 
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


silx/test/utils.py
```diff
@@ -243,6 +239,57 @@ def temp_dir():
         shutil.rmtree(tmp_dir)
 
 
+# Simulate missing library context
+class EnsureImportError(object):
```
t20100: `ensure_import_error` should be better here as it is not in `silx.gui`.

silx/test/utils.py
```diff
@@ -243,6 +239,57 @@ def temp_dir():
...
+    """
+    def __init__(self, name):
+        """
+
+        :param str name: Name of module to be hidden (e.g. "h5py")
+        """
+        self.module_name = name
+
+    def __enter__(self):
+        """Simulate failed import by setting sys.modules[name]=None"""
```
t20100: You can simplify the `if else` with:
`self._backup = sys.modules.get(self.module_name)`

and use `if self._backup is None:` in `__exit__`.

silx/test/utils.py
PiRK: Even for a class? Is there a specific good practice for naming context manager classes ?

silx/test/utils.py
PiRK: I thought there could be a case where sys.modules[self.module_name] exists and is is already None, if someone already tried to import and it genuinely failed. Better not to delete it in such a case?
.

silx/test/utils.py
t20100: My bad, I read too fast.

I could not found a guideline for that... and looking at Python doc does not help, you find both CamelCase and `_` (especially with contextlib which allows to define a context manager with a function, but even with classes...):
- https://docs.python.org/3/library/contextlib.html#using-a-context-manager-as-a-function-decorator
- https://www.python.org/dev/peps/pep-0343/

silx/test/utils.py
t20100: OK, I didn't considered this case.

PR #1409, opened 2017-12-12
[nxdata] Support older NXdata specification

### LINKED ISSUE #1407: [View] NxData not recognized
Data produced by pyFAI-difmap not recognized/interpreted by "silx view"

PiRK:
Closes #1407 
This PR adds support for legacy NXdata specifications, if there is no `group@signal` attribute and/or no `group@axes` attribute.  

 - [x] support @signal attribute on dataset (only main signal @signal=1, other indices > 1 are ignored)
 - [x] support @axes attribute on signal dataset
 - [x] support @axis attribute on individual axis datasets

### DIFF
--- a/silx/io/nxdata.py
+++ b/silx/io/nxdata.py
@@ -491,7 +491,7 @@ class NXdata(object):
                 axes_dataset_names = []
                 numbers = [a[0] for a in numbered_names]
                 names = [a[1] for a in numbered_names]
-                for i in ndims:
+                for i in range(ndims):
                     if i in numbers:
                         axes_dataset_names.append(names[numbers.index(i)])
                     else:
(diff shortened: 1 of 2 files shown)


vallsv:
Let me know when it is ready, we also can test a dataset from soleil.

vasole:
@PiRK 

I would not ignore indices higher than 1. It is quite common to have multiple signals on the same axes by setting the signal attribute as 1,2,3,4,.... The typical case is showing several counters as function of a motor. That is also a feature that we have to consider to implement when exporting our plots to HDF5.

PiRK:
OK. Then should I add a new `signals` attribute, while keeping `signal` for the default/main signal?

vallsv:
Looks good.

Issue #1416, opened 2017-12-19
[io] is_specfile is not working for binary file in Python3


vallsv:
```
silx.io.specfile.is_specfile("debug_blank_mask.npy")

---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-4-7ac3925a4693> in <module>()
      1 for f in filenames:
      2     print(f)
----> 3     print(silx.io.specfile.is_specfile(f))
      4 

silx/io/specfile.pyx in silx.io.specfile.is_specfile()

/usr/local/lib/python3.5/codecs.py in decode(self, input, final)
    319         # decode input (taking the buffer into account)
    320         data = self.buffer + input
--> 321         (result, consumed) = self._buffer_decode(data, self.errors, final)
    322         # keep undecoded input until the next call
    323         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0x93 in position 0: invalid start byte

PR #1419, opened 2017-12-19
[io] is_specfile PY3 bug when reading binary data

PiRK:
This solves the immediate issue (#1416).
 
But the current test to find whether a file is a specfile is not very well suited for investigating large binary files.
The issue is that we assume the data is text, and we loop `for line in f:` to test for a `#f` or `#s` in the first 10 lines. This means that we might end up reading an entire data file to find out that it is not a Specfile, if there are not at least 10 `\n` characters in the file.


PiRK:
There is no perfect solution. I don't think that we can guarantee that the first  character (after stripping whitespaces) will always be a #. I also don't think we can assume that a specfile must only contain ascii characters, if we want to support files generated by third party scripts.

vasole:
There should not be any stripping of white spaces at the beginning of the lines. We support specfile and not something resembling it.

Concerning the problem. What about an fread of a maximum of 2500 bytes maximum and apply the 10 lines rule to it? For instance, one should only have to look for "\n#F" ot "\n#S" in that maximum of 2500 bytes.

vallsv:
It sounds reasonable

PiRK:
Done.

PR #1421, opened 2017-12-19
fix histogram action in the case contain some infinite value.

payno:
Using thomas function `min_max` instead of numpy.nanmin and numpy.nanmax


vallsv:
LGTM

PR #1431, opened 2017-12-21
[opencl] CBF compression

t20100:
This PR adds OpenCL compression to the `ByteOffset` class through an `encode` method with some tests.

There is still some TODOs (see in source file) for the API, reusing buffers, etc.. and a sample code.
Code in the test class can be reused across different tests.

### DIFF
--- a/doc/source/modules/opencl/codec_cbf.rst
+++ b/doc/source/modules/opencl/codec_cbf.rst
@@ -1,0 +1,9 @@
+
+.. currentmodule:: silx.opencl.codec
+
+:mod:`byte_offset`: Byte Offset compression/decompression
+---------------------------------------------------------
+
+.. automodule:: silx.opencl.codec.byte_offset
+   :members: ByteOffset
+   :show-inheritance:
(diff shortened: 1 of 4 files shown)


t20100:
I've added:
- an `encode_to_bytes` method
- support of pyopencl Array as argument of `encode*` methods
- record of OpenCL events
- use of semaphore
- sample code in docstring
- benchmark logs in tests + refactor tests to share code. 

One point still needs to be addressed: reusing buffers from one run to the next one and whether to share buffer with decompression..

t20100:
TODO: When an output buffer is given by the user, check if it is large enough and raise an exception if not.

t20100:
Added:
- raise exception if out buffer is too small and returns a sub-region array if out is provided.
- Reuse buffers (small performance improvement)
- Avoid device to device copy if out is not provided by creating a sub-region array.

Tested with Nvidia GPU and AMD CPU drivers on Debian 8.
On my machine, it is slower than fabio on CPU but 4x faster on GPU (including copy both ways).

t20100:
Latest updates:
- Added ByteOffset to the documentation
- Make `raw_size` and `dec_size` optional `__init__` arguments as they are not needed for compression and it avoids creating decompression buffers when performing compression.
  Only `dec_size` is mandatory for decompression.

PR #1434, opened 2018-01-03
[plot] Colormap

t20100:
This PR uses `Colormap.getColormapRange` method to initialize matplotlib colormap used in the backend.
This avoids copies of code doing the same stuff, and the code preparing matplolib colormap was doing it poorly regarding autoscale of only one boundary (which is now available from the colormap dialog).


vallsv:
LGTM

PR #1441, opened 2018-01-04
Improve colormap dialog

vallsv:
Few improvements on the colormap dialog before reworking the viewer.

- Reuse the concept of `setDataRange`
- Use the `setHistogram` API instead of curves
- Display the data range in the plot
- Provide an example to be able to test the dialog
- Provide feature to auto set convenient min/max values
- Rework the plot interaction

colormap dialog example

### DIFF
--- a/examples/colormapDialog.py
+++ b/examples/colormapDialog.py
@@ -1,0 +1,195 @@
+# coding: utf-8
+# /*##########################################################################
+#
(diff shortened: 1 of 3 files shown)


payno:
I think that the histogram of values and the range in the colormap dialog could be of different color with some transparency in order to increase comprehension when both are meld.

Otherwise look really nice.

payno:
LGTM

PR #1442, opened 2018-01-05
[plot SaveAction] Save as NXdata

PiRK:
This PR implements:

 - [x] save active curve as NXdata
 - [x] save active image as NXdata
 - [x] save active scatter as NXdata

TODO: 

 - [x] support saving graph title, and display it in our NXdataViewer 
 - [x] support saving and parsing NXdata with RGBA images
 - [x] find out why "hdf.nxs" is always appended to the file name, and fix it (in *nexusformat.NXobject.save()*).
  - [x] Add `@NX_class` =  `NXentry` / `NXroot` and `@default` attributes to generated files. 
 - [ ] In save dialog, add two checkboxes: *"Save this NXdata as default for this NXentry"* and *"Save NXentry as default for NXroot"*
   
  
- [ ] provide a way to add the generated NXdata to an existing NXentry in an existing HDF5 file.


PiRK:
It seems that there are no *nexpy/nexusformat* packages in Debian. Maybe we will have to move nexusformat out of `third_party/_local`, to make sure it's not deleted during packaging.

PiRK:
Status: 
A plot can now be saved as NXdata to a new file (saved as "/entry/data"), or an existing file. 
In case of saving to an existing file, I still need to implement a dialog to provide the output NXentry and the output NXdata groups. Currently a random entry is selected (the first one found) and a new NXdata group is created. Waiting for #1413.

Only the active curve, scatter or image is saved. If we want to implement a "save all curves" we need to discuss the output format: one NXdata per curve, or a single NXdata with multiple signals? If we want a single NXdata group per plot, how to handle curves with different number of samples, different axes...? Using multiple NXdata seems cleaner, but it does not allow to save the state of the plot (single plot with multiple curves -> multiple plots).

vasole:
We wanted to have a file dialogue allowing to specify HDF5 file and path for this release supporting the construction of a URI and also able to take it. The saving to existing HDF5 files needs it.

PiRK:
I think we should merge this PR and create an issue for the remaining TODOs (mainly add a HDF5 data selection dialog for writing the NXdata group at a specific location of an existing file). 
While its sitting there, it keeps accumulating conflicts because smaller PRs are being merged to improve silx.io.nxdata.

payno:
LGTM

PR #1444, opened 2018-01-05
[CI] Fix Travis-Linux download of OpenCL SDK

t20100:
Yet another update of AMD SDK download.


t20100:
This restores previous download URL, certificate seems to be renewed.

PR #1446, opened 2018-01-05
Fix popup checks on HDF5 tree

### LINKED ISSUE #1426: [silx view] Right click on the item tree looks to create an invisible popup menu
The next mouse click is disabled, as if a popup have to be closed first.
But no popup is visible.

vallsv:
Close #1426


payno:
LGTM

PR #1449, opened 2018-01-08
More checks on setup/bootstrap subprocess execution/result

vallsv:
It should help to catch faster error instead of spending time to reach it


payno:
LGTM

PR #1460, opened 2018-01-10
[plot] Fix console dock widget not showing

### LINKED ISSUE #1459: [plot] PlotWindow issue with dock widget
@mretegan catched an issue with the IPython console dock widget of `PlotWindow`.
To reproduce: create a `Plot1D` widget, open the legend dock widget from the option menu and then the console... which doesn't show.

t20100:
This PR fix an issue of `PlotWindow` which was not always shown when activated from the Option menu

Closes #1459

### DIFF
--- a/silx/gui/plot/PlotWindow.py
+++ b/silx/gui/plot/PlotWindow.py
@@ -315,8 +315,8 @@ class PlotWindow(PlotWidget):
                 custom_banner=banner,
                 parent=self)
             self.addTabbedDockWidget(self._consoleDockWidget)
-            self._consoleDockWidget.setVisible(True)
-            self._consoleDockWidget.visibilityChanged.connect(
+            #self._consoleDockWidget.setVisible(True)
+            self._consoleDockWidget.toggleViewAction().toggled.connect(
                 self.getConsoleAction().setChecked)
 
         self._consoleDockWidget.setVisible(isChecked)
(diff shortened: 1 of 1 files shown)


mretegan:
Works only the first time I enable the iPython console. If I disable it and then re-enable it it has the same issue as before.

t20100:
My bad, this should be a better fix.

It now registers to `QDockWidget.toggleViewAction` rather than `visibilityChanged` as the later also signals when the `QDockWidget` is shown but "hidden" behind another dock widget.

mretegan:
I works fine now. 

Not related to this, but a small cosmetic change. Would it be possible to add two spaces to the "Options" label on line 251. The down arrow overlaps with the text and is a bit annoying :grin:.

vallsv:
Sounds related to the style. I don't think it is a good idea to artificially tune that in this way.

mretegan:
I agree with you, it is style related, but it doesn't mean it looks nice by default. And this is not limited to macOS. Also native macOS apps don't really have such "tool buttons". In any case, I've ended up using css to change the style of it. Maybe we can talk again during the next code camp. From my point of view, the issue can be closed as it is.

vallsv:
I try to say that style related is not easy to fix, cause Qt support many different styles. For example here on Debian 8 using Cleanlooks style, i don't have this problem. But it exists about 10 styles, then it is not an easy.

BTW, i think to would be better to move this button into the toolbar on top. I am quite sure it makes more sens there. Which should also fix your style issue.

t20100:
Let's discuss it in a issue: I opened issue #1464 for moving option menu to the top

PiRK:
LGTM

PR #1463, opened 2018-01-11
[plot3d] Scene/parameter tree for SceneWidget

t20100:
This PR provides a `QTreeView` (in `ParamTreeView` module) for browsing `SceneWidget` content through a model.
To test it, use examples/plot3dSceneWindow.py.

I've made a few workarounds to make it work with PyQt5, PyQt4 and PySide.

It provides a single tree for both the scene content and the parameters of each objects.
@vallsv proposed to split both and it would be better, but I propose to do it in another PR (the model machinery can be updated to support this).

There are also a few remaining issues that I propose to solve separately:
- Setting globally properties of all scene items (e.g., colormap)
- Missing synchronization between Plot3DWidget properties (e.g., background color) and the tree.
- Object/properties not represented in the QTreeView (e.g., 3x3 matrix item transform, Mesh object, ...)


t20100:
Related to #1311

t20100:
Screenshot of the ParamTreeView:
paramtreeview

PiRK:
LGreatTM

PR #1466, opened 2018-01-11
make a simpler @creator, to avoid file leaks

PiRK:
Appending the complete command to the @creator attr can cause important file leaks when the output file is used to incrementally save a long series of scans.

This PR simplifies the creator to *"silx convert (v0.7.0-dev0)"*. In case multiple versions of silx are used to append to the same file (probably not a good idea), the creator can still be modified, but this will not happen often :  *"silx convert (v0.7.0-dev0); silx convert (v0.7.0)"*


vallsv:
LGTM

PR #940, opened 2017-07-07
[WIP] Doc for OpenCL modules

pierrepaleo:
This PR adds documentation of the API of the OpenCL modules.

PR #1467, opened 2018-01-12
[doc] Minimal bootstrap of silx.opencl doc

t20100:
This PR adds a minimalistic silx.opencl doc.
It is not complete, but provides a placeholder.

It builds over PR #940


kif:
OK for doc

PR #1472, opened 2018-01-15
[doc] Fix doc build warnings

t20100:
This PR fixes sphinx documentation build warnings (tested with Sphinx 1.2.3 and 1.6.5).

Most warnings were due to reference leading to 2 targets and a few typos.
In particular, with Sphinx 1.6.5:
- Using `list` for argument/returned values type had 2 targets in silx and linked to `specfilewrapper.list`.. I use `List` (and `List[..]`) instead as in Python3.5 type hinting.
- Using `function` instead of `callable` had the same issue.


vallsv:
LGTM

PR #1474, opened 2018-01-15
Better support for Python without importlib

vallsv:
The use of `__import__` was not working cause the behaviour between `importlib` and `__import__` is not the same.

It fix run_tests when using a very old Python (before 2.7)


payno:
LGTM

PR #1475, opened 2018-01-15
[DataViewer] view NXentry like a NXdata

### LINKED ISSUE #1471: [silx view] Support NXentry@default when present
We already support NeXus NXdata plots. However, we do not support a very handy feature of NeXus that is NX@default. 

NXentry foresees a "default" attribute to be set to the default NXdata associated to the entry.

Support of that feature can greatly improve user experience since at the top level of the tree the user can visualize the main information associated to an entry with very little interaction. That would reduce the amount of user interaction.

PiRK:
Closes #1471.
If a NXentry provides a @default pointing to a valid NXdata group, we can open it with an appropriate NXdata viewer.

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -97,8 +97,7 @@ class DataInfo(object):
         self.isComplex = False
         self.isBoolean = False
         self.isRecord = False
-        self.isNXdata = False
-        self.isNXentryWithNXdata = False
+        self.hasNXdata = False
         self.shape = tuple()
         self.dim = 0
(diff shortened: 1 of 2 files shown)


vasole:
It works. However, for completeness, the root level also accept an attribute "default" that tells the default entry and if that entry contains a "default" NXdata it would be great to visualize it. Can that be added or I open a new pull request?

PiRK:
I will do it in this PR.

PiRK:
Done.

vasole:
LGTM. Thanks.

PR #1479, opened 2018-01-16
[plot3d] Update scene parameter tree

t20100:
This PR adds a few features and a fix to plot3d parameter tree:

- Enable/disable scatter2D symbol and line properties depending on visualization mode
- Add display/edit of 3x3 matrix transform of nodes
- Separate autoscale for colormap min and max in the model (and the tree view)
- Add cutplane clip values below min parameter
- Fix cut plane interpolation parameter that was not properly synchronized.

Related to #1311


payno:
LGTM

PR #1486, opened 2018-01-17
Colormap read only

### LINKED ISSUE #1483: [colormap] Create a readonly property
The complex view expect a colormap fixed between `[-Pi, Pi]`, it would be very nice to create a colormap that the `ColormapDialog` can't edit.

Maybe it is a usecase that can be reused.

payno:
should close #1483


t20100:
Just a minor style comment: It is recommended to inherit user-defined exceptions from `Exception`: https://docs.python.org/3.6/library/exceptions.html#Exception.

Otherwise LGTM

vallsv:
LGTM

Issue #1493, opened 2018-01-19
[complexview] Colormap dialog histogram not consistant


vallsv:
The colormap histogram is not consistent when using a mix between amplitude and phase.

The colormap range is [-Pi, Pi], while the histogram is the amplitude.

PR #1496, opened 2018-01-19
[plot] Refactoring of ComplexImageView

t20100:
This PR provides a refactoring of `ComplexImageView` widget.
This provides:
- A new plot item: `ImageComplexData`
- The `ComplexImageView` uses this item and part of its API is now a proxy to this item.

This should ease treatment specific to complex (e.g., #1493)

### DIFF
--- a/silx/gui/plot/ComplexImageView.py
+++ b/silx/gui/plot/ComplexImageView.py
@@ -41,144 +41,11 @@ import numpy
 
 from .. import qt, icons
 from .PlotWindow import Plot2D
-from .Colormap import Colormap
 from . import items
+from .items import ImageComplexData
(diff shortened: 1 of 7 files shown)


t20100:
TODO: The `getData` needs caching

t20100:
Caching added to `getData`

vallsv:
LGTM

vallsv:
It would be better to have a reflexivity between `getData/setData`, but as an internal object it's fine.

t20100:
I'm not happy with `ImageComplexData.get|setData` either.
The only point for this solution is that it is compatible with `ImageData.getData` returning a float32 2D array... so with code using it.

PR #1499, opened 2018-01-22
[plot] Fix ColormapDialog for complex images

### LINKED ISSUE #1493: [complexview] Colormap dialog histogram not consistant

t20100:
Merge #1496 first.

This PR adds a specific case to the plot ColormapAction for complex images: If the image is in a mode combining a colormap for the phase and transparency for the amplitude, the histogram is set from the phase in the ColormapDialog.

closes #1493

### DIFF
--- a/silx/gui/plot/actions/control.py
+++ b/silx/gui/plot/actions/control.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2004-2017 European Synchrotron Radiation Facility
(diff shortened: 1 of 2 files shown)


vallsv:
Now the histogram looks good, but the displayed image looks wrong from amp+phase or log(amp)+phase.

t20100:
I've fixed the phase+amplitude display regression (introduced in PR #1496).

vallsv:
LGTM

PR #1500, opened 2018-01-22
[convert] no extra root level

### LINKED ISSUE #1498: [silx convert] Unnecessary group at top level created converting SPEC file
Using the command ``silx convert align.spec -o align.h5`` or ``silx convert align.spec`` the generated HDF5 file contains a top level directory named align.spec that should not be there.

PiRK:
Closes #1498.

Change default behavior : don't add an extra group at the root, unless explicitly requested. 
Remove `--no-root-group` option, add `--add-root-group` option.

After this PR, users should ensure that they are not merging scans with the same scan number.

At the moment the only proper way to merge EDF files is for file series, with the `--file-pattern` option. `silx convert *edf` should only be used with option `--add-root-group`. The same goes for appending EDF files to an existing file. This is because `fabioh5` currently assigns the name `scan_0` to all entries.

### DIFF
--- a/doc/source/applications/convert.rst
+++ b/doc/source/applications/convert.rst
@@ -26,12 +26,16 @@ Usage
 Options
 -------
 
+::
+
   input_files           Input files (EDF, TIFF, SPEC...). When specifying
                         multiple files, you cannot specify both fabio images
                         and SPEC files. Multiple SPEC files will simply be
                         concatenated, with one entry per scan. Multiple image
                         files will be merged into a single entry with a stack
                         of images.
+
+
   -h, --help            show this help message and exit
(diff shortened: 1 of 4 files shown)


vasole:
- ``silx convert *edf`` should have the same behavior as specifying a list of edf files by using --file-pattern
- we should check that the list of files have the same extension (it can be no extension) to make sure no different types are mixed and raise an exception as soon as possible
- as it is written now, if a file is missing in the file pattern, it is skipped with no error and no warning. At least one of the should be there (error or warning).

PiRK:
The only remaining issue is your last point: print an error when a number is missing in a file series. I'm still trying to figure out a good way to do that, without adding too much complexity to the code.

PR #1531, opened 2018-01-29
Load hdf5plugin before h5py

### LINKED ISSUE #1528: [silx view] hdf5plugin not installed message
Both, silx.io and fabio import h5py. Therefore when the application tries to import hdf5plugin, the later cannot be imported because h5py is already imported.

That is desired behavior when somebody is using the library, but not for the silx view application.

The try/except with the warning (or without) around the import of hdf5plugin should be moved before any of silx.io or fabio are imported.

vallsv:
Closes #1528


vasole:
LGTM. Thanks.

PR #1534, opened 2018-01-29
[convert & spech5] Use utf-8 text

PiRK:
This PR introduces a change in the way `spech5` exposes strings (bytes -> unicode) and in the way `convert` writes them to HDF5 (ASCII -> utf-8).

### DIFF
--- a/silx/app/convert.py
+++ b/silx/app/convert.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 # /*##########################################################################
-# Copyright (C) 2017 European Synchrotron Radiation Facility
+# Copyright (C) 2017-2018 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


PiRK:
Now, when we run a `silx convert` command, everything should written consistently in utf-8 in the output HDF5 file. When reading it back with h5py & PY3, we should see standard strings (unicode) rather than bytes.

```
 python bootstrap.py silx convert ../../data/31oct98.dat -o ../../data/31oct98.py2.h5
 python3 -m PyMca5.PyMca.PyMcaMain 31oct98.py2.h5
```
screenshot - 01302018 - 09 36 02 am


```
python3 bootstrap.py silx convert ../../data/31oct98.dat -o ../../data/31oct98.py3.h5
python3 -m PyMca5.PyMca.PyMcaMain 31oct98.py3.h5
```
screenshot from 2018-01-30 09-36-41

Before this PR, the scan title was bytes:
screenshot from 2018-01-30 09-37-59

vasole:
Just to be consistent, I guess the encode call should use utf-8 instead of ascii as argument.

PR #1536, opened 2018-01-30
[doc] IO tutorials

### LINKED ISSUE #1490: [doc] More documentation for silx.io.convert
There is already material in http://www.silx.org/doc/silx/latest/Tutorials/specfile_to_hdf5.html

We should split this tutorial and make the second part more generic, add a few more realistic examples, and add information about the new features (file series).

Would be good to have it before the next silx training, mid-march.

PiRK:
Reorganize IO tutorials, split _Reading SPEC as HDF5_ and _Converting various data files_.

Closes #1490


vallsv:
LGTM

PR #1538, opened 2018-01-30
[convert] fix bug with None @file_name

### LINKED ISSUE #1537: [convert] file series fail because of None in attrs
When merging a file series with *silx convert*, the process now fails because the "file" attribute *@filename* is *None*. 
This causes the following error:
`TypeError: Object dtype dtype('O') has no native HDF5 equivalent`

I think that previously it didn't crash because the was a coercion of all attributes to a dtype *numpy.string_*, so we ended up with a *@filename=b"None"*. 

I'm not sure if this is a `fabioh5` bug or a `silx.io.convert` bug.

In `silx.io.convert`, this patch fixes the issue:

```
diff --git a/silx/io/convert.py b/silx/io/convert.py
index 3af2c44..a2839fd 100644
--- a/silx/io/convert.py
+++ b/silx/io/convert.py
@@ -165,8 +165,9 @@ class Hdf5Writer(object):
         root_grp = h5f[self.h5path]
         for key in infile.attrs:
             if self.overwrite_data or key not in root_grp.attrs:
-                root_grp.attrs.create(key,
-                                      infile.attrs[key])
+                if infile.attrs[key] is not None:
+                    root_grp.attrs.create(key,
+                                          infile.attrs[key])
```
 
If we want to fix this In `fabioh5`, we could set the `@filename` attribute to "", or "None", or drop it altogether.

PiRK:
Closes #1537 

In a file series, don't fill the file_name attribute.

Also, add another layer of checking attribute types in `silx.io.convert`, to ensure they are encoded as utf-8 in the HDF5 output, so we can always read them back as unicode. This was not necessarily the case with `fabioh5` objects converted with python2.

### DIFF
--- a/silx/app/test/test_convert.py
+++ b/silx/app/test/test_convert.py
@@ -155,7 +155,7 @@ class TestConvertCommand(unittest.TestCase):
             if sys.version < '3.0':
                 title12 = title12.encode("utf-8")
             self.assertEqual(title12,
-                             "1 aaaaaa")
+                             "aaaaaa")
 
             creator = h5f.attrs.get("creator")
             self.assertIsNotNone(creator, "No creator attribute in NXroot group")
--- a/silx/io/convert.py
+++ b/silx/io/convert.py
@@ -52,9 +52,11 @@ Text strings are written to the HDF5 datasets as variable-length utf-8.
 """
 
 import logging
+import numpy
 
 import silx.io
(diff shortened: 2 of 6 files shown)


vasole:
- Drop the scan number from the title when converting SPEC files
- Use variable length (utf-8) string arrays for file_header and scan_header.

PR #1541, opened 2018-01-31
[CI] Xorg rather than Xvfb on travis linux

### LINKED ISSUE #1521: [CI] Travis-Linux python3.5/pyqt5 seg fault
There is a seg fault on Travis with python3.5 and pyqt5.10 (looks it was working with pyqt5.9.2).
From @vallsv investigation it arrives when creating the QApplication....

t20100:
This workaround a seg fault in Qt most probably due to a bug in Qt 5.10.0 when RANDR extension is missing.

Closes #1521


vallsv:
LGTM

Issue #1473, opened 2018-01-15
Grid Image view. Desirable feature in silx?


vasole:
Several beamlines (particularly ID13 but also ID21) have inquired about the possibility to display multiple images side by side.

The first question is of course is if we want that in silx or it is responsibility of the applications to achieve that as most applications take the easy solution of browsing a stack.

The point is that I was wondering if with the shared axes feature of silx it would be simple enough to get a table or a grid containing widgets to have simultaneous visualization of several images in such a way that when one zooms in one image the zoom is applied to all of them, the colormap could be changed to all of them, ...

t20100:
There is already a demo for axes synchronization: `examples/syncaxis.py`.
At first, we can add one that is closer to a real use case.

vallsv:
In case, that's something we could provide on `silx view` by selecting many datasets, for only some special cases like curves or images. But it really depend on there use case. If it match something really generic or not.

vasole:
The use case is an NXdata containing several images. For instance to visualize the different elemental maps (Fe, Co, Ni, ...) associated to a fluorescence map.

vasole:
Just to make it clear, to be able to toggle from the usual stack view of images to a grid of images.
It could be even implemented as an action with a pull down menu specifying the desired layout ( (2, 1), (2, 2), ...)

vasole:
There are two needs:

- Gallery of thumbnails 
- Side by side image comparison

vasole:
Question replaced by #2243

PR #1542, opened 2018-01-31
[WIP]GridImageWidget

PiRK:
addresses #1473 

TODO: 

 - [x] improve axis synchronization (issue when setting keepAspectRatio in one plot only)

Questions:

 - do we want to synchronize colormaps too? Or is it desirable to be able to compare the same frame in multiple colormaps side by side?
 - do we need all the Plot2D features, or should we use a more minimalistic plot widget?
    
     - should we get rid of the toolbars alltogether?
     - do we need the mask? do we want mask synchronization?
     - do we want the profile tools?
     - do we want the X Y Data widget at the bottom of each plot?
 
screenshot - 01312018 - 04 59 21 pm

### DIFF
--- a/examples/imageGrid.py
+++ b/examples/imageGrid.py
@@ -1,0 +1,18 @@
+
+
+import sys
+import numpy
+from silx.gui import qt
+from silx.gui.plot.GridImageWidget import GridImageWidget
+
+app = qt.QApplication(sys.argv)
+
+giw = GridImageWidget()
+giw.setNCols(3)
+giw.setNRows(2)
+
+giw.setFrames(numpy.random.rand(50, 150, 200))
+
+
+giw.show()
+app.exec_()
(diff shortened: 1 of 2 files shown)


PiRK:
Ideas for future improvement:

 - Ideally, there should be no plot specific toolbar, and less need to synchronize the plots' states (aspect ratio policy...) using their signals. The toolbar should be unique for the whole grid widget, and some actions should simply apply to all plots (zoom, pan, mask, showing colorbar...) while some other actions should allow to toggle/select whether they apply to a single plot or all plot (modifying colormap, profile in single image or stack mode...).
 - We need horizontal and vertical sliders in case there are too many plots to fit on the screen
 - a plot should only be updated if it is visible

vasole:
I would aim at the widget being an alternative display of a stack of images. In that sense, the functionality (and therefore the toolbar) should be the same as the one for a stack of images. In other words, we should be able to pass from one view to the other.

PiRK:
I'm closing this for now. It was just a proof of concept. It would require a lot more work to have something clean and usable.

PR #1550, opened 2018-02-05
[CI] Use a different version of pyopencl depending on wheel availability

t20100:
I made the wheel for pyopencl 2018.1 for python3.6, let's try to use it....


vallsv:
LGTM

PR #1558, opened 2018-02-06
[plot3d] Add bounding boxes to SceneWindow items

t20100:
This PR:
- Adds a bounding box to each data item.
- Separates text color from foreground color (used for bounding boxes)
- Updates the model to add bounding box and text color control in the tree view.
- Adds a convenient `Item3D.root` method to retrieve the root of a tree of item.

Related to #1311


t20100:
Added a fix of the bounding box drawing when the size on one dimension is zero.

t20100:
- Enabled bounding boxes for groups
- Added a group with axes, ticks and labels
- Moving some methods from SceneWidget to the scene root group.
- Fixed a problem with position of labels along the axes

Ready for review.

vallsv:
LGTM

Issue #1455, opened 2018-01-09
[plot] Mouse zooming is not always working in PlotWidgets


UUSim:
I'm using several `PlotWidgets`, which are integrated into a window. Zooming using the scroll wheel doesn't always work. Only after zooming using the left mouse button (i.e. drawing a zoom box), or panning using the keyboard, the zooming using the wheel works properly.

My initial though was that focus isn't propagated correctly, however the left and right mouse buttons work as expected, as does panning using the keyboard.

Do you have any idea how to fix this problem?

Btw: I'm using `PyQt5` and `opengl` as plot backend.

t20100:
Hi,

I would also had first looked at a focus problem, and as is I have no other idea.
Could you tell us on which platform you tested it?
Do you have some code to reproduce the issue?
Did you try with the matplotlib backend?

t20100:
BTW, zoom on wheel only works if the mouse in the area is plotted (it does not work when the mouse is in the plot margins).

UUSim:
I have investigated the issue more closely and it turns out focus isn't the problem in this issue. The mouse wheel event is always received by `PlotWidget.onMouseWheel`, but when it is forwarded to the event handler, nothing happens.

I found out that zooming only no longer works after right-clicking (which opens the context menu).
Upon right-clicking the `PlotWidget` enters the `RightClick` state (leaving `Idle` state), which doesn't propagate the mouse wheel event (for zooming).

Somehow, the `RightClick` state is not left when the right mouse button is released.
While trying the matplotlib backend I noticed that the state switches to `Idle` **when I move the mouse** after `RightClick` state is left.

This state switch originates from  `BackendMatplotlib.py, line 765 in _onMouseMove`.
The OpenGL backend is supposed to do the same, in `BackendOpenGL.py, line 391 in mouseMoveEvent`, but this function is never called.

<sub>My current workaround for going to `Idle` is by triggering another state switch by left-clicking.</sub>

Bottom line so far: 2 issues
* `RightClick` state is not left after releasing button (opengl backend)
* `RightClick` state is left only after the mouse is moved (matplotlib backend)

UUSim:
I've found a solution to the problem: enable mouse tracking for the `OpenGLWidget's` underlying `GLWidget`.
See: [QWidget api docs](http://doc.qt.io/qt-5/qwidget.html#mouseTracking-prop)

[...]

PR #1563, opened 2018-02-08
Fixed opengl backend not receiving mouse move events, causing state lockups

UUSim:
Fixes issue #1455.

Matplotlib backend always sends mouse move events. Opengl backend by default does not.
This fix configures Qt to send the mouse move events to the opengl backend.


t20100:
LGTM, thanks for the fix!

PR #1564, opened 2018-02-08
[plot] Removing right click management from ClickOrDrag interaction

t20100:
This PR removes right click management from plot interaction.
Handling of right click in interaction is no more needed since we decided to keep right click free for context menu and this was useless.

Related to #1455 and #1563.

### DIFF
--- a/silx/gui/plot/Interaction.py
+++ b/silx/gui/plot/Interaction.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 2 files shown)


t20100:
Thanks to @sole, this is actually not a good idea: even if all implemented interactions do not use right button, there is still some signals for it that PyMca is using, so closing for now.

vasole:
Just a word of caution.

One thing is that we do not implement any action on right click and a different one is to prevent users from catching events associated to right clicks.

PR #1575, opened 2018-02-09
[plot3d] Add selection support to SceneWidget

t20100:
Merge #1570 (cherry-picked here)

This PR adds a single item selection management to `SceneWidget` (through a `SceneSelection` object accessible with `SceneWidget.selection()`.
A corresponding selection model is available to sync the treeview selection with it.
Upon selection, Item bounding box is shown and color is changed by the selection management. I am not convinced by this mode but it is simple, so this can be disabled and the API leaves room to implement other modes of highlight.

Related to #1311

### DIFF
--- a/silx/gui/plot3d/ParamTreeView.py
+++ b/silx/gui/plot3d/ParamTreeView.py
@@ -46,6 +46,7 @@ from silx.third_party import six
 
 from .. import qt
 from ..widgets.FloatEdit import FloatEdit as _FloatEdit
+from ._model import visitQAbstractItemModel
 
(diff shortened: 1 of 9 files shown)


t20100:
By the way, item selection cannot be done by clicking on an item in the SceneWidget, it can only be done from the TreeView or through the API.

vallsv:
There is a conflict :-/ and I dont really know what to do.

vallsv:
Else it looks fine

t20100:
Conflict fixed

PR #1577, opened 2018-02-09
Fix colormap dialog range

### LINKED ISSUE #1572: [gui] Error while playing with colormap dialog
I try to move manually the min/max outside of the min of the histogram, and then i apply auto on the min to force this issue. Clicking back to manual create the issue (maybe the previous manual value was memorized, and the check min<max is not applied).

```
ERROR:silx.gui.qt._qt:<type 'exceptions.ValueError'> Can't set vmin and vmax because vmin >= vmax vmin = 0.00102488, vmax = -11.4602   File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/ColormapDialog.py", line 880, in _updateMinMax
    colormap.setVRange(vmin, vmax)
  File "/workspace/valls/silx.git/build/lib.linux-x86_64-2.7/silx/gui/plot/Colormap.py", line 344, in setVRange
    raise ValueError(err)
```

vallsv:
Closes #1572

### DIFF
--- a/silx/gui/plot/ColormapDialog.py
+++ b/silx/gui/plot/ColormapDialog.py
@@ -63,7 +63,7 @@ from __future__ import division
 
 __authors__ = ["V.A. Sole", "T. Vincent", "H. Payno"]
 __license__ = "MIT"
-__date__ = "19/01/2018"
+__date__ = "09/02/2018"
 
 
 import logging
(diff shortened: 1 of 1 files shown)


vallsv:
It could be also nice to rework the abstraction of the signals emitted by the min/maxValue, but it is not very critical.

PR #1593, opened 2018-02-13
fix PY3 bug when generating NXdata examples in hdf5widget

PiRK:
Usual issues with sequences of unicode not handled by h5py.

### DIFF
--- a/examples/hdf5widget.py
+++ b/examples/hdf5widget.py
@@ -325,7 +325,7 @@ def get_hdf5_with_nxdata():
     g2d0 = g2d.create_group("2D_regular_image")
     g2d0.attrs["NX_class"] = "NXdata"
     g2d0.attrs["signal"] = "image"
-    g2d0.attrs["auxiliary_signals"] = "image2", "image3"
+    g2d0.attrs["auxiliary_signals"] = str_attrs(["image2", "image3"])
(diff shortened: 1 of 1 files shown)


vasole:
You should prepend a "u" in front of all those attributes in order to make sure you get the same output in python2 and python3.

vasole:
Example:

``gd1.attrs["title"] = "x, y, values scatter with asymmetric y_errors"``

should become

``gd1.attrs["title"] = u"x, y, values scatter with asymmetric y_errors"``

PiRK:
It may work for single strings in this specific case, but it crashes with tuples of strings:
`gd1.attrs["axes"] = u"x", u"y"

It also crashes with single strings if, instead of assigning the value with `=` we use the `.create` method:
`gd1.attrs.create("title", u"x, y, values scatter with asymmetric y_errors"`

Initially all these strings had a `"b"` prepended, which worked in all these cases, but it produced ASCII encoding.

PiRK:
h5py is very picky with its implicit `unicode` -> `UTF-8` conversion. I found that the only method that works in absolutely all cases for writing utf-8 is 
```
gd1.attrs["axes"] = numpy.array(whatever,
                                                     dtype=h5py.special_dtype(vlen=six.text_type))
```

vasole:
I already reported that. I was explicitly referring to the single string case. Whenever one has more than one string one has to pass by the numpy.array with special type.

PR #1607, opened 2018-02-13
[fix #1592] Not a fix but a first step on solving #1592

### LINKED ISSUE #1592: [plot] Plot2D widget's ColorBar disappear
To reproduce:
```
from silx import sx
w = sx.imshow()
w.hide()
w.show()
```
and the colorbar is not visible.

payno:
[ColorbarDialog] Work around for the #1592 issue.
I don't know why but the show event when using the following code was called only once if we called had the call to self._viewAction.setChecked() in the `hideEvent()` and in `showEvent()`

### DIFF
--- a/silx/gui/plot/ColorBar.py
+++ b/silx/gui/plot/ColorBar.py
@@ -72,6 +72,7 @@ class ColorBarWidget(qt.QWidget):
         self._viewAction = None
         self._colormap = None
         self._data = None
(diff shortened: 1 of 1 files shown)


vallsv:
Looks strange. I understand the plot deconnection, but i don't see the point for other things. But i just check this diff.

vallsv:
Closed by #1627

PR #1616, opened 2018-02-14
[save_NXdata] write NX_class = "NXentry" when creating entry

### LINKED ISSUE #1595: Issue save as Nx data
/36145412-33315426-10a9-11e8-92f2-45215dd1dc74.png)
When opening this sino.edf (seems to not be limited to this file) file and saving it with 'image as NX data' it save it.

But when opening the generated file it give me the message 'NXroot group provides...malformed' at the root level as show in the image.

screenshot - 02132018 - 11 32 01 am

PiRK:
closes #1595 
This also forces the PY2 encoding of attrs to utf-8


vallsv:
LGTM

PR #1618, opened 2018-02-14
[DataViews] fix numpy 1.14 warning

PiRK:
`numpy.complex` is actually just a reference to the builtin python complex.  `numpy.complexfloat` is the correct ancestor to use when testing for complex data with `issubdtype`


vallsv:
LGTM

PR #1619, opened 2018-02-14
[plot] Improve backend picking

### LINKED ISSUE #1617: [plot gl] Slow down due to unnecessary picking
With the OpenGL backend (at least) there is a slow down when displaying many curves due to the fact that on mouse move, a complete picking is performed to update markers

t20100:
This PR changes plot backend picking API to avoid useless picking with OpenGL backend by filter by item kind before picking items rather than after...

closes #1617


vallsv:
LGTM

PR #1622, opened 2018-02-14
[NXDataView] update troubleshooting message in InvalidNXdataView

### LINKED ISSUE #1594: Label not updated in invalidNXdataView
When running `examples/hdf5widget.py`, creating NXdata examples, viewing `invalid/invalid NXentry` then `invalid/invalid NXdata`, the error message in the label is not properly updated (but it is updated when opening the groups in the opposite order).

PiRK:
The message must be reinitialized every time a new data is provided, because the view is reused between two invalid NXdata groups that have different issues. Closes  #1594


vallsv:
LGTM

PR #1624, opened 2018-02-14
Add pythonic callback to DataFileDialog to custom selectable data

### LINKED ISSUE #1599: [datafiledialog] It would be nice to allow to select an NXentry
Request from Manolo

vallsv:
Allow to specify even more the objects the dialog can return.

Can be used like that
```
def customFilter(obj):
    if "NX_class" in obj.attrs:
        return obj.attrs["NX_class"] in [b"NXentry", u"NXentry"]
    return False
dialog = DataFileDialog(self)
dialog.setFilterMode(DataFileDialog.FilterMode.ExistingGroup)
dialog.setFilterCallback(customFilter)
```

Closes #1599

### DIFF
--- a/examples/fileDialog.py
+++ b/examples/fileDialog.py
@@ -31,7 +31,7 @@ from __future__ import absolute_import
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "12/02/2018"
+__date__ = "14/02/2018"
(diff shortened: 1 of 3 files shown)


vallsv:
Here a PR with an example to only select NXentries. It can be useful for you @srio.

PR #1626, opened 2018-02-15
Fix icon examples

### LINKED ISSUE #1623: [sample code] animatedicons.py not working if mng is not supported

vallsv:
- Merge animated and non-animated icons into the same example
- `animatedIcons.py` could be used to help testing different kind of animation class, but i remove it
- It is there on the history, in case.

Closes #1623


vallsv:
There is not much changes in the icons.py one. I will remove the other one.

vallsv:
PR updated

PR #1631, opened 2018-02-19
[sx] Add ginput

### LINKED ISSUE #1614: [plot] Implement ginput method
The ginput function is quite convenient for simple interactive scripts.

https://matplotlib.org/devdocs/api/_as_gen/matplotlib.pyplot.ginput.html

t20100:
This PR adds `sx.ginput` that uses a PlotWidget to get a set of points given by the user.
The function plugs an object on a PlotWidget to handle the interaction and use its status bar to provide indication to the user.
It operates by default the most recent visible plot created with either `sx.plot` or `sx.inshow`.

It blocks until one of the following occurs: the user select all the required points, the PlotWidget gets hidden or a timeout is triggered.
It returns a list of (x,y) coordinates of all the points.
Points can be undo with either Delete, backspace or Ctrl-Z.
Points can only be set while in zoom mode.

I didn't succeeded so far in writing an interactive test for it... 

Closes  #1614

### DIFF
--- a/doc/source/modules/sx.rst
+++ b/doc/source/modules/sx.rst
@@ -14,6 +14,9 @@ The following functions plot curves and images with silx widgets:
 - :func:`plot` for curves
 - :func:`imshow` for images
 
+The :func:`ginput` function handles user selection on those widgets.
+
+
 .. note:: Those functions are not available from a notebook.
 
 :func:`plot`
@@ -26,6 +29,11 @@ The following functions plot curves and images with silx widgets:
 
 .. autofunction:: imshow
 
+
+:func:`ginput`
+++++++++++++++
+
+.. autofunction:: ginput
 
 3D plot functions
 -----------------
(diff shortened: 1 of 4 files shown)


vasole:
It would be convenient to give back a list of objects with compatibility with matplotlib.

If those objects are accessed as a list they get the same information as being just (x, y) but leaving the door open to provide more information (xPixel, yPixel, data, ...)

t20100:
I can provide objects that subclass tuple and give more information through additional getters

t20100:
Also need to fix support of draggable markers: selected points data does not get updated...

t20100:
It now returns a list of result objects that inherits from 2-tuple but provide additional information about the top-most item in the plot at that position through additional methods: `getItem`, `getData`, `getIndices`

I fixed consistency of returned positions by removing the fact that markers are draggable.
If draggable markers are needed, it can be done, but it is simpler this way and the user can always undo the last point.

I added how-to use examples in `ginput` docstring.

vasole:
Great.

I was expecting that when calling ginput, the widget to which ginput refers should be shown on top:

``
w.show()
w.raise_()
``

PR #1637, opened 2018-02-21
[nxdata] Improve logging

### LINKED ISSUE #1636: [NXdataView] No warning if old specification is met
I got the following report:

``
I confirm that with Silx version  0.7.0-dev0 the plot loads automatically when clicking the "plot_1" NXdata node.
Nevertheless, the console output shows 66 times the same warning:
WARNING:silx.io.nxdata:NXdata warning: NXdata group does not define a signal attr. Testing legacy specification.
``

Indeed, that WARNING should be an INFO and the complain should arrive if none of the NXdata specifications are met.

PiRK:
Closes #1636 

Log an info, and not a warning, when old spec is met. Now, warning is only printed in cases when `is_valid_NXdata()` returns `False`, to help troubleshooting.

I also improved the warning prefix, to mention the name of the problematic group, to avoid seeing the exact same message multiple time when opening different groups.

### DIFF
--- a/silx/io/nxdata.py
+++ b/silx/io/nxdata.py
@@ -62,6 +62,7 @@ def _nxdata_warning(msg, group_name=""):
     *"NXdata warning: "*
 
     :param str msg: Warning message
+    :param str group_name: Name of NXdata group this warning relates to
     """
     warning_prefix = "NXdata warning"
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM

PR #1643, opened 2018-02-21
[plot3d] Add alternative interaction when ctrl is pressed

### LINKED ISSUE #1131: [plot3d] Add a different mode on Ctrl pressed
As now interaction is only with left button, it would be nice to temporarily switch mode with modifier key, e.g., to have pan while Ctrl pressed when in rotate mode.

t20100:
This PR adds support of a second interaction when Ctrl  (command on macOS) is pressed:
- For scene rotation, Ctrl switches to scene panning
- For scene panning, Ctrl switches to scene rotation
- For plane panning (so far only available in ScalarFieldView), Ctrl switches to scene rotation.

closes #1131

### DIFF
--- a/silx/gui/plot/Interaction.py
+++ b/silx/gui/plot/Interaction.py
@@ -82,8 +82,6 @@ __date__ = "18/02/2016"
 
 import weakref
 
-from silx.third_party import enum
-
 
 # state machine ###############################################################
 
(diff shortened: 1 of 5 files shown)


vallsv:
Why don't you use id keys from Qt?

t20100:
`silx.gui.plot3d.scene` was so far not depending on Qt, but there is no real need for it, so I can use key codes from Qt.

t20100:
Updated key codes + refactor `setInteractiveMode` method in `Plot3DWidget` and `ScalarFieldView`

vallsv:
LGTM

PR #1645, opened 2018-02-21
[sx] Do not populate the global namespace with pylab

### LINKED ISSUE #1633: [sx] Module should not import pylab

t20100:
This PR is related to #1633, but instead of removing pylab, it enables it but disable `from numpy|matplotlib import *` which avoids confusion between namespaces...

closes #1633


vallsv:
LGTM

PR #1646, opened 2018-02-22
[DataViewer] Add a NXDATA_MODE id

PiRK:
This is needed to  be able to replace the view using `DataViewer.getViewFromModeId`, `DataViewer.removeView` and `DataViewer.addView`


vallsv:
Just few things:
- Remove the `#TODO`
- Call `__updateAvailableViews` always something was changed (from composite or not)

PR #1651, opened 2018-02-22
[plot3d] Add pan selected plane interaction to SceneWindow

t20100:
This PR adds support for pan interaction of the selected plane (clip plane or cut plane) in `SceneWidget`.
It adds a QAction in `SceneWindow` to toggle this interaction mode.
This QAction is not in the public API yet as I am not very happy with the way it is done, but the functionality is available:

The user can pan the selected plane in the parameter tree with the mouse.
For now, if the user press Ctrl and drag the mouse, it rotates the scene... In the future I might use Ctrl to interactively rotate the plane...

I also optimized the model structure as during plane interaction most of the time was spent in `QObject.parent()` up to the point that interaction was not smooth!


vallsv:
LGTM

PR #1652, opened 2018-02-22
[nxdata] bug when reading back unicode and testing for str type (PY2)

PiRK:
Found a bug when trying to run `examples/hdf5Widget.py`, `NXdata/1D_spectrum`. Single unicode string attr was not cast as list because I'm testing for type `str`.


vallsv:
It would be better to test strings using `six.string_types`, instead of doing that.
Anyway, let's merge it.

PR #1654, opened 2018-02-23
[fabioh5] interpret MCA data in EDF files as spectra

PiRK:


vasole:
@vallsv 

It is OK from my side.

PR #1659, opened 2018-02-26
Explicite clean up of list

vallsv:


vallsv:
This `clear` API is not available on Python2...

PR #1671, opened 2018-02-26
New example: plotUpdateFromThread2d.py

fangohr:
Based on plotUpdateFromThread.py, create 2d-fake data and plots as Image.

Fake data is 2d Gaussian peak that shifts slightly from update to update, superimposed with weak noise.

### DIFF
--- a/examples/plotUpdateFromThread2d.py
+++ b/examples/plotUpdateFromThread2d.py
@@ -1,0 +1,128 @@
+# coding: utf-8
+# /*##########################################################################
+#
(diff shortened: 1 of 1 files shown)


t20100:
Thanks for the sample code!

If you provide `resetzoom=False` to `addImage`, then it is possible to zoom in/out and pan the image while it is updated, otherwise zoom is reset each time the image is updated.
Then you need to call `Plot2D.resetZoom()` after displaying the first image to set the limits of the plot to that of the image.

PR #1678, opened 2018-02-28
[plot] Change addimage replace argument default value to False

### LINKED ISSUE #1555: [plot] addImage replace=True default argument is missleading
I keep forgetting about the `replace=True` in `PlotWidget.addImage` and get missleaded when displaying more than one image in the plot.

Now that there is the plot object oriented API and a `setData` on the image object, can't we drop the `replace=True` argument ?

t20100:
This PR change addImage replace argument default value and update code accordingly.

closes #1555

### DIFF
--- a/examples/scatterMask.py
+++ b/examples/scatterMask.py
@@ -109,7 +109,7 @@ class MaskScatterWidget(qt.QMainWindow):
         self._plot.addImage(image, legend=self._bgImageLegend,
                             origin=(xscale[0], yscale[0]),
                             scale=(xscale[1], yscale[1]),
-                            z=0, replace=False,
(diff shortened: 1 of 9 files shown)


vallsv:
Just a small comment about the change on `NXdataWidgets`: `replace=True` is done anyway if the `legend` already exists no? `replace` can be removed no?

t20100:
@vallsv, `replace=True` and using the same `legend` is not the same if there is multiple images in the plot...
I haven't checked the whole class, so I took a conservative approach.
If there is just one image, yes it can be removed.

vallsv:
Let's merge then

Issue #1035, opened 2017-08-14
[debian] pyqt5 should take precedence over pyside


picca:
Hello, I am working on the jessie-backports of the Debian packaging.

during the jessie build, pyqt5 AND pyside are installed.

Then I end up wit this error

```
INFO:silx.gui.qt._qt:PySide.QtOpenGL not available
Traceback (most recent call last):
  File "run_tests.py", line 369, in <module>
    unittest.defaultTestLoader.loadTestsFromNames(options.test_name))
  File "/usr/lib/python2.7/unittest/loader.py", line 130, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python2.7/unittest/loader.py", line 115, in loadTestsFromName
    test = obj()
  File "/<<BUILDDIR>>/silx-0.5.0+dfsg/.pybuild/pythonX.Y_2.7/build/silx/test/__init__.py", line 49, in suite
    from . import test_sx
  File "/<<BUILDDIR>>/silx-0.5.0+dfsg/.pybuild/pythonX.Y_2.7/build/silx/test/test_sx.py", line 73, in <module>
    from silx.gui import qt
  File "/<<BUILDDIR>>/silx-0.5.0+dfsg/.pybuild/pythonX.Y_2.7/build/silx/gui/qt/__init__.py", line 54, in <module>
    from ._qt import *  # noqa
  File "/<<BUILDDIR>>/silx-0.5.0+dfsg/.pybuild/pythonX.Y_2.7/build/silx/gui/qt/_qt.py", line 164, in <module>
    from ._pyside_dynamic import loadUi  # noqa
  File "/<<BUILDDIR>>/silx-0.5.0+dfsg/.pybuild/pythonX.Y_2.7/build/silx/gui/qt/_pyside_dynamic.py", line 44, in <module>
    from PySide.QtUiTools import QUiLoader
ImportError: No module named QtUiTools
E: pybuild pybuild:256: test: plugin custom failed with: exit code=1: env PYTHONPATH=/<<BUILDDIR>>/silx-0.5.0+dfsg/.pybuild/pythonX.Y_2.7/build GPU=False SILX_OPENCL=False SILX_TEST_LAW_MEM=True xvfb-run -a --server-args="-screen 0 1024x768x24" python2.7 run_tests.py -v
```

Will it be possible for silx to prefer pyqt5 over pyside.

vallsv:
[...]

PR #1686, opened 2018-02-28
[gui] Change PyQt preference order

t20100:
This PR tries to load PyQt5 first, then PyQt4, PySide and PySide2.

Doc is updated accordingly to acknowledge pitfalls with IPython.
sx module is also updated to perform imports as %pylab but to avoid the initialization of Qt by IPython.


t20100:
closes #1035

vallsv:
LGTM

PR #1690, opened 2018-02-28
[doc] Add image update from thread sample code to doc

t20100:
This PR adds the new sample code plotUpdateImageFromThread.py to the documentation.
It also rename curve and image update from thread sample code to have more meaningful names


vallsv:
LGTM

PR #1691, opened 2018-02-28
[build] Avoid symbol collision with manylinux1

### LINKED ISSUE #1070: [build] manylinux build subject to symbol name collision
When generating manylinux wheels, by default the C symbols within an extension looks to be resolved globally (so you can overload a function in a python extension with an already loaded function with the same name.
On Debian Python build is using `-Bsymbolic-functions` ld option which is not available in manylinux OS.

This problem occured in pyFAI where `crc32` C function of pyFAI was replaced by the zlib crc32 function: https://github.com/silx-kit/pyFAI/issues/649

numpy has NPY_NO_EXPORT and NPY_VISIBILITY_HIDDEN C macros to deal with this kind of issue.

We should take care of this or we probably end-up with funny seg faults.

t20100:
This PR changes build option on unix to avoid possible symbol collision with manylinux build.
It is a copy-paste from pyFAI setup.py

closes #1070

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 # coding: utf8
 # /*##########################################################################
 #
-# Copyright (c) 2015-2017 European Synchrotron Radiation Facility
+# Copyright (c) 2015-2018 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


kif:
In pyFAI we had to use different options for python 2 & 3 ... how did you manage this ?

kif:
LGTM

PR #1696, opened 2018-03-01
Make sure the spectrum and channels length match.

### LINKED ISSUE #1692: [spech5] Silx convert of MCAs may generate files with wrong number of channels
When converting a SPEC file with MCAs, the channels field and the data field do not have the same number of channels.

vasole:
Closes #1692


t20100:
Closed as superseeded by PR #1697

PR #1697, opened 2018-03-01
[0.7][spec] spech5 mca bug

### LINKED ISSUE #1692: [spech5] Silx convert of MCAs may generate files with wrong number of channels

t20100:
Superseed PR #1696 (It is the same, but on 0.7 branch)
Closes #1692


vallsv:
This PR breaks the spec unittests

vallsv:
LGTM

PR #1700, opened 2018-03-01
[gui] Add a shared QPrinter instance in silx.gui

t20100:
This PR adds a `silx.gui.printer` module with functions to access a shared `QPrinter` instance.
It is used in `PrintAction` in `plot` and `plot3d` and in `PrintPreview`.

It also switch from `printer` property to `getPrinter` method in plot PrintAction.

To make `PrintAction` more generic, it is possible to add a `setPrinter` in order to allow using a different printer from the default one of silx. If desired, let me know.


vallsv:
LGTM

PR #1702, opened 2018-03-02
[plot] Split PlotWindow toolbar and refactor silx.gui.plot.PlotTools

### LINKED ISSUE #1162: [plot] Split PlotWindow toolbar to smaller toolbars

t20100:
This PR adds a new `silx.gui.plot.tools` package.
Content of `silx.gui.plot.PlotTools` has moved there and is now depreacted.
It also contains a few "ready to use" toolbars:
- for interactive modes
- for copy, save, print
- for basic plot config for curves
- for basic plot config for images

The first 2 toolbars are used in `PlotWindow`.
It is not possible to further refactor `PlotWindow`, `Plot1D` and `Plot2D` toolbars without breaking the API and I don't think it worth it.

`examples/plotWidget.py` sample code was update to illustrate how to create an ad hoc PlotWindow-like widget by reusing toolbars, actions and the colorbar.

closes  #1162


vallsv:
LGTM

PR #1712, opened 2018-03-05
[CI] Fix travis-macOS CI env

t20100:
This should fix macOS build, broken after update of brew making python3 the default.


vallsv:
Still not working. I restarted the process on Mac, just in case

vallsv:
LGTM

PR #1714, opened 2018-03-05
[plot] rework ROI curve widget with object (Fix 1711)

payno:
- [x]  add the ROI Object
- [x] reset the API
- [x] move listener into the ROITable instead of the CurvesROIWidget
- [x] <s>Move the ROI object somewhere else ?</s>

The roles between CurvesROIWidget and ROITable are:

* ROITable: display ROIs and insure connection with the plot (curves) and rois
* CurvesROIWIdget define behavior to have and clicking 'add ROI' and 'remove ROI'

### DIFF
--- a/doc/source/modules/gui/plot/roi.rst
+++ b/doc/source/modules/gui/plot/roi.rst
@@ -1,11 +1,14 @@
-.. currentmodule:: silx.gui.plot
+.. currentmodule:: silx.gui.plot.CurvesROIWidget
 
 :mod:`CurvesROIWidget`: ROI from curves
 =======================================
 
+
 .. |roiWidgetImage| image:: img/roiwidget.png
    :height: 400px
    :align: middle
+
+.. automodule:: silx.gui.plot.CurvesROIWidget
 
 You can access to the ROIWidget from a Plot window by :
 
@@ -14,4 +17,28 @@ You can access to the ROIWidget from a Plot window by :
 
 |roiWidgetImage|
 
-.. automodule:: silx.gui.plot.CurvesROIWidget
+
+
+
+:class:`ROI` class
+------------------
+
+.. autoclass:: ROI
+   :show-inheritance:
+   :members:
+
+
+:class:`CurvesROIWidget` class
+-------------------------------
+
+.. autoclass:: CurvesROIWidget
+   :show-inheritance:
+   :members:
+
+
+:class:`ROITable` class
+-----------------------
+
+.. autoclass:: ROITable
+   :show-inheritance:
+   :members:
--- a/silx/gui/plot/CurvesROIWidget.py
+++ b/silx/gui/plot/CurvesROIWidget.py
@@ -159,6 +159,7 @@ class CurvesROIWidget(qt.QWidget):
         self.loadButton.clicked.connect(self._load)
         self.saveButton.clicked.connect(self._save)
 
+<<<<<<< HEAD
         self._middleROIMarkerFlag = False
         self._isConnected = False  # True if connected to plot signals
         self._isInit = False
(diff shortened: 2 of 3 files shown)


payno:
PR #1713 should be merged before.

payno:
- [ ]  add the ROI Object
- [ ] reset the API
- [ ] move listener into the ROITable instead of the CurvesROIWidget
- [ ] move the marker management into the ROI object ?

payno:
All the 'standard ROI functionalities' have been moved to the ROITable.
The CurvesROIWidget only contains the behavior to have when requesting for a new ROI according to the existing ROIs.

current behavior modifications:

CurvesROIWidget + ROITable
- getRois return the dict of ROI **Object**
- setRois require a list of ROI object. If is a dict of ROI will deal with it but signal the deprecation warning

ROITable:

Removed:
    - fillFromROIDict function
    - nameslot
Deprecated:

    - getROIListAndDict. Even if deprecated returning ROI object

Added:
    - getRois
   - save
   - load
   - showAllMarkers
   - setMiddleROIMarkerFlag

In CurvesROIWidget I would like to remove the calculateROIs which has been deprecated since the v

    @deprecation.deprecated(replacement="calculateRois",
                            reason="CamelCase convention")
    def calculateROIs(self, *args, **kw):
        self.calculateRois(*args, **kw)

payno:
Should the ROI object link to markers objects representing him (could avoid duplication of information such as draggable, color....) ?

payno:
There is some approximation actually made by the calculation of netCounts and Raw Counts. 

For example in the case of the following curve : 

roi

We have RawCounts giving 8.0 and netCounts giving 2.0 instead of 4.0 and 1.0

I think you can change it to call the numpy.trapz. This shouldn't be too long no ?
What to you think @vasole, @t20100 

The new algorithm (first draw on the commit 3991be0db92090d0d5877623098037d5c5debbf7 ) would be something like:
- compute the area of the curve in the ROI (between the Xs included in the ROI)
- compute the integral between the quadrilateral defined by the curve intersection, ROI boundaries and the x axis to get the background.
- actually there is no interpolation made to get the correct y value of the intersection between the curve and the ROI boundaries. We are simply taking the closest. That might be interesting to add a few more operation here no ?

payno:
This PR is close to be complete.
Two questions remains:

- should we have to CurvesRoiWidget to insure full compatibility
- I think the ROI should be moved out from the widget but where can we add it. Consecutively what could be merge between the ROI '1D' and the newest ROI '2D' from @t20100 to have at least a common base class even if not a lot seem generic.

payno:
Now classes look like:


diagcurvesroiwidget

payno:
Maybe to generalize for n d we could have something like:

diagramcurvesroiwidget2

vasole:
A word of caution. In the example you give, the correct results (as of today) should be 8 for the raw counts and 4 for the net counts.

I'm not against having different algorithms available. It can be useful to cross check fit results for instance.

payno:
Agree for the different algorithm (at least having areas). For the net counts it gives 2.0 which is correct from my point of view because raw is 8 and background is 6*1.

vasole:
@payno 

``For the net counts it gives 2.0 which is correct from my point of view because raw is 8 and background is 6*1.``

Correct, the contribution of each channel at 2 is 2-1.

payno:
I think having a ROI is good but we should discuss with @t20100 to define 1D/2D/3D ROI common base class / API

t20100:
Yes, if it's possible, it would make sense to have similar API for all ROIs.

payno:
@t20100 :
After looking the `RegionOfInterest` Object I guess we can start with a simple class (_BaseRegionOfInterest) for example which will be something like:

_BaseRegionOfInterest(qt.QObject) containing:

- QSignal sigChanged (or sigRegionChanged) as you prefer
- name (which will replace the curren t_label from RegionOfInterest)
- self._editable
- we might also add the _id that I am using in the 1D ROI

PR #1716, opened 2018-03-06
[plot] Unselect active curve when clicking on empty space

### LINKED ISSUE #1612: [plot] Deselecting a curve
Very often I click on some "empty" region of the plot canvas to disable the active curve. Would it be possible to implement this behaviour?

t20100:
This PR allows to unselect active curve with a left click in empty area of the plot.
It does not interfere with drag operation

closes #1612


PiRK:
It seems to be working.

PR #1722, opened 2018-03-07
[plot] Remove save RGB image as TIFF from SaveAction

t20100:
This PR removes the option to save a data image as a RGB TIFF from the plot SaveAction.
There is still an option to save the image data as TIFF, this PR only removes the option to save an image with colormap applied as TIFF.

It was using Qt to save such image and TIFF support is optional... so it might not be there
As there is still an option to save an image with colormap applied as RGB PNG (always available in Qt), I don't see it as a big loss of functionality...
Is it OK to drop it?


vasole:
It is OK with me.

t20100:
If this is not missing to anyone, then ready for review...

Issue #1119, opened 2017-09-12
[plot] Add stats information.


vasole:
Beamlines require simple statistics available (max, min, std, ...) for curves and images.

payno:
What stats do the user need ?
- min 
- max
- std
- ???

vasole:
At the very least what PyMca is already offering. For instance, position of the max, center of mass, ...

t20100:
closed by #1723

PR #1723, opened 2018-03-07
[plot] Curves stats widget

payno:
Deal with stats:

   - [x] min (+ index and coordinates)
   - [x] max (+ index and coordinates)
   - [x] std
   - [x] mean
   - [x] delta (max - min)
   - [x] COM. To be checked for Scatter.
   - [x] add unit test for those stats

API

   - [x] Define API for defining new API and format.

Doc
   - [x] Add sphinx documentation
   - [ ] <s>Create tutorial for defining new Stat </s>   # I guess an example is enough
   - [x] add an example on how to define your own stats


payno:
Relative to #1119.

payno:
This has been made using a QTableWidget.

This way we can display stats on several curves, images... or only the active one.
We are taking the legend as the ID without taking into account the kind. So if a legend is duplicated will probably fail in one way or another but the kind should be removed from API for the next release no ? 

The interface change from PyMCA (not a dedicated widget for the active item). If this modification is an issue we can create a second widget displaying the same information more PyMCA looks like

payno:
tmp

payno:
PyMCA like widget:

tmp

vasole:
The table seems more extendible.

vasole:
Table should use the version allowing cell copying in silx.gui.widgets

payno:
Should we always display the 'kind' column in the StatsTable or only when a conflict on legend can appear ?
I think the second option is the best one. Will save some space

t20100:
+1

payno:
Actually when a row is selected on the Table we are updating the current item to show the correspondent item to the user.
But I am not sure we wan't this behavior or we might want to turn it off in some case ?

vasole:
I do not anticipate any major issue with that behavior. Any other opinion against or in favor?

PiRK:
I like it.

t20100:
Sounds good, anyway, a method to toggle it can be added later.

payno:
I think stats should be moved to math.

silx/gui/plot/StatsWidget.py
```diff
@@ -0,0 +1,516 @@
...
+                "When visible data is activated we will process to a simple "
+                "filtering of visible data by the user.\nThe filtering is a "
+                "simple data sub-sampling.\nNo interpolation is made to fit "
+                "data to boundaries."
+            )
+
+    def __init__(self, parent=None, plot=None, stats=None):
+        self._stats = stats
+        if stats is None:
+            self._stats = stats
```
t20100: This `if` statement has no effect

t20100:
A more generic API than `getAllHistograms` and `getAllScatters` should come soon (See issue #1826), so in the mean time I would use `_getItems`.

silx/gui/plot/stats/stats.py
```diff
@@ -0,0 +1,447 @@
...
+from silx.gui.plot.items.image import ImageBase as ImageItem
+from silx.gui.plot.items.scatter import Scatter as ScatterItem
+from silx.gui.plot.items.histogram import Histogram as HistogramItem
+from silx.math.combo import min_max
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class Stats(dict):
```
t20100: Probably better to use an `OrderedDict` to keep the order statistics were added to it

payno:
statswidget

PR #1725, opened 2018-03-09
[plot3d] Fix model when colormap has no name

t20100:
The case where the colormap has no name (i.e., defined with a color LUT) was not handled in the parameter tree model


payno:
LGTM

PR #1727, opened 2018-03-12
[CI] Add pinned requirements file

t20100:
This PR proposes to add a file `ci/requirements-pinned.txt` for temporary version pinning or additional dependencies.

As we test with pre-released version of our dependencies, we regularly fall into issues with such versions.
While this is the purpose of testing, the idea here is to **temporarily** pin a version/add a dependency in a single file until the issue is fixed upstream.


vallsv:
LGTM

Issue #1740, opened 2018-03-20
wheels does not claim the minimal version of numpy required


jonwright:
In [3]: silx.io.specfile.SpecFile("31oct98.dat")
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-8e08275efa8d> in <module>()
----> 1 silx.io.specfile.SpecFile("31oct98.dat")

/home/jon/Downloads/silx-training/silx/io/specfile.pyx in silx.io.specfile.SpecFile.__cinit__()

NameError: name 'SF_ERR_NO_ERRORS' is not defined

(xenial)jon@localhost:~/Downloads/silx-training$ python    
Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import silx
>>> silx.version
'0.7.0'

jonwright:
Seems to be a problem of the pypi wheel not matching the xenial system numpy. Fixed by recompiling silx.

kif:
On Tue, 20 Mar 2018 04:09:53 -0700
Jon Wright <notifications@github.com> wrote:

> Seems to be a problem of the pypi wheel not matching the xenial system numpy. Fixed by recompiling silx.

Wheels are built to be installed on "up-to-date" systems (typically
virtual-envs where you can make a "pip install --upgrade").

Cheers,

-- 
Jérôme Kieffer

jonwright:
In which case there is a bug in the documentation: 
http://www.silx.org/doc/silx/latest/install.html
"Linux
If NumPy is not installed on your system, you need to install it first, preferably with the package manager of your system"

t20100:
As the wheels impose a specific version of `numpy`, they should claim it in the `install_requires`.

vasole:
Why do they impose a particular version of numpy?

vasole:
Besides that, it seems we have an error on the handling of an error....

PiRK:
> Besides that, it seems we have an error on the handling of an error....

I had a quick look, and I don't see any obvious problem. `SF_ERR_NO_ERRORS` is defined as a module attribute, so unless there is a subtlety I don't understand for *cdef* classes,  *SF_ERR_NO_ERRORS* should be available in `__cinit__`.

t20100:
A few tests:
[...]

PR #1749, opened 2018-03-22
[cython] Minimize use of cimport numpy

t20100:
This PR cleans up usage of `cimport numpy` by:
- removing unnecessary numpy includes
- always using `cimport numpy as cnumpy` to identify usage (when mixed with `import numpy`)
- avoid using `cnumpy.ndarray` as argument/variable type

Related to #1740.

This PR does NOT change `silx/third_party/scipy_spatial`.


PiRK:
LGTM

PR #1752, opened 2018-03-22
[GroupDialog] dialog to select a group in a HDF5 tree

PiRK:
Todo list :
 - [x] implement dialog
 - [x] use it in `SaveAction` (save as NXdata to an existing group)
 - [x] screenshot of widget to gallery


PiRK:
- [ ] Add a line edit to specify the name (path relative to the currently selected group) of a new group to be created. Use case: specify the name of the NXdata group to be created in an existing entry.

vallsv:
Here is few comments:
- The example is not fixed with the last API
- You should use the same API as the already existing dialogs (for exaple: `selectedDataUrl` for the object, or `selectedUrl` for the string)

vallsv:
I would prefer a common API with other dialogs, but LGTM.

PR #1758, opened 2018-03-27
[plot] Add class to handle interactive selection

t20100:
This PR adds a class `InteractiveSelection` to the plot tools.
This class controls interaction to request input from the user. The input can be points, rectangles, polygons, lines , horizontal and vertical lines.
It is possible to undo an input with Ctrl-Z.

It also change the structure of tests by adding a silx.gui.plot.tools.test package.

This is a first step, that it is possible to extend in different ways (not for this PR, but here to provide a list of ideas):
- Implement redo
- [Done] Add edition of input points/shapes
- [Done] Add text markers to each shape
- [Done] Allow to give the text to print in for each marker
- Return 'rich' information by performing a picking under the points (or what is inside shapes)
- [Done] Use it in `sx.ginput`
- Snapping to curves


t20100:
Added:
- Add edition of input points/shapes
- Add text markers to each shape
- Allow to give the text to print in for each marker

t20100:
Added a Table widget to view the current selection and manipulate it.
API updated.

vallsv:
Just few comments:
- The architecture of `InteractiveSelection` looks good to me as manager of ROIs
- It would be good to be able to have a `addSelection` with a single object as argument (already created by the user)
- Maybe you could rename `addSelection` into `createInteractiveSelection`?
- I am not very comfortable with `updateStatusMessage` and the message signals. Maybe because this class also manage the creation or ROIs while it should maybe not be it's responsibility. But i dont have clear idea, there's maybe nothing better.

vallsv:
Is that for silx 0.8?

t20100:
Hopefully yes...

t20100:
- Added use in `sx.ginput`
- Splitted class into a selection/ROI manager and stuff more dedicated to be used from the console as a blocking function.

t20100:
I renamed things from `Selection` to `RegionOfInterest` and did a few fixes and changes so that I can use it for the profile interaction.

As it is now, there is a single RegionOfInterest class for all kind of shape (rectangle, polygon...). It would be more convenient to have one subclass for each, but for me that can wait.

Ready for review.

t20100:
BTW, PR #1870 should be merged first.

vallsv:
Few comment about the GUI.
- I think it would be good to set a row selection to the tableview `setSelectionBehavior(qt.QAbstractItemView.SelectRows)`, or no selection at all
- And it would be better to move the icons as toolbar on top of the widget
- And remove the title "Add ROIs", but add little "+" in the icons.

Let me know if you want a patch.

t20100:
Yes wouldn't mind a patch

vallsv:
- Typo `@ enum.unique` -> `@enum.unique`?
- It could be good to split the widget (widget, toolbars...) and the model (manager and object) into 2 modules (it's a big one)
- The delete icon could go to the end of the row
- The delete column could not have the title (it can be a smaller column without)
- Add a tooltip to the delete button

vallsv:
I will send a patch with another PR.
LGTM.

PR #1766, opened 2018-04-03
Add a marching squares algorithm to silx.image

vallsv:
This PR provides a marching squares algorithm to `silx`. `skimage` also provides a marching squares algorithm but without the expected features. It is based on a first implementation provided in `pyFAI`, with improvement to use OpenMP and reordering points in polygons at a lower cost.

Features
=======

Some features was needed by pyFAI in order to speed up calibration and calibration rendering.

- Provides a compatible API with `skimage` (`silx.image.marchingsquares.find_contours(image, level)`)
- An efficient implementation  (for common images this impl is x2 faster than skimage)
- Allow to request many level contours to the same image (for our worst use case (gradient of 4096x4096 pixels, 1000 levels) is about x20 faster than `skimage`, cause we uses a data cache, efficient for gradient images)
- Supports mask (for complex use cases this impl is x20 faster than an implementation based on `skimage` and accurate, cause we don't have to post-process the data)
- Allow to request image pixels instead of polygons (the implementation is about x10 faster than the one from `pyFAI` or one based on `skimage`)

silx-marching-squares-mask

Results
======

With our worst use case: 1442 rings inside an image size of 4096x4096
```
find_pixels
===========
skimage
38.6 s ± 112 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
pyFAI
47.5 s ± 259 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx
24.8 s ± 90.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx + openmp
5.09 s ± 185 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx + minmax cache
2.22 s ± 29.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx + minmax cache + openmp
1.28 s ± 197 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

find_contours
============
skimage
38.5 s ± 177 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
matplotlib
34.3 s ± 94.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx
24.6 s ± 81.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx + openmp
4.66 s ± 37.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx + minmax cache
2.05 s ± 7.39 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
silx + minmax cache + openmp
733 ms ± 11.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

- As result this implementation is much faster than others for our use cases.
- OpenMP looks to be useful
- `find_pixels` could be optimized if we could use `unordered_set` from C++11. This implementation uses a lot sets.
- `find_pixels` could be optimized using another implementation i work on (scaninsert from https://github.com/vallsv/esrf_py_tests/tree/master/iso_contour). But not finished. The current result is already quite good.

Todo
====

- [x] skimage impl
- [x] Cython impl
- [x] Example
- [x] Should we move it to `silx.image.marchingsquares` with a `s`?
- [x] Documentation
- [x] Implement `find_pixels`
- [x] Check `find_pixels` speed up compare to the pyFAI one
- [x] Check `find_contours` speed up compare to the matplotlib one in pyFAI
- [x] Clean up the FIXME
- [x] Compute min/max using cython/openmp instead of numpy
- [x] Try to reproduce a bug in the merge poly (merge of 2 poly creating a closed poly but connecting the wrong points together: creating a shape like `8` instead of `0`)
- [x] Fix a mask bug
- [x] There is a bug when using minmax at the merge of the segment (it looks like it's a problem to compute the minmax in the y-axis)
- [ ] <s>Try to finalize the `scanmerge` algo (less dependent of the data)</s> Not much time. Maybe one of this days.
- [x] Uses QGroupBox in the example
- [ ] <s>Support an external image in the example?</s> Not much interesting
- [x] Fix image input using the trick of @t20100 (continuous native float 32bits)


t20100:
marching square**S** seems to be the dedicated name.

vallsv:
Ready for review

payno:
LGTM

PR #1768, opened 2018-04-04
[nxdata] fix bug with undefined axis

### LINKED ISSUE #1767: [nxdata] bug when assigning None to an array of strings
The following two lines cause a bug in python 2 when parsing a NXdata group that has attributes stored as ASCII:
https://github.com/silx-kit/silx/blob/master/silx/io/nxdata.py#L358
https://github.com/silx-kit/silx/blob/master/silx/io/nxdata.py#L593

The issue is that h5py `attrs.get` returns numpy arrays of `numpy.string_`, and when assigning `None` to such an array, numpy converts it to the string `"None"`.

```
>>> import numpy
>>> arr = numpy.array(["toto", "tata", "tutu"], dtype=numpy.string_)
>>> arr[2] = None
>>> arr
array(['toto', 'tata', 'None'], dtype='|S4')
>>> arr[2] is None
False
```

PiRK:
Closes #1767 (issue with None assigned to arry of strings)

Also modifies the controversial function `get_attr_as_string` to always return unicode, rather than native python str (bytes in PY2 and unicode in PY3).

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -35,7 +35,7 @@ from silx.gui import qt, icons
 from silx.gui.data.TextFormatter import TextFormatter
 from silx.io import nxdata
 from silx.gui.hdf5 import H5Node
-from silx.io.nxdata import get_attr_as_string
+from silx.io.nxdata import get_attr_as_unicode
 from silx.gui.plot.Colormap import Colormap
 from silx.gui.plot.actions.control import ColormapAction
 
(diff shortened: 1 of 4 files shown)


vasole:
Honestly I would avoid the six import from third party just to define the special dtype. 

The code below makes very explicit what is going on under the scenes.

```
import sys
if sys.version_info < (3,):
    text_dtype = h5py.special_dtype(vlen=unicode)
else:
    text_dtype = h5py.special_dtype(vlen=str)
```

PR #1769, opened 2018-04-04
[gui] IPython console only supports qtconsole

### LINKED ISSUE #1763: silx view aborts when opening IPython console
With silx 0.7.0 on debian8 (rnice8 at least) silx view aborts when one tries to open the embedded IPython console from the GUI.

To reproduce: visualize some curves with silx view and open IPython console (Options->Console).

Error message:
```
QWidget: Must construct a QApplication before a QPaintDevice
Aborted
```

On another debian8 computer silx view breaks while loading IPython widget tries to load PyQt4 while silx view loads PyQt5.

t20100:
This PR removes support of loading IPython widget from IPython.
We now only support loading it from `qtconsole`.

This is inline with using PyQt5 as the main Qt binding (as old version of IPython does not probe it).
This de facto drops the support of IPython console widget on debian 8.

closes #1763


PiRK:
I made a small pull request on your branch https://github.com/t20100/silx/pull/13

Issue #1757, opened 2018-03-23
Possible to show date/time stamps on the X-Axis


titusjan:
Hello all. 

I am looking around for various plot libraries and am in the process of selecting one for my project. One of my requirements is to plot dates on the X-Axis. This doesn't seem possible in Silx. 

Would it be possible to implement functionality so that dates or times are displayed next to the ticks the X-axis? I am willing to make pull requests (if I choose to use silx for my project).

I see two approaches:

1. Let silx accept numpy arrays of Python [datetime](https://docs.python.org/2/librarydatetime.html#datetime-objects) objects and detect the array dtype when drawing ticks.
2. Add an option to the PlotWidget and use this option to decide if number or dates at the ticks will be drawn.

I think option 1 is preferable. I tried to make a plot with  array of datetime objects as x-coordinates and I get the following error.

```
Traceback (most recent call last):
  File "/Users/kenter/prog/external/silx/bootstrap.py", line 222, in <module>
    run_file(target, argv)
  File "/Users/kenter/prog/external/silx/bootstrap.py", line 104, in run_file
    execfile(filename, module_globals, module_globals)
  File "/Users/kenter/prog/external/silx/bootstrap.py", line 77, in execfile
    exec(code, globals, locals)
  File "/Users/kenter/work/gssw/src/egse/python/gssw/quicklook/trials/silx_test.py", line 89, in <module>
    main()
  File "/Users/kenter/work/gssw/src/egse/python/gssw/quicklook/trials/silx_test.py", line 83, in main
    makePlot1D()
  File "/Users/kenter/work/gssw/src/egse/python/gssw/quicklook/trials/silx_test.py", line 53, in makePlot1D
    curve = plot1D.addCurve(x=x, y=y, legend='curve')  
  File "/Users/kenter/prog/external/silx/build/lib.macosx-10.7-x86_64-3.5/silx/gui/plot/PlotWidget.py", line 708, in addCurve
    self.resetZoom()
  File "/Users/kenter/prog/external/silx/build/lib.macosx-10.7-x86_64-3.5/silx/gui/plot/PlotWidget.py", line 2577, in resetZoom
    self._forceResetZoom(dataMargins=dataMargins)
  File "/Users/kenter/prog/external/silx/build/lib.macosx-10.7-x86_64-3.5/silx/gui/plot/PlotWidget.py", line 2502, in _forceResetZoom
    ranges = self.getDataRange()
[...]

PR #1770, opened 2018-04-09
Implemented issue 1757 show date/time stamps on the X-Axis

titusjan:
As promised the PR that aims to implement issue #1757. 

I've adapted the second alternative and added a `setTimeSeries` method to indicate the X-axis values should be interpreted as timestamps. An example use is:

```
import datetime as dt
import numpy as np

def makePlot1D():

    base = dt.datetime.today()
    dates = [base - dt.timedelta(days=x) for x in range(0, 50)]

    x = np.array([d.timestamp() for d in dates], dtype=np.uint64)

    np.random.seed(seed=1)
    y = np.random.random(x.shape) * 12 - 3

    plot1D = Plot1D(backend='mpl')
    xAxis = plot1D.getXAxis()
    xAxis.setTimeSeries(True)

    curve = plot1D.addCurve(x=x, y=y, legend='curve')  
    plot1D.show()


def main():
    app = QApplication([])
    makePlot1D()
    app.exec_()

if __name__ == "__main__":
    main()

```

Most of the new code handles the tick positions. I looked at your implementation and adapted the Nice Numbers algorithm for date/times, so that ticks are placed at meaningful intervals and not, for example, every 1000 seconds.

This is a first version, I expect that you have some comments on style or other things. Please let me know what you think.

I have two questions:
1) Do you want me to implement it for the y-axis as well?
2) How to best handle time-zones. Shall I add a `setTimeZone` method to the plot?


t20100:
Great! I'll look at it soon.

t20100:
For the Y axis, it is up to you. If you don't add it now, it can be added later if needed.
For the time-zone, I don't have a clear idea. I was initially seeing it as an Axis attribute, but I guess there is little chance one would like to use 2 different timezones on 2 different axes! How is day-light saving supposed to be handled (if supported) ?

I made a small PR on your branch to advertise `dateutil` as a dependency + minor changes.
Here are a few comments:

For the `Axis.is|setTimeSeries` API, it would be better to make the API ready for other tick modes.
Something like `Axis.get|setTickMode` which takes values from an enum (DEFAULT, TIME_SERIES)). This also leaves the door open to add support of user defined tickers.
The backend API is fine as is as it can be changed more easily if needed.

As `niceNumGeneric` is quite close to `_niceNum`, both functions could be combined.

`PlotWidget.addCurve` could be extended to accept list of datetime as `x` input and do the conversion.

BTW, don't hesitate to put your name or institution/company in the copyright of the files.

titusjan:
I have merged your PR but now there is a conflict in the `requirements.txt` file. I'm not sure how to proceed here. Can you fix it or do I have to do it?

And thank you for your feedback, I will respond in detail next week. Unfortunately I didn't have the time to delve in to them.

I have a quick question: do you already know when you will release version 0.8? (I.e. do I have to hurry to get it in?)

vallsv:
Hi. I did the merge for you using the github gui.

@t20100 is not around this days. But i can aswer for the release. There a new one every 3 months. silx 0.7 was on march, then you can expect 0.8 for june.

titusjan:
Hi @t20100, I updated my PR with your feedback. Can you take a look again?

> For the Y axis, it is up to you. If you don't add it now, it can be added later if needed.

Ok, I left it out. I don't think many users will want to plot dates on the Y-Axis

> For the time-zone, I don't have a clear idea. I was initially seeing it as an Axis attribute, but I guess there is little chance one would like to use 2 different timezones on 2 different axes! 

I agree that you don't want two different time zone on two different axes, but I still made `setTimeZone` a method of the axes objects. This because it's closely related to the TickMode, which is also an attribute of the axes. If desired we could always later make a `setTimeZone` method on the PlotWidget, that sets the time zone for all its axes. 

> How is day-light saving supposed to be handled (if supported) ?

If you use None as the axes `timeZone` (the default) the PlotWidget will use [datetime.timestamp()](https://docs.python.org/3.6/library/datetime.html?highlight=timezone#datetime.datetime.timestamp) to internally convert to floats. This uses the platform's `mktime()` C-function which is time zone aware, so DST is taken into account. 

If you use `"UTC"` as the `setTimeZone` parameter, the times are in UTC, which doesn't have DST.

I think this covers 99.9% off all use cases. If you want to have dates in a TZ that is not local or UTC you must make your own class that derives from [datetime.tzinfo](https://docs.python.org/3.6/library/datetime.html?highlight=timezone#tzinfo-objects), or find a 3rd party implementation since this isn't part of the Python standard library.

> For the Axis.is|setTimeSeries API, it would be better to make the API ready for other tick modes.
> Something like Axis.get|setTickMode which takes values from an enum (DEFAULT, TIME_SERIES)). This also leaves the door open to add support of user defined tickers.
> The backend API is fine as is as it can be changed more easily if needed.

I implemented this. I didn't change it in the backend because I then didn't know where to define the enum.

> As niceNumGeneric is quite close to _niceNum, both functions could be combined.

I didn't do this before because they give slightly different results and I didn't want to change existing code. I changed it now and I don't see significant differences. I left the old implementation as a comment so that it can be reverted back easily. I also think the old code is more clear what happens. Please take a look.

BTW, do you have a copy of Paul Heckberts article on this? It's from "Graphics Gems", Academic Press, 1990 but I can't find this online.

> PlotWidget.addCurve could be extended to accept list of datetime as x input and do the conversion.

I implemented this but this assumes all elements in the array have the same time zone. Also that all the curves are time series. Let me know what you think.

t20100:
I made a PR on your branch with a few minor updates.

For the conversion in `addCurve`, I was just thinking of doing the conversion from datetime to numbers without setting the axis timezone and switching it to time series (i.e., only `x = [d.timestamp() for d in x]`).
Also `datetime.timestamp` is only available in Python3 and we still aim at supporting Python2, so it would need a fallback.

Otherwise it looks good to me. Thanks for this nice feature!
I would propose to leave the conversion in `addCurve` for later and merge this PR.

titusjan:
> For the conversion in addCurve, I was just thinking of doing the conversion from datetime to numbers without setting the axis timezone and switching it to time series (i.e., only x = [d.timestamp() for d in x]).

Ok, I have removed the automatic setting of the tick mode to time series, and also the automatic setting of the time zone. The caller now has to do this explicitly.

I have backported the `timestamp` functionality from Python 3.5. The PR now also works in Python 2.7.

I think it is ready to merge, unless you have some comments about these last two changes. Please let me know.

PR #1774, opened 2018-04-11
Remove cythonized files

vallsv:
This PR remove all the cythonized files from the Git repository.

The configuration of the continuous integration files was updated to include cython as dependency.

The `setup.py`'s targets `sdist` and `clean` was updated.

- `python setup.py clean` will now also remove cythonized files
- `python setup.py sdist` will react as we expect for the project: regenerate all the cython files without OpenMP to provided them as source to pypi.


kif:
does sdist generate the C-files ?

vallsv:
No, but i can try.

I think i will try something like: `python setup.py sdist --cythonized`, instead of only `sdist`.

vallsv:
BTW there is a pbm with the qhull compilation here.

t20100:
I think we should add cython to `setup_requires` in the `setup.py`... eventually except for `sdist` if the C files are embedded...

t20100:
I'm checking for QHull, looks like an issue with different version of cython.

t20100:
PR #1776 is fixing the qhull issue...

kif:
looks great to me.

PR #1776, opened 2018-04-12
Update from upstream (commit 45f14e1) to fix issue with cython 0.28

t20100:


vallsv:
Thanks for the fast fix.
LGTM

Issue #1775, opened 2018-04-11
Exception when reading NXdata (possibly malformed) in silx view


kif:
```
ERROR:silx.gui.qt._qt:<class 'KeyError'> "Unable to open object (Object 't' doesn't exist)"   File "/usr/lib/python3/dist-packages/silx/app/view.py", line 200, in displayData
    self.__dataViewer.setData(data)
  File "/usr/lib/python3/dist-packages/silx/gui/data/DataViewerFrame.py", line 166, in setData
    self.__dataViewer.setData(data)
  File "/usr/lib/python3/dist-packages/silx/gui/data/DataViewer.py", line 494, in setData
    self.__updateView()
  File "/usr/lib/python3/dist-packages/silx/gui/data/DataViewer.py", line 413, in __updateView
    self.__updateAvailableViews()
  File "/usr/lib/python3/dist-packages/silx/gui/data/DataViewer.py", line 394, in __updateAvailableViews
    info = DataViews.DataInfo(data)
  File "/usr/lib/python3/dist-packages/silx/gui/data/DataViews.py", line 120, in __init__
    nxd = nxdata.get_default(data)
  File "/usr/lib/python3/dist-packages/silx/io/nxdata.py", line 881, in get_default
    return NXdata(default_data)
  File "/usr/lib/python3/dist-packages/silx/io/nxdata.py", line 358, in __init__
    if dsname is not None and "long_name" in self.group[dsname].attrs:
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/kieffer/debian8/h5py-2.6.0/h5py/_objects.c:2408)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/kieffer/debian8/h5py-2.6.0/h5py/_objects.c:2365)
  File "/usr/lib/python3/dist-packages/h5py/_hl/group.py", line 166, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/kieffer/debian8/h5py-2.6.0/h5py/_objects.c:2408)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/kieffer/debian8/h5py-2.6.0/h5py/_objects.c:2365)
[...]

PR #1778, opened 2018-04-17
[nxdata] refactoring

PiRK:
This PR transforms the `silx.io.nxdata` module into a package, with 3 modules: `parse`, `validate` and `write`.

The `validate` module will need a major refactoring if we want to be able to improve and maintain it. The current validation function `is_valid_nxdata` has become too large, and still it fails to catch important issues (like missing axes #1775).


PiRK:
I think that we should merge this, there are already important patches in the middle of the refactoring (https://github.com/silx-kit/silx/pull/1778/commits/44a7586b9b9453b46aaeed2e616dd4e72fea0e17, https://github.com/silx-kit/silx/pull/1778/commits/4ebed9ee911d146622c3f1cb68dcca823d800075 and https://github.com/silx-kit/silx/pull/1778/commits/741564269487006195ff4110d99fa57134f1bfa5). 

The remaining work, properly refactoring the validation, requires a separate pull request and will take some time.

vallsv:
LGTM

PR #1793, opened 2018-04-24
Move Colormap and ColormapDialog outside of plot

### LINKED ISSUE #1557: Move ColormapDialog inside silx.gui.dialog

vallsv:
This PR rework colors to expose then outside of the plot module.

- Create a module for colors `silx.gui.plot.colors`
- Merge and deprecate `silx.gui.plot.Colormap` inside `colors`
- Merge and deprecate `silx.gui.plot.Colors` inside `colors`
- Move `silx.gui.plot.ColormapDialog` to `silx.gui.dialog.ColormapDialog`

There is still a dependency to `matplotlib`, but @t20100 also want to remove it and use our own code.

The module looks a little big, it could be nice to move `Colormap` outside of the module, and still expose it through this module, but it's also fine as it is.

Closes #1557


t20100:
As you want for moving `Colormap` in a separate module, for me it is fine like that.
We can move it later, as I think it is anyway good to have it available in the `colors` module.

Issue #1746, opened 2018-03-21
Add load/save settings for silx view


t20100:
Discussion started in #1744

vasole:
Needed:

- Defaults for silx view (including geometry of the application itself) automatically saved
- A way for silx view to bypass the saved settings
- A way to provide defaults for silx:
    - Default colormap (linear, log, viridis, grey, temperature, ...)
    - Default image origin (up, down)
    - Default fit functions and fit configurations
    - Default binding (matplotlib or OpenGL)
    - ...

- A way for applications to set the silx defaults with the possibility to ask silx widgets to use the defaults saved as silx settings or using the defaults imposed by the application.

vallsv:
All the features are not there, like fitting or the bypass, but it could be already enough for the 0.8 milestones?

vasole:
Is it too difficult to just add the bypass (--fresh or -f in analogy with SPEC and PyMca flags)? I had to add it in PyMca because of people saving the settings in one terminal and then using the program from a different one therefore leading to huge inconveniences (for instance when switching from dual screen to single screen).

The fit it can wait.

vallsv:
- No it is not difficult to add a `--fresh` and it is much safer.
- It is also easy to store ini file always (i check the Qt API, Qt take care of the path).

vallsv:
Hmmm, well, a bypass and a fresh is not the same usage...
- `fresh` will open the file, clear it, and save it at the end (in case something is buggy, it can help)
- `bypass` will not use user preference at all

Are we talking about 2 features?

t20100:
Wouldn't `fresh` be enough?

vallsv:
Well, i will do it at least.

vallsv:
Done by #1844. If something is missing we still can open a new issue. I think about fit configuration, which is very special i think.

PR #1800, opened 2018-04-25
Add user preferences to silx-view 

vallsv:
Thing which are part of user preferences
- Main window location
- Default colormap
- Plot backend

This informations are private to the application. I do not think about sharing this informations to the library in any way.

Missing things
- Default Y-origin

Relative to #1746


t20100:
Good to have a way to provide general settings for the lib at runtime and one on top for `silx view` application persistent settings.
I have doubts on the y-axis orientation in `silx.config` as it configures multiple widgets: I'm not sure we need the same setting for all, but at the same time having one setting for each sounds heavy...

vallsv:
Here is the final version. Only cosmetic changes. I have no other proposal for the y-axis. But this could be changed.

Issue #1744, opened 2018-03-21
[plot] Change Y axis orientation default for image widgets


t20100:
During the last training it was suggested to display by default images with Y axis going downward.

We can do this at different level: `sx.imshow` or for each image plot widget: `Plot2D`, `ImageView`, `ComplexImageView`, `StackView`.
Doing it at the widget level would break current behavior (and no deprecation warning is possible)...

t20100:
To me, the question here was more: did we choose the default Y axis orientation for images because it's the same as for curves or because it covers enough cases. BTW, matplotlib `imshow` looks to have Y axis downward by default.

Now that there is a silx application we can consider have settings.
But I would use it for the application and eventually for `sx` module but not for widgets where each application should decide.

vasole:
I guess we chose it as default for Plot because it covers all cases.

For imshow by sure we can take matplotlib's default.

For ScatterView by sure it is like Plot and that makes me doubt about applying the imshow behavior to Plot2D.

vallsv:
Settings for `silx view` looks to be a good idea.

t20100:
I created issue #1746 for discussing `silx view` settings.

vasole:
I would just make imshow compatible with matplotlib setting but I did not take into account that the matplotlib defaults can be different too (use of rcparams).

PR #1801, opened 2018-04-25
[plot] Allow to set origin = upper or lower in sx.imshow

t20100:
This PR allows to pass` origin='lower'` or `'upper'` to `sx.imshow` in order to toggle the Y axis orientation.
It DOES NOT change the default Y axis orientation (upward) in order to avoid inconsistence with all plot widgets and other plot functions in `sx`... So it remains inconsistent with matplotlib default.

I think it is how we said we wanted to address #1744


vallsv:
LGTM

PR #1804, opened 2018-04-26
[plot gl] Overcome float32 precision 

### LINKED ISSUE #1772: Wrong results when plotting small x-range with large offset with Open GL backend
When making PR #1770 I ran into the issue that many values are grouped together when plotting a data with a small (x) range and a large offset. The problem only occurs when using the OpenGL backend, the Matplotlib backend works fine. 

The following code demonstrates the problem:
```

import datetime as dt
import numpy as np

if 1:
    BACKEND = 'gl' # this gives incorrect results
else:
    BACKEND = 'mpl' # this works

def makePlot1D():

    # Make large value with a relatively small range by creating POSIX time stamps.
    base = dt.datetime.today()
    dates = [base - dt.timedelta(seconds=x) for x in range(0, 2500, 20)]

    x = np.array([d.timestamp() for d in dates], dtype=np.uint64)
    np.random.seed(seed=1)
    y = np.random.random(x.shape) * 12 - 3

    print('x[0:10]: {}'.format(x[0:10]))

    plot1D = Plot1D(backend=BACKEND)
    xAxis = plot1D.getXAxis()

    curve = plot1D.addCurve(x=x, y=y, legend='curve', symbol='o')
    plot1D.show()


def main():
    app = QApplication([])
    makePlot1D()
    app.exec_()

if __name__ == "__main__":
    main()

```

When using the `gl` backend the output looks as follows.

silx_opengl

The `mpl` backend *does* give the correct results.

silx_matplotlib

I suspect that somewhere in the `gl` backend 32-bits floats are used instead of 64-bit floats.

t20100:
This PR overcomes the float32 precision issue occurring for data with difference only visible with double precision. This is due to sending float32 to OpenGL and using data coordinates as coordinates for the display.

It is not complete and not tested (TODOs: error bars, images, shapes, log scale,...) but it works for the case of #1772.

closes #1772

### DIFF
--- a/silx/gui/plot/backends/BackendOpenGL.py
+++ b/silx/gui/plot/backends/BackendOpenGL.py
@@ -652,15 +652,13 @@ class BackendOpenGL(BackendBase.BackendBase, glu.OpenGLWidget):
                      plotWidth, plotHeight)
         gl.glEnable(gl.GL_SCISSOR_TEST)
 
-        gl.glViewport(self._plotFrame.margins.left,
-                      self._plotFrame.margins.bottom,
-                      plotWidth, plotHeight)
+        gl.glViewport(0, 0, self._plotFrame.size[0], self._plotFrame.size[1])
 
         # Prepare vertical and horizontal markers rendering
         self._progBase.use()
         gl.glUniformMatrix4fv(
             self._progBase.uniforms['matrix'], 1, gl.GL_TRUE,
-            self._plotFrame.transformedDataProjMat.astype(numpy.float32))
+            self.matScreenProj.astype(numpy.float32))
         gl.glUniform2i(self._progBase.uniforms['isLog'], isXLog, isYLog)
         gl.glUniform1i(self._progBase.uniforms['hatchStep'], 0)
         gl.glUniform1f(self._progBase.uniforms['tickLen'], 0.)
@@ -701,9 +699,9 @@ class BackendOpenGL(BackendBase.BackendBase, glu.OpenGLWidget):
                                            align=RIGHT, valign=BOTTOM)
                             labels.append(label)
 
-                        xMin, xMax = self._plotFrame.dataRanges.x
-                        vertices = numpy.array(((xMin, yCoord),
-                                                (xMax, yCoord)),
+                        width = self._plotFrame.size[0]
+                        vertices = numpy.array(((0, pixelPos[1]),
+                                                (width, pixelPos[1])),
                                                dtype=numpy.float32)
 
                     else:  # yCoord is None: vertical line in data space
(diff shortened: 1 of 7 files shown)


titusjan:
Thanks for doing this. I briefly tested it as it is now and I can confirm that this resolves my issue #1772.

It did crash, however, when I selected a cross hair cursor and moved the cursor around for a while. It then gave the following error.

```
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/_glutils/OpenGLWidget.py", line 234, in paintGL
    parent.paintGL()
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/plot/backends/BackendOpenGL.py", line 546, in paintGL
    self._paintFBOGL()
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/plot/backends/BackendOpenGL.py", line 491, in _paintFBOGL
    self._renderPlotAreaGL()
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/plot/backends/BackendOpenGL.py", line 882, in _renderPlotAreaGL
    isXLog, isYLog)
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/plot/backends/glutils/GLPlotCurve.py", line 1262, in render
    self.lines.render(matrix, isXLog, isYLog)
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/plot/backends/glutils/GLPlotCurve.py", line 442, in _renderSolid
    self.yVboData.setVertexAttrib(yPosAttrib)
  File "/home/pepijnk/src/extern/silx_t20100/build/lib.linux-x86_64-3.5/silx/gui/_glutils/VertexBuffer.py", line 210, in setVertexAttrib
    c_void_p(self.offset))
  File "/usr/lib/python3/dist-packages/OpenGL/latebind.py", line 61, in __call__
    return self.wrapperFunction( self.baseFunction, *args, **named )
  File "/usr/lib/python3/dist-packages/OpenGL/GL/VERSION/GL_2_0.py", line 426, in glVertexAttribPointer
    contextdata.setValue( key, array )
  File "/usr/lib/python3/dist-packages/OpenGL/contextdata.py", line 58, in setValue
    context = getContext( context )
  File "/usr/lib/python3/dist-packages/OpenGL/contextdata.py", line 41, in getContext
    """Attempt to retrieve context when no valid context"""
OpenGL.error.Error: Attempt to retrieve context when no valid context
```

As you say above, you are still working on this so some errors are still to be expected. Just wanted to let you know.

t20100:
Thanks for the testing and the feedback.
As it is now it clearly needs more work and validation.

t20100:
At last, reworked plot OpenGL backend to extend precision support by:
- subtracting an offset from the curves data before sending them to the GPU (which implied applying axes log scale to data before sending them)
- doing CPU matrix operations in float64
- displaying shapes, markers, crosshair in pixel coordinates (doing data to pixel conversion on the CPU)

This also includes quite a few bug fixes (in particular with log scaled axes) and code clean-up (remove axes scales shader code).

@titusjan I could not reproduce your issue, under which conditions did it occur?

It still needs a bit of testing.
BTW, this PR does improve precision support for images.

titusjan:
@t20100 Thanks for putting in the effort. I really appreciate it.

Do you mean the issue with the cross hair? I'm afraid I can't remember anything next to what I typed earlier. Next Tuesday I will be at work again and I can try to reproduce it again.

t20100:
Yes I was wondering about the crosshair issue.

t20100:
I propose to review it as is and do extra fixes in other PRs.

vallsv:
I will merge it this afternoon after checking it is working with https://github.com/silx-kit/silx/issues/680. This dataset is known to be unaccurate in float32.

vallsv:
LGTM

PR #1811, opened 2018-05-02
[plot] Use origin=upper as default for imshow

### LINKED ISSUE #1744: [plot] Change Y axis orientation default for image widgets

t20100:
closes #1744


vallsv:
LGTM

PR #1813, opened 2018-05-02
[PlotWindow] Deprecate pan and zoom actions

PiRK:
Use a deprecation phase for `PlotWindow.panModeAction` and `PlotWindow.zoomModeAction`.
Addresses https://github.com/vasole/pymca/issues/220

The code will also be modernized on the PyMca side, but it is a good idea to maintain the previous API for a couple of releases.


vallsv:
LGTM

PR #1816, opened 2018-05-03
[nxdata] always fetch a copy of attrs

PiRK:
same patch as #1814, but applied on the refactored code (after #1778).


vallsv:
LGTM

PR #1836, opened 2018-05-17
[nxdata] Make debugging information available when parsing

PiRK:
Up to now, attempting to parse an invalid NXdata group caused a TypeError. Debugging information about why the group was invalid NXdata was simply logged. 

This PR changes this behavior, the  error should no longer be raised on object instantiation, but only when trying to access the actual invalid data. A new `is_valid` is also added to the object.

Previously one would need to write:
```
try:
    nxd = NXdata(group)
except TypeError:
    signal = None
else: 
    signal = nxd.signal
```

The same behavior would now be achieved using:
```
nxd = NXdata(group)
signal = None if not nxd.is_valid else  nxd.signal
```

Additionally, the object will store error messages for malformed NXdata groups.

```
if not nxd.is_valid:
    for i, err_msg in enumerate(nxd.issues):
        print("malformation %d: %s" % (i, err_msg))
```


PiRK:
Note : this can be considered an small API change, for users who relied on a TypeError being raised for malformed NXdata. I'm hoping that very few users started using the *NXdata* class, and that most of them tested the validity before creating the object, rather than catching the error.

vallsv:
LGTM

PR #1844, opened 2018-05-23
[silx view] Improve user preferences

vallsv:
- Implement `--fresh` to start the application from fresh user preferences
- Allways use ini file, for all platforms (Windows will use APPDATA directory)

Relative to #1746


PiRK:
LGTM

Issue #1829, opened 2018-05-14
[convert] silx convert tries to write float128 while it is not supported by h5py


t20100:
There is an issue with `silx convert` on Debian8 with float128 values.

`fabioh5` converts values to `float128` when needed, but the support of `float128` in `h5py` only starts with v2.7.0 (https://github.com/h5py/h5py/pull/749)... while Debian8 provides v2.5.0.

vallsv:
Then what to do?
Is it better to cancel the conversion? Or just use `float64` which can generate dirty result?
The header is anyway also stored as text.

vallsv:
We also can cancel the cast and store an array of string insteaqd of float64 or float128.

t20100:
I would store as strings to be conservative. And maybe if one day someone needs to add an option to enable unsafe conversion...

vasole:
Considering the header is written as text, I would just convert to float64. The header is still there...

vallsv:
Mostly done by #1856. As result header should usually not contains anymore float128. But it could.

The case where numpy provides float128 and h5py do not support it is not taken into account. It have to be fixed in convert. But we can consider it is fine as it is (an error is better than a lose of precision).

Reassigned to @PIRK.

PR #1856, opened 2018-05-24
Fix fabioh5 float conversion

### LINKED ISSUE #1830: [io] fabioh5 conversion to float does not check the exponent
In `fabioh5`, the conversion from string to the appropriate floating type takes into account the number of digits but not the value of the exponent to decide which floating type should be used.

vallsv:
- Take care of exponent to choose floating-point size
- Use `numpy.longdouble` if available (instead of `float128`/`float96`)
- Fallback using `float64`
- Create a dedicated module `silx.utils.number`

Relative to #1829
Closes #1830

### DIFF
--- a/silx/io/fabioh5.py
+++ b/silx/io/fabioh5.py
@@ -951,7 +951,13 @@ class File(commonh5.File):
         if first_file_name is not None:
             _, ext = os.path.splitext(first_file_name)
             ext = ext[1:]
-            use_edf_reader = ext in fabio.edfimage.EdfImage.DEFAULT_EXTENSIONS
+            edfimage = fabio.edfimage.EdfImage
+            if hasattr(edfimage, "DEFAULT_EXTENTIONS"):
+                # Typo on fabio 0.5
+                edf_extensions = edfimage.DEFAULT_EXTENTIONS
+            else:
+                edf_extensions = edfimage.DEFAULT_EXTENSIONS
+            use_edf_reader = ext in edf_extensions
         elif first_image is not None:
             use_edf_reader = isinstance(first_image, fabio.edfimage.EdfImage)
         else:
--- a/silx/io/fabioh5.py
+++ b/silx/io/fabioh5.py
@@ -44,6 +44,7 @@ import numpy
 from . import commonh5
 from silx.third_party import six
 from silx import version as silx_version
+import silx.utils.number
 
 try:
     import h5py
@@ -106,7 +107,6 @@ class FrameData(commonh5.LazyLoadableDataset):
             attrs = {"interpretation": "image"}
         commonh5.LazyLoadableDataset.__init__(self, name, parent, attrs=attrs)
         self.__fabio_reader = fabio_reader
-
 
     def _create_data(self):
         return self.__fabio_reader.get_data()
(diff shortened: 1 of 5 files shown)


silx/io/fabioh5.py
```diff
@@ -82,6 +83,68 @@ def supported_extensions():
...
+
+def min_numerical_convertible_type(string):
+    """
+    Parse the string and return the minimal numerical type while fit for a
+    convertion.
+
+    :param str string: Representation of a float/integer with text
+    :raise ValueError: When the string is not a numerical value
+    :retrun: A numpy numerical type
+    """
```
kif: I would strip the string at the input

silx/io/fabioh5.py
```diff
@@ -82,6 +83,68 @@ def supported_extensions():
...
+            digits = number + max(decimal, exponent)
+        elif exponent < 0:
+            exponent = -exponent
+            digits = max(number, exponent) + decimal
+        else:
+            digits = number + decimal
+
+        if digits <= 3:
+            # A float16 is accurate with about 3.311 digits
+            return numpy.float16
```
kif: float16 is not standard in numpy and may cause a lot of conversion error. I would start with float32

kif:
my main concern is about float16 which are not handled by CPU and need to be packed/unpacked for every single operation on them

vallsv:
Thanks for the feed back.
I will fix that.

t20100:
The way to deal with the exponent overestimates the minimum required type:
```
In [5]: min_numerical_convertible_type(str(numpy.float32(1.1e30)))
Out[5]: numpy.float128
```
It should be considered independently of the number of digits: does the exponent value fits the range of values provided by a specific type? (e.g., exponent in [-37, 37] for float32 (to keep using "normalized")

kif:
On Thu, 24 May 2018 05:07:01 -0700
Thomas VINCENT <notifications@github.com> wrote:

> BTW, this functions seems of general interest, what about moving it somewhere like `silx.utils`?

I agree with Thomas twice (about exponant values as well)

t20100:
I think there is a much simpler way: convert string to `longfloat` and then check if it can be safely cast to a dtype:

```
    value = numpy.longfloat(string)
    for dtype in (numpy.float16, numpy.float32, numpy.float64):
        if numpy.can_cast(value, dtype):
            return dtype
    return numpy.longfloat
```

BTW, `numpy.longfloat` should always be defined (either float128, float96 or float64)

t20100:
Looks fine.

There is a corner case where it overestimates the dtype, it is for numbers like: 0.00[...]0001

```
In [17]: min_numerical_convertible_type('0.000001')
Out[17]: numpy.float32
In [18]: numpy.float16('0.000001')
Out[18]: 1e-06
```

vallsv:
Yes, i did not take tralling 0 after the ".".

Anyway i also only support float16 containing 3 digits without exponent, as it should not be often use. But it could be considered as a normal case without problem.

vallsv:
According to https://github.com/silx-kit/silx/issues/680 float64 looks to be enough.

vallsv:
It's fine now. The precision was too strong. A `float64` looks to be always enough for the result of `time.time()`.

vallsv:
```
In [2]: numpy.longdouble
Out[2]: numpy.float64

In [3]: numpy.longdouble == numpy.float64
Out[3]: False
```
Well well...

PR #1837, opened 2018-05-17
Colormap implementation in silx

t20100:
This PR provides a cython/OpenMP implementation of colormap computation (Related to #1648 ).
This implementation provides linear, log, sqrt and arcsinh (Related to #1212 ).

This PR contains only the implementation of the colormap and not its usage in silx.gui.plot which will be done in another PR.

PR #1858, opened 2018-05-24
[plot] Use silx colormap implementation in the plot

### LINKED ISSUE #1317: [plot] Colormap sampling subject to rounding errors

### LINKED ISSUE #1648: [plot] Speed up colormap calculation
The display of dynamic scatter plots is very slow even using the OpenGL backend.

- It should be fixed at the very least for that backend.
- Since we can send the colormaps in most cases, it would be convenient to calculate ourselves the colormap instead of relying on matplotlib.

t20100:
This PR makes use of silx colormap in the plot added in PR #1837: closes #1648
It also makes sampling of color LUT with 256 (uint8) values avoiding duplicates: Closes #1317

Merge #1837 first as this PR builds on it.


vallsv:
LGTM

Issue #1765, opened 2018-04-03
[mask] getSelectionMask returns an array where there is no mask


vallsv:
```
In [1]: from silx import sx

In [2]: plot = sx.imshow()
Out[2]: <silx.gui.plot.PlotWindow.Plot2D at 0x7f70a67c4218>

In [3]: plot.getMaskToolsDockWidget().getSelectionMask()
Out[3]: array([], dtype=uint8)
```

I am supposed to work with images, and here i have to test an array with a dim==1. It would be more convenient to return `None` if there is no mask.

vasole:
The safe approach is to return an array with non-selected points everywhere and with the same shape as the active image(or item). The problem in your case seems to be that there is no active item, not the fact of having an empty array.

t20100:
I though we were always returning a 2D array (even if empty), but it is not the case, so indeed None would be convenient if there is no image to draw the mask on.

vasole:
As I see things:

- If there is no active item, None is much better than an empty array
- If there is an active item, the safest approach is to have the returned mask with the same shape (and size) as the active item shape.

In PyMca I ended up accepting None as equivalent of a mask with zeros everywhere so, an array with same size as the item but zero filled.

vasole:
Added the 0.8 milestone to make sure we decide something as coding does not seem to be the problem.

t20100:
PR #1867 makes sure a 2D array is always returned.
Anyway it is better to do it this way internally, and personally I expected it this way.

Now we have to choose for the API between an empty 2D array and None.

t20100:
We decided to return None and accept None in `setSelectionMask`.

PR #1867, opened 2018-05-25
[plot] make image mask always return 2D array

t20100:
This PR makes image mask `getSelectionMask` always return 2D arrays.
Related to #1765


t20100:
Also fixes reset of mask when removing image while the mask widget is hidden.

vallsv:
Before merging we have to check that PyFAI is still working with.

vallsv:
No problem with PyFAI, but i send you a PR to fix an example.

vallsv:
LGTM

PR #1868, opened 2018-05-28
[plot3d] Add missing method to access Global scene parameters widget

t20100:
This PR adds a method that was missing to access a widget of the `SceneWindow` main window.


vallsv:
LGTM

PR #1870, opened 2018-05-29
[plot] Add select-draw interaction mode to PlotWidget

t20100:
This PR adds a `select-draw` interactive mode which provides items (e.g., markers) selection and dragging and shape drawing interaction at the same time.
This would be useful for PR #1758.


vallsv:
I don't really understand, but it looks fine.

PR #1873, opened 2018-05-31
[plot] Make set|getSelectionMask support/return None

### LINKED ISSUE #834: [mask] setSelectionMask can't process return value of getSelectionMask()

### LINKED ISSUE #1765: [mask] getSelectionMask returns an array where there is no mask

t20100:
This PR changes the returned value of `getSelectionMask` to None when there is no mask and makes `setSelectionMask` accept None as input to reset the mask.

closes #1765
closes #834

Please check if it breaks anything in applications before merging.
Maybe not merge for v0.8?

### DIFF
--- a/examples/findContours.py
+++ b/examples/findContours.py
@@ -465,8 +465,6 @@ class FindContours(qt.QMainWindow):
         if mask is None:
             if self.__useMaskButton.isChecked():
                 mask = self.__plot.getMaskToolsDockWidget().getSelectionMask()
-                if mask is not None and mask.size == 0:
-                    mask = None
 
         self.__image = image
         self.__mask = mask
--- a/examples/scatterMask.py
+++ b/examples/scatterMask.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 # /*##########################################################################
-# Copyright (c) 2016-2017 European Synchrotron Radiation Facility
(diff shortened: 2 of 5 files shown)


vallsv:
We can merge it if there is no problem on PyMCA

PiRK:
It will be working with PyMca too, with a minor fix (https://github.com/vasole/pymca/pull/245).

payno:
I am just checking this on tomogui.

payno:
Good for me.

Issue #1872, opened 2018-05-31
[plot] Make sure there is no circular references between plot widgets and associated tools


t20100:
It seems there is places where strong reference are kept instead of using weakref.
This can be the root of issues like "seg fault" at the end of the application and accessing a PlotWidget while the underlying C++ object has been destroyed..

@dnaudet, if you have simple sample code to reproduce that would be nice.

dnaudet:
Quick example: in my app I use weakmethod to keep a ref to a callback, so that it doesn't get called when the object it belongs to is deleted  (and so I don't have to explicitly keep track of it myself). This doesn't work with PlotWidget.
(note that all slots/signals get disconnected, because the underlying C++ object is gone, but the python wrapper is still there)

```python
import weakref
import datetime
from functools import partial

import numpy as np

from sip import ispyowned, isdeleted

from silx.gui import qt as Qt
from silx.gui.plot import PlotWindow
from silx.utils.weakref import WeakMethod


def ref_destroyed(obj, name=None):
    print('Weakref "{0}" destroyed! {1}'.format(name, obj))


class Scheduler(object):
    def __init__(self, callback):
        self._timer = Qt.QTimer()

        self._cb = WeakMethod(callback)

        self._timer.timeout.connect(self._timeout)
        self._timer.start(1000)

    def _timeout(self):
        cb = self._cb()
        if cb:
            cb()


class MyWid(Qt.QWidget):
    def someSlot(self, *args, **kwargs):
        print('MyWid SLOT!', ispyowned(self), isdeleted(self))


class PlotWithSlot(PlotWindow):
    def someSlot(self, *args, **kwargs):
        print('PlotWithSlot SLOT!', ispyowned(self), isdeleted(self))
        self.addCurve(np.arange(3) + datetime.datetime.now().second, np.arange(3))


class MainWindow(Qt.QMainWindow):
    def __init__(self, *args, **kwargs):
        super(MainWindow, self).__init__(*args, **kwargs)

        mdi = Qt.QMdiArea()
        self.setCentralWidget(mdi)

        subWin = MyWid()
        mdi.addSubWindow(subWin)

        plotWSlot = PlotWithSlot()
        mdi.addSubWindow(plotWSlot)

        self._weakQWwid = weakref.ref(subWin,
                                      partial(ref_destroyed, name='QWidget'))
        self._weakPlotWid = weakref.ref(plotWSlot,
                                        partial(ref_destroyed,
                                                name='plotWSlot'))

[...]

PR #1874, opened 2018-05-31
[plot] Fix circular reference in plot widgets

t20100:
This PR fixes circular reference at Python level in `PlotWidget`, `PlotWindow`, `Plot1D` and `Plot2D` by using `weakref`.

Related to #1872

### DIFF
--- a/silx/gui/plot/ColorBar.py
+++ b/silx/gui/plot/ColorBar.py
@@ -31,7 +31,9 @@ __date__ = "24/04/2018"
 
 
 import logging
+import weakref
 import numpy
+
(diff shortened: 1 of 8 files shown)


t20100:
Updated CurvesROIWidget and SyncAxis use of weakref.

vallsv:
I am afraid of the changes on the syncaxis as it was the result of a lot of tries and errors. But let see if some code complains.

t20100:
Tests were borken, apparently because the circular reference in the PlotWidget was needed... Let's see how it does, but to me the lazy removal of deleted axis should help.

Issue #1875, opened 2018-06-01
[plot3d] numpy.version check


vallsv:
The code contains
```
    if numpy.version.version < '1.8.0':
```
But
```
In [7]: "1.12.1" < "1.8.0"
Out[7]: True

In [8]: "1.14.0" < "1.8.0"
Out[8]: True
```

`pkg_resources` could be used
```
pkg_resources.parse_version(numpy.version.version) < pkg_resources.parse_version("1.8.0")
```

It could be nice to check all the code checking version.
`pkg_resoruces` was already a dependancy, i think.

t20100:
`pkg_resources` is a dependency as it comes with `setuptools`.

This  check was for a fallback for Debian7... Do we still want to support for Debian 7?
As we don't develop nor test on debian 7 I guess there are places where there are some issues...

PiRK:
About Debian 7, FYI:

> Subject: Debian 7 Long Term Support reaching end-of-life
> Fri,  1 Jun 2018 17:42:46 +0000 (UTC)
> debian-announce@lists.debian.org
> 
> The Debian Long Term Support (LTS) Team hereby announces that Debian 7
> "Wheezy" support has reached its end-of-life on May 31, 2018, five years
> after its initial release on May 4, 2013.
> 
> Debian will not provide further security updates for Debian 7. A subset
> of Wheezy packages will be supported by external parties. Detailed
> information can be found at Extended LTS [1].
> 
>     1: https://wiki.debian.org/LTS/Extended
> 
> The LTS Team will prepare the transition to Debian 8 "Jessie", which is
> the current oldstable release. The LTS team will take over support from
> the Security Team on June 17, 2018.

t20100:
timely...

vasole:
closed by #1876

PR #1876, opened 2018-06-01
[plot3d] Fix numpy version checking by removing it: Drop support of debian 7

t20100:
This PR remove a fallback for debian 7 and the associated version check which was erroneous.

This drops the support of the numpy version of debian 7.


t20100:
closes #1875

PR #1877, opened 2018-06-04
[StackView] Fix bug in clear()

PiRK:
The valueChanged signal needs to be disconnected before clearing the slider range..


vallsv:
LGTM

PR #1884, opened 2018-06-05
[plot] Update ROI

t20100:
This PR:
- Moves the max number of ROI management to `InteractiveRegionOfInterestManager` subclass
- Add the ROI to the list at the beginning of draw interaction so one can follow ROIs update while they are drawn.


vallsv:
LGTM

PR #1894, opened 2018-06-08
Clean up qhull wrapper in third_party

t20100:
This PR remove code from third_party qhull wrapper that is not used.


vallsv:
LGTM

PR #1897, opened 2018-06-08
[plot] Scatter profile tool

### LINKED ISSUE #1613: [plot] Extraction of profiles in scatter plots
- Delaunay? Combination of griddata and meshgrid?

t20100:
This PR adds a `ScatterProfileToolBar` which provides profile display on the "active" scatter plot.
I added a `silx.gui.plot.tools.profile` package for it and the profile toolbars for images and stack (and the ProfileMainWindow widget) should be moved there. And it would be nice to have all profile toolbar sharing the same base class (`_BaseProfileToolBar`).

It is in use in the `ScatterView` widget.

The profile computation is based on scipy LinearNDInterpolator and if it is not available it uses QHull delaunay + matplotlib linear interpolator (which have the drawback of not releasing the GIL during initialisation computation which freezes the GUI...).
The initialisation of the interpolator is done in a thread (on my PC it takes around 15 seconds for  10**6 points.

closes #1613

This profile tool can be extended with the support of other interpolators: Nearest (faster init), Inverse distance weighting.


vallsv:
It would be nice to split your unittest in several tests. It looks to be very big, which can be complicate to fix it is case of problem.

Anyway, it looks fine.

PR #1899, opened 2018-06-08
Fix test failed with PySide due to delegate being deleted

t20100:
Fix test failing with PySide


vallsv:
LGTM
Thanks for the fix

PR #1901, opened 2018-06-08
Fix few things

vallsv:
- Try to remove widget without parents (it should not change anything)
- Do not store signals from items as it is own by the items anyway

### DIFF
--- a/silx/gui/plot3d/SFViewParamTree.py
+++ b/silx/gui/plot3d/SFViewParamTree.py
@@ -1024,13 +1024,13 @@ class IsoSurfaceAddRemoveWidget(qt.QWidget):
         layout.setContentsMargins(0, 0, 0, 0)
         layout.setSpacing(0)
 
-        addBtn = qt.QToolButton()
+        addBtn = qt.QToolButton(self)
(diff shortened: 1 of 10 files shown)


silx/gui/plot3d/SFViewParamTree.py
```diff
@@ -155,7 +155,7 @@ def slotfn(*args, **kwargs):
                 for sigIdx, signal in enumerate(signals):
                     slot = gen_slot(sigIdx)
                     signal.connect(slot)
-                    slots.append((signal, slot))
+                    slots.append(slot)
```
t20100: `self.__slots` is used in `_disconnectSignals` which unpacks `signal` and `slot`... but `_disconnectSignals` doesn't seem to be used anywhere, so it can be removed...

vallsv:
My point is signals are anyway connected to a widget, then if you don't store the ref, the object will not be released anyway.

But it can be safer to revert this patch.

vallsv:
Oups, yes, you are right. My bad. It have to be fixed.

vallsv:
Rebased, with commit about signals removed.

PR #1902, opened 2018-06-11
Use Qt widget origin (top) rather than matplotlib origin (bottom)

t20100:
This PR changes the frame used for y pixel coordinates with matplotlib to use the same as Qt.

matplotlib pixel coordinates origin is the lower-left corner of the figure (i.e., widget), while Qt widgets (and the OpenGL plot backend) origin it their upper-left corner.

This worked fine as pixel coordinates were received from and interpreted by matplotlib.
There was still an issue: The `PlotWidget` methods `pixelToData`, `dataToPixel` and `getPlotBoundsInPixel` do not have a consistent behavior between both backends (one using lower-left origin and the other upper-left origin)...

### DIFF
--- a/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/silx/gui/plot/backends/BackendMatplotlib.py
@@ -833,15 +833,30 @@ class BackendMatplotlib(BackendBase.BackendBase):
 
     # Data <-> Pixel coordinates conversion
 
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM

vallsv:
BTW it is maybe more readable to have a transformation with both x and y, than only y. `qtx, qty = f(mplx, mply)`. But it is also fine like it is.

PR #1903, opened 2018-06-11
[plot] Fix position info update when active image changes

### LINKED ISSUE #1682: [plot] With Plot2D, data value in bottom bar is not updated when image change
With Plot2D, data value in bottom bar is not updated when image change if the mouse does not move

t20100:
Merge #1902 first.

This PR adds an `updateInfo` method to `PositionInfo` widget and makes use of it in `Plot2D` to update the displayed info when the active image changes.

closes #1682

While it would be worth to refactor/improve the `PositionInfo` widget, this PR only fixes the active image update issue.


t20100:
Reworked so that `Plot2D` uses `PositionInfo` from `PlotWindow`

vallsv:
LGTM

PR #1904, opened 2018-06-11
Clear silx view

vallsv:
The PR clear the models when closing silx view.

This avoid to trust Python for releasing objects, which can introduce unpredictable behaviors.

Relative to https://github.com/silx-kit/fabio/issues/220


t20100:
LGTM.
This PR apparently fixes https://github.com/silx-kit/fabio/issues/220

PR #1928, opened 2018-06-18
[silx convert] don't load complete file series into memory

### LINKED ISSUE #1925: [silx convert] A stack of EDF files should not be loaded into memory
There must be a threshold on the memory used for the conversion. One could use extendable datasets or create a dataset of the known final size.

PiRK:
The goal of this PR is to read and write one frame at a time when converting a EDF file series into a HDF5 3D dataset.

Closes #1925


vallsv:
LGTM

PR #1929, opened 2018-06-18
Add sum profile

payno:


vallsv:
Here is a patch to opimize the SVG: https://github.com/vallsv/silx/commit/b93474955d39f94a52430762ec3124b875bc2399

payno:
As we are gone for it, maybe adding the median can also be interesting...

vallsv:
LGTM

PR #1944, opened 2018-06-19
[plot] Refactor PositionInfo widget snapping mode

### LINKED ISSUE #1611: [plot] Snapping to point
Would it be possible for this to be available without the need of having the curve enabled?

t20100:
This PR allows a better control of the snapping of the position used by the `PositionInfo` widget through a `setSnappingMode` method.
It also adds support of snapping on scatter plots.
When used inside `PlotWindow`, it remains consistent with the previous automated snapping mode.

closes #1611

### DIFF
--- a/silx/gui/plot/PlotWindow.py
+++ b/silx/gui/plot/PlotWindow.py
@@ -259,7 +259,13 @@ class PlotWindow(PlotWidget):
                     converters = None
                 self._positionWidget = tools.PositionInfo(
                     plot=self, converters=converters)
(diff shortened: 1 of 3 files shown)


t20100:
@mretegan, if you have some time to test it, feedback would be welcomed!

vallsv:
Hi @mretegan, could you try again with this patch?

No idea why it is not working for you, but as windowHandler expect a native widget, it is many not the case on your platform.

mretegan:
Now it works. Thanks!

vallsv:
Let's merge it then

PR #1948, opened 2018-06-20
[sx] Add sx.enable_gui() function to load silx widgets from the notebook

### LINKED ISSUE #1130: [sx] Add possibility to load sx from the notebook with Qt
As is, if running in the notebook, `silx.gui` is not loaded.
It would be ideal to detect if running locally or remotely and load `silx.gui` when running locally...
Alternatively (simpler) the `sx` module can check if Qt is already loaded and then load `six.gui`.

t20100:
closes #1130


vallsv:
LGTM

Issue #1906, opened 2018-06-12
[plot] COM computation enhancements


t20100:
Stats widget COM computation is missing a few things:
- Y coord is missing for scatter plot data
- For images, scale and origin are not taken into account
- Maybe check that the sum of weights is not 0 to avoid divide by zero warning (e.g. for `sx.plot((-1, 0, 1))`)

PR #1950, opened 2018-06-21
Fix 1906

payno:


payno:
During doing it I have some doubt about consistency to 'order' results when computing stats.
For now it is ordering as (x, y). As it should be displayed. But to be more numpy compliant maybe having it y, x would be better. And manage it in the display. But I an not sure this is consistent inside the entire library.

silx/gui/plot/stats/stats.py
```diff
@@ -426,23 +439,41 @@ def __init__(self):
...
+            xOrigin, yOrigin = context.origin
 
-            ycom = numpy.sum(yData * dataYRange) / numpy.sum(yData)
-            xcom = numpy.sum(xData * dataXRange) / numpy.sum(xData)
+            denoY = numpy.sum(yData)
+            if denoY == 0.:
+                ycom = 0.
+            else:
+                ycom = numpy.sum(yData * dataYRange) / denoY
+                ycom = (ycom - yOrigin) * yScale + yOrigin
```
t20100: I think it is more: ycom = ycom * yScale + yOrigin

silx/gui/plot/stats/stats.py
```diff
@@ -426,23 +439,41 @@ def __init__(self):
...
+
+            denoX = numpy.sum(xData)
+            if denoX == 0.:
+                xcom = 0.
+            else:
+                xcom = numpy.sum(xData * dataXRange) / denoX
+                xcom = (xcom - xOrigin) * xScale + xOrigin
```
t20100: Idem

t20100:
There is also an issue with the coordinates of min and max when many points have this value.
An image can have many points with the same value, and this case leads to issues: To reproduce, try to display the stats of an image like: `numpy.ones(1000).reshape(10, 100)` or even `numpy.ones(10**6).reshape(1000, 1000)`

t20100:
As it is now, for images, coord min. and coord max. provides the coordinates of a corner of the pixel.
Wouldn't it be better to return the center of the pixel? so adding a 0.5 pixel offset.

t20100:
Switching back from visible range to full data range is not taken into account.

t20100:
I made a PR with a few fixes.

Also, when denominator is 0, it is better to return an invalid result (infinity, NaN).

closes #1906

silx/gui/plot/stats/stats.py
payno: No, if i am correct we have to move to the data reference to apply the scale then to go back to the plot reference.

payno:
> As it is now, for images, coord min. and coord max. provides the coordinates of a corner of the pixel.

Wouldn't it be better to return the center of the pixel? so adding a 0.5 pixel offset.
Don't know. Could be an option

> There is also an issue with the coordinates of min and max when many points have this value.
An image can have many points with the same value, and this case leads to issues: To reproduce, try to display the stats of an image like: numpy.ones(1000).reshape(10, 100) or even numpy.ones(10**6).reshape(1000, 1000)

We can display only the first coordinates or a QList of coordinates. Or having both (into two classes).

vasole:
To provide the center of the pixel, we should also draw the pixels at the center. At this moment, pixel 0 is represented with its center at 0.5. So, at this point I would not change anything.

When we represent an image as a scatter plot, there is an effective shift because the symbol is drawn at the actual coordinates. So, if something is to be changed at a certain point, is the visualization of the images with a shift of 0.5 pixels, the calculations will not change.

t20100:
LGTM, can be merged once CI is happy

t20100:
BTW, closes #1906

vallsv:
Everything's fine

Issue #1796, opened 2018-04-25
[plot] Scatter mask pencil tool


t20100:
As is the scatter mask pencil tool is no always useful as its minimum size is 1 in data coordinates which might not be suited (e.g. data range < 1).

When using log scale axes, the pencil circle is no more a circle as it is drawn in data coordinates rather than screen coords.

t20100:
Issue moved to 0.9 as PR #1957 provides a temporary not ideal but working solution for the pencil size issue.

PR #1957, opened 2018-06-26
[plot] Scatter mask pencil

t20100:
This PR makes the scatter mask tool pencil size relative to the extent of the scatter data.
This makes this tool more usable with scatter plots of different data range

related to #1796

To me, this is a temporary solution for the 0.8 release. The ideal solution is that the pencil shape is defined in screen coordinates rather than in data coordinates.


vallsv:
It would be better to use `dataExtent` insteadof `_data_extent`, but it is fine as it is.
LGTM

t20100:
yes indeed, and this is also true for other private attributes

Issue #1642, opened 2018-02-21
[plot] Profile should use coordinates not just indices


vasole:
- When extracting a horizontal or vertical profile the x axis should correspond to the original coordinates
- When extracting a line profile one has two possibilities a) to use as x axis the coordinates corresponding to longest projection or b) to use the distance between the points.

mretegan:
Yeah, that would be useful.

t20100:
For v0.8, do it for horizontal and vertical profile at least

t20100:
Move to v0.9 as PR #1959 implements it for horizontal and vertical profile.

Free line profile still needs to be addressed.

PR #1959, opened 2018-06-27
Image profile coordinates

t20100:
This PR scales horizontal and vertical profile x coordinates according to image origin and scale.
It does nothing for free line profile and nothing for stack profile.

related to #1642


vallsv:
LGTM

Issue #1961, opened 2018-06-27
[appveyor] Impossible to import anymore PyQt5


vallsv:
```
PyQt5            5.10.1       
```
The version of PyQt5 did not change since February.

It could be a problem with Windows VC Redistribuable dependancy or something like that? Maybe not installed in latest Appveyor env. No idea.

Here is the output from the tests
```
python run_tests.py --installed -v --no-opencl --qt-binding %QT_BINDING%
INFO:silx.setup:Use setuptools
INFO:silx.setup:Use setuptools.setup
WARNING:run_tests:resource module missing
Traceback (most recent call last):
  File "run_tests.py", line 431, in <module>
    import PyQt5.QtCore  # noqa
ImportError: DLL load failed: The specified module could not be found.
```

t20100:
It's related to numpy v1.15.0rc1. I reproduced locally, and see https://github.com/numpy/numpy/issues/11431.

t20100:
closed by #1970

PR #1970, opened 2018-06-28
[ci] update appveyor config

t20100:
Attempt to fix appveyor issue

### DIFF
--- a/ci/appveyor.yml
+++ b/ci/appveyor.yml
@@ -12,7 +12,7 @@ notifications:
   on_build_failure: false
   on_build_status_changed: true
 
-os:
+image:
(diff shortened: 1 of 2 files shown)


t20100:
related to #1961

payno:
LGTM

charris:
If anyone here is in a position to do so, we (NumPy) would appreciate a bisection to discover which commit caused this problem.

t20100:
@charris, I reproduced the issue with `import numpy; import PyQt5.QtCore` and it disappear when I build numpy from source (see https://github.com/numpy/numpy/issues/11431#issuecomment-401041624)

PR #1977, opened 2018-06-28
[convert]preserve dtype for file series

### LINKED ISSUE #1976: [convert] when converting a file series, the dtype is not preserved
Converting a series of EDF frames in uint16 (unsigned short) creates a 3D volume in float32. That creates an output file twice the size of the input dataset.

PiRK:
Close #1976 

When initializing the output dataset prior to writing frames, the dtype must be specified.


vallsv:
LGTM

PR #1979, opened 2018-07-02
Icon for unsupported ROIs

vallsv:
This PR allow to use costom ROIs in the ROI manager.
A default display will be provided.


t20100:
The icon looks to have an embedded PNG image, it would be better to avoid it for optimum compatibility.
We had issues with icons with images, so it is best to avoid it in new icons

t20100:
I made a PR on your branch with an optimized svg file

vallsv:
Oups, maybe i missing to remove my shadok template.
It would be better to fully remove it from the history, i guess it's a big image.

t20100:
It's probably nothing compared to the history of Cython generated files...

vallsv:
Yes you are right. Anyway, this PR is now cleaned up from the RAW image.

PR #1980, opened 2018-07-02
Remove engine param as we only have a single engine

vallsv:
As discussed with @t20100 a long time ago, our scipy implementation is here but hidden. We only have one engine, then it is maybe not really needed to provide an attribute for that.


t20100:
LGTM

PR #1981, opened 2018-07-02
[CI] Install scipy from brew

t20100:
scipy is not yet available as wheel from pypi for Python3.7.
This is a an issue for CI on macOS.
This PR pre-install scipy through homebrew in this case.


vallsv:
LGTM

PR #1991, opened 2018-07-03
Few fixes for 0.8 release

t20100:
- Add qt svg to debian dependencies (for print preview)
- Fix init of logging in imageview.py example
- add python 3.7 to classifiers


vallsv:
LGTM

PR #1995, opened 2018-07-04
Compare 2 images

vallsv:
This PR provide a widget to compare 2 images.
```
from silx.gui.plot.CompareImages import CompareImages
window = CompareImages()
window.setData(data1, data2)
window.setVisible(True)
```

- Visualization mode: vertical/horizontal line, blue/red channel composition
- Alignment mode: center/origin of images, stretch images, , or use sift
- Supports intensity images, RGB, RGBA

```
./bootstrap.py examples/compareImages.py joconde.png joconde2.png
```

TODO
-------

- [ ] Use fabio/silx/hdf5 to load images from the example
- [ ] Use PlotWidget instead of Plot2D
- [ ] Remove useless plot tools
- [ ] Provide an API to custom the OpenCL context
- [ ] Display transparent borders instead of black border (for aligned images)

Sift improvement:
- [ ] Allow to align intensity images with RGB images 
- [ ] Allow to align 2 images with different sizes
- [ ] Allow to align an RGBA image
- [ ] Extract keypoints and align image in a single pass (the python API do not allow that)

compareimages


payno:
Nice widget. Will be really usefull. Can be an application from my point if view.

A few questions/observations that might be of some interest:

- a tooltip or a widget displaying extra information can be usefull with pixel value of img1, img2, and/or diff of values.
- an option to choose between global intensity normalization (which is the case now) and 'local/individual image intensity' normalization can be usefull in some case
- I prefer your 'gray compare mode ' to the 'single channel compare mode'

vallsv:
The code can be reviewed.

On a side note, i think it is needed to add an API to custom the OpenCL context. And a default one should be provided by silx. And we should provide a dialog to select it. Then for another PR i guess.

t20100:
It's a pity the image plot items and the backend does not provide something like `get|setVisibleRect`.
For the OpenGL backend it would avoid to update the image when the separation marker is moving, as it only needs to update which part is displayed... (maybe for later)

t20100:
Just a question on the structure: why not making the `CompareImages` a MainWindow and adding toolbars there ?
Also it's probably useful to be able to access the underlying plotWidget from the public API.

silx/gui/plot/CompareImages.py
```diff
@@ -0,0 +1,1141 @@
...
+        action.setIconVisibleInMenu(True)
+        action.setCheckable(True)
+        action.setShortcut(qt.QKeySequence(qt.Qt.Key_C))
+        action.setProperty("mode", VisualizationMode.COMPOSITE_RED_BLUE_GRAY)
+        menu.addAction(action)
+        self.__brChannelModeAction = action
+        self.__visualizationGroup.addAction(action)
+
+        icon = icons.getQIcon("compare-mode-rbneg-channel")
+        action = qt.QAction(icon, "Yellow/cyan compare mode (substractive mode)", self)
```
PiRK: subtractive (sans le s)

silx/gui/plot/CompareImages.py
```diff
@@ -0,0 +1,1141 @@
...
+        self.__alignmentAction = qt.QAction(self)
+        self.__alignmentAction.setMenu(menu)
+        self.__alignmentAction.setIconVisibleInMenu(True)
+        self.addAction(self.__alignmentAction)
+        self.__alignmentGroup = qt.QActionGroup(self)
+        self.__alignmentGroup.setExclusive(True)
+        self.__alignmentGroup.triggered.connect(self.__alignmentModeChanged)
+
+        icon = icons.getQIcon("compare-align-origin")
+        action = qt.QAction(icon, "Align images on there upper-left pixel", self)
```
PiRK: their

vallsv:
- Yes, you are right about the `MainWindow`, but i really dont like the idea to use it for a widget. Then i did a kind of cheat, which is definitly not good.
- For the `setVisibleRect` it is not really big issue cause we don't really need to be efficient here. But i ask myself before starting if 3D widget and its scene could be used here, instead of a 2D plot :-)

t20100:
The plot3d widgets for now only provides interaction for 3D content which would sounds weird if you only display 2D images...

vallsv:
And yes, there is already a `getPlot` method.

payno:
LGTM

PR #1997, opened 2018-07-05
[gui] PySide2 support

t20100:
This PR upgrades supportof PySide2 in silx.
It fixes a few issues, modifies a bit the tests and adds PySide2 to travisCI config

PR #1998, opened 2018-07-05
[gui] Deprecate PySide Qt binding

t20100:
Merge PR #1997 first!

This PR:
- adds a deprecation warning when loading PySide,
- moves PySide to the last probed Qt binding
- removes PySide from the documentation.

Are you OK with this?


vasole:
OK with me,

vallsv:
LGTM then

Issue #2003, opened 2018-07-06
[silx view]Error on drag and drop


PiRK:
When I drag and drop a file into `silx view`, I get an error 

```
ERROR:silx.gui.qt._qt:<type 'exceptions.AttributeError'> 'Hdf5LoadingItem' object has no attribute 'h5Class'   File "/users/knobel/.local/lib/python2.7/site-packages/silx/gui/hdf5/NexusSortFilterProxyModel.py", line 202, in data
    if self.__isNXnode(item):
  File "/users/knobel/.local/lib/python2.7/site-packages/silx/gui/hdf5/NexusSortFilterProxyModel.py", line 93, in __isNXnode
    class_ = node.h5Class
```

But after closing the error popup, it works fine.

t20100:
Closing, it should be fixed by PR #2004

PR #2004, opened 2018-07-06
Fix bunch of GUI issues

vallsv:
- Fix files created with new HDF5 lib and loaded with old HDF5 lib
- Fix ndim compound types
- Fix display of threaded load of files

#2002
#2003


t20100:
LGTM

PR #2008, opened 2018-07-11
Replace numpy.fromstring with numpy.frombuffer when reading binary data

PiRK:
This addresses a deprecation warning in numpy 1.14. See https://github.com/numpy/numpy/pull/9487


vallsv:
LGTM

PR #2013, opened 2018-07-12
[SaveAction] Activate "curves" filter even for single curve

PiRK:
Until now, we activated "curve" filters if there was an active curve or a single curve, and we activated "curve**s**" filters when there was at least curve.

There is an issue if one wants to add a function that can save indifferently one or more curves. The only solution with silx <= 0.8 is to add the filter twice.

This PR activates "curves" filters for  plot with one or more curves. As far as I can tell, the only function currently provided for multiple curves can already handle one or more curves.


vallsv:
LGTM

PR #2018, opened 2018-07-13
Add support of NaN in cpp median filter

### LINKED ISSUE #2016: support of NaN in median filter
I need to ignore NaNs while doing a median filter.
I haven't found such feature (not available in scipy except through generic_filter+numpy.nanmedian which is really slow).

t20100:
This PR adds support of NaNs by ignoring them in cpp median filters.

In case of an even number of values on which to do the median, it gets the highest of the 2 central values.

closes #2016

### DIFF
--- a/silx/math/medianfilter/include/median_filter.hpp
+++ b/silx/math/medianfilter/include/median_filter.hpp
@@ -220,6 +220,7 @@ void median_filter(
                 T min = 0;
                 T max = 0;
                 getMinMax(window_values, min, max, window_end);
+                // NaNs are propagated through unchanged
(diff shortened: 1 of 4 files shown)


t20100:
Added some optimizations:
- Remove indirect in window_values and use `nth_element` without `cmp` argument.
- Avoiding handling border specific stuff for most of the image. This part can be further optimized, but already provides a speed-up.

While more optimization could be done, this makes this median filter comparable to alternatives, so
closes #2019

Benchmark of the current master:
medfilt-master-noopenmp

Benchmark of this pull request:
medfilt-new-noopenmp

payno:
Optimizations look very nice.

PR #2032, opened 2018-07-17
Fix few other warnings

vallsv:


PiRK:
Looks like the non-deprecated unittest methods are not available in python 2.7.*.
That's the perfect excuse for dropping python 2.

vallsv:
Oho crap...

kif:
many tests failed in python2

PR #2034, opened 2018-07-18
Add iso level slider normalization setting in ScalarFieldView param tree

t20100:
This PR adds the option to set a normalization (linear or arcsinh) to the sliders controlling the level of isosurfaces in the ScalarFieldView parameter tree.


vallsv:
LGTM

PR #2036, opened 2018-07-18
Use MDI window for GUI tests (WIP)

vallsv:
This PR tries to reduce some inconvenience of GUI testing, by using an only one MDI window as container for all GUI tests.

It reduce blinking, then allow to move the tests manually, or reduce them. It could maybe speed up a little the tests.

- This feature do not working with PyQt4. Some widget are not released at the end of each tests.
- Only dialog tests was updated

Can be tested using
```
python ./run_tests.py silx.gui.dialog.test.suite -v
```


vallsv:
Outdated, and pytest-xvfb does the job

PR #2037, opened 2018-07-18
Deprecate pyqt4

### LINKED ISSUE #2015: [gui] Mark PyQt4 as deprecated
Expected end of support is in a year or so.

t20100:
This PR marks PyQt4 as deprecated, removes it from the documentation and also replaces a few occurrence of PySide with PySide2.

closes #2015


vallsv:
LGTM

PR #2038, opened 2018-07-18
[plot]change condition for choosing the dock widget area

### LINKED ISSUE #2011: [Plot] DockWidgets should account for plot aspect ratio
In PyMca, the legends where shown at the right of the plot if the plot widget was wider than taller.

It would be desirable to have the same behavior in silx.

PiRK:
Use the legacy pymca condition w > 1.25*h to initialize docks in the RightDockWidgetArea instead of the BottomDockWidgetArea.

Closes #2011.
Closes https://github.com/vasole/pymca/issues/277


vallsv:
LGTM

PR #2039, opened 2018-07-18
Stop interaction when hiding roi tools dock widget

### LINKED ISSUE #2033: plotInteractiveImageROI example issue when hiding ROI tool
When the ROI dock widget is hidden, the interactive mode is not reset, this sounds weird as one can continue to draw ROIs while the tool is hidden.

t20100:
This PR improves ROI sample code when hiding the roi tools dock widget

closes #2033

### DIFF
--- a/examples/plotInteractiveImageROI.py
+++ b/examples/plotInteractiveImageROI.py
@@ -99,8 +99,17 @@ widget.setLayout(layout)
 layout.addWidget(roiToolbar)
 layout.addWidget(roiTable)
 
+def roiDockVisibilityChanged(visible):
+    """Handle change of visibility of the roi dock widget
+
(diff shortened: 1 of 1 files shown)


vallsv:
Maybe it could be integrated inside the tool itself.

Here is a code i unsed on pyFAI which is supposed to catch visibility change. It could be checked?
```
    def eventFilter(self, widget, event):
        result = super(AbstractCalibrationTask, self).eventFilter(widget, event)
        if event.type() == qt.QEvent.Show:
            self.widgetShow.emit()
        elif event.type() == qt.QEvent.Hide:
            self.widgetHide.emit()
        return result

```

t20100:
Yes it works, not sure it is simpler though

Issue #1447, opened 2018-01-08
Display legend in plot


UUSim:
How can I display a legend with all curves in a PlotWidget?

(Similar to the legend functionality in matplotlib? [Legend guide](https://matplotlib.org/users/legend_guide.html))

vallsv:
Hi,

We don't have this concept on legend in silx, but the default Plot1d provide a widget containing the list of the curves. You can find it on the bottom of the widget: `Options`, then check `Legends`.
![screenshot plotwidget legend](https://user-images.githubusercontent.com/7579321/34674630-9540959c-f47e-11e7-9bd1-5d037b452091.png)

If you use a real `PlotWidget` you can create your legend in this way:
```
import silx.gui.plot
plot = silx.gui.plot.PlotWidget()
plot.addCurve([1,2,3],[0,1,2],legend="foo")
legend = silx.gui.plot.LegendSelector.LegendsDockWidget(plot=plot)

plot.show()
legend.show()
```

If it is not enough for you we can show you how to hack the matplotlib backend of our widgets. But this part will not be stable.

UUSim:
Thanks a lot for the explanation. This is not exactly what I'm looking for, although it might be useable after some tweaking:
- Reducing the visual space taken by the widget.
- Allow for curve aliases (show a different name for a curve)
- Ideally, the legend should be included when performing an export to png.

On a side note: We are using the `opengl` backend, so using the matplotlib legend is not an option.

vasole:
If I am not mistaken, the legends widget support a context menu allowing to rename a curve.

vasole:
In PyMca one distinguish display and final output. For the later one allows full customization. I cannot imagine the user customizing everything for each plot.

From this plot:

![image](https://user-images.githubusercontent.com/5636251/42622501-d2857b10-85c0-11e8-8149-ac9f9383045e.png)

One reaches this one where everything can be customized for the final publication quality output.

![image](https://user-images.githubusercontent.com/5636251/42622643-24e6d9f8-85c1-11e8-8a22-e9cd9938f08b.png)

UUSim:
This issue is fixed since the PR #2041 has been merged and released.

Thanks guys!

PR #2041, opened 2018-07-20
Curve legends widget

### LINKED ISSUE #2009: Pymca-like display of legends?
Users would like to see legends like in PyMCA, under Plot1D without the need to click on 'options' then 'legends', and with a compact view.

t20100:
This PR adds a `CurveLegendsWidget` to the plot associated tools widgets.
The idea is to have a widget taking less space that `LegendSelector` and (at least as it is) with less functionalities (i.e., no possibility to change the curves).

curvelegendswidget

To acheive this, it uses a `FlowLayout` and reuse `LegendSelector` icon widget.

Todos:

- [x] Documentation
- [x] Tests
_- Improve FlowLayout size management and alignment (kept for later:  #2059)_

related to  #1447
closes #2009


t20100:
Idea from @vallsv add active curve selection from the legend widget

vasole:
What about implementing signals in line with sigLegendClicked and sigLegendContextMenuRequested? That would allow the application to implement whatever they want.

t20100:
Good idea, that would allow extensions

t20100:
Added:
- `sigCurveClicked` signal which provides the curve object corresponding to the clicked legend.
- `curveAt(QPoint)` method which returns the curve corresponding to the Legend widget at the given position. This makes it easy to use QWidget's customContextMenuRequested signal (see plotCurveLegendsWidget.py for sample code) and is even more generic (this is used for the `sigCurveClicked`)

I opened issue #2059 to do " Improve FlowLayout size management and alignment" later, so ready for review.

t20100:
Update: instead of hiding the legend when the curve is hidden, it is now displayed with 'disabled' style.
Example updated with show/hide curve context menu option and active curve set/unset.

PiRK:
LGTM.

PR #2048, opened 2018-07-25
[gui] Add a range slider with background image

### LINKED ISSUE #1324: [gui] Add a range slider
This would be useful to have  in our widget zoo.
@dnaudet wrote one in xsocs with a colormapped histogram background: https://gitlab.esrf.fr/kmap/xsocs/blob/master/xsocs/gui/widgets/RangeSlider.py

t20100:
This PR adds a `RangeSlider` to `silx.gui.widgets`:

rangeslider

The initial version of this slider comes from xsocs (thanks to @dnaudet ).
I updated the API to make it somewhat closer to QAbstractSlider and QML RangeSlider API. I updated the behavior so that the slider resolution is independent from the background image resolution.

closes #1324

TODOs:

- [x] Documentation, sample code
- [x] Tests


vallsv:
Very nice. I need something like that soon :-)

t20100:
Check if possible to rework number of steps (make it optional and by default use pixel as step) ?

t20100:
- Updated so that number of steps is optional
- Added documentation and a few tests.

Ready for review.

vallsv:
LGTM

PR #2049, opened 2018-07-26
[stackView]Fix perspective change in setData

PiRK:
The condition `perspective != self._perspective` was always False (attribute already updated when reaching this code).

Thus, the drop-down list was not updated properly.


t20100:
I made a PR on your branch with more fixes: Add get|setPerspective and getCalibrations, fix title updates.

PiRK:
I don't remember exactly our discussion, when we did these changes, but I think that preserving the perspective is a feature we wanted to add. So I will remove the test that is failing because of the perspective not being reinitialized. (https://github.com/silx-kit/silx/pull/2049/commits/7cdcca5348ee1f3cbbcbb89cb0fd53cae30236fe)

PR #2050, opened 2018-07-26
Ship examples within lib

vallsv:
This PR ship examples with the lib without changing anything to the current project directory tree. This looks to nice as examples are still exposed at the very first level. pyqtgraph example are also provided like that.

- [x] Ship the examples with the lib
- [x] Allow to execute example from the bootstrap
- [ ] There is a problem with `fft.png`. this resource is not stored in the package. I tries many ways without finding any solution. `package_data` looks to be unsupported by `numpy.distutils`, but i can be wrong. Anyway, I don't think it's a big issue in this case.

Examples can be imported like that
```
import silx.examples
```
And tested like that
```
./bootstrap.py examples/imageview.py

or

./bootstrap.py -m silx.examples.imageview
```


vallsv:
I just fixed the conflict.

t20100:
I made a PR on your branch to copy fft.png during build.
It also removes `examples/setup.py` which is not needed.

Otherwise, looks good to me!

vallsv:
Nice. Looks more easy than things i tried.

t20100:
I remember having a hard time with it for `silx.resources`...

PR #2052, opened 2018-07-30
Make bootstrap and run_tests work with multiprocessing on Windows

t20100:
This PR makes bootstrap.py and run_tests.py works with multiprocessing application on Windows where care is needed for things to occur in `if __name__ == "__main__"` section or one ends up with new processes creating new processes infinitely....


vallsv:
LGTM

PR #2060, opened 2018-08-02
Make silx.sx doc build without display

t20100:
This is the case for our nightly builds (and maybe also for debian)


vallsv:
LGTM

PR #2064, opened 2018-08-02
Add extras_require to setup.py

### LINKED ISSUE #2031: [build] Use extra_requires
This allows to define optional dependencies and to install them from pip:
`pip install silx[gui]` for example or `pip install silx[all]` instead of relying on an external requirements file.

t20100:
This allows for : `pip install silx[io]` or even `pip install .[io]`
Available "extras": `io`, `opencl`, `gui`, `all`
`gui` and `all` are using PyQt5.

The extras_require somewhat duplicate the requirements.txt file

I also added Cython to setup_requires as now it is needed for the build.

closes #2031


t20100:
As discussed, changed to a single `full` extra target.
Also removed cython from `setup_requires` as it is not needed with source distribution.

PiRK:
LGTM. Tests are  failing due to PySide (cf #2091)

PR #2071, opened 2018-08-14
Add selection table for compare imgs

payno:


payno:
@vallsv are you ok with the modification of your original example to pack the `UrlSelectionTable` ?
The current example (compareImages) will only deal with single frame file. It might be interesting to extend it if needed. But if we want to make a more complete example then maybe an application will have to be created.
For now I don't intend to do more on this.

vallsv:
LGTM

PR #2072, opened 2018-08-16
Fixes AttributeError in Colorbar.getLegend

titusjan:
There is a small bug in the `Colorbar.getLegend`. You can reproduce it with the following example.

```
from silx.gui.plot.ColorBar import ColorBarWidget

app = QApplication([])
colorbar = ColorBarWidget()
colorbar.setLegend('Hi there')
print(colorbar.getLegend())
```

This PR fixes the bug.


payno:
Thanks for the fix.

PR #2094, opened 2018-09-03
display the Nexus class name into the tooltip & datatype for HDF5groups

### LINKED ISSUE #2090: [viewer] Support nexus class attributes in the tree
Nexus defines the content of a group by using "class-attributes".
It would be nice to have those class attributes displayed in the tree, for example in the "Type" column

kif:
close #2090


PiRK:
LGTM

PR #2104, opened 2018-09-05
[MaskWidget] Save mask as hdf5

### LINKED ISSUE #1945: [plot] Add hdf5 support to load/save mask
We support EDF (and a few others) to load/save mask from the mask tool widgets, but not HDF5... this is kind of missing...

PiRK:
Support for hdf5 for loading and saving masks.

Closes #1945 

A new dialog is needed to select a dataset (existing or new) in a HDF5 file. The code can be reused from existing GroupDialog used to save plots as NXdata. But it needs to be improved in order to support a *read/open* mode, and to select a group while specifying a new dataset name in *write/save* mode.

### DIFF
--- a/silx/gui/dialog/DatasetDialog.py
+++ b/silx/gui/dialog/DatasetDialog.py
@@ -1,0 +1,105 @@
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2018 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


PiRK:
Ready to be merged. 

To test it, you can use `python bootstrap.py examples/imageview.py` and play around with the mask tool.

t20100:
I have a small comment on the display of the current selection and messages.
I would display messages like "Select a dataset" instead of "Current Selection" rather than below and not in light gray.

Otherwise LGTM

PR #2105, opened 2018-09-05
Add fat binary generation for silx

t20100:
This PR adds the necessary script to create a fat binary of silx based on pyinstaller.

A few questions:
- Only provide silx view? as it is now, it can only be used from the command line silx view
- Generate a fat binary with appveyor?

TODO: Document the process, add pyinstaller to requirements-dev


vasole:
Under windows you can go a step further by making/preparing an NSIS installer.

In PyMca:

https://github.com/vasole/pymca/blob/master/nsisscript.nsi.in

that is modified by:

 https://github.com/vasole/pymca/blob/master/cx_setup.py

vasole:
The goal of that would be to provide silx view also as an application accessible from the start menu.

t20100:
To answer myself:
- It now contains 2 .exe (`silx.exe` and `silx-view.exe`) in the same folder.
- I think doing it on appveyor is overkill: consume useless resources + most probably fail often (when pyinstaller receipes need to be update). It should be well enough if it is done for releases.

I added some documentation on how to proceed.

I created issue #2106 to add creation of a Windows installer as it is done in PyMca, which would be really nice.
For now it produces a folder that can be zipped. A zip with the produced Windows fat-binary is available here: http://www.silx.org/pub/silx/silx-0.9.0dev+581c298.zip

PR #2107, opened 2018-09-06
[plot3d] Add a special item for the top group of the scene

t20100:
This PR changes the way to handle the top-level group of the SceneWidget so that applied transforms are taken into account in the axes.
Before applying a transform at the top level had no visible effect, while now it is visible in the value of the axes.


PiRK:
LGTM

PR #2111, opened 2018-09-07
Fix opening spec file without scans

### LINKED ISSUE #2110: [io] SegFault when opening spec file without scans
`silx.io.specfile.SpecFile` segfault if there is no scan in the file.

As it is now, it looks for `#S` or `#F` in the first 10 lines of the file...
Looking for `#S` in the entire file (as pymca does) - and may be also look for `#F` (spec's spec to be checked) - turns the segfault into a python exception.

t20100:
This PR proposes a patch for the C specfile library to avoid segfault when opening a spec file without scans. It also fixes a small memory leak.

As I am not familiar with specfile, I think a lot of proof-reading is needed.

closes #2110

### DIFF
--- a/silx/io/specfile/src/sfinit.c
+++ b/silx/io/specfile/src/sfinit.c
@@ -1,5 +1,5 @@
 # /*##########################################################################
-# Copyright (C) 1995-2017 European Synchrotron Radiation Facility
+# Copyright (C) 1995-2018 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


vasole:
- OK at the PyMca side. No crash when accessing directly.
- No crash with silx.

@PiRK 
- The silx.io.specfilewrapper module misbehaves under windows (at least with the test file used) reporting methods missing when they are there.

vasole:
@PiRK 

Forget my previous comment. The problem is specfilewrapper provides access to SpecFile (silx) and Specfile. In My opinion specfilewrapper should import SpecFile as _SpecFile to prevent confusion.

The direct binding of PyMca returns an empty list when accessing allmotors() of an empty scan while the one of silx returns Scan not found error.

PR #2113, opened 2018-09-07
Add PlotWidget.getItems

t20100:
This PR adds a `PlotWidget.getItems` method which returns all items of the plot.

In the long term, this is a replacement for `getAllCurves`, `getAllImage` and any missing `getAll*` methods.
One can then filter out what he wants from the returned tuple.

Use case: PR #2112 add some methods to marker items, but there is no way from the public API to retrieve the marker items....


vallsv:
In case of filtering the result, an iterator should be better here no? That's now a common thing with Python3.

t20100:
We can use an iterator indeed.

PR #2114, opened 2018-09-11
Improve support of Qt designer with PySide2

t20100:
This PR adds the support of the `Line` class which is produced by the Qt designer but is not supported by PySide2 (it's not a Qt widget).
With this, one can use Line in the Qt designer and use PySide2


payno:
LGTM

PR #2115, opened 2018-09-11
Fix 2027

payno:


vasole:
Start from a clean branch. The first commit should not be there.

PR #2116, opened 2018-09-12
add platform_machine restriction on pyopencl in requirements.txt

### LINKED ISSUE #2070: Problem at Debian side or at our side?
@picca 

Unless we have made a mistake, pyopencl is an optional dependency for silx. However, it seems to have been set somewhere as a hard dependency thus leading to unavailability of depending packages on some platforms:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=905348

t20100:
closes #2070


kif:
LGTM

Issue #2120, opened 2018-09-19
Clean-up try/except import of fabio and h5py


t20100:
As fabio and h5py are strong dependencies, this kind of special case is no longer needed

alexmarie78:
Hello,
I'm trying to deal with this issue.
Is it "import fabio" or "import h5py" replacing "try/except" blocks ? 
Or does it need something else ?

t20100:
Great!

It's not just replacing:
```
try:
    import fabio
except ImportError:
    fabio = None
```
with: `import fabio`, it is also cleaning-up the usage of it such as removing `if fabio is None: ...`, remove test skipping (e.g., https://github.com/silx-kit/silx/blob/master/silx/io/test/test_fabioh5.py#L50), etc...

Maybe it is worth splitting this if it's a lot of changes (e.g., going submodule by submodule, or one for `fabio` another one for `h5py`)

alexmarie78:
Thanks ! 
I'll do that.

alexmarie78:
Hello,
While cleaning #2120, I ran through these errors:
- [1]

ERROR: test_matching (silx.opencl.sift.test.test_matching.TestMatching)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/experiences/diffabs/mariea/Github/silx/silx/build/lib.linux-x86_64-2.7/silx/opencl/sift/test/test_matching.py", line 162, in test_matching
    delta = abs(res_sort - siftmatch).max()
ValueError: operands could not be broadcast together with shapes (491,2) (488,2)


- [2]

ERROR: testH5pyNotInstalled (silx.app.test.test_convert.TestConvertCommand)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/experiences/diffabs/mariea/Github/silx/silx/build/lib.linux-x86_64-2.7/silx/utils/testutils.py", line 227, in wrapper
    result = func(*args, **kwargs)
  File "/home/experiences/diffabs/mariea/Github/silx/silx/build/lib.linux-x86_64-2.7/silx/utils/testutils.py", line 183, in __exit__
    expected_count, logging.getLevelName(level), count))
RuntimeError: Expected 1 ERROR logging messages, got 0

These two errors (in python2 run_tests.py) prevent me from doing a PR, but I don't know how to deal with it

t20100:
Hi,

What's your OS? and do you know what are the available opencl drivers?

alexmarie78:
[...]

PR #2121, opened 2018-09-19
Add h5py and fabio as strong dependencies

### LINKED ISSUE #2024: Add h5py and fabio as "strong" dependencies
Those dependencies are available the same way as silx (i.e., debian package, wheels on pip for the same targets) and are already more or less requested for `silx.io` and so for all parts depending on it in `silx.gui`.
This would also spare us the management of missing dependency in modules which is barely tested.

t20100:
.. and update doc accordingly
The associated code clean-up is not part of this PR (see issue #2120)

closes #2024

### DIFF
--- a/ci/requirements-pinned.txt
+++ b/ci/requirements-pinned.txt
@@ -1,1 +1,2 @@
 PyOpenGL     # PyOpenGL pre-release broken on Windows (#2074)
+pybind11     # For pyopencl 2018.2 to build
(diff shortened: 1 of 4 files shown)


t20100:
Ready for review.

This PR also pinpoint pyopencl to 2018.1.1 until issues with 2018.2 get fixed

kif:
LGTM

PR #2124, opened 2018-09-20
Implement active curve selection mode legacy

### LINKED ISSUE #2117: [gui] Lost active curve on plot interaction
#1612 has the side effect that when selecting a marker the active curve is lost. That also happens with ginput.

It should be possible to toggle off that behavior without having to inherit PlotWidget.

It is very cumbersome in applications associating actions to single mouse clicks in the plot area.

vasole:
closes #2117


vasole:
Set to work inprogress to restore old behavior of addCurve setting the added curve as active iif it is the only curve and the active curve selection mode is AlwaysOne

vasole:
Ready to merge.

t20100:
I would use lower case strings ('alwaysone' and 'atmostone') to be consistent with other places in the plot where strings are used as enum (e.g., yaxis='right', 'left', colormap and color names, interaction modes...)

vasole:
Changes implemented.

t20100:
I looked further at it, there is no need to store both active curve handling and the mode, in Qt it is usually stored as a single "selection mode" property.
I'm preparing a PR on your branch with this and other things to discuss (e.g.: in always one mode, why just setting active curve if there is only one?)

t20100:
I made a PR on your branch with some propositions.

vasole:
Why setting active curve if there is only one? To clearly show the active curve color and not the color of the curve.

There can be actions associated to a curve becoming active like showing its calibration.

t20100:
I still don't get why a curve should be activated only when there is one curve in "alwaysone" mode.
To me it is miss-leading to call the mode "alwaysone" and have some cases where you have some curves visible but no active curve..

t20100:
I propose to add a "none" mode to disable the active curve handling, because this PR propose to have (at least) 3 active curve handling modes: no active curve selection, "alwaysone" or legacy one from PyMca and the current one ("atmostone").
I would not use 2 methods to enable/disable active curve handling to select to mode, as it makes setting the mode and getting it more complicated than it should (as an example, in Qt selection modes include a "no selection' mode).

vasole:
I was afraid of breaking someone else code and therefore I took the conservative approach of adding something without touching the rest. 

So you prefer to have 'none' as equivalent of disabling active curve handling, legacy as the ancient behavior and atmostone as the current one. No problem.

Where I think we differ is on the behavior when the active curve handling is toggled.

- In none, I guess if there is an active curve, there will be a call to setActiveCurve(None) to make sure the curve recovers its look.
- In atmostone basically I have no opinion. I would do nothing on activation, even if there is a single curve visible but it is not "my mode".
- In legacy mode, on activation I want to make a curve active if there is only one visible and nothing if there are more than one curve. Therefore my test with two curves in which on activation there was no active curve.

t20100:
I agree with those behaviors, and it should be now what's in the modifications I'm proposing.
BTW, what's in my PR should be fully backward compatible with the previous behavior of silx regarding `is|setActiveCurveHandling`.

t20100:
To me `is|setActiveCurveHandling` is superseded by `set|getActiveCurveSelectionMode` so not needed anymore.. I removed it from the doc so we can deprecate it when we want.

t20100:
LGTM
Best if someone else is doing the review and merge.

PiRK:
LGTM

Issue #2022, opened 2018-07-16
[plot3d] picking of position in items of SceneWidget


t20100:
Add functionality to get value/position of an item (image, scatter, isosurface...) at a pixel position in the scene.

t20100:
closed by #2129

PR #2129, opened 2018-09-28
Picking in 3D scene

t20100:
This PR provides an API to perform picking in a 3D scene from the items.
One can call `item.pickItems(xInPixel, yInPixel, condition)` on a group item to get an iterator of `PickingResult` objects.

A `PickingResult` object provides access to the picked item, the positions picked (in either scene or object frame of reference), indices in the item's data array, and picked data of corresponding data point if applicable.

related to #2022


t20100:
- Added a `PositionInfoWidget` displaying the closest picked item data point value and position upon double-click (used in plot3dSceneWindow.py example)
- Updated documentation

t20100:
Added tests + a few fixes

t20100:
Ready for review.
There is still a few things to do, but this PR is already large and the main feature is there.
I'd prefer to continue to work on it in other PRs.

The `plot3dSceneWindow.py` sample code can be used to test the picking.

In the plot3d API, it adds:
- `silx.gui.plot3d.SceneWidget.SceneWidget.pickItems(x, y, condition)` which yields `silx.gui.plot3.items._pick.PickingResult` objects.
- `silx.gui.plot3d.tools.PositionInfoWidget.PositionInfoWidget` which displays picked position and data value upon double click in the associated `SceneWidget`.

Remaining things:
- Provide alternative ways to handle the picking: a dedicated mode with a button, perform picking in whole scene/only in selected item(s)
- Add picking of 3D meshes
- Improve picking of scatter plots in wireframe/surface mode
- Benchmark and optimizations

PR #2151, opened 2018-10-05
Improve file synchronization

### LINKED ISSUE #298: [Hdf5Widget] Improve synchronize
Synchronization
- should not change expanded nodes
- should not change selected items

### LINKED ISSUE #2083: [silx view] Add a toolbutton to reload the hdf5 file

### LINKED ISSUE #2087: [silx view] synchronize (reload) h5 file is broken
There is a typo in the indentation leading to an exception (see patch below) and the content of the tree is not updated anyway.
Doing a remove and opening again the file works.

Patch:
```
--- a/silx/app/view/CustomNxdataWidget.py
+++ b/silx/app/view/CustomNxdataWidget.py
@@ -944,10 +944,10 @@ class CustomNxdataWidget(qt.QTreeView):
             if edited:
                 item.setAxesDatasets(datasets)
 
-        dataset = item.getSignalDataset()
-        newDataset = self.__replaceDatasetRoot(dataset, removedRoot, loadedRoot)
-        if dataset is not newDataset:
-            item.setSignalDataset(newDataset)
+            dataset = item.getSignalDataset()
+            newDataset = self.__replaceDatasetRoot(dataset, removedRoot, loadedRoot)
+            if dataset is not newDataset:
+                item.setSignalDataset(newDataset)

```

vallsv:
- Close and reload files (hack allowing to synch file)
   - As result, virtual HDF5 or displayed data are not updated but removed
   - This issue needs a bigger understanding and rework to be fixed
- Returns expanded tree and selection
- Icon to sync a file (to work, the root of the tree have to be selected)

Closes #298 (not part of the treeview, but part of silx view)
Closes #2083 
Closes #2087

### DIFF
--- a/silx/app/view/Viewer.py
+++ b/silx/app/view/Viewer.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "05/10/2018"
+__date__ = "08/10/2018"
 
 
 import os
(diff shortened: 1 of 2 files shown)


t20100:
The behavior of the icon is misleading, it would be better to make it synchronize the file of selected items rather than not doing anything when the selected item is not a file.

Otherwise, it looks to work fine and is good to have until we have "in-place" synchronization.

vallsv:
You are right, i was lazy. I will fix the icon.

vallsv:
It's done.

If a subitem is selected the full file is updated.

PR #2153, opened 2018-10-08
HDF5 tree nexus class format issue

t20100:
This PR fixes an issue I got with python 3.4, h5py 2.7.1, PyQt5, where `__nx_class` was already a `str`.

### DIFF
--- a/silx/gui/hdf5/Hdf5Item.py
+++ b/silx/gui/hdf5/Hdf5Item.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2017 European Synchrotron Radiation Facility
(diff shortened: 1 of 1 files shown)


vallsv:
Did you check if this code is also used somewhere else? I am pretty sure it is a copy paste.

vallsv:
I think it is better to use `six` as it will be easier to remove Python2 code.

t20100:
I didn't checked if it is a copy-paste, I just fixed the issue I encountered in `silx view` which prevented using it.

vallsv:
And the `try-except-everything` can be replaced by `obj.attrs.get("NX_class", None)`, which is safer

kif:
I confirm I copy-pasted some code from PierreK

PR #2156, opened 2018-10-08
Add a silx.gui.qt.inspect module

t20100:
This PR adds a `silx.gui.qt.inspect` module which provides 3 functions to check the state of a wrapped C++ QObject: `isValid`, `createdByPython` and `ownedByPython`.

It makes API consistent across PyQt* and PySide* by taking PySide2 names.
For PySide(1), it does its best to load shiboken... but it might not be available (I could not find it on Debian8).


vallsv:
LGTM

PR #2158, opened 2018-10-09
Update dependencies warning

### LINKED ISSUE #2023: Lazy-display of missing dependencies warning
Idea: avoiding warnings of missing dependency (using INFO instead) until one is actually using the feature in which case an exception should be raised.

t20100:
This PR removes duplicated warnings about `pyopencl` not being available.

The only remaining warning about a missing dependency in silx is when importing `silx.opencl` when `pyopencl` is missing... which makes sense to me (To me it should even raise an `ImportError`).

This PR also reworks the way `CompareWidgets` and `silx.image.medianfilter` are probing the availability of OpenCL as `ImportError` is not raised.

To me it closes #2023

### DIFF
--- a/silx/gui/plot/CompareImages.py
+++ b/silx/gui/plot/CompareImages.py
@@ -46,11 +46,10 @@ from silx.third_party import enum
 
 _logger = logging.getLogger(__name__)
 
-try:
-    from silx.image import sift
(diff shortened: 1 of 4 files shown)


vallsv:
LGTM

PR #2164, opened 2018-10-10
Hide displayed tools when the plot is not anymore visible

### LINKED ISSUE #2109: [NXdata] and histogram
There are two bug, the first one is easy, the second more nasty:
* [ ] The histogram button exists in the default image viewer but not in the NXdata image viewer. It would be nice it could be consistent.
* [ ] The histogram only applies when it is called for the first time. It gets updated when changing the image to another image, but not when changing to an NXdata image display.

vallsv:
The PlotAction manage to displayed tool, and hide it if expected.

Applyed on histo, medfilt, and fit actions

Closes #2109

### DIFF
--- a/silx/gui/data/NXdataWidgets.py
+++ b/silx/gui/data/NXdataWidgets.py
@@ -26,7 +26,7 @@
 """
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
-__date__ = "24/04/2018"
+__date__ = "10/10/2018"
 
 import numpy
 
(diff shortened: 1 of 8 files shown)


t20100:
`sigVisibilityChanged` should be added to `PlotWidget` rather than `PlotWindow` as plot actions should work with `PlotWidget`.

Otherwise it's OK for me.

vallsv:
Yes you are right.

t20100:
To complete it, `showEvent` and `hideEvent` still need to be move in `PlotWidget`

vallsv:
Oups, you are right again.

vallsv:
It's done now

PR #2173, opened 2018-10-11
Add alpha channel to ScatterView

maurov:
It adds the possibility to give a transparency channel in addition to the value.

A minimal example to see its behaviour:

```
import numpy as np
from silx import config
from silx import sx
from silx.gui.plot.ScatterView import ScatterView

samps = 1000
levels = 100
fwhm = 50
cen = 500

x = np.random.rand(samps)*levels
y = np.random.rand(samps)*levels
v = np.random.rand(samps)*levels
x0, y0 = x[cen], y[cen]
a = np.exp(-4*np.log(2) * ((x-x0)**2 + (y-y0)**2) / fwhm**2)

config.DEFAULT_COLORMAP_NAME = 'viridis'

#without transparency
s1 = ScatterView()                                                
s1.setData(x, y, v)
s1.show()
#with transparency
s2 = ScatterView()
s2.setData(x, y, v, alpha=a)
s2.show()

```

### DIFF
--- a/silx/gui/plot/ScatterView.py
+++ b/silx/gui/plot/ScatterView.py
@@ -273,7 +273,7 @@ class ScatterView(qt.QMainWindow):
         :return: Colormap currently in use
         :rtype: ~silx.gui.colors.Colormap
         """
-        self.getScatterItem().getColormap()
+        return self.getScatterItem().getColormap()
 
     # Control displayed scatter plot
 
--- a/silx/gui/plot/ScatterView.py
+++ b/silx/gui/plot/ScatterView.py
@@ -277,7 +277,7 @@ class ScatterView(qt.QMainWindow):
 
     # Control displayed scatter plot
 
-    def setData(self, x, y, value, xerror=None, yerror=None, copy=True):
+    def setData(self, x, y, value, xerror=None, yerror=None, alpha=None, copy=True):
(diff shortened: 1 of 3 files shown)


vasole:
I would support these cases:

- Alpha between 0 and 1 -> Float
- Alpha as integer in the range [0, 255]
- Alpha given as color (I think that was already in)

Please do nothing and wait for @t20100 opinion.

Should we allow something similar for images at same Z order? After all we aim at having most of the functionalities common between scatter plots and images.

PiRK:
Note that,, AFAIK, this will add transparency to possibly existing transparency set with `item.setAlpha` (https://github.com/silx-kit/silx/blob/master/silx/gui/plot/items/core.py#L708), which currently only supports a float in range 0--1. We need to keep the APIs consistent, so we should also change `setAlpha` if we decide to support integers in range 0--255.

t20100:
I think it is fine with alpha in [0, 1].
I'll add a few checks in the setter and change the blending of alpha with `rgbacolors` (using multiply).

t20100:
I made a pull request on your branch (https://github.com/maurov/silx/pull/2) which:
- fixes the test
- changes the way to apply the transparency (by multiplying it)
- adds some checks/conversion of `alpha` in `setData`

PR #2177, opened 2018-10-11
[FitConfig]fix bug when clicking Default button

### LINKED ISSUE #2174: [fit gui] Pressing default gives an error
Asking to load the default parameters after selecting a FWHM gives an error

image

image

PiRK:
Hopefully closes #2174 (still to be tested)

### DIFF
--- a/silx/gui/fit/FitConfig.py
+++ b/silx/gui/fit/FitConfig.py
@@ -138,7 +138,7 @@ class TabsDialogData(TabsDialog):
 
         self.default = {} if default is None else default
 
-        self.buttonDefault.clicked.connect(self.setDefault)
+        self.buttonDefault.clicked.connect(self._resetDefault)
         # self.keyPressEvent(qt.Qt.Key_Enter).
 
     def keyPressEvent(self, event):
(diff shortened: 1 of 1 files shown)


PiRK:
It fixes the crash, but does not seem to do what it is supposed to do. To be investigated, and checked against the PyMca behavior.

PiRK:
> To be investigated, and checked against the PyMca behavior.

PyMca does not have (or does not show?) a default button.

PiRK:
After investigation, what the default button does is to restore the values as they were when the dialog was open. It is like a Cancel but without closing the window. 

This is not exactly what I expected. I expected that it would restore the hardcoded default values, the ones that appear the first time the widget is opened. But the dialog is recreated each time it is called, and the current configuration is then set as default. 

I'm not sure what would be the best fix at this stage :
 - remove the button?
 - rename it "Undo changes"?
 - fix it to actually do what I initially intended to do (restore with this dict https://github.com/silx-kit/silx/blob/master/silx/math/fit/fitmanager.py#L89)? In that case, the name could be changed to "Restore default", as it seems the current name "Default" is confusing and can be interpreted as "save current values as default".

vasole:
At this point the safest is to change to Undo Changes

payno:
LGTM

PR #2178, opened 2018-10-11
Use silx.config.DEFAULT_COLORLIST in PlotWidget

### LINKED ISSUE #2167: [plot] expose PlotWidget.colorList to config
If one wants to change the default list of colors in `PlotWidget` the way to proceed is:

```
PlotWidget.colorList = MYCOLORS
```

it would be useful to expose it to `silx.config.DEFAULT_COLORLIST`

maurov:
Fix #2167

### DIFF
--- a/silx/_config.py
+++ b/silx/_config.py
@@ -81,3 +81,32 @@ class Config(object):
 
     .. versionadded:: 0.8
     """
+
+    DEFAULT_COLORLIST = ['#000000', #black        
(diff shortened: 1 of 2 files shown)


t20100:
Good feature to configure.

I made a small PR on your branch to rename the config variable (https://github.com/maurov/silx/pull/1).
The number of config variables is expected to grow so, the more specific the names are the better.

PR #2180, opened 2018-10-11
Opencl fft

kif:
Re-work Pierre's code

### DIFF
--- a/silx/opencl/codec/byte_offset.py
+++ b/silx/opencl/codec/byte_offset.py
@@ -37,7 +37,7 @@ __authors__ = ["Jérôme Kieffer"]
 __contact__ = "jerome.kieffer@esrf.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "24/10/2017"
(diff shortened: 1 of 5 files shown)


kif:
Let's see if it works everywhere

pierrepaleo:
Looks good. In the tests :

- [Line 90](https://github.com/silx-kit/silx/pull/2180/files#diff-37c19655baa63e2fc756eee085a5cac3R90): We can use `numpy.fft.rfft2(self.data)` instead of `numpy.fft.fft2(self.data)[:, :self.shape[-1] // 2 + 1]`
- The test precision is low (`1e-3 * max`), but IIRC the "error" might be large due to the computations done in float32 (whereas the first thing done by numpy FFT is to cast input data to float64). I don't remember having tested with `double_precision=True`.

In the FFT Class, the code logic seems to imply that complex input data might not be "detected" properly as complex data (`self.real_fft = not(force_complex)`). Perhaps the input data type (real/complex) should be checked.

kif:
@pierrepaleo : thanks for the tip

pierrepaleo:
This work was extended on #2278.

PR #2181, opened 2018-10-11
[plot] Add more options for active curve style

### LINKED ISSUE #2119: [plot] Need for different ways to hightlight active curve.
Discussed long ago.

Need of alternative methods to highlight the active curve: for instance active color could be set to None and line width could be expressed as an integer factor to be applied to the line width to achieve highlight without change of color. However, line symbol has to be modified too (if the curve has no line it can be difficult to know it is the active curve).

t20100:
This PR adds the option to set not only the colour, but also line style, line width, symbol and symbol size.
This adds `PlotWidget.get|setActiveCurveStyle` methods.

TODO:

- [x]  Add tests

closes #2119

### DIFF
--- a/doc/source/modules/gui/plot/items.rst
+++ b/doc/source/modules/gui/plot/items.rst
@@ -30,10 +30,10 @@ Curve
              getXLabel, getYLabel,
              getLineWidth, setLineWidth, getLineStyle, setLineStyle,
              isHighlighted, setHighlighted, getHighlightedStyle, setHighlightedStyle,
-             getCurrentColor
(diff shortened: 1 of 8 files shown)


t20100:
Question: `set|getActiveCurveColor` are superseded by `set|getActiveCurveStyle`.
Is it OK to deprecate them?

t20100:
- Added a few tests.
- Deprecated `get|setActiveCurveColor`
- Fix default argument value and equality of `CurveStyle` object.

Ready for review.

PR #2183, opened 2018-10-12
Improve datapanel header

### LINKED ISSUE #2131: [silx view] The header of the data display is sometime very big
When a file is loaded using absolute path, the full path is displayed.

This widget is not elided, as result the application take a lot of space and can't be reduced.

Beedback from Peter ID16A

vallsv:
Closes #2131


payno:
Very useful. LGTM

PR #2205, opened 2018-10-22
Rework the import error message for testing lib

vallsv:
- The real exception backtrace is helpful


t20100:
Why do you changed to use `PROJECT_NAME` global variable? Using `project_name` argument (which is set to `PROJECT_NAME`) seems better.

vallsv:
I did not notice that. It's a copy past from pyfai or fabio.

vallsv:
Here it is fixed. Thanks for the feed back.

PR #2208, opened 2018-10-22
[test] Only test remaining widgets created by Python

t20100:
This PR proposes to only check for remaining widgets that were created by Python after each test.


payno:
LGTM

PR #2209, opened 2018-10-23
Add option to toggle the profile option button

mretegan:
It defaults to no button (the previous behavior). I've put this on top of master, but it would be nice to have it in 0.9.0 if is not too late.

### DIFF
--- a/silx/gui/plot/PlotToolButtons.py
+++ b/silx/gui/plot/PlotToolButtons.py
@@ -36,7 +36,7 @@ The following QToolButton are available:
 
 __authors__ = ["V. Valls", "H. Payno"]
 __license__ = "MIT"
-__date__ = "27/06/2017"
+__date__ = "23/10/2018"
 
 
 import functools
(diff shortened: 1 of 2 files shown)


vasole:
I think a cleaner solution would be to have the W (for the width) to have the possibility to toggle between W and W with a bar on top to indicate the mean in the same manner it is done for the toggle the keep aspect ratio.

That is what I always wanted for PyMca. It does not take any additional space ((the W is already there) and does not hinder any functionality.

mretegan:
What about the spin box? Would it be placed to the right of the button?

vasole:
No, the look and feel would be identical to version 0.8.0 The W to indicate the width would toggle between W and "W bar"

t20100:
Sorry, it's too late for 0.9.0, the release is already prepared..

payno:
This is True that the GUI is not very userfriendly here.
First to hide the action it is to go for something like:

```python
plot.profile.actions()[-1].setVisible(False)
```
But an API to hide this action is also missing. You can also redefine your own Toolbar.

We already freeze the branch for the release so I guess this will stay like this for the next release (0.9.0) sorry about it.
But we will have time to discuss about the modification to bring to the interface for 0.10.

t20100:
A `ProfileToolBar.getProfileOptionToolButton` method would be more Qt-style and more flexible than an extra `__init__` argument.
You can then do: `myProfileToolBar.getProfileOptionToolButton().setVisible(False|True)`

vasole:
I repeat that instead of appearing and disappearing buttons is better to use the same approach as with the aspect ratio. Perhaps you do not understand what I mean. We just need a W and a W bar in the same manner as the average icon:

https://github.com/silx-kit/silx/blob/master/silx/resources/gui/icons/math-average.png

mretegan:
@t20100 

> A `ProfileToolBar.getProfileOptionToolButton` method would be more Qt-style and more flexible than an extra `__init__` argument.
> You can then do: `myProfileToolBar.getProfileOptionToolButton().setVisible(False|True)`

In my code I've followed the approach used in `PlotWindow`, and probably that wan't the best idea as that code covers some old use case. Something like the implementation of `SceneWindow` would have been a better example.

t20100:
Yes we now try to have a more or less Qt-like API and avoid to have a lot of argument in `__init__`.
We should have some written guidelines so it is simpler to step in.

payno:
the PR #2303 will give access to the profile action in order to hide it:

```python
myProfileToolBar.getProfileOptionToolAction().setVisible(False|True)
```

But a question remain: should we hide this option by default ?

It would also be nice to find better icons for mean and sum ?

I was thinking something like:


mean_img

sum_img

note: #2303 will close this PR

t20100:
I'd leave the default as it is, important thing is that someone can change it easily if needed..

t20100:
Closing as superseded by PR #2303

PR #2211, opened 2018-10-23
[plot] Remove some old deprecated methods/arguments

### LINKED ISSUE #1677: [plot] Use deprecation logger in PlotWidget
As it is now, it is using PlotWidget logger

t20100:
This PR removes some (not all) deprecated (at the very least since a year) methods/signals/arguments in `PlotWidget` and `PlotWindow`.

As a side effect, it closes #1677


payno:
LGTM

PR #2212, opened 2018-10-24
Use Cython language_level=3

### LINKED ISSUE #2191: Take care of cython 'language_level'
With Cython 0.29, this warning is showing up:

`Cython/Compiler/Main.py:367: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release!****`

At least set language_level to 2 (simple but short term solution), and best set it to 3 and make sure it works.

t20100:
Makes it work with cython language_level=3, issues with absolute/relative imports of .pxd files.

closes #2191

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -588,7 +588,8 @@ class BuildExt(build_ext):
             from Cython.Build import cythonize
             patched_exts = cythonize(
                 [ext],
-                compiler_directives={'embedsignature': True},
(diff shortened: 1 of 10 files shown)


t20100:
Care is needed about byte/str, str/unicode as now Cython code uses python3 syntax even when run in python2 (i.e., str means unicode when run in python2)

PR #2223, opened 2018-11-05
Refactor colormap

### LINKED ISSUE #1859: [gui] Remove strong dependency on matplotlib in silx.gui.colors
It would be better if strong dependency on matplotlib remains in plot and plot3d packages.
As it is now silx.gui.colors relies on matplotlib for the colormap.
Not sure how to deal with that... have default colormaps moved there + lazy loading of others

### LINKED ISSUE #2217: Colormap: getColormapLUT is not working for our own colormaps
- `colormap.getNColors(None)` is working but could have the wrong number of colors
- cause `getColormapLUT` returns `None`
- then 256 colors are requested

Then there is no way to access to the stored raw LUT.

It's a small issue, mostly with no consequences.

t20100:
This PR reworks `silx.gui.colors.Colormap` implementation and deprecates `silx.gui.plot.matplotlib.Colormap` (which is no longer used within silx).
It:
- Removes the strong dependency to matplotlib in `silx.gui.colors`: closes #1859
- Improves `Colormap.getNColors`  to return the same number of colors as the LUT (and not a hard-coded 256): closes #2217

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -154,23 +154,9 @@ def cursorColorForColormap(colormapName):
     return _COLORMAP_CURSOR_COLORS.get(colormapName, 'black')
 
 
-DEFAULT_COLORMAPS = (
-    'gray', 'reversed gray', 'temperature', 'red', 'green', 'blue')
-"""Tuple of supported colormap names."""
-
-DEFAULT_MIN_LIN = 0
-"""Default min value if in linear normalization"""
-DEFAULT_MAX_LIN = 1
-"""Default max value if in linear normalization"""
-DEFAULT_MIN_LOG = 1
-"""Default min value if in log normalization"""
-DEFAULT_MAX_LOG = 10
-"""Default max value if in log normalization"""
-
-
 # Colormap loader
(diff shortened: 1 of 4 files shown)


vallsv:
```
        colormapObject = Colormap(name='red',
                                  colors=numpy.array([[1., 0., 0.],
                                                      [0., 1., 0.],
                                                      [0., 0., 1.]]),
```

We could talk how you we have to use the object, cause if `colors` are provided, the name at least could not be used (here there is an inconsistency). It should not be a problem for a custom lut to have no name. You provide colors, it's a custom LUT, else you provide a LUT name but it have to exists.

vallsv:
Ok, no pbm, it's fine. Colors are ignored (backward compatibility)

t20100:
BTW, it's now  more restrictive in that you cannot create a colormap that has no color LUT (which was possible before).

t20100:
`name` and `colors` arguments both set the color LUT... There should be only one argument instead of 2, but that would change the API...

vallsv:
LGTM

PR #2228, opened 2018-11-06
Add the possibility to download an external resource in a subdir

t20100:
This PR adds the possibility to load external resources in a subdirectory


vallsv:
LGTM

vallsv:
Maybe using `os.path.exists` is semantically better, but it's fine.

vallsv:
No, you need nothing to be able to create a directory. You can't create a directory if it's a file. Then you will not receive the error you expect. But it's a detail.

PR #2229, opened 2018-11-06
Robustness on colormap

vallsv:
- Avoid to to use both `name` and `colors` in the constructor (`colors` is ignored + deprecation warning)
- Now `Colormap(colors=ndarray)` do not create anymore a default gray LUT
- `imshow` supports `Colormap` inside the argument `cmap`
- Improve the `Colormap` object API (`copy`, `__eq__`, `setFromColormap`)
- Get ride of dict in `ColormapDialog` to avoid exception when using colormap with custom LUT
- Some small clean up

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -31,9 +31,10 @@ __authors__ = ["T. Vincent", "H.Payno"]
 __license__ = "MIT"
 __date__ = "06/11/2018"
 
-from silx.gui import qt
 import numpy
 import logging
+
+from silx.gui import qt
 from silx.math.combo import min_max
 from silx.math.colormap import cmap as _cmap
 from silx.utils.exceptions import NotEditableError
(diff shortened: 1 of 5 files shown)


vallsv:
I don't think it's a good idea, cause it will only be a constraint on our side without feature.

But if we add an API to add custom colormaps to our predefined set it make sense. This could be an interesting feature.

t20100:
@vasole, then you can override an already existing colormap, it would indeed be better to allow to register extra colormaps.

@vallsv, I don't think an extra `deepcopy` is needed, colormap LUT is small enough so we can copy it.

vallsv:
I fix feedback about deepcopy, yes it's pointless.

For the name i would prefer not to do it. Also there is often confusion between LUT and Colormap. The colormap name is not the name of the colormap, but the name of the LUT. That's a big difference. I dont think it is the right place to register LUT names.

PR #2239, opened 2018-11-09
Improve silx view loading

### LINKED ISSUE #2225: [silx view] Openning a file should also display it
Feedback from Jerome

### LINKED ISSUE #2238: [silx view] Fabio file not released
While loading an EDF file.

I don't know yet what's happen.
```
Exception ignored in: <bound method File.__del__ of <_io.FileIO name='../data/edf/payno/ref0001_0050.edf' mode='rb' closefd=True>>
  File "/home/valls/.local/lib/python3.5/site-packages/fabio/fabioutils.py", line 470, in __del__
  File "/home/valls/.local/lib/python3.5/site-packages/fabio/fabioutils.py", line 474, in close
AttributeError: 'NoneType' object has no attribute 'close'
```

vallsv:
- `silx view` now display the file opened on the data view (previously it was only displayed on the side tree)
- `fabioh5` provides an `NXdata` containing the image (this NXdata is defaulted by the root and the entry)
- As result loading an EDF file displays an image (sounds original)
- Fix bug, as some fabio files was not closed

Closes #2238 
Closes #2225

### DIFF
--- a/silx/app/view/Viewer.py
+++ b/silx/app/view/Viewer.py
@@ -829,7 +829,7 @@ class Viewer(qt.QMainWindow):
                 menu.addAction(action)
 
             if silx.io.is_file(h5):
-                action = qt.QAction("Remove %s" % obj.local_filename, event.source())
+                action = qt.QAction("Close %s" % obj.local_filename, event.source())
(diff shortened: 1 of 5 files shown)


vallsv:
Here is a patch

PR #2246, opened 2018-11-19
Support background color

### LINKED ISSUE #2220: [plot] API to custom background color
Could it be possible to provide an API to custom 2 different background colors?

```
import matplotlib.pyplot as plt
fig = plt.figure()
fig.patch.set_facecolor('cyan')
ax = fig.add_subplot(1, 1, 1)
ax.set_facecolor('blue')
```
screenshot from 2018-10-30 13-32-18

vallsv:
Add `setBackgroundColor` and `setDataBackgroundColor` to the plot API

- `setBackgroundColor` allow to custom the full background of the widget (including the background bellow the plot data)
- `setDataBackgroundColor` allow to custom the background bellow the plot data only

Closes #2220 

The OpenGL backend is not part of the PR

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -29,7 +29,7 @@ from __future__ import absolute_import
 
 __authors__ = ["T. Vincent", "H.Payno"]
 __license__ = "MIT"
-__date__ = "09/11/2018"
+__date__ = "19/11/2018"
 
 import numpy
 import logging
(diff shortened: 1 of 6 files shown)


t20100:
I made a PR on your branch with the OpenGL implementation and a patch in `PlotWidget` to set the dirty flags so that the plot is updated when changing the background colors.

Otherwise LGTM.

vallsv:
Thanks a lot, i did not expect it for now :-)

t20100:
That was really simple, no need to add it to pending stuff...

PR #2247, opened 2018-11-19
Sync center of axis

### LINKED ISSUE #2243: [plot] Provide different synchronization modes
Synchronization based on axes min and max gives troubles when widgets do not have exactly the same size and aspect ratio is selected.

vallsv:
This PR add another way to sync axes.

- `syncCenter` allow to synchronize axes to the value displayed in there center
- `syncZoom` allow to sync the pixel size of axes, reaching displayed pixels used to display the axes

As result `limits` of all axes can be different according to the size of the widget. Fixed limits was creating a lot of inconsistencies in case of use of fixed aspect ratio, or in case a lot of widgets are synchronized (widget size could not always match). Both constrains are still useful, depending on the context.

As it is mostly designed to sync Plot2D (displaying images), only a subset was implemented.

- Can only be used with linear axes
- Can be used to constrains `center` or `center+zoon`, but not `zoom` only

Closes #2243

### DIFF
--- a/examples/syncPlotLocation.py
+++ b/examples/syncPlotLocation.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2016-2017 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


vallsv:
3 functions could be moved into `Axes`
- `getCenter`
- `getLimitsFromCenter`
- `getRangeInPixel` or `getPixelSize`

vallsv:
The test was fine, only the test profile fails

PR #2264, opened 2018-11-30
[doc] add automatic screenshots

payno:
## Policy

- command `python setup build build_screenshot` will generate doc and screenshots
- sphinx directive `.. snapshotqt::` (inherited from sphinx 'Figure')

   ```sphinx
   * - .. snapshotqt:: widgets/img/FrameBrowser.png
         :width: 110px
         :align: center

         from silx.gui.widgets.FrameBrowser import FrameBrowser
         widget = FrameBrowser()
         widget.setRange(0, 10)
         widget.show()
   ```

- directive also has the :script: option, this case no content (source code) should be given but a script to generate it.
In this case the path to the script is given from the root directory.

```sphinx
   * - .. snapshotqt:: plot3d/img/SceneWidget.png
          :height: 150px
          :align: center
          :script: examples/plot3dSceneWindow.py
```

- screenshot executable are stored in silx/doc/screenshot/...
- image generated by script at silx/doc/screenshot/x/y/screenshot.py will be stored in script at silx/doc/screesnhot/x/y/img/screenshot.png
- if we run `python setup build build_doc` it will also link with the image silx/doc/screesnhot/x/y/img/screenshot.png etc

## TODO:

- [x] update snapshot directive
- [x] create new setup.py command
- [x] convert gallery.rst ..image to ..
- [x] python 2 / python3 compatiblity
- [x] avoid qt dependancy if build_doc is called
- [ ] ~~avoid multiple screenshot generation~~
- [ ] ~~add an option to generate only one specific screenshot ?~~ -> overkilled
- [x] add a CI scenario for generating screenshots


payno:
The more I think about it more I think we would better have a command / option to generate screen-shot and overwrite existing one instead of having screen-shots generated each time.
This way we could generate screen-shots during the release and avoid adding any other dependency.

payno:
Missing materials:

- data used for PlotWidget and PlotWindow screenshots

t20100:
I made a small PR with minor change.

LGTM, nice way to do it.

PR #2266, opened 2018-11-30
Improve drawmask selection2

### LINKED ISSUE #1270: [plot] Mask: Use different color for the current selection and the mask
As it is, when drawing an opaque mask, it is not possible to see the currently drawn polygon (and in particular the starting point) as the same color is used to display both the mask and the drawn polygon.

vallsv:
Here is a patch to allow to display dotted lines in 2 colors.

- The Matplotlib backend is based on a custom `_DoubleColoredLinePatch` customing the rendering method. And other expected methods.
- Tested only on matplotlib 3.0.2, and Debian8 one
- There is no OpenGL backend
- `line`/`vline`/`hline` do not uses patches, then there is no special rendering for this.
    - It could be fixed

screenshot from 2018-11-30 18-44-22

Closes #1270

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -29,7 +29,7 @@ from __future__ import absolute_import
 
 __authors__ = ["T. Vincent", "H.Payno"]
 __license__ = "MIT"
-__date__ = "19/11/2018"
+__date__ = "30/11/2018"
 
 import numpy
 import logging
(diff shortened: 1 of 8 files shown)


vallsv:
I rework the thing.
The rendering is not the same (was better before) but it's more convenient to use default mpl styles.

vallsv:
I will fully rework this PR using the new master

PR #2267, opened 2018-12-03
include requirements.txt in requirements-dev.txt

payno:
I also saw that `scipy` was on the requirements.txt but from my point of view it should be in requirements-dev.txt no ?


vallsv:
LGTM

PR #2270, opened 2018-12-03
[doc]Fix sphinx errors and warnings 

PiRK:
Sphinx was unable to create link to `plot3d.items` without the full path (confusion with `plot.items`)

@t20100 : I'm not sure about https://github.com/silx-kit/silx/commit/9c82c19e929930e645161462804a79207e0c7b89 . Maybe it would be better to import/expose `CurveStyle` also in `silx/gui/plot/items/__init__.py` instead of providing the correct path in the incorrect doc page.


t20100:
@PiRK yes it would be better to provide it in the `__init__.py`

PR #2275, opened 2018-12-11
Unpin pyopengl and pyopencl versions

t20100:
This PR removes version pin-pointing for pyopengl and pyopengl as new versions are available.

### DIFF
--- a/ci/info_platform.py
+++ b/ci/info_platform.py
@@ -8,10 +8,14 @@ __license__ = "MIT"
 
 
 import sys
+import platform
 
 print("Python %s bits" % (tuple.__itemsize__ * 8))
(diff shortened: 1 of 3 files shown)


vallsv:
Nice try ;-)

t20100:
Should be fine now.

I fixed an issue in `requirements.txt`: pyopencl was not installed on Windows (and so related tests were not running) because the machine is called `AMD64` while it is called `x86_64` under Linux... Why making simple things simple?

t20100:
Actually tests requiring OpenCL are not run on appveyor.

t20100:
A wheel is missing for windows, python2.7, otherwise it should be fine.

t20100:
Should be fine now, `pyopencl` is still pin-pointed on Windows but anyway it is not tested there...

vallsv:
LGTM

PR #2278, opened 2018-12-13
Multi-backend FFT

pierrepaleo:
## About

This PR adds a FFT with multiple backends: numpy, FFTW, OpenCL and CUDA. The idea is to have the same API for all the underlying backends. All the boring work (plans computation, axes, data type, memory alignment) is taken care of.

The numpy backend serves as a reference to compare the 3 other against. 
Therefore, the numpy backend is tested separately. 

All comments/suggestions on the API are welcome.

## Features 

  - 1D -> ND transforms (all what the underlying backend can support)
  - Batched transforms
  - R2C, C2C modes, with float32 and float64 data types
  - For GPU backends, input and/or output can be a device array.

## Currently not supported

  - In-place transforms
  - Hermitian transforms
  - For CUFFT backend, batched transform where batch is other than dimension 0.
  - Nice-to-have: transforms with "easy" zero-padding

### Simple FFT with numpy
```python
import numpy as np
from scipy.misc import ascent
from silx.math.fft import FFT
img = ascent().astype(np.float32)

F = FFT(data=img, backend="numpy") # automatically chooses R2C transform
img_f = F.fft(img)
```

### Using FFTW
```python
F = FFT(data=img, backend="fftw", num_threads=4) 
img_f = F.fft(img)
# do some operation of img_f ...
F.ifft(img_f)
```

### Using OpenCL
```python
import pyopencl.array as parray
F = FFT(data=img, backend="opencl")
# All the Host <-> Device copies are handled under the hood
img_f = F.fft(img) # by default, result is a numpy array
# Input and/or output can be device array as well
d_in = parray.to_device(F.queue, img)
d_out = parray.zeros(F.queue, F.shape_out, dtype=F.dtype_out)
F.fft(d_in, output=d_out)
```

### Using CUDA
```python
import pycuda.autoinit
import pycuda.gpuarray as gpuarray 
F = FFT(data=img, backend="cuda")
d_in = gpuarray.to_gpu(img)
d_out = gpuarray.zeros(F.shape_out, F.dtype_out)
F.fft(d_in, output=d_out) # CUFFT is twice faster than clfft for R2C transforms

### DIFF
--- a/silx/math/fft/basefft.py
+++ b/silx/math/fft/basefft.py
@@ -52,11 +52,11 @@ def check_version(package, required_version):
 
 
 class BaseFFT(object):
+    """
+    Base class for all FFT backends.
+    """
     def __init__(self, **kwargs):
-        """
-        Base class for all FFT backends.
-        """
-        self.get_args(**kwargs)
+        self.__get_args(**kwargs)
 
         if self.shape is None and self.dtype is None and self.data is None:
             raise ValueError("Please provide either (shape and dtype) or data")
@@ -65,12 +65,12 @@ class BaseFFT(object):
             self.dtype = self.data.dtype
         self.user_data = self.data
         self.data_allocated = False
-        self.calc_axes()
-        self.set_dtypes()
-        self.calc_shape()
+        self.__calc_axes()
+        self.__set_dtypes()
+        self.__calc_shape()
 
 
-    def get_args(self, **kwargs):
+    def __get_args(self, **kwargs):
         expected_args = {
             "shape": None,
             "dtype": None,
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: silx/math/fft/basefft.py [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ###########################################################################*/
+import numpy as np
+
+
+
+def check_version(package, required_version):
```
t20100: You can use `pkg_resources.parse_version`  instead (https://setuptools.readthedocs.io/en/latest/pkg_resources.html#parsing-utilities)

t20100:
I made a PR with typos in docstrings, otherwise LGTM

PR #2281, opened 2018-12-17
Clean-up 3rd party

### LINKED ISSUE #2274: py3-violating exception in silx/third_party/_local/concurrent_futures/_base.py, line 414
I am trying to build a conda package for silx 0.9.0, and doing that runs into a problem of:

```
compiling .pyc files...
  File "lib/python3.6/site-packages/silx/third_party/_local/concurrent_futures/_base.py", line 414
    raise exception_type, self._exception, self._traceback
                        ^
SyntaxError: invalid syntax
```

To be clear, `silx` 0.9.0 imports and appears to run fine for me.

I'm left wonderign how others have created a conda package,  but it seeems like this should be fixed by replacing this with `raise exception_type(self._exception, self._traceback)`, I  believe.

t20100:
This PR  removes `six`, `enum` and `concurrent.futures` backports packages from `third_party` and adds dependencies to those packages (only with python 2 for `enum` and `concurrent.futures`).

closes #2274


vallsv:
Debian packaging files was not updated? Maybe there something to do around?

t20100:
No the packages are already listed there... except that looking at it, `six` is missing from the `Build-Depends` but it is in the dependencies.

PR #2282, opened 2018-12-17
[PrintPreviewDialog]Fix resizing in PyQt5 

### LINKED ISSUE #2279: [PrintPreview]Resizing figure does not work properly
See https://github.com/vasole/pymca/issues/383

The first resizing operation (dragging to bottom-right corner of a figure) works fine, then the results seem to become random. 

This can be replicated using `examples/printPreview.py`

PiRK:
Closes #2279 
This method seems to work better with PyQt5.


PiRK:
While fixing this bug, I noticed that the *keepAspectRatio* configuration parameter was only used when initially sending the SVG item. 
Now this parameter is also set in the resizable *QGraphicsRectItem*, and users can actully resize the graphics freely if they explicitly untick the corresponding tickbox in the configuration dialog.

PR #2289, opened 2018-12-18
[build] Add six to debian Build-Depends

t20100:
related to https://github.com/silx-kit/silx/pull/2281#issuecomment-448132356


vallsv:
LGTM

PR #2290, opened 2018-12-18
Support old setuptools

### LINKED ISSUE #2288: [pip] install_requires ==python2.7 is not supported an old pip
It would be better to append this items in the list in case of Python2.7.

t20100:
Boring but working.
Support old version of `setuptools`. It should fix the nightly build of debian packages

closes #2288


vallsv:
LGTM

PR #2293, opened 2018-12-18
Fix NX_class decoding issue

### LINKED ISSUE #2292: NX_class name decoding error
I get this traceback when trying to open a file with silx view: 

```
'utf-8' codec can't decode byte 0xb2 in position 1: invalid start byte   File "/Users/marius/Library/Python/3.7/lib/python/site-packages/silx/gui/hdf5/Hdf5TreeModel.py", line 515, in data
    return node.dataType(role)
  File "/Users/marius/Library/Python/3.7/lib/python/site-packages/silx/gui/hdf5/Hdf5Item.py", line 401, in dataType
    text = self.nexusClassName
  File "/Users/marius/Library/Python/3.7/lib/python/site-packages/silx/gui/hdf5/Hdf5Item.py", line 375, in nexusClassName
    self.__nx_class = str(self.__nx_class, "UTF-8")
```

The type of the self.__nx_class object is <class 'numpy.ndarray'>.

PiRK:
This PR fixes a @NX_class decoding issue by using a safe formatter rather than trying to guess every possible input type.

Closes #2292

### DIFF
--- a/silx/gui/hdf5/Hdf5Item.py
+++ b/silx/gui/hdf5/Hdf5Item.py
@@ -364,14 +364,12 @@ class Hdf5Item(Hdf5Node):
     def nexusClassName(self):
         """Returns the Nexus class name"""
         if self.__nx_class is None:
-            self.__nx_class = self.obj.attrs.get("NX_class", None)
-            if self.__nx_class is None:
-                self.__nx_class = ""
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM :-)

PR #2295, opened 2018-12-19
[mask] If single mask mode, set only 1 level when loading

### LINKED ISSUE #2284: [mask] State of the mask inconsistent when loaded from numerical values
I try to load a mask from an image containing `[[0,1,2,3,4,5,6,7,8,9,10]]`.

Using the pencil with set or unset sometimes do not change anything.

The rendering looks strange.

Maybe i am wrong.

t20100:
closes #2284


vallsv:
LGTM, thanks a lot

PR #2298, opened 2018-12-19
Synchronize ColorBar with plot background colors

vallsv:
This PR fix the `ColorBar` background according to the plot state.

Relative to comment https://github.com/silx-kit/silx/issues/2220#issuecomment-448607217

It also fixes small issues and typo


payno:
LGTM, 

note: CI failure on python 3.6 does not come this PR but from the pyside2 latest release.

Issue #2300, opened 2018-12-19
[gui] PrintPreview Allow to provide comment and title 


vasole:
https://github.com/vasole/pymca/issues/385

The API is still there but the silx implementation simply ignores it.

The simplest way to recover the functionality I see is to modify PrintPreviewToolButton method _plotToPrintPreview  by checking if the plot has a method to return the graphics item, the title and the comment instead of just doing the work on its own.

Would it be acceptable? I could easily code it and it would not break anything by implementing that.

Any other idea?

PiRK:
I first thought that the simplest method would be the make the  `PrintPreviewToolButton._plotToPrintPreview` public so that applications cans just add additional info (comments) to the `self.printPreviewDialog.addSvgItem(...)` call. But actually it is not that simple, because `_plotToPrintPreview` calls `_getSvgRendererAndViewbox` (which calls `_getViewBox`...). so if someone wanted to overload `plotToPrintPreview` he would need to know to call a few private method, or need to reimplement them. At the end, it is equivalent to reimplementing the entire toolbutton on the application side.

It would be the same issue with your idea of letting the application plot returning the graphics item, it would come to rewriting most of the toolbutton's code on the application side anyway. 

Making `_plotToPrintPreview` check if the plot has a method to provide only additional information (title and comments, but not the graphics item) would be the solution with the least amount of code duplication.

t20100:
I think it is better to add some methods one can override in the `PrintPreviewToolButton` to provide the title and comment (e.g., `getTitle`, `getComment` or `computeTitleAndComment`) rather than expecting a method to be defined in the plot.

PiRK:
Good idea.

PR #2301, opened 2018-12-20
[PrintPreview]API for comment and title

PiRK:
This addresses #2300 by providing two methods that can be implement in a subclass to provide custom comment and title to be printed below and above the plot.

### DIFF
--- a/examples/printPreview.py
+++ b/examples/printPreview.py
@@ -42,22 +42,38 @@ from silx.gui import qt
 from silx.gui.plot import PlotWidget
 from silx.gui.plot import PrintPreviewToolButton
 
+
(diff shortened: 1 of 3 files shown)


PiRK:
TODO:

 - [ ] *commentPosition="CENTER"* seems to be broken.
 - [ ] the text is not scaled together with the graph when resizing interactively

PiRK:
The comment alignment was actually not implemented in the original code (PyMca), and what I did on the silx side to implement it had no effect. The  QGraphicsTextItem has a width identical to the graph (we scale it to ensure this), so centering the item is the same as aligning it to the left.

Tomorrow I will try to figure out a solution to align the text inside the QGraphicsTextItem, maybe with HTML tags.

payno:
LGTM

PR #2308, opened 2018-12-21
Add a function to create external dataset for .vol file

### LINKED ISSUE #1741: Ready .vol files
PyHST is generating .vol files. Those are binary files which have to be associated with a .vol.info and/or a .vol.xml files to get the contain shape.

Those files seems to be used a lot and it would be interesting to have silx or fabio reading them.
But the problem from my point of view is the file association that has to be made to read them correctly.

Here is an example of a file set:

D2_H2_T2_h_multipag_1079.vol   -> contains binary data
D2_H2_T2_h_multipag_1079.vol.info -> contains information to read the data 
D2_H2_T2_h_multipag_1079.vol.xml -> contains the same information as the .info (at least on this case)

payno:
close #1741

### DIFF
--- a/silx/io/test/test_utils.py
+++ b/silx/io/test/test_utils.py
@@ -640,6 +640,105 @@ class TestGetData(unittest.TestCase):
         self.assertRaises(IOError, utils.get_data, url)
 
 
+def _h5_py_version_older_than(version):
+    v_majeur, v_mineur, v_micro = h5py.version.version.split('.')[:3]
+    r_majeur, r_mineur, r_micro = version.split('.')
+    return v_majeur >= r_majeur and v_mineur >= r_mineur
+
+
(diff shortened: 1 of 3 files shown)


vasole:
Just a comment. Virtual Dataset and External Dataset are very different things in HDF5 jargon.

t20100:
I made a few updates to the function docstring.

Also, this seems more generic than .vol files from pyhst, this apply to any raw data.
Maybe renaming `vol_to_h5_external_dataset` to `rawfile_to_h5_external_dataset` would make it clear.

Finally, if the `.vol.info` file is usually the `.vol` file name + `.info` suffix, you could (easily) guess it when both `shape` and `info_file` are None...

payno:
Yes, in fact I will create both functions 'vol_to_h5_external_dataset' and 'rawfile_to_h5_external_dataset'
and the first one will try to make the assumption on .vol.info file

PR #2313, opened 2018-12-21
Fix colorbar background

### LINKED ISSUE #2305: Issue with colorbar background
The colorbar background of the PlotWindow in black because the "invalid" QColor return is black.
Setting the `PlotWindow` background color to white by default (which is actually the case in the backend) would fix the issue.

vallsv:
Closes #2305


t20100:
LGTM

PR #2315, opened 2019-01-04
Respect left alignment.

vasole:
- Remove print statement
- The comment was never at left position but centered when asking LEFT position


vasole:
closes #2316

vallsv:
LGTM

Issue #2296, opened 2018-12-19
Support foreground color


titusjan:
I would like to make plots with white axes, lables, and grid lines on a black background. I noted that in issue #2220 a `setBackgroundColor` method was added to the `PlotWidget` class, but I can't  find a corresponding `setForegroundColor` to set the color of the axes, labels and grid lines. 

Is it currently possible to set the foreground color? If not, could you please implement this?

vallsv:
Hi,

I am pretty sure it is not possible to change the text color yet. I will not have time to do anything around, but maybe somebody else?

t20100:
No foreground color yet, but it should be not to hard to do:
- Adding `BackendBase.setForegroundColor` method and implementing it for both backend
- For the OpenGL backend, the color parameter is already coded in the shader but it is set to black in `GLPlotFrame.py`. It's possible to make it settable and use it in  `GLPlotFrame.render` and also set the color of all `Text2D` instance and then call this from `setForegroundColor`.
- For the matplotlib backend calling the right matplotlib methods to set axes, tick, labels and title color.

@titusjan if you feel like doing it it will be welcomed, I also have quite some stuff already planed for next release.

titusjan:
I had some free time so I made a PR. Please take a look.

t20100:
closed by #2317

PR #2317, opened 2019-01-04
Foreground color

titusjan:
This PR implements foreground color (issue #2296).

I implemented a separate `foregroundColor` and `gridColor`, similar as the existing `backgroundColor` and `dataBackgroundColor`. It is implemented for both backends (OpenGL en Matplotlib).

Note that setting the `dataBackgroundColor` makes the grid lines invisible in the GL backend (the MPL backend works fine). This bug was already present before this PR and I didn't see an obvious fix.

If fixed another bug though: the `backgroundColor` didn't work in the MPL backend.

I added `bg_color.py` to the `examples` directory for if you want to test it. Feel free to remove this demo script if it's no longer necessary.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -21,7 +21,7 @@ import logging
 import collections
 from argparse import ArgumentParser
 
-#logging.basicConfig()
+logging.basicConfig()
 logger = logging.getLogger("bootstrap")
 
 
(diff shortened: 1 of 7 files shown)


vallsv:
Hi. Thanks for the PR.
I clean up few little things, @t20100 should check your PR later.

PR #2319, opened 2019-01-07
Added test of read return value.

alexmarie78:
In order to fix the warning.


vallsv:
Nice catch

t20100:
Thanks for the fix!

PR #2326, opened 2019-01-10
Make setup.py compatible with numpy 1.16.0rc1|2

t20100:
This PR makes updates to `setup.py` to work with `numpy.distutils` from v1.16.0rc[1|2].

It also uses pre-release of numpy and cython during build not just during testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
               - BUILD_COMMAND=sdist
               - QT_BINDING=
               - SILX_OPENCL=1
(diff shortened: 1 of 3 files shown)


vallsv:
Looks fine, waiting for the end of the integration.

BTW what about defining `--pre` in the matrix configuration, then we could enable of disable it.

t20100:
Yes good idea (same for `-q`).

t20100:
Done.

This PR should be reviewed/merged quickly as numpy 1.16.0 is out and build fails with it....

PR #2331, opened 2019-01-11
2327 opencl kahan

### LINKED ISSUE #2327: [OpenCL] Kahan summation failing with POCL and x87
This issue comes from pyFAI:
https://github.com/silx-kit/pyFAI/issues/974

As Kahan summation is also used in silx for statistics calculation, we should address it here first.
The solution is to enforce the copy in a 32bit memory of a (longer) register of the temporary buffer, but only for x87 engine as it probably has performance penalty.

kif:
close #2327

### DIFF
--- a/silx/opencl/processing.py
+++ b/silx/opencl/processing.py
@@ -41,7 +41,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "27/02/2018"
+__date__ = "11/01/2019"
 __status__ = "stable"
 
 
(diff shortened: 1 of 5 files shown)


pierrepaleo:
Looks good. Perhaps we could factor [send_buffer()](https://github.com/silx-kit/silx/blob/c66df3dee54338e3a88fc8ac4f2e9ddb4d2115cc/silx/opencl/statistics.py#L149) method in OpenCLProcessing in the future, as other processing methods might benefit from the casting done on device ?

kif:
On Mon, 14 Jan 2019 00:06:33 -0800
Pierre Paleo <notifications@github.com> wrote:

> Looks good. Perhaps we could factor [send_buffer()](https://github.com/silx-kit/silx/blob/c66df3dee54338e3a88fc8ac4f2e9ddb4d2115cc/silx/opencl/statistics.py#L149) method in OpenCLProcessing in the future, as other processing methods might benefit from the casting done on device ?

It could make sense ... but probably in another issue/PR

-- 
Jérôme Kieffer
tel +33 476 882 445

PR #2332, opened 2019-01-11
Update plot background/foreground color handling

t20100:
This PR follows PR #2246 and #2317 and rework background/foreground color.

It also fixes the display of the plot grid when the data background color is set with the OpenGL backend (issue initially discussed in PR #2317).


vallsv:
About
```
        :param color: The new RGB(A) color.
        :type: Union[List[int],List[float],QColor,None]
```
why dont you merge it as
```
        :param Union[List[int],List[float],QColor,None] color: The new RGB(A) color.
```
? I dont care, but in case there is a reason.

vasole:
The documentation generator contemplates and creates separated fields for the description and the type.

vallsv:
LGTM

PR #2333, opened 2019-01-11
[plot] Implement line bg color for shapes

t20100:
This PR follows PR #2266 and:
- adds implementation in the OpenGL backend of line background color,
- moves `get|setLineBgColor` from `LineMixIn` to `Shape` item as it is only implemented here for now, and
- adds a warning in matplotlib backend where it is not available (i.e., `line`, `hline` and `vline` shapes)
- fixes `hline` and `vline` shapes in the OpenGL backend.

The implementation of the dash line background color in the OpenGL backend is different than that of the matplotlib backend in that it fills the holes of the dash line with the second color rather than being a true background color (different behavior when there is some transparency).
This can be easily changed, but I think it is more the desired behavior.

### DIFF
--- a/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/silx/gui/plot/backends/BackendMatplotlib.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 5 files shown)


vallsv:
>  but I think it is more the desired behavior.

Yes it is.

Issue #2302, opened 2018-12-20
[warning] silx/io/specfile/src/sfinit.c


alexmarie78:
warning on compilation : 
```
static void
sfAssignScanNumbers(SpecFile *sf) {

  int i;
  char *ptr;
  char buffer[50];
  char buffer2[50];

  register   ObjectList *object,
                        *object2;
  SpecScan              *scan,
                        *scan2;

  for ( object = (sf->list).first; object; object=object->next) {
        scan = (SpecScan *) object->contents;

        lseek(sf->fd,scan->offset,SEEK_SET);
        read(sf->fd,buffer,sizeof(buffer));
        buffer[49] = '\0';

        for ( ptr = buffer+3,i=0; *ptr != ' ';ptr++,i++) buffer2[i] = *ptr;

        buffer2[i] = '\0';

        scan->scan_no = atol(buffer2);
        scan->order   = 1;
        for ( object2 = (sf->list).first; object2 != object; object2=object2->next) {
            scan2 = (SpecScan *) object2->contents;
            if (scan2->scan_no == scan->scan_no) scan->order++;
        }
  }
}
```
The return of read() and lseek() are not checked. We want to fix this warning, shall we stop the for loop  when the return value is -1 (failed) or shall we do something else. The return type of the C function is void. So maybe we need to setup an errno somewhere ?

t20100:
To me the clean way to improve it is that `sfAssignScanNumbers` returns a status code based on lseek/read return code and that `SfOpen2` and `SfUpdate` which are calling it check the returned value and in case of error forward it through their `int * error` argument.

The conservative handling of lseek/read error event would be to set the returned status as error, skip the current scan but continue the loop.

vasole:
@t20100 

Perhaps it is not necessary to be conservative and one can stop the loop. I guess the returned values were never checked because they were not expected to fail given the fact the scan limits have been previously determined.

t20100:
Yes sounds reasonable.
This case should only occur if the file is clipped in the meantime which should not occur under normal condition.

PR #2336, opened 2019-01-14
Added a test to check read() return value.

alexmarie78:
Trying to resolve warning from silx/io/specfile/src/sfinit.c #2302 issue.

### DIFF
--- a/silx/app/view/Viewer.py
+++ b/silx/app/view/Viewer.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
-__date__ = "09/11/2018"
+__date__ = "15/01/2019"
 
 
 import os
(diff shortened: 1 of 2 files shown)


vasole:
The function should return an error (instead of being void) with that error taken into account by the calling functions.

vasole:
... or not? Still to decide what to do in case of error. We do not want that a problem with one scan preventing us from accessing all the previous scans.

vallsv:
Is there a way to create a file which will use this return branch? It could help.

alexmarie78:
Hello,
I was working on my fork repository, thus I don't know why my last commit :
"Removed "try/except" blocks for h5py and fabio." is here instead of in my fork repo.
And this thing prevents me from open a new PR with the last commit.

t20100:
Looks you're working on the master branch of your local repository.
if you start working on multiple pull requests (PR) at once, you should do the development in a different branch for each PR.
You can look at scikit-image documentation for a brief description of the process: http://scikit-image.org/docs/dev/contribute.html

alexmarie78:
Hello,
Thanks for all your answers, i'm working on it.

PR #2343, opened 2019-01-15
Fix docstring types in PlotWidget

t20100:
Fix docstring typo spotted in https://github.com/silx-kit/silx/pull/2332#issuecomment-453931525.


vallsv:
LGTM

PR #2347, opened 2019-01-16
Add an env variable for full dependencies

### LINKED ISSUE #2341: Adding a build option for packaging purpose
For debian packaging purpose, @picca would like a build option to add all dependencies as `install_requires` to ease the packaging.

t20100:
Not very nice but simple and working

closes #2341

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -921,6 +921,12 @@ def get_project_configuration(dry_run):
         'full': full_requires,
     }
 
+    # Here for packaging purpose only
+    # Setting the SILX_FULL_INSTALL_REQUIRES environment variable
+    # put all dependencies as install_requires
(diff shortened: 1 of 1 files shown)


vallsv:
I agree it's horrible.
I would prefer a command line argument.
What kind of command line does he use?

t20100:
It should be during the build. It is the distuil command generating dist info with a require.txt file.
I first implemented a global `--with-full-install-requires` argument... but it's not really nice either as you have to handle it before passing on to setup, and it makes changes that are silx specific spread across the setup.py and so it is less easier to reuse elsewhere...
If you feel to propose a better solution don't hesitate.
For me this one is really not nice, but also only 2 lines at a single place.

t20100:
I don't like it either, it just does the job.

PR #2348, opened 2019-01-16
Fix bug with histogram bounds with log scaled X axis

### LINKED ISSUE #2237: [plot] Wrong autozoom with histogram and log axis
The autozoom with histogram is done well (left image), while axis is not switched to log scale (right image).

- The negative part on the X-axis is not display (that's ok)
- But the part between 0..22 on the y-axis should be displayed

screenshot from 2018-11-09 13-22-32

t20100:
closes #2237


vallsv:
LGTM

PR #2353, opened 2019-01-16
Trying to fix issue from #2120

alexmarie78:
I can't import hdf5plugin as the computer doesn't have it installed.
I don't have the rights to install it, thus I can't start "run_tests.py" optimally.
Maybe someone can have a try ?

### DIFF
--- a/examples/compareImages.py
+++ b/examples/compareImages.py
@@ -41,11 +41,7 @@ from silx.gui.widgets.UrlSelectionTable import UrlSelectionTable
 
 _logger = logging.getLogger(__name__)
 
-try:
-    import fabio
-except ImportError:
-    _logger.debug("Backtrace", exc_info=True)
-    fabio = None
+import fabio
 
 try:
     import PIL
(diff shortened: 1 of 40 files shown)


vasole:
The code below does not require any rights

```
pip install --user
```

alexmarie78:
Hi, I'll correct that, thanks.
About hdf5plugin, the issue is while I run "run_tests.py", there are few importError caused by hdf5plugin is missing. And by "rights" I meant that I can't install on this computer as it is not my personal one.

t20100:
Having a quick look at it, changing:
```
try:
    import fabio
except ImportError:
    fabio = None
```
to `import fabio` is not the only change to do.
Every place where the availability of `fabio` or `h5py` is checked can be cleaned-up.
E.g., for `silx/examples/compareImages.py` line 134:
```
    if fabio is None and PIL is None:
        raise ImportError("fabio nor PIL are available")

    if fabio is not None:
        try:
            return fabio.open(filename).data
        except Exception:
            _logger.debug("Error while loading image with fabio", exc_info=True)
```
can be replaced with:
```
    if PIL is None:
        raise ImportError("fabio nor PIL are available")

    try:
        return fabio.open(filename).data
    except Exception:
        _logger.debug("Error while loading image with fabio", exc_info=True)
```

Also `run_tests.py` should be kept as it is. It is an exception as we share it across multiple projects so it is best to keep it generic.

It is best to make small Pull Requests with only a few file changed (e.g., all examples, then submodule by submodule): it will get merged faster in the master.

alexmarie78:
Splitted in two patch set, one for fabio and the other one for h5py.
Hope it will be okay for you

t20100:
Good, splitting is better for us, thanks!

PR #2355, opened 2019-01-17
Allow to drop URI from hdf5 tree to external applications

vallsv:
- Implement `url` from H5Node (object provided to user on tree interaction)
- Implement generic mimetypes for item of the tree
    - `text/uri-list` was NOT implemented as the behaviour with most application was not very nice
    - `text/plain` containing the URL, convenient to drag and drop thing to any application
    - `application/x-silx-uri` also containing the same URL, convenient for application that really expecting an URI from silx

The Qt drop code is easy to write. And silx provides API to read the URI in case:
```
# Parsing the URL
import silx.io.url
dataurl = silx.io.url.DataUrl(url)
# dataurl provides access to URL fields

# Read it as a numpy array
import silx.io
dataset = silx.io.get_data(url)

# Open an hdf5 object (URL targetting a file or a group)
import silx.io
with silx.io.open(url) as h5:
    # ...your stuff...
```

Closes #2355

### DIFF
--- a/examples/dropZones.py
+++ b/examples/dropZones.py
@@ -1,0 +1,134 @@
+#!/usr/bin/env python
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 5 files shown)


vallsv:
`./bootstrap.py examples/dropZones.py` could be used to test this feature.
You have to open `silx view` on a side.

payno:
LGTM

PR #2356, opened 2019-01-17
[plot] Rework of StatsWidget

t20100:
This PR updates to StatsWidget module with:

- Added proxy methods instead of using attributes to do the proxies
- Updated docstings
- Added `getPlot` and use of a `weakref` to the `PlotWidget`.


payno:
LGTM

PR #2360, opened 2019-01-18
Fix 2358

### LINKED ISSUE #2358: [stats] create 'line' stats widget
pymca like widget which display the statistic on one or two lines with stat name and value.

payno:
PR to create a 'pymca' like widget for displaying the statistics.

Some points have to be clarify

* this widget is dedicated to a given kind because we can have several active item.
* should this widget have a control (at least optional) to display statistic on the full data or only on the plot viewport

TODO:

- [x] deal with plot viewport (add control like the _OptionsWidget). Maybe we should limit ourself to a single parameter in the constructor to keep this widget 'compact'.
- [x] merge StatsLineWidget file and plotStats
- [x] create a common class between StatsLineWidget & StasWidget (MixIn or Base class)
- [x] add unit-test

close #2358


payno:
tmp

payno:
@t20100 should we include this by default in the plotWindow as the StatsWidget ?

vasole:
It is just cosmetics, but usually one puts the mean prior to the standard deviation.

The order COM, Mean, STD is more appropriate.

vallsv:
BTW it would better not to use a `QLineEdit` as it is not editable. A `QLabel` is enough i guess. You can custom the flags to allow to select the text with the mouse. And if you are afraid of the fixed size of the `QLabel`, i created an elidable version on pyFAI.

payno:
move to PR #2376

PR #2363, opened 2019-01-25
Add support of plot3d widgets in StatsWidget

### LINKED ISSUE #2021: [plot3d] Add a widget to display some statistics on SceneWidget content
Idea is to have something similar to the plot stats widget, sharing API/code whenever possible.
Issue is that for 3D computation probably needs to be done in a separated thread

t20100:
This PR makes the `StatsWidget` compatible with `silx.gui.plot3d` `SceneWidget` and `ScalarFieldView` content.

Some things are missing (to be done later):
- The `StatsWidget` can be optimized to avoid unneeded computation (issue #2367, #2368)
- For 3D, the coordinates of the computed stats are the array coordinates (i.e., no transformation is applied) (issue #2364).
- For 3D, the computation of the stats on data currently visible in the viewing frustum is not available... but is it really needed?. And in the same idea, the clipping plane is not taken into account. (issue #2365)

To do so, I did a refactoring of the `StatsWidget` to stop using plot legend and kind and use items instead.

closes #2021


payno:
Nice rework.
Minor modifications PR before merge: https://github.com/t20100/silx/pull/20
Otherwise LGTM

t20100:
I merged your branch.

PR #2366, opened 2019-01-25
math.fft: make explicit the supported batched transforms in cufft

pierrepaleo:
This PR adds a small fix for the cufft backend in `silx.math.fft`.
The "generalized" batched transform has still to be written (for now it only supports batched transform along the "slow" dimension).


vallsv:
Empty stat segfault. I restrarted CI on Travis Python 3.7

Issue #2346, opened 2019-01-16
Yet another issue with nightly build


t20100:
Looks to be due to loading different version of PyQt during the doc generation:
```
[...]
/home/tester/silx/doc/source/modules/gui/console.rst:7: WARNING: autodoc: failed to import module u'silx.gui.console'; the following exception was raised:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/sphinx/ext/autodoc.py", line 529, in import_object
    __import__(self.modname)
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/console.py", line 95, in <module>
    from qtconsole.rich_ipython_widget import RichIPythonWidget
ImportError: No module named qtconsole.rich_ipython_widget
/home/tester/silx/doc/source/modules/gui/data/dataviewer.rst:4: WARNING: autodoc: failed to import module u'silx.gui.data.DataViewer'; the following exception was raised:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/sphinx/ext/autodoc.py", line 529, in import_object
    __import__(self.modname)
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/data/DataViewer.py", line 30, in <module>
    from silx.gui.data import DataViews
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/data/DataViews.py", line 40, in <module>
    from silx.gui.dialog.ColormapDialog import ColormapDialog
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/dialog/ColormapDialog.py", line 76, in <module>
    from ..plot import PlotWidget
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/plot/__init__.py", line 63, in <module>
    from .PlotWidget import PlotWidget  # noqa
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/plot/PlotWidget.py", line 50, in <module>
    from .backends.BackendMatplotlib import BackendMatplotlibQt
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/plot/backends/BackendMatplotlib.py", line 47, in <module>
    from ..matplotlib import FigureCanvasQTAgg
  File "/home/tester/silx/build/lib.linux-x86_64-2.7/silx/gui/plot/matplotlib/__init__.py", line 95, in <module>
[...]

PR #2371, opened 2019-01-25
remove debug packages from debian 8 build

t20100:
This PR removes debian 8 debug packages from the nighlty build.


kif:
LGTM

t20100:
Related to #2346

PR #2372, opened 2019-01-25
plot3d items: add ColormapMesh and add indices to Mesh

t20100:
This PR mainly adds `indices` to `Mesh` and a `ColormapMesh` item for the `SceneWidget`.
It also improves the handling of data range in the model which has some issue for `Scatter2D|3D` and `ColormapMesh`.

### DIFF
--- a/silx/gui/plot3d/_model/items.py
+++ b/silx/gui/plot3d/_model/items.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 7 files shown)


t20100:
- `Colormap` synchronization reworked thanks to @vallsv 
- Added picking of meshes with indices (still need double checking, see issue #2378).

Ready for review.

vallsv:
LGTM, tested with pyFAI

Issue #2007, opened 2018-07-10
Missing Peak, COM, FWHM information in Plot1D


mguijarr:
Users would like to see the same information as displayed by PyMCA in Plot1D.

payno:
Part of it (COM, maybe peak, depending on what is behind it for you) is already included in the latest release (0.8.0).

This is part of the 'silx.gui.plot.StatsWidget.'
An example on how to use it and how to define your own statistics are available in examples/plotStats.

But fron an instance of Plot1D you should be able to display it calling:
`
plot.getStatsWidget().parent().setVisible(True)
`

mguijarr:
Great! Good to know - id10 users will be happy maybe.
I am going to activate it by default in our live scan display tool based on silx (Flint).

vallsv:
BTW using that to display it is not very safe
```
plot.getStatsWidget().parent().setVisible(True)
```

It would be better to use the dedicated action if some, or the `setVisible` should be overrided on our side.

t20100:
You can also add the stats widget along with the plot widget in your code (same applies for the legend widget).

vasole:
@mguijarr 

Is the FWHM actually  needed? The correct way to obtain a FWHM is to have a fit.

mguijarr:
Hi @vasole yes it is still a demand from users.

We offer to calculate fwhm with a function in BLISS (but it is not shown on the silx plot),
as spec was doing, and it is not based on a fit (I do not know the calculation we do, though -- I can
have a deeper look...).

What would be really great for us is to have some functions in silx to calculate everything spec was
doing before, and to be able to have those values displayed on 1D plots. This way we could call the functions from BLISS if we need to get a value in the Python code, or people could see the
values directly from the plot widget.

payno:
Hi @mguijarr 

In fact silx is offering a structure to display any statistic you want to compute.
See (examples/plotStats).

So if you already have the function computing the FWHM you just have to define the specific statistic from something like:

```python
class FWHM(StaBase):
   ...
   def calculate(self, context):
      # compute and return result
```

Then this Statistic class will work with a StatsWidget to manage the display and connection with silx Plots.

But I don't think the goal of silx is to contain all those kinds of statistic calculation.
Otherwise we will end with thousands of those classes.
So I guess you should have this dedicated code on your side. But we can help for the implementation if needed.

payno:
I would say this is done. @vallsv have you heard of any related need on bliss side ?

PR #2376, opened 2019-01-29
[gui][stats] add pymca like widget for statistics

payno:
This class will be the base class of any widget displaying silx's statistic from a silx plot.


payno:
tmp

payno:
Example of how to use the BasicGridStatsWidget

```python
class FWHM(StatBase):
    """
    Simple calculation of the Full width half maximum
    """
    def __init__(self):
        StatBase.__init__(self, name='fwhm', compatibleKinds=('curve',))

    def calculate(self, context):
        # TODO
        pass

...

my_stats = DEFAULT_STATS
my_stats.add(stat=FWHM(), formatter=StatFormatter())
plot = Plot1D()

...

stats_widget = BasicGridStatsWidget(parent=None, stats=my_stats, plot=plot)
stats_widget.show()

```

payno:
this should close #2007

payno:
@mguijarr

I hope this will solve issue to include the statistics regarding curves on a Plot1D.
And you might want to  include an instance of `BasicGridStatsWidget` on your `LivePlot1D`.

A full example script is given below.

Please be aware that the fwhm implementation is just given as an example and is very 'raw'. But I guess you already have your own implementation of the fwhm.
This widget is limited to a specific kind.. Here 'curve'.

The current example will display only the statistic regarding the active curve.
Be also aware that displayed statistics will not take into account the viewport. So statistic will be made on the entire curve, no matter the viewport. (To change the behavior you can use the `statsOnVisibleData` option).

tmp

payno:
```python
from silx.gui import qt
from silx.gui.plot.stats.stats import StatBase
from silx.gui.plot.StatsWidget import BasicGridStatsWidget, DEFAULT_STATS
from silx.gui.plot.stats.statshandler import StatFormatter
import numpy
import logging

_logger = logging.getLogger()


class FWHM(StatBase):
    """
    Simple calculation of the Full width half maximum
    """
    def __init__(self):
        StatBase.__init__(self, name='fwhm', compatibleKinds=('curve',))

    def calculate(self, context):
        """
        warning: given if the function to be follows a normal distribution
        """
        y_max_value = context.max - context.min
        hm = y_max_value / 2.0
        index = numpy.argmax(context.values)
        # if has several index, return a numpy array
        if isinstance(index, numpy.ndarray):
            _logger.warning('more than one peak found fail to compute FHWM')
            return None

        left_index = (numpy.abs(context.values[0:index] - hm)).argmin()
        right_index = (numpy.abs(context.values[index:-1] - hm)).argmin()
        right_index = right_index + index

        # TODO: find the real intersection of the full fwhm to remove the
        return context.data[0][right_index] - context.data[0][left_index]


if __name__ == '__main__':
    _qapp = qt.QApplication([])
    my_stats = DEFAULT_STATS
    my_stats.add(stat=FWHM(), formatter=StatFormatter())
    plot = Plot1D()

    x = numpy.linspace(-2, 5, 5000)
    sigma = 0.5
    mu = 1.5
    y = 1.0 / (sigma * numpy.sqrt(2.0*numpy.pi)) * numpy.exp(-(x-mu)**2) / (2.0 * sigma**2)
    plot.addCurve(x=x, y=y, legend='normal-curve')
    plot.setActiveCurve('normal-curve')
    plot.show()

    stats_widget = BasicGridStatsWidget(parent=None, stats=my_stats, plot=plot)
    stats_widget.show()
    _qapp.exec_()
```

vasole:
I would suggest the order COM mean STD instead of STD mean COM. It is much more common to see mean followed by STD than the other way around.

vasole:
FAIL: testSetAnotherPlot (silx.gui.plot.test.testStats.TestStatsWidgetWithCurves)
  File "C:\projects\silx\venv_test\lib\site-packages\silx\gui\plot\test\testStats.py", line 434, in testSetAnotherPlot
    self.assertEqual(self.widget.rowCount(), 1)
AssertionError: 0 != 1

PR #2380, opened 2019-01-30
Fix random test failure with PySide2

t20100:
This PR should fix random test failure of testEmptyStatsWidget with PySide2


payno:
Nice catch

Issue #2383, opened 2019-02-04
FFT/OpenCL tests require user input to select the device


kif:
this locks the build ...

kif:
```
test_numpy_fft (silx.math.fft.test.test_fft.TestNumpyFFT) ... ok
test_numpy_fft (silx.math.fft.test.test_fft.TestNumpyFFT) ... ok
Choose platform:
[0] <pyopencl.Platform 'Intel Gen OCL Driver' at 0x7fed50181540>
[1] <pyopencl.Platform 'Intel(R) OpenCL' at 0x30d7490>
[2] <pyopencl.Platform 'Portable Computing Language' at 0x7fed4c975020>
```

pierrepaleo:
Should be fixed by #2384.

Issue #2375, opened 2019-01-29
silx.gui.plot breaks with pyfftw


alexbjorling:
I just cloned the `silx` master branch to get access to a new feature.

There's some problem with `pyfftw` as shown below. If I uninstall `pyfftw` there's no problem. If I revert to `silx==0.9.0` with `pyfftw` installed, it works too.

```
In [1]: import silx.gui.plot
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-8a6d86e68a2f> in <module>()
----> 1 import silx.gui.plot

/home/alex/.local/lib/python2.7/site-packages/silx/gui/plot/__init__.py in <module>()
     61 
     62 
---> 63 from .PlotWidget import PlotWidget  # noqa
     64 from .PlotWindow import PlotWindow, Plot1D, Plot2D  # noqa
     65 from .items.axis import TickMode

/home/alex/.local/lib/python2.7/site-packages/silx/gui/plot/PlotWidget.py in <module>()
     50 from .backends.BackendMatplotlib import BackendMatplotlibQt
     51 
---> 52 from ..colors import Colormap
     53 from .. import colors
     54 from . import PlotInteraction

/home/alex/.local/lib/python2.7/site-packages/silx/gui/colors.py in <module>()
     37 from silx.gui import qt
     38 from silx import config
---> 39 from silx.math.combo import min_max
     40 from silx.math.colormap import cmap as _cmap
     41 from silx.utils.exceptions import NotEditableError

/home/alex/.local/lib/python2.7/site-packages/silx/math/__init__.py in <module>()
     38 from .histogram import HistogramndLut  # noqa
     39 from .medianfilter import medfilt, medfilt1d, medfilt2d
---> 40 from .fft import fft

/home/alex/.local/lib/python2.7/site-packages/silx/math/fft/__init__.py in <module>()
      6 __date__ = "12/12/2018"
      7 
----> 8 from .fft import FFT

/home/alex/.local/lib/python2.7/site-packages/silx/math/fft/fft.py in <module>()
     24 #
     25 # ###########################################################################*/
---> 26 from .fftw import FFTW
     27 from .clfft import CLFFT
     28 from .npfft import NPFFT

/home/alex/.local/lib/python2.7/site-packages/silx/math/fft/fftw.py in <module>()
     28 from .basefft import BaseFFT, check_version
     29 try:
---> 30     import pyfftw
     31     __have_fftw__ = True
[...]

PR #2384, opened 2019-02-04
FFT/clfft: choose best device based on ocl class

pierrepaleo:
This PR aims at solving #2383. It also removes the non-lazy import of FFT in `silx.math.__init__.py`, fixing #2375.


kif:
LGTM

PR #2387, opened 2019-02-05
Change image rendering in matplotlib backend

### LINKED ISSUE #2185: [plot] Glitch on image rendering
There is a difference of rendering between a pure matplotlib image and the silx plot. The silx plot introduce artefact.

glitch-modestimage

Here is the matplotlib code
```
import fabio
im = fabio.open("mynicedata.edf").data
imshow(log(im), cmap=cm.inferno, interpolation="nearest")
```

Maybe it come from `ModestImage`. As `ModestImage` was updated on the master, this issue have to be checked again.

Feedback from Jon Wright.

t20100:
This PR removes the use of `ModestImage` from the matplotlib backend.
Instead, it uses a smaller `Image` class providing a fast path when setting `uint8` RGBA images and it uses the resampling of matplotlib.

It also removes some useless compatibility code in `addImage`.

It would be nice if you can check for performance change.
From what I tested (4kx4k), it seems similar.
What might differ is when changing the data or colormap, it might be slower for very large images compared to `ModestImage` as it first converts the whole image to RGBA instead of doing it for each rendering for the displayed part...

closes #2185


vallsv:
Looks to be fine on pyFAI

vasole:
Well, I find a bit fast the approval of the PR. The problems are not to be expected with static images but with applications dealing with online visualization.

vallsv:
We wanted to use it at the pyFAi workshop (i agree it was not needed to merge it to use it).
But i think it is also good to test it with people at the code camp.
At worst we can revert it.

Issue #2390, opened 2019-02-05
[ROIs] Buggy behavior introduced in current master


vasole:
The color of the default full region markers is buggy. When the markers are not movable the color should be black and the middle marker should not appear.

vasole:
It works with silx 0.8.0 and 0.9.0 but not with current master.

vasole:
closed by #2391

PR #2391, opened 2019-02-05
fix 2390 - ICR roi color and middle marker visibility

payno:
if ROI is ICR then set back default marker color to black and hide middle - marker


vasole:
- Default color is black for the ICR
- Default color is blue for the other markers
- Default color for the middle marker is yellow
- Default behavior is no middle marker

payno:
close #2390

vasole:
Not fixed by latest commit.

What is not respected is the behavior requested by a call  ``setMiddleROIMarkerFlag(True)``

vasole:
The behavior seems correct now. The deprecation warning corrected with #2389 is present in the code.

PR #2394, opened 2019-02-06
Add cxi extension to supported files

### LINKED ISSUE #2385: [silx view] Add *.cxi to supported HDF5/NeXus file
`.cxi` is a standard file format for coherent X-ray imaging file, it would be nice to have it in the supported format of silx view.
It is a HDF5 file and apparently NeXus compliant.

t20100:
Also use OrderedDict to keep format order.

closes #2385


vallsv:
LGTM

PR #2395, opened 2019-02-06
Make profile coordinates use plot coords + rework title

### LINKED ISSUE #1642: [plot] Profile should use coordinates not just indices

t20100:
This PR updates the image/stack profile tool to use coordinates from the initial plot as X coordinates in the profile.

closes #1642

### DIFF
--- a/silx/gui/plot/Profile.py
+++ b/silx/gui/plot/Profile.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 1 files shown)


payno:
Work for me

PR #2397, opened 2019-02-06
Update CurvesROIWidget.py

vasole:
- The method fillFromROIDict does not conflict with anything and it is very explicit. It should not be deprecated.
- If the method setRois wants to issue a deprecation warning when receiving a dict and not a Roi object, then fillFromROIDict should take care of generating that object. Until then, no warnings.


vasole:
The API setRois, getRois, calculateRois was introduced with the idea to leave the old API untouched for compatibility. The idea was to remove the old API from the documentation therefore making sure newcomers would use the new API.

PR #2425, opened 2019-02-12
[silx convert] Allow to use a filter id for compression

vallsv:
The command line argument is converted to int, if possible.

### DIFF
--- a/silx/app/convert.py
+++ b/silx/app/convert.py
@@ -25,7 +25,7 @@
 
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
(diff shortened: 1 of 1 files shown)


vasole:
I think the accompanying documentation should be improved too. For instance by showing the usage with some typical filters id (GZIP, LZ4,..) and providing a link to the HDF5 group page where the filter ids are stored:

https://support.hdfgroup.org/services/filters.html

vallsv:
I was also thinking about adding mapping (id<->name) on hdf5plugins? At least filters exposed.

Issue #2434, opened 2019-02-13
getpass.getuser() is broken on containerized environment


kif:
The solution is to fall back on default user-ID instead of the name

vallsv:
BTW i read that there is a way to create users on docker, with should also solve the solution.
```
docker run -v etcvol:/etc python..... useradd -u `id -u` $USER
```

kif:
This impacts Jan at Desy

PR #2440, opened 2019-02-13
Fix getuser

vallsv:
Some re-factorisation on top of the patch from Jerome

Improve #2434


kif:
There is a problem as os.getuid does not exist on windows ...

vallsv:
Interesting, but `getlogin` exists on Windows, then it should be fine.

PR #2441, opened 2019-02-13
Improve mask tools GUI

vallsv:
Few small changes on the threshold tools to improve usability.

It was difficult to know how to use it.

- Split other tools on another group (make the group consistent)
- Setup the text button to know what it will do
- Select a tool by default (the `bellow` tool)
- Hide unused configuration widgets
- Plus some clean up on the layouts

On a side note application could need to add new tools (for pyFAI we could want to add filter designed to remove cosmic rays). With the new group it is more easy to add this kind of buttons.

screenshot from 2019-02-13 14-44-37


t20100:
+1 for a more modular approach and thresholding improvement.

Can't the treshold stuff and histogram-based mask (issue #2442) be the same thing?

PR #2445, opened 2019-02-14
Add a tool to mask ellipses

vallsv:
This PR implements a mask tool using ellipses.

It allows to create in a single pressed button an ellipse aligned to the axis.

It is supported by both scatter and image mask.


t20100:
Looking at how other tools are providing drawing ellipse interaction: Gimp and Inkscape provide 2 interactions (toggled with a modifier key):
- The first and last point of the drawing drag interaction are 2 corners of the bounding box aligned with the X and Y axes.
- The first point of the drawing drag interaction is the center and the last point a corner of the bounding box aligned with the X and Y axes.

So I would say that providing one of those would be enough.

Another option would be to place the center at the first point of the drawing drag interaction and to use the last point as a point on the drawn ellipse where the tangent of the ellipse is perpendicular to the segment [center, last point].
Doing this, the position of the mouse lies on the ellipse (as opposed to controlling the bounding box).
Yet, it has to be tested to see if it is better than controlling the bounding box.

vallsv:
I aim your third proposal. I think it is the most convenient. It feels a little strange to use, though.

vallsv:
As you asked for, @jangarrevoet if you want to try, a release of silx is planed next week.

jangarrevoet:
@vallsv I will give it a try.

PR #2447, opened 2019-02-14
Update default symbol

t20100:
This extends and modify a bit what's been introduced in PR  #2419.
It makes sure `silx.config.DEFAULT_PLOT_SYMBOL` is used:
- in `plot3d` scatter items
- as the default symbol for `sx.scatter` and `sx.points3d` functions.

There is a change of behavior from the initial PR: What was set by the initial PR was the symbol of curves only, while here it changes the symbol of scatters.
@maurov, is that OK for you?
If you need to set a default symbol for curves as well, I would be in favor of introducing a `silx.config.DEFAULT_PLOT_CURVE_SYMBOL` (default to '') because 2 different default are needed for scatter and curves...

### DIFF
--- a/silx/_config.py
+++ b/silx/_config.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2019 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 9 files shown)


vasole:
My opinion is that one needs to be able to set a different default symbol for curves and scatter.

t20100:
Default for curves and scatter are different.
The default for curves is hard-coded to no symbol, while that of scatters can be set through `silx.config`.
It's easy to make it also settable for curves, but I would only do it if it is needed... because we can add many things in `silx.config`.

maurov:
@t20100 I agree with @vasole that we should add in the config both default symbol for scatters and curves. `silx.config` will inevitably grow. Personally, I would like to have it similar to `matplotlib.rcParams`, permitting to dynamically change the default settings in a python script or interactively from the python shell.

t20100:
It is a bit more complicated with curves because there is a special mechanism in `PlotWidget` with `PlotWidget.setDefaultPlotPoints` which changes the default symbol...
I'll make something specific for that, but I think we should deprecate this mechanism, it should only be used by the action changing curve style and so it should be implemented there.

t20100:
Added settable default PlotWidget curve symbol.
Ready for review.

Issue #2406, opened 2019-02-11
[silx view] Make NXdata view richer: complex field, 3D view...


t20100:
As it is now, if we try to put display a 2d image of complex, it is displayed as an image... how far can we go with nexus, and far do we want to go into having a rich interpretation of the NXdata?

t20100:
Closed by #2450 #2455 #2456 and #2457

PR #2450, opened 2019-02-15
Add a NXdata for complex images

t20100:
This PR adds a specific view for NXdata with complex image signal.

related to #2406

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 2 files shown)


t20100:
The feature is there, but it is slow when browsing a 3D stack compared to complex view outside NXdata view... This needs to be investigated.
Also there is a lot of duplicated code at least between image and complex image... and maybe other type of data in NXdata view widgets.

I would merge it as it provides the feature, and improve it later.

t20100:
closes #2453

vallsv:
LGTM

PR #2454, opened 2019-02-15
OpenCL convolution

pierrepaleo:
## About

This PR adds a general-purpose OpenCL convolution in silx.

### Use cases
- [x] 1D <-> 1D
- [x] 2D <-> 2D
- [x] Batched 1D on 2D data
- [x] Separable (2D->1D) on 2D data
- [x] 3D <-> 3D
- [x] Batched 1D on 3D data
- [x] Batched 2D on 3D data
- [x] Separable (3D->1D) on 3D data
- [x] Separable (3D->2D) on 3D data
- [ ] Batched separable (2D->1D) on 3D data

### Additional internal work

- <del> Modify kernel and/or axes on-the-fly </del> --> not relevant, lot of compiled code depend from axes.
- [x] Allow for profiling with the use of OpenCL events
- [x] Bypass use of `d_tmp` when memory is lacking (only done for nonseparable transforms)
- <del>Put convolution kernel in constant memory</del> --> not relevant with textures
- [x] Use "textures" when possible
- [x] Refactor `convolution.cl` to use less code duplication
- [ ] Benchmark (eg. against arrayfire)
- [x] Add boundaries handling other than periodic
- [ ] Automatic composition with `ImageProcessing` for pre-processing
- [ ] Support OpenCL Buffers (not only pyopencl Arrays)
- [x] Built-in default kernels (ex. Gaussian)
- [ ] Nonseparable convolution with non-square kernels
- [ ] [Bonus] Input/output strides ?

### DIFF
--- a/doc/source/modules/opencl/convolution.rst
+++ b/doc/source/modules/opencl/convolution.rst
@@ -1,0 +1,10 @@
+
+.. currentmodule:: silx.opencl
+
+:mod:`convolution`: Convolution
+-------------------------------
+
+.. automodule:: silx.opencl.convolution
+    :members: Convolution, gaussian_kernel
+    :show-inheritance:
+    :undoc-members:
--- a/doc/source/modules/opencl/index.rst
+++ b/doc/source/modules/opencl/index.rst
@@ -11,6 +11,9 @@
    sift/index.rst
    fbp.rst
    sinofilter.rst
+   processing.rst
+   convolution.rst
+   statistics.rst
    medfilt.rst
    codec_cbf.rst
 
--- a/doc/source/modules/opencl/processing.rst
+++ b/doc/source/modules/opencl/processing.rst
@@ -1,0 +1,10 @@
+
+.. currentmodule:: silx.opencl
+
+:mod:`processing`: Processing
+-------------------------------
+
+.. automodule:: silx.opencl.processing
+    :members: OpenclProcessing, KernelContainer
+    :show-inheritance:
+    :undoc-members:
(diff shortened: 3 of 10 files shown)


pierrepaleo:
I think this PR can be merged. Other features will be implemented in further works.  
Some basic benchmarks suggest that this implementation matches the ArrayFire performances only when using textures. Further speed optimizations in kernels can be achieved (like pre-fetching in shared memory), at the expanse of readability.

Here are some notes about possible remaining work. 

## Additional nice-to-have features

* Composition with the `ImageProcessing` class. For example, one could propose on-the-fly color->grayscale conversion. Difficulty: easy
* Use case: Batched separable (2D->1D) on 3D data. This might be useful, for example, when processing color images with separable convolution. Difficulty: moderate (it is likely to break some code genericness).
* Support pyopenCL Buffers (not only Array): this might be useful when composing with other `OpenclProcessing`. Difficulty: easy
* Nonseparable convolution with non-square slices: For now, when doing a nonseparable convolution, all the kernels dimensions must be the same. When not the case, zero-padding of the kernel has to be done. The kernel must be modified to support this case. Difficulty: moderate.
* Input/output strides: sometimes it is interesting to do "convolution + downsampling" in one step, and its transpose (upsampling + convolution). To this end, all kernels/templates must be rewritten, preferably in another file. Difficulty: hard (notably handling the cases where the images dimensions are not appropriate to downsampling).

payno:
There is no link to the 'Convolution' class in the documentation.
But this is also the case for other opencl module, I don't know if this is intentional or not.

payno:
Look good to me.
I don't have any constructive criticism to made but this PR is here for a while now.
Work on my PC.

Issue #2415, opened 2019-02-11
[silx view] Add a NXdata 3d view


t20100:
So that for a NXdata with 3D data, one has 2 views: the current NXdata view with a stack viewer and a 2nd view with a 3D view.

vallsv:
The better thing to do is to create a new main feature: A view that can expose sub views. In this way, i would be really clean to filter the views. I could do that.

```
3D data -> is catched by an abstract 3D nx view -> which provide a 2D view and a 3D view
```

vallsv:
After some tests, the patch is bigger than expected (and not yet ready).
It's safer to provide it for the 0.11 release.

If the feature is needed for 0.10, we still can implement it using a dedicated view (as proposed first):
- The 3D on the NXview
- The 2D (based on 3D data) and another view

PR #2455, opened 2019-02-15
Add a 3d scalar field view to the NXdata views zoo

t20100:
As it is, it is never used as the stack has more priority...
To be effective, this feature needs to be combined with PR #2433 so that both stack and 3d views are available for NXdata.

Also, the 3d view has the same priority as the stack.
Maybe when interpretation=vertex, one wants to favor the 3d view...

related to #2415

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2019 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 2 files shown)


t20100:
related to #2406

vallsv:
LGTM

vasole:
@t20100 @vallsv 

Be careful, interpretation="vertex" means that a three-dimensional array is to be considered as the coordinates of a point. For a scalar field, the interpretation of the signal should be "scalar" or "scaler" so that the 3D array cannot be interpreted as a stack of images or a map of curves.

vallsv:
We do not enforce anything cause we was not sure.
I'll fix it as you asked for.

t20100:
For visualizing 3D scalar fields, both 3D view and stack of images can be useful.

PR #2456, opened 2019-02-15
Support 2dview for 3dnxdata

### LINKED ISSUE #2415: [silx view] Add a NXdata 3d view

vallsv:
This PR allow to display an NXdata containing a volume using a 2D view or a 3D view.

This is the result of #2455 and #2433 

As commended https://github.com/silx-kit/silx/pull/2455#issuecomment-464116563, the NXdata interpretation have to be "scalar" (or no interpretation), the signal must have 3dim, and axes must have the right dimension (according to the signal).

The default selected view is the 2D one, as it is faster and it do not have dependency with OpenGL.

screenshot from 2019-02-15 18-08-54

Closes #2415

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -69,6 +69,7 @@ NXDATA_XYVSCATTER_MODE = 74
 NXDATA_IMAGE_MODE = 75
 NXDATA_STACK_MODE = 76
 NXDATA_VOLUME_MODE = 77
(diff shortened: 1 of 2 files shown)


vallsv:
@vincefn here is what you asked for. If you have time to check it with your data, before the silx release, it could be useful.

payno:
merge this PR before looking at PR #2457 which is related.

PR #2457, opened 2019-02-16
Silx/support 2dview for 3dnxdata

vincefn:
Hi @vallsv (@vasole), thanks for adding this. Some additional changes in this PR:

- In NXData.interpretation:
  - "scaler" is not in "allowed_interpretations" even though it's the official NeXus keyword (I agree scalar should be better though !!)
  - if 'scaler' (or any other not-allowed value) is used for interpretation, the logger instruction (line 384 `if interpretation not in allowed_interpretations`) fails because of the 'None' in allowed_interpretation (`ERROR:silx.gui.qt._qt:<class 'TypeError'> sequence item 0: expected str instance, NoneType`)

- in NXData.is_volume(): the valid interpretation should be 'vertex', not scalar or scaler ? I left scaler and scalar as valid values, not sure what is intended

- I added a separate _NXdataComplexVolumeAsStackView to handle 3d complex data. It could be merged in _NXdataVolumeAsStackView  as well but may be more confusing.

- in ArrayVolumePlot use absolute value for complex arrays

One remaining issue is that opening the NXdata 3D view seems long (>10s for a 256**3 dataset) compared to the standard Cube view - I assume that's because of tricks used in the display in the Cube view (there's a delay too to compute the isosurface, but after the 3D interface pops up), but that's a minor issue.

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -69,6 +69,7 @@ NXDATA_XYVSCATTER_MODE = 74
 NXDATA_IMAGE_MODE = 75
 NXDATA_STACK_MODE = 76
 NXDATA_VOLUME_MODE = 77
+NXDATA_VOLUME_AS_STACK_MODE = 78
 
 
 def _normalizeData(data):
@@ -1654,6 +1655,8 @@ class _NXdataStackView(DataView):
 class _NXdataVolumeView(DataView):
     def __init__(self, parent):
         DataView.__init__(self, parent,
+                          label="NXdata (3D)",
+                          icon=icons.getQIcon("view-nexus"),
                           modeId=NXDATA_VOLUME_MODE)
         try:
             import silx.gui.plot3d  # noqa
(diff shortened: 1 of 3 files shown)


vasole:
@vincefn 

In your modifications you have put vertex as interpretation for volume. That is *not* what you want. Vertex is to be used when having to interpret a multidimensional array as a set of (x, y, z) or (z, y, x) values. What you have is a scalar field in three dimensions. In that case the interpretation of the signal has to be "scaler". Unfortunately "scalar" was misunderstood by the NIAC colleagues that only considered interpretation as being of interest for detectors and they officialized the term "scaler" instead of the proposed "scalar".
 
The interpreation attributed was created precisely to know how to deal with cases like a three dimensional array:

MCA Map -> (n_rows, n_colums, n_channels) -> "spectrum"
Image stack -> (n_images, n_rowx, n_channels) -> "image"
Regular scalar field associated to axes dim_0, dim_1, dim_2 -> ((dim_0, dim_1, dim_2) -> "scaler"

The case "vertex" is to be used, for instance, when you have (n_values, 3) to indicate that those n_values are to be taken as coordinates or vectors (in groups of three).

vincefn:
> In your modifications you have put vertex as interpretation for volume. That is _not_ what you want. Vertex is to be used when having to interpret a multidimensional array as a set of (x, y, z) or (z, y, x) values. What you have is a scalar field in three dimensions. In that case the interpretation of the signal has to be "scaler". Unfortunately "scalar" was misunderstood by the NIAC colleagues that only considered interpretation as being of interest for detectors and they officialized the term "scaler" instead of the proposed "scalar".
> 
> The interpreation attribute was created precisely to know how to deal with cases like a three dimensional array:
> 
> MCA Map -> (n_rows, n_colums, n_channels) -> "spectrum"
> Image stack -> (n_images, n_rowx, n_columns) -> "image"
> Scalar field associated to regular axes dim_0, dim_1, dim_2 -> ((dim_0, dim_1, dim_2) -> "scaler"
> 
> The case "vertex" is to be used, for instance, when you have (n_values, 3) to indicate that those n_values are to be taken as coordinates or vectors (in groups of three).

OK... - but http://download.nexusformat.org/sphinx/design.html would _really_ benefit from being updated, as it states "vertex (3-D data)" and "scaler (0-D data)". And with the same logic, 'image' should also be 'scaler'. That's very confusing specs...

vasole:
True. The representation of a single 2D array would be the same as scaler or as image provided the axes coordinates are monotonic and regularly spaced. A mesh scan is not an image but a scatter plot although if one assumes the motor positions as the nominal ones and not the measured ones the result is equivalent.

What I find confusing is to have to deal with the particular case of regular grids instead of dealing with the generic ones.

I have always advocated to use:

-  (n_points,) or (n_points, 1) for  "scaler"
-  (n_points, n_channels) for  "spectrum"
-  (n_points, n_rows, n_columns) for  "image"
-  (n_points, 3) for "vertex" 
 
repeating n_points the coordinates instead of having to deal with unnatural 3D shapes  for scaler and spectrum. The interpretation attribute would be then redundant.

Following that reasoning, the default representation for a 3D array should be a stack of images.
If one wants to visualize that as a scalar field in a regular grid, then one has to add the "scaler" attribute and if it is a map of 1D data one would have to add the "spectrum" one.

vallsv:
Ok, then ill remove `vertex` for the volume visualization and merge this PR.

Thanks for the support of complex.

vallsv:
Was already fixed. Thanks for it. Then i merge.

Issue #2471, opened 2019-02-25
[compareImages] add an option to skip resetZoom in setData, setImage1...


payno:
Having an option to skip resetzoom in setData, setImage1...

t20100:
I don't know well the API of this widget, but what about adding a `resetZoom` method and not doing reset zoom as part of `set*` methods unless nothing was displayed before ?

payno:
Yes, it might be less constraining.

payno:
closed by #2475

PR #2475, opened 2019-02-26
Compare images modifications for tomwer

payno:
- [x] add an option skip automatic plot reset when setting images
- [x] add a 'raw' A - B mode
- [x] check behavior of Colormap with CompareImages


payno:
close #2471

t20100:
You might want to rename `activateAutoResetZoom` to `setAutoResetZoomActivated`, or both getter and setter to `is|setAutoResetZoom` for consistency with other method names in the module and the package.

PR #2476, opened 2019-02-27
Docstring decorator

### LINKED ISSUE #2443: Add a decorator to duplicate docstring for overridden and proxy methods
This would be convenient when we want to override a method or to make a proxy method while keeping the same docstring.

t20100:
This PR provides a `@docstring` decorator that allows to duplicate a docstring from a function or method.
This is convenient when overriding method or making proxy for object composition.

Usage:
```
from silx.utils.proxy import docstring

class B(A):
    @docstring(A)
    def overridden_method(self):
        pass
```

This PR makes use of it throughout silx.

closes #2443


vallsv:
LGTM

PR #2480, opened 2019-03-04
[ImageStack] Add ImageStack.py file

payno:
Screenshot from 2020-01-15 16-24-07

### DIFF
--- a/examples/imageStack.py
+++ b/examples/imageStack.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


t20100:
I made a PR on your branch with a few small things: renaming `getPlot` to `getPlotWidget` and avoiding to hide the whole plot while waiting.

A first few comments on the API: 
- Instead of using inheritance and override `getUrlLoader` to change the `UrlLoader`, you can provide a `setUrlLoader` method. BTW, is this tuning needed for now?
- I'd keep `getPreviousUrl`, `getNextUrl`, `getNNextUrls` and `getNPreviousUrls` private and add a `getUrls` returning the tuple of URLs and a `getCurrentUrlIndex` methods.
- Adding `getNPrefetch` would complete the `setNPrefetch`.
- Don't you need a signal to know when the current URL changes?

t20100:
Also I'm wondering whether it would better or not to put the list of URLs in a `QDockWidget`. That would be inline with the other `PlotWidget` tools (e.g., mask).

payno:
>     * Instead of using inheritance and override `getUrlLoader` to change the `UrlLoader`, you can provide a `setUrlLoader` method. BTW, is this tuning needed for now?
> 

Yes, the idea is to replace some code in tomwer where we have some different file format like .vol from pyhst... and we need to have a different loader.

>     * Don't you need a signal to know when the current URL changes?

Yes I should provide this information :smiley:

payno:
I think I should adapt the url loader to allow him the possibility to load several url.
Will be usefull for speed up, avoiding some file opening / closing and avoiding massive number of threads in some cases

payno:
> I think I should adapt the url loader to allow him the possibility to load several url.
> Will be usefull for speed up, avoiding some file opening / closing and avoiding massive number of threads in some cases

could be done internally but this is not the first goal of it. Might be done later because will bring some complexity in it and might be really 'application' specific.

payno:
> Also I'm wondering whether it would better or not to put the list of URLs in a `QDockWidget`. That would be inline with the other `PlotWidget` tools (e.g., mask).

Good question.
- It is True that it is not convenient at all to different 'sources' for the Dock Widget and it can be confusing for the users.
- In the ImageStack widget I also think the UrlList is really a 'core' feature that is meant to be the center of the intereaction with users. So it has from my point of view a different status than others possible DockWidgets...

PR #2482, opened 2019-03-05
[debian packaging] Add setuptools to dependencies

t20100:
It was missing


t20100:
nbsphinx was also missing

PR #2486, opened 2019-03-05
[plot3d] Add scene fog

### LINKED ISSUE #2414: For scatter plot3d please add alpha for depth
When viewing points on a 3D scatter plot it is difficult to see the 3D depth. Can you add an "alpha" color for the depth so the points that are further away from the viewer are fainter (e.g. like distance fog : https://en.wikipedia.org/wiki/Distance_fog )

Thanks!

t20100:
This PR adds the option to render the scene with linear fog.

As it is, it is an on/off linear fog across the whole scene.
It is possible to add different kind of fog, to allow to set the range over which to apply it or to specify a fog color.

To try it, use `examples/plot3dSceneWindow.py`, and in the `Object parameters` dock widget on the right, expand the `Settings` node and toggle the `Fog` checkbox.

closes #2414

### DIFF
--- a/silx/gui/plot3d/Plot3DWidget.py
+++ b/silx/gui/plot3d/Plot3DWidget.py
@@ -279,7 +279,9 @@ class Plot3DWidget(glu.OpenGLWidget):
         :raise ValueError: If mode is not supported
         """
         mode = self.FogMode.asMember(mode)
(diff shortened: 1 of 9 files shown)


payno:
Nice

PR #2492, opened 2019-03-08
Move external resources

vallsv:
Move this class to be able to copy-paste the module easily.

- Move the module
- Create `get_file_and_repack` for the fabio usecase
- Update `getdir` to avoid to decompress files into the data_home root (no mix with downloaded files)


vallsv:
We could discuss the need of `get_file_and_repack`. I think it is not a nice function. Then we could remove it or rework it with a more obvious behaviour.

t20100:
To me, this is more internal utilities than the public API, so it's OK if `get_file_and_repack` is not really nice.
It can be improved later.

PR #2495, opened 2019-03-08
Add sample code on how to update a plot3d widget from a thread

t20100:
This PR adds an example on how to update a 3D scatter plot from a thread.
There was already such example for 1D/2D plots, this is adding one for 3D data.

This PR also fixes an issue when there is items in the scene but not containing any data (as initially in the sample code).


vallsv:
LGTM

PR #2500, opened 2019-03-11
Plot3d picking mode

### LINKED ISSUE #2140: [plot3d] Rework picking interaction
As it is now with `PositionInfoWidget` picking is done on double click.
Some alternative ways to handle picking interaction: a dedicated mode with a button, perform picking in whole scene/only in selected item(s)...

t20100:
This PR adds an action to toggle plot3d's `PositionInfoWidget` picking.
When enabled, the picking is performed on left button click (it was a double click before).
The picking mode works in conjunction with pan and rotate interaction.

closes #2140


t20100:
Can someone have a look at this? I would need to build on that.

vallsv:
LGTM

PR #2504, opened 2019-03-12
load hdf5plugin before silx in silx view to make sure it is loaded first

t20100:
This PR makes sure `hdf5plugin` is loaded before anything else is loading `h5py` in silx view.
Without this, the bitshuffle/lz4 plugin is not available in silx view even if installed (tested on macOS).

### DIFF
--- a/silx/app/view/main.py
+++ b/silx/app/view/main.py
@@ -73,13 +73,6 @@ def mainQt(options):
     # Import most of the things here to be sure to use the right logging level
     #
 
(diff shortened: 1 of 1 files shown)


vasole:
Can that one go into 0.10.1 too?

vallsv:
Cause importing `silx` importing `h5py`? That's it?

payno:
I agree it should also go into 0.10.1 ...

Issue #2323, opened 2019-01-09
[plot3d] Add item to display 3D complex field as colormapped isosurface


t20100:

t20100:
What remains is displaying colored isosurface, e.g., isosurface in amplitude colored with phase.

PR #2512, opened 2019-03-20
[plot3d] Add complex 3D volume item to SceneWidget

t20100:
This PR adds a `ComplexField3D` item to the `plot3d.SceneWidget` to display 3D array of complex.
It extends the `ScalarField3D` item by allowing to select a different display mode (i.e., real part, imaginary part, amplitude, phase, amplitude**2) for isosurfaces and the cut plane.

It does NOT provide yet:
- the display of a color-mapped isosurface e.g., an isosurface in the amplitude and colored with the phase.
- the integration in `silx view`.

There is also some work to make the `plot.items` and `plot3d.items` API regarding complex consistent (`getVisualizationMode` in `plot` and `getComplexMode` in `plot3d`).
All this will be done in other PRs.

related to #2323

### DIFF
--- a/silx/gui/plot3d/SceneWidget.py
+++ b/silx/gui/plot3d/SceneWidget.py
@@ -267,7 +267,7 @@ class SceneSelection(qt.QObject):
             assert isinstance(parent, SceneWidget)
 
             if item.root() != parent.getSceneGroup():
-                self.setSelectedItem(None)
+                self.setCurrentItem(None)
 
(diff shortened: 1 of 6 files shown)


vallsv:
I try to visualize a 3D complex data using silx view and it looks to be filtered as float32. Is that normal? Or it's not the way to test it?

t20100:
It is not yet integrated in `silx view`. coming up.

PR #2514, opened 2019-03-22
Script to generate a HDF5 sample

vallsv:


vasole:
It would be nicer if it would generate the correct output.

PR #2517, opened 2019-03-22
Hdf5 creator script

### LINKED ISSUE #2515: Script to generate HDF5 samples fails under windows
As a reminder, float128 not supported under windows.

image

### LINKED ISSUE #2516: Script to generate HDF5 samples. Irreproducible results.
The code generates different output under Python 2 and Python 3. It should aim at generating utf-8 strings, not bytes.

Generated with Python 2:

image

Generated with Python 3:

image

Example of appropriate output (no bytes anywhere):

image

t20100:
This PR updates to `create_df5_types` tool.
It should fix the issue with float128 under Windows (where it will print that it cannot create the dataset).
Little and big endian datasets are produced for int, uint, float and complex types (this provides more test cases and avoids error messages 'name already exists').
The support of python2 is explicitly dropped, and list of str in attributes are stored as unicode.

closes #2515
closes #2516


vallsv:
LGTM, thanks a lot

Issue #2518, opened 2019-03-22
Improve SceneWindow usability when used in silx view as 3D array viewer


t20100:
Usability of it in this specific case can be improved:
- Add a way to set plan // to each axis rather than editing the plane equation.
- Open tree when loading, eventually not displaying part of it (through proxy model)
- Improve selection of plane to pan (actually need to select the plane in the parameter tree).. maybe select this mode when the plane is made visible.

vincefn:
Nice to be able to see the complex data ! Some suggestions:
- rename the type of display from 'Absolute' to 'Amplitude'
- it may be better to initially display the amplitude by default
- the option 'Values<=min' works nicely, but it would be good if it was instead 'amplitude<=min' (I'm aware that may be tricky), especially for the phase
- I'm not sure the Iso-surface of the phase data will be used a lot but it could be good to mask the values where the amplitude is below a threshold. Of course that's less important than the phase colouring of the amplitude iso-surface
- crazy idea: it could be interesting to be able to shrink an iso-surface by a few pixels (especially for the phase colouring of the iso-surface)

t20100:
Thanks for the feedbacks!

t20100:
After PR #2523 gets merged, it remains:

- eventually not displaying part of the proxy model (through proxy model)
- the option 'Values<=min' works nicely, but it would be good if it was instead 'amplitude<=min' (I'm aware that may be tricky), especially for the phase
- I'm not sure the Iso-surface of the phase data will be used a lot but it could be good to mask the values where the amplitude is below a threshold. Of course that's less important than the phase colouring of the amplitude iso-surface
- crazy idea: it could be interesting to be able to shrink an iso-surface by a few pixels (especially for the phase colouring of the iso-surface)

PR #2519, opened 2019-03-22
[silx view] Use SceneWindow for 3D to enable 3D complex field visualization

t20100:
Merge PR #2512 first!

This PR replaces the `ScalarFieldView` widget in `silx view` 3D plotting with `SceneWindow`.
This brings support for 3D complex field visualization.

The pitfall is that `SceneWindow` is more generic and thus the parameter tree is more complicated...
Usability of it in this specific case can be improved, see #2518.

related to #2323


vallsv:
LGTM

PR #2523, opened 2019-03-26
[plot3d] 3D volume visualization/parameter tree improvements

t20100:
This PR makes some improvements of ``silx view`` 3D volume view and ``SceneWidget`` model.
It:
- Adds a drop down menu in the parameter to set the cut plane // to each axis
- Expands the parameter tree when loading
- Sets the interaction mode to plane panning and select the cut plane when it is made visible.
- Initialize the plane position as going through the center of the dataset
- Renames the complex mode from ``Absolute`` to ``Amplitude``
- Uses ``Amplitude`` as the initial complex mode
- Renames ``Values<=Min`` to ``Show <=Min``

related to #2518

### DIFF
--- a/silx/gui/data/_VolumeWindow.py
+++ b/silx/gui/data/_VolumeWindow.py
@@ -32,7 +32,7 @@ __date__ = "22/03/2019"
 import numpy
 
 from ..plot3d.SceneWindow import SceneWindow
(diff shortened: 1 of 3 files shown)


payno:
LGTM

PR #2524, opened 2019-03-26
Some performance improvements

schooft:
When using the OpenGL backend of the Plot2D widget, the profile window still uses the default matplotlib backend. Additionally, the profile is updated every time the image changes, even when the profile window was never shown.

The first commit adds the backend parameter to the ProfileMainWindow and forwards to it the backend of the main plot. This already leads to a noticeable speedup.

The second commit updates the profile only when the profile window is visible. As I haven't spent much time looking for side effects, this approach might be too simple.

Unfortunately, the test suite segfaults on my Debian 9 installation. Therefore, I have run only the Profile tests.

### DIFF
--- a/silx/gui/plot/Profile.py
+++ b/silx/gui/plot/Profile.py
@@ -419,7 +419,8 @@ class ProfileToolBar(qt.QToolBar):
         """
 
         if self._profileWindow is None:
-            self._profileMainWindow = ProfileMainWindow(self)
+            backend = type(plot._backend)
+            self._profileMainWindow = ProfileMainWindow(self, backend=backend)
 
         # Actions
         self._browseAction = actions.mode.ZoomModeAction(self.plot, parent=self)
--- a/silx/gui/plot/ProfileMainWindow.py
+++ b/silx/gui/plot/ProfileMainWindow.py
@@ -35,8 +35,15 @@ __date__ = "21/02/2017"
 class ProfileMainWindow(qt.QMainWindow):
     """QMainWindow providing 2 plot widgets specialized in
     1D and 2D plotting, with different toolbars.
+
     Only one of the plots is visible at any given time.
+
+    :param qt.QWidget parent: The parent of this widget or None (default).
+    :param Union[str,Class] backend: The backend to use, in:
+                    'matplotlib' (default), 'mpl', 'opengl', 'gl', 'none'
+                    or a :class:`BackendBase.BackendBase` class
     """
+
     sigProfileDimensionsChanged = qt.Signal(int)
     """This signal is emitted when :meth:`setProfileDimensions` is called.
     It carries the number of dimensions for the profile data (1 or 2).
--- a/silx/gui/plot/ProfileMainWindow.py
+++ b/silx/gui/plot/ProfileMainWindow.py
@@ -51,13 +51,14 @@ class ProfileMainWindow(qt.QMainWindow):
     """Emitted when the method to compute the profile changed (for now can be
     sum or mean)"""
 
-    def __init__(self, parent=None):
+    def __init__(self, parent=None, backend=None):
         qt.QMainWindow.__init__(self, parent=parent)
 
         self.setWindowTitle('Profile window')
         # plots are created on demand, in self.setProfileDimensions()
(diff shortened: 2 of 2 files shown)


t20100:
Thanks for the PR!

Indeed, it makes sense to use the same backend by default for the profile widget.
I made a small comment in the code regarding the implementation.

BTW, if you want to use the OpenGL backend for all `PlotWidget`, you can set it globally with:
```
import silx
silx.config.DEFAULT_PLOT_BACKEND = 'gl'
```

It's also better not to update the profile window if it is hidden, but you have to make sure it is synchronize it when it gets shown again. I'll check that.

Could you tell us where the tests segfault on your machine? (`python run_tests.py -v` to print the name of each tests)

schooft:
Thanks for the feedback, I updated the PR.

The relevant output for the segfault:
```
testAlignemnt (silx.gui.plot.test.testCompareImages.TestCompareImages) ... WARNING:silx.opencl.common:Last chance to get an OpenCL device ... probably not the one requested
INFO:silx.opencl.processing:Compiling file ['sift/sift', 'sift/convolution', 'sift/preprocess', 'sift/algebra', 'sift/image', 'sift/gaussian', 'sift/reductions', 'sift/memset', 'sift/orientation_gpu', 'sift/orientation_cpu', 'sift/descriptor_gpu1', 'sift/descriptor_cpu', 'sift/descriptor_gpu2'] with options -D WORKGROUP_SIZE=4096
5 warnings generated.
WARNING:py.warnings:/usr/lib/python3/dist-packages/pyopencl/cffi_cl.py:1470: CompilerWarning: Non-empty compiler output encountered. Set the environment variable PYOPENCL_COMPILER_OUTPUT=1 to see more.
  "to see more.", CompilerWarning)

INFO:silx.opencl.processing:0.009MB are needed on device: pthread-Intel(R) Core(TM) i5-6500T CPU @ 2.50GHz,  which has 18742.559MB
Segmentation fault
```

I did not knowingly install anything OpenCL related.

t20100:
Hi,

Looking a bit deeper at it, the profile is not computed if the profile main window is not visible when the image is changed: ``updateProfile`` gets called but ``_createProfile`` returns immediately (https://github.com/silx-kit/silx/blob/master/silx/gui/plot/Profile.py#L673), so the profile computation is not done.
Yet, the profile is cleared (which does not do anything if already empty) and the title and axes label are updated which triggers the refresh machinery in any case. The title and axes label should check if they are actually changed and only perform things in this case. With this, a call to ``updateProfile`` would not actually change anything.

Did you notice any performance improvements with the 2nd commit modification in your case?

As it is implemented, each time the profile is updated, the profile main window is shown and when closed the profile ROI is cleared, so the check of visibility should not be needed.

schooft:
Sorry, I saw a significant improvement with the second commit for the Debian 9 version and just applied it on master without investigating further. Indeed, on master, there is no performance improvement for this change (and therefore also not for the first commit as long as one does not open the profile window). Should I delete the second commit?

t20100:
Yes it's better to remove 2nd commit. As it is, it does not cope with the case where the profile plot widget is provided by the user rather than using the default profile main window.
Improving the update of plot title and axes label will turn what's done in updateProfile to virtually nothing , but that's another PR..

schooft:
I deleted the second commit.

vallsv:
Hi. I added another comment about `if plot._backend is None`. Is there any reason to check that? Else i think it is not needed.

schooft:
That is because `backend = type(None)` will cause an error later in `if backend is None` in `_setBackend` in PlotWidget.py:294, because `type(None) is None` is false.

vallsv:
Yes i see. But I mean `_backend` can't be `None`. `_setBackend` will always set something, else will be aborted. Ontologically, a plot can't work without a backend (that's a constraint, from the `__init__` of the plot).

schooft:
I updated the PR.

vallsv:
LGTM, i merge it with with an additional commit for documentation.

vallsv:
Thanks for the PR

t20100:
Thanks for the contribution!

Issue #2534, opened 2019-03-28
Backprojector: error when padding size is odd


pierrepaleo:
Since v0.10, Backprojector uses Real-to-complex (R2C) FFT for Filtering, and a padding strategy different from v0.9: 

  - In <= 0.9, the padded sinogram size was computed as `size_padded = next_power_of_two(size)`.  
  - From 0.10, it is computed as `size_padded = next_friendly_power(size)`, where `next_friendly_power(n)` gives the closest number in the form `(2**a) * (3**b) * (5**c) * (7**d)`. This avoids excessive memory consumption. 

However in some rare cases, the padded size obtained with the second scheme gives an odd number. Odd numbers are not recommended when using R2C transforms.

Therefore, the padding scheme should not allow odd numbers.

PR #2536, opened 2019-03-29
Backprojection: disallow odd numbers when computing FFT plan

pierrepaleo:
This PR aims at fixing #2534 .

### DIFF
--- a/silx/opencl/backprojection.py
+++ b/silx/opencl/backprojection.py
@@ -154,6 +154,9 @@ class Backprojection(OpenclProcessing):
         self.extra_options = {
             "cutoff": 1.,
             "use_numpy_fft": False,
+            # It is  axis_pos - (num_bins-1)/2  in PyHST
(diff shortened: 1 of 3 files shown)


pierrepaleo:
As mentioned in #2534, the `next_friendly_power(n)` cannot be "any" number in the form `(2**a) * (3**b) * (5**c) * (7**d)` :

  - When using R2C transforms, it is safer to use even numbers (so we should enforce `a >= 1`) 
  - When computing the "Ram-Lak" filter of size `n2 = next_friendly_power(n)`, `n2//2` should also be even (so we should enforce `a >= 2`) ; otherwise the Filter maximum is not at the correct location. Because of this, the final reconstructed slice has an offset in its gray levels. The function  `silx.opencl.sinofilter.compute_ramlak_filter` should be rewritten to avoid this.

pierrepaleo:
The last commits incidentally fix a regression that happened during the 0.10 refactoring.

payno:
LGTM

Issue #2538, opened 2019-04-01
Memory Leak in PlotWidget


schooft:
It seems, there exists a small memory leak in the PlotWidget, see the following example code:

```python
from PyQt5.QtCore import Qt, QTimer
from PyQt5.QtWidgets import QApplication, QWidget
from silx.gui.plot import Plot2D
import numpy as np
import gc


# from http://python.6.x6.nabble.com/Hunting-down-memory-leaks-tp1792250p1792260.html
def gcHistogram():
    """Returns per-class counts of existing objects."""
    result = {}
    for o in gc.get_objects():
        t = type(o)
        count = result.get(t, 0)
        result[t] = count + 1
    return result


def diffHists(h1, h2):
    """Prints differences between two results of gcHistogram()."""
    print("----------diff------------")
    for k in h1:
        if k not in h2:
            print("missing in h2:", k)
        elif h1[k] != h2[k]:
            print("%s: %d -> %d (%s%d)" % (
                k, h1[k], h2[k], h2[k] > h1[k] and "+" or "", h2[k] - h1[k]))


class MyWidget(QWidget):
    def __init__(self, parent=None):
        super().__init__(parent=None)

        self.plot = None
        self.hist = None

        self.timer = QTimer()
        self.timer.timeout.connect(self.newPlot)
        self.timer.timeout.connect(self.printObjects)
        self.timer.start(1000)

    def newPlot(self):
        if self.plot:
            self.plot.close()
        self.plot = Plot2D(backend="opengl")
        self.plot.setAttribute(Qt.WA_DeleteOnClose)
        self.plot.addImage(np.random.rand(1000, 1000))
        self.plot.show()

    def printObjects(self):
        gc.collect()
        hist = gcHistogram()
        if self.hist:
            diffHists(self.hist, hist)
        self.hist = hist


def main():
    app = QApplication([])

    widget = MyWidget()
    widget.show()

    app.exec_()


main()
```

Its output eventually looks like this:
```
----------diff------------
<class 'PyQt5.QtGui.QOpenGLContext'>: 16 -> 17 (+1)
<class 'weakref'>: 8980 -> 9007 (+27)
<class 'tuple'>: 17350 -> 17353 (+3)
<class 'ctypes.c_void_p'>: 17 -> 18 (+1)
<class 'dict'>: 17369 -> 17370 (+1)
```

In case of the matplotlib backend, only the `weakref`s accumulate over time.

t20100:
[...]

PR #2539, opened 2019-04-05
Avoid object leak at destruction of PlotWidget with OpenGL backend

t20100:
This PR avoids to keep reference to objects related to a given OpenGL context after this was destroyed.
It replaces the use of ``dict`` with ``WeakKeyDictionary`` to circumvent this.

It also includes the renaming of 2 functions used internally.

related to #2538

### DIFF
--- a/silx/gui/_glutils/Context.py
+++ b/silx/gui/_glutils/Context.py
@@ -50,7 +50,7 @@ def getCurrent():
     created in.
 
     :return: Platform specific OpenGL context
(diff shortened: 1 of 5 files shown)


t20100:
This now fixes the leak of ``c_void_p`` and ``dict``.
Along with PR #2544, it looks to fix all object leak from issue #2538, it's closing #2538.

@schooft thanks for the investigation!

schooft:
Thanks for all the work!

vallsv:
LGTM

PR #2544, opened 2019-04-12
Fix plotaction parent

### LINKED ISSUE #2538: Memory Leak in PlotWidget

vallsv:
Non-parented `QAction` creates memory leak.

This following code can be used to help to catch this problems.
```
def assert_qaction_parent():

    from PyQt5.QtWidgets import QAction
    qt_qaction_constructor = QAction.__init__

    def my__qaction__constructor(self, *args, **kwargs):
        parent = None
        # Maybe not a very accurate way to reach the parent
        if "parent" in kwargs:
            parent = kwargs["parent"]
        if len(args) != 0:
            parent = args[-1]
        if len(args) + len(kwargs) != 0:
            # Looks like it is used intetrnally like that
            assert(parent is not None)
        return qt_qaction_constructor(self, *args, **kwargs)

    QAction.__init__ = my__qaction__constructor

```

Closes #2538

### DIFF
--- a/silx/gui/plot/PlotWindow.py
+++ b/silx/gui/plot/PlotWindow.py
@@ -29,7 +29,7 @@ The :class:`PlotWindow` is a subclass of :class:`.PlotWidget`.
 
 __authors__ = ["V.A. Sole", "T. Vincent"]
 __license__ = "MIT"
-__date__ = "21/12/2018"
+__date__ = "12/04/2019"
 
 import collections
 import logging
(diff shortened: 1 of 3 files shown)


vallsv:
I also fixed a toolbar parent. But maybe there was a reason for a `None` parent. I suggest to add a comment in this case. That's not a common case.

t20100:
PyQt4 is unhappy with ``parent=self``... otherwise LGTM

vallsv:
I think I will fix it by removing the Qt4 test :-)

PR #2547, opened 2019-04-18
pyopencl: Dont use numpy.int32 for workgroup param

pierrepaleo:
This PR attempts to fix a bug occurring with pyopencl 2014.x, crashing the continuous integration on edna-site.


pierrepaleo:
Should be good... I just tested it with pyopencl 2014.1 and 2018.2.5.

kif:
Looks good.

PR #2548, opened 2019-04-29
python 3.8 compatibility

t20100:
This PR makes `silx` compatible with python3.8 which provides abstract base classes only in `collections.abc` (and no longer in `collections`).

This could be simpler if we'd drop python2.7 (as it does not provide a `collections.abc` module)...


t20100:
Merge PR #2549 first (this is cherry-picked here)....

payno:
LGTM

PR #2549, opened 2019-04-29
Simplify matplotlib init and make compatible with matplotlib3.1

t20100:
This PR reworks the initialization of `matplotlib` to make it compatible with `matplotlib` 3.1 and also to simplify it.


payno:
LGTM

PR #2552, opened 2019-04-29
Add an Enum subclass to silx.utils

t20100:
This PR adds a subclass of `enum.Enum` in `silx.utils` with a few extra convenient class methods.
It makes use of it in `silx.gui.plot3d` and removes some code duplication.

### DIFF
--- a/silx/gui/plot3d/Plot3DWidget.py
+++ b/silx/gui/plot3d/Plot3DWidget.py
@@ -274,7 +274,7 @@ class Plot3DWidget(glu.OpenGLWidget):
         :param Union[str,FogMode] mode: The mode to use
         :raise ValueError: If mode is not supported
         """
-        mode = self.FogMode.asmember(mode)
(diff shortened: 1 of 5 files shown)


payno:
The silx `Enum` class seems nice but what is the question related to the `question` label ?

t20100:
yes the questions:

- Do we want to extend the `Enum` type?
- Is it a good idea to use such `Enum` in APIs in order to allow to set a property which value is in an `Enum` from either the `Enum` member or its value (as I started to do in `plot3d`)?

payno:
For me having our own extension of Enum seems useful. Until it stays light. Having a common implementation of some utils functions will make the code more 'generic':
- members
- from_value (especially for configuration files)
- ...

Usage will be up to the developers I guess.

t20100:
So useful to more than one person...

renamed `asmember` to `from_value`.

Issue #2554, opened 2019-05-02
plot and plot3d API consistency


t20100:
There is some inconsitencies between `plot` and `plot3d` API (complex and visualization mode API) to work on.

Issue #2324, opened 2019-01-09
[plot] Add display of scatter plots as a surface.


t20100:
As is available in plot3d

PR #2555, opened 2019-05-02
[plot] add surface visualization to scatter item

t20100:
This PR extends the 'scatter' item of the `PlotWidget` to support Delaunay triangulation based display of scatter plots.

It adds `set|getVisualization` and `supportedVisualizations` methods to `scatter` items.
It avoids code duplication with `plot3d` through a `triangulation` function.

Related to #2324 but not closing it:
- OpenGL backend implementation is not done yet (to do in another PR).
- `get|setVisualization` API needs to be reworked for consistency with `plot3d` (issue #2554)

### DIFF
--- a/silx/gui/plot/_utils/delaunay.py
+++ b/silx/gui/plot/_utils/delaunay.py
@@ -1,0 +1,72 @@
+# coding: utf-8
+# /*##########################################################################
+#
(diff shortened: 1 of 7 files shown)


t20100:
Made some clean-up of `scatter` API (removed linewidth and linestyle which ended up not being used).

vallsv:
LGTM

PR #2560, opened 2019-05-03
[plot] Allow to use plot OpenGL without matplotlib

t20100:
This PR makes `matplotlib` an optional dependencies when using the `PlotWidget` OpenGL backend.

I'm not really for optional dependencies, but it's actually a small change set (actually there were places where it was already handled).

### DIFF
--- a/silx/_config.py
+++ b/silx/_config.py
@@ -38,7 +38,7 @@ class Config(object):
     .. versionadded:: 0.8
     """
 
-    DEFAULT_PLOT_BACKEND = "matplotlib"
+    DEFAULT_PLOT_BACKEND = "matplotlib", "opengl"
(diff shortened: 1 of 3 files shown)


vallsv:
I think it would be better to remove the warning on the root of the `PlotWidget` module.

And maybe, the default backend have to be checked more carefully if matplotlib is not here. Instead of just an import error, a better exception. What do you think?

t20100:
Warning at import time removed (left as debug log).
Reworked the backend loading to:
- raise RuntimeError if a backend cannot load
- allow to provide a list of backends to try to load (by default 'matplotlib' and then 'opengl')

vallsv:
LGTM

PR #2561, opened 2019-05-03
[plot] OpenGL implementation of scatter plot displayed as surface

### LINKED ISSUE #2324: [plot] Add display of scatter plots as a surface.

t20100:
This PR adds the implementation of `addTriangles` method to the plot OpenGL backend, thus allowing to display scatter plots as surfaces with the OpenGL backend.

Picking still needs to be done.

closes #2324


t20100:
Ready for review.
Quick benchmark: usable up to 100K points with OpenGL on my PC (picking is a bit slow which shows off when using it for hover interaction with `PositionInfo` widget .This can be optimized, as it is now, it is reusing 3D triangles picking).

vallsv:
LGTM

PR #2575, opened 2019-05-09
plot/plot3 complex visualization mode API consistency

t20100:
This PR makes `plot` and `plot3d` items API consistent concerning complex data visualization mode.
It moves the `ComplexMixIn` from `plot3d` to `plot` and uses it in `ImageComplexData` plot item.
API of `ComplexImageView` is updated to be synchronized with `ComplexMixIn`.
Backward compatibility is kept (with deprecation warnings).

Also by using `silx.utils.Enum` it is now possible to set the complex mode with a `str` (e.g., 'phase') as well as an `Enum` member.


t20100:
related to #2554

vallsv:
LGTM

PR #2584, opened 2019-05-15
add a setVisible method to ROI objects

PiRK:
I have a use case where I need multiple sets of ROIs (one set for each image in a list of images) associated with a single PlotWidget. When I change the image in the plot widget, I want to show only the ROIs specific to the new image.

This PR adds a public API to class `RegionOfInterest` to show or hide it in the plot:
`roi.setVisible(False)`

Alternative APIs that would also work for me, in case you don't like this one:

`roimanager.setRoisVisible(False)`   (this works because I have one ROI manager per image)

`item.setVisible(False) for item in roi.getPlotItems()`

### DIFF
--- a/silx/gui/plot/items/roi.py
+++ b/silx/gui/plot/items/roi.py
@@ -321,12 +321,14 @@ class RegionOfInterest(qt.QObject):
             if self._labelItem is not None:
                 self._labelItem._setLegend(legendPrefix + "label")
                 plot._add(self._labelItem)
+                self._labelItem.setVisible(self.isVisible())
 
         self._items = WeakList()
         plotItems = self._createShapeItems(controlPoints)
         for item in plotItems:
             item._setLegend(legendPrefix + str(itemIndex))
             plot._add(item)
+            item.setVisible(self.isVisible())
             self._items.append(item)
             itemIndex += 1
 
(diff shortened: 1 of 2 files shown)


t20100:
This makes sense... I would have expected it to be already done.
Your API sounds good as it is inline with other `PlotWidget` items.

t20100:
I made a PR on your branch to use `is|setVisible` in `RegionOfInterestTableWidget`.

There is an issue when calling `setEditable` while the ROI is not visible (it becomes visible again).

Also (maybe in another PR), it would be good to have a signal for the change of visibility: Either a `sigItemChanged` as for other plot items (maybe the best), or a `sigVisibleChanged`.. or eventually through the already exisiting `sigRegionChanged`... It needs a bit of thinking.

PiRK:
I'm not sure if making the ROI visible when `setEditable(True)` is called is a bug or a feature. But for sure, when  `setEditable(False)` is then called it should be hidden again if still not visible.

PiRK:
The latest commits sets the visibility when the items are first created. So if they are deleted and added back for any reason, they will take into account the current visibility flag.

t20100:
Sounds like the right option.

`setEditable(True)` should not change the visibilty. It is best to split things. If one wants this feature, he can implement it... If you implement it by default, it is not possible to remove it if you don't want it.

t20100:
LGTM, let's do the proper signal implementation in another PR.
I created issue #2587 for it.

PR #2586, opened 2019-05-16
[plot] ignore exotic mouse buttons in GL backend

### LINKED ISSUE #2585: [plot] pressing exotic mouse buttons cause errors
The plot seems to have issues when used with a gaming mouse (accidental clicks)

```
  File "/home/pierre/.local/lib/python3.7/site-packages/silx/gui/plot/backends/BackendOpenGL.py", line 385, in mousePressEvent
    btn = self._MOUSE_BTNS[event.button()]
KeyError: PySide2.QtCore.Qt.MouseButton.ForwardButton

  File "/home/pierre/.local/lib/python3.7/site-packages/silx/gui/plot/backends/BackendOpenGL.py", line 385, in mousePressEvent
    btn = self._MOUSE_BTNS[event.button()]
KeyError: PySide2.QtCore.Qt.MouseButton.TaskButton

  File "/home/pierre/.local/lib/python3.7/site-packages/silx/gui/plot/backends/BackendOpenGL.py", line 385, in mousePressEvent
    btn = self._MOUSE_BTNS[event.button()]
KeyError: PySide2.QtCore.Qt.MouseButton.BackButton

```

Accepting these buttons in `BackendOpenGL.mousePressEvent` will probably cause issues downstream: https://github.com/silx-kit/silx/blob/6cfaaf05d87e2f1d978fcc0dce06b74111e5f270/silx/gui/plot/PlotEvents.py#L62

Maybe they should be ignored.

PiRK:
Close #2585 
~~This provides interaction with 3 new mouse buttons in the gl backend.~~
Edit: Ignore all buttons press or release events, besides "left", "middle" and "right".

As far as I can tell, when using the mpl backend, these buttons are filtered out by matplotlib. They never reach BackendMatplotlib (using matplotlib 3.0.3).

### DIFF
--- a/silx/gui/plot/PlotEvents.py
+++ b/silx/gui/plot/PlotEvents.py
@@ -59,7 +59,7 @@ def prepareDrawingSignal(event, type_, points, parameters=None):
 def prepareMouseSignal(eventType, button, xData, yData, xPixel, yPixel):
     """See Plot documentation for content of events"""
     assert eventType in ('mouseMoved', 'mouseClicked', 'mouseDoubleClicked')
-    assert button in (None, 'left', 'middle', 'right', 'back', 'forward', 'task')
+    assert button in (None, 'left', 'middle', 'right')
 
     return {'event': eventType,
             'x': xData,
--- a/silx/gui/plot/PlotEvents.py
+++ b/silx/gui/plot/PlotEvents.py
@@ -59,7 +59,7 @@ def prepareDrawingSignal(event, type_, points, parameters=None):
 def prepareMouseSignal(eventType, button, xData, yData, xPixel, yPixel):
     """See Plot documentation for content of events"""
     assert eventType in ('mouseMoved', 'mouseClicked', 'mouseDoubleClicked')
-    assert button in (None, 'left', 'middle', 'right')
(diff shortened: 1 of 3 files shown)


t20100:
Let's wait for game developers to ask for it...
In the mean time, it is best to ignore those events early rather than propagating them.

PiRK:
I'm not very familiar with QEvent. Do we just need something like

    def mouseReleaseEvent(self, event):
        if event.button() not in self._MOUSE_BTNS:
            return
        ...

Or do we need to reject the event some other way (event.ignore()) ? I'm not sure what the implications of this are. I found it easier to just add the 3 buttons than to ignore them.

t20100:
Yes, but that might cause issues later on... and there is a lot more buttons (https://doc.qt.io/qt-5/qt.html#MouseButton-enum)

Your code is correct, you can also call the default implementation (but AFAIK it should be the same):

```
def mouseReleaseEvent(self, event):
    if event.button() not in self._MOUSE_BTNS:
        return super(XXX, self).mouseReleaseEvent(event)
   ...
```

t20100:
LGTM thanks for the fix.

PR #2588, opened 2019-05-16
[gui][plot] Add image information

### LINKED ISSUE #1862: Widget with image information
E. Boller from id19 think it would be nice to add the 'active' image dimensions to the list of information displayed by the Plot2D.
Adding it to the existing displayed information or creating a new widget with this information. Third option could be to have it on a tool-tip.

payno:
Goal of this PR is to fix #1862 .
My first approach will be to have a getter / setter functions relative to the ImageInformationWidget as the getPositionInfoWidget from the PlotWindow.
This way people will be able to modify it simply. Like for adding the data type (rgb ...). But no option will be available from the constructor.

We could have a default widget (ImageInformationWidget?) if we found a common usage otherwise we can just keep the position in the layout. @t20100 what do you think about this ?


payno:
image

t20100:
If it's integrated in `Plot2D`, I would just put a getter there: if it does not fill someone needs, it is best to do write another one.

I'm wondering if this is not similar to stats widget and could not be integrated there?

Also, at an implementation detail level, I would not make a `BasePlotInformation` base class that is inherited only once.
That being said, it is probably worth to have a common base class for all `PlotWidget` surrounding tools in the same idea.

Finally, I would rename `setPlot` to `setPlotWidget`. Both are used here and there in `silx.gui.plot` which is not really nice and should be unified in the long term and `setPlotWidget` provides information of the class it takes as input.

payno:
Might be simpler to include this in PLot2d  'posInfo'

payno:
This has been modified to be embed in the 'posInfo' widget.

Screenshot - 05242019 - 01:22:00 PM

t20100:
I made a commit with some fixes, is it OK for you?
LGTM

payno:
My bad, I inverted the first two and the last two.
Looks find

PR #2593, opened 2019-05-22
Improve handling of QApplication reference

### LINKED ISSUE #2592: [ci] silx holds a reference to QApplication irrespectively of being owner or not
See line 142 in https://github.com/silx-kit/silx/blob/master/silx/gui/utils/testutils.py

It should use a weakref.proxy if the application is already existing.

t20100:
This PR improves the way `silx` is dealing with reference to the `QApplication` by only keeping a reference when it created it.
Keeping reference to `QApplication` is known to create some issues when the application ends.

It is done by clean-up unnecessary references, and only keeping a ref when it was created.
Also tests for `silx.sx` were moved outside `silx.sx` to avoid the creation of `QApplication` by `sx` at import...

closes #2592

There is no change of examples or sample code in modules about how the `QApplication` is handled.


payno:
LGTM

PR #2597, opened 2019-05-27
[Plot] common API for draggable item

t20100:
This PR adds a common API (`drag` method) to move `DraggableMixIn` items.
It also make the base class of markers available.

This is to prepare for larger modifications of the plot.


t20100:
Add missing base item class `drag` method.

Ready for review.

PR #2602, opened 2019-05-27
[plot] Refactor picking

t20100:
This PR provides a refactoring of the `PlotWidget` picking machinery to make it more generic.
The initial need for this comes from displaying scatter plot as images (for density plot) which requires a split between plot items and the backend that was not possible with the previous picking implementation.

`PlotWidget` provides 2 methods: `_pick` and `_pickTopMost` to do the picking (to become public at some point and probably return some `PickingResult` objects as in `plot3d` and `sx.ginput`).
Plot items provide a `pick` method returning `None` if not picked and otherwise the indices of the picked data points.


vallsv:
I only test on an own plot demo. Curves on Y2 are selected, that's great.
I restart the tests

vallsv:
I also checked the ROI example. Looks to be fine.

Issue #2609, opened 2019-05-28
[CI] warning/deprecation warnings during tests


t20100:
Looking at CI logs, there are quite a few:
- [x] deprecation warnings (from numpy, python, h5py)
- [x] deprecation warnings from silx (testing legacy API?)
- [x] runtime warnings

...that would be worth cleaning-up.

Example:
https://travis-ci.org/silx-kit/silx/jobs/538142935

Also from debian CI (initially raised by @picca https://github.com/silx-kit/silx/issues/2605#issuecomment-496455219):
https://salsa.debian.org/science-team/silx/-/jobs/183495

t20100:
PR #2610, #2611,  #2612, #2614 and #2617 should address all deprecation warnings and most (but not all) irrelevant runtime warnings.

So this can be closed once all those PRs are merged.

PR #2612, opened 2019-05-28
Remove silx deprecation

t20100:
This PR removes deprecation warning from `silx` while running the tests.

related to #2609


payno:
LGTM

PR #2625, opened 2019-06-07
[OpenCL] CSR utilities

pierrepaleo:
This PR aims at bringing sparse matrix compaction/decompaction in the Compressed Sparse Row (CSR) format with OpenCL.

Work in progress:
- [x] CSR compaction ("dense" array -> sparse array)
- [x] CSR decompaction (sparse array -> dense array)

Questions:
  - Do we implement generic matrix-vector, matrix-matrix in this processing ?
  - Do we include more specialized code (ex. decompaction of `op(csr1, csr2)` where `op` is an elementwise operation) ?

### DIFF
--- a/silx/opencl/sparse.py
+++ b/silx/opencl/sparse.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2019 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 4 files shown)


kif:
On my side I am refactoring the CSR multiplication in pyFAI so that it performs the sparse matrix multiplication simultaneously  on the signal and on the normalization.

pierrepaleo:
Sparse matrix-vector product is available in PyFAI and will be integrated in silx in another PR.

silx/opencl/sparse.py
```diff
@@ -0,0 +1,319 @@
...
+"""Module for data sparsification on CPU/GPU."""
+
+from __future__ import absolute_import, print_function, with_statement, division
+
+__authors__ = ["P. Paleo"]
+__license__ = "MIT"
+__date__ = "07/06/2019"
+
+import numpy as np
+import pyopencl.array as parray
```
kif: While there are no "official" guidelines, most of the silx code uses numpy with 5 letters. This could be done in a separated PR but it would be cool to converge to a unified code style.

silx/resources/opencl/sparse.cl
```diff
@@ -0,0 +1,40 @@
+kernel void densify_csr(
```
kif: License is missing + doc + list of compile time directives

silx/opencl/sparse.py
```diff
@@ -0,0 +1,319 @@
...
+    def check_sparse_arrays(self, arrays):
+        """
+        Check that the provided sparse arrays are compatible with the current
+        context.
+
+        :param arrays: tuple of numpy.ndarray or pyopencl.array.Array
+            Tuple of length 3. It contains the arrays "data", "indices", "indptr"
+        """
+        assert len(arrays) == 3
+        data, ind, indptr = arrays
```
kif: What about using a namedtuple for this ?

silx/opencl/sparse.py
pierrepaleo: OK thanks for the tip. While I find this guideline debatable (it is not uncommon to find `np` for numpy in Python codebases), I will try to stick to it in the future.

silx/opencl/sparse.py
pierrepaleo: In the API, I would prefer using simple data structures for user convenience. But we might use named tuple in the internal mechanisms - is this what you suggest ?

kif:
> But we might use named tuple in the internal mechanisms - is this what you suggest ?

Yes, their overhead is neglectable and makes things easier to explain
when supporting users.

kif:
Thanks for the corrections

Issue #2029, opened 2018-07-16
One error and tons of warning on PPC64le


kif:
```
WARNING:py.warnings:/pwrwork/rech/gjy/uwk35yz/anaconda3/lib/python3.6/importlib/_bootstrap_external.py:426: ImportWarning: Not importing directory /pwrwork/rech/gjy/uwk35yz/anaconda3/lib/python3.6/site-packages/mpl_toolkits: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)

WARNING:py.warnings:/pwrwork/rech/gjy/uwk35yz/anaconda3/lib/python3.6/importlib/_bootstrap_external.py:426: ImportWarning: Not importing directory /gpfs16l/pwrwork/rech/gjy/uwk35yz/anaconda3/lib/python3.6/site-packages/mpl_toolkits: missing __init__
  _warnings.warn(msg.format(portions[0]), ImportWarning)

WARNING:run_tests:Test silx 0.9.0-dev0 from /gpfs16l/pwrhome/rech/grpgen/uwk35yz/workspace/silx/build/lib.linux-ppc64le-3.6/silx
WARNING:silx.test:silx.sx tests disabled Skipped by command line
WARNING:py.warnings:/gpfs16l/pwrhome/rech/grpgen/uwk35yz/workspace/silx/build/lib.linux-ppc64le-3.6/silx/io/utils.py:86: DeprecationWarning: invalid escape sequence \*
  """

WARNING:py.warnings:/pwrwork/rech/gjy/uwk35yz/anaconda3/lib/python3.6/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)

WARNING:py.warnings:/gpfs16l/pwrhome/rech/grpgen/uwk35yz/workspace/silx/build/lib.linux-ppc64le-3.6/silx/io/configdict.py:150: DeprecationWarning: invalid escape sequence \,
  sstr = sstr.replace("\,", ",").replace("^@", ",")

WARNING:py.warnings:/gpfs16l/pwrhome/rech/grpgen/uwk35yz/workspace/silx/build/lib.linux-ppc64le-3.6/silx/io/configdict.py:194: DeprecationWarning: invalid escape sequence \,
  if sstr.count(",") == sstr.count("\,"):

[...]

PR #2626, opened 2019-06-11
Uses a more precise value to ensure warning is raised

### LINKED ISSUE #2606: [debian][0.10.1] error during build
Hello, I have uploaded silx into Debian experimental but I get error on the arm64 architecture.

https://buildd.debian.org/status/fetch.php?pkg=silx&arch=arm64&ver=0.10.1%2Bdfsg-1%7Eexp1&stamp=1559037411&raw=0

t20100:
This PR uses a value to test conversion that does not fit even in a 128 bits mantissa to make sure the warning is raised with any kind of `float128` dtype (which is not necessarily a padded 80 bits float depending on the platform e.g., ppc64, arm64).

closes #2606

### DIFF
--- a/silx/utils/test/test_number.py
+++ b/silx/utils/test/test_number.py
@@ -148,7 +148,8 @@ class TestConversionTypes(testutils.ParametricTestCase):
         self.skipIfFloat80NotSupported()
         if pkg_resources.parse_version(numpy.version.version) <= pkg_resources.parse_version("1.10.4"):
             self.skipTest("numpy > 1.10.4 expected")
-        value = "1000000000.00001013332"
+        # value does not fit even in a 128 bits mantissa
+        value = "1.0340282366920938463463374607431768211456"
         func = testutils.test_logging(number._logger.name, warning=1)
         func = func(number.min_numerical_convertible_type)
         dtype = func(value)


kif:
Do you still want to test on an ARM64 ?

kif:
I resurecting my pine64, I am running the tests...

kif:
For info, this test
silx.utils.test.test_number.TestConversionTypes.testLosePrecisionUsingFloat80

passes on aarm64 prior to the patch. It also passes after the patch. Note for the release, the default version of FabIO from debian9 is not compatible anymore.

kif:
Final note:
numpy.finfo(numpy.float128) returns

aarm64: finfo(resolution=1e-33, min=-1.18973149536e+4932, max=1.18973149536e+4932, dtype=float128)
x86_64: finfo(resolution=1e-18, min=-1.18973149536e+4932, max=1.18973149536e+4932, dtype=float128)
ppc64le finfo(resolution=1e-31, min=-1.79769313486e+308, max=1.79769313486e+308, dtype=float128)

Conclusion: long double are messy!

t20100:
Thanks for the tests!

t20100:
related to #2029

PR #2631, opened 2019-06-17
[plot] scatter item shared Delaunay triangulation/interpolator

### LINKED ISSUE #2594: [plot] Avoid redundent delaunay computation in scatter items
In `silx.plot`, both the scatter item "solid" visualization and the scatter profile are computing delaunay triangulation... while this could be done once and stored in the scatter item.

t20100:
This PR moves the management of Delaunay and interpolator into plot scatter item, so that it is shared by both rendering and profile to avoid duplicated long computation.

This implementation is based on Python's Future and a "greedy" thread pool executor.

closes #2594

### DIFF
--- a/silx/gui/plot/_utils/delaunay.py
+++ b/silx/gui/plot/_utils/delaunay.py
@@ -38,35 +38,25 @@ import numpy
 _logger = logging.getLogger(__name__)
 
 
-def triangulation(x, y, dtype=numpy.uint32):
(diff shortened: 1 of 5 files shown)


t20100:
This PR as well as master has issues with `scipy` > 1.1.0 for the profile...

t20100:
Issue is related to macOS (old) version and scipy wheels (see https://github.com/scipy/scipy/issues/9751)

payno:
LGTM

PR #2644, opened 2019-07-01
snap scatter profile to points

### LINKED ISSUE #2643: [ScatterView] Show actual point coordinates
The data field is filled when we are close to a data point. When that happens, the X and Y should be those of the point and not those of the mouse position.

t20100:
This PR snaps `ScatterView` profile to points: if there is a `Data` and `Index` displayed, then the `X` and `Y` info is that of the data point rather than that of the mouse. Otherwise, the mouse coordinates is used.

When the scatter plot is displayed as a surface, the snapping is with the scatter plot point closest to the mouse when inside the convex hull.

closes #2643


payno:
LGTM

t20100:
PR #2648 should fix the CI issue with `PySide2`

Issue #2664, opened 2019-07-04
[FitWidget] dismiss pushButton on FitWidget


physwkim:
After click the dismiss button on FitWidget to close it, then open the window again, the FitWidget is not visible properly(empty widget). However, if I close it by using the x button on the upper right, it work ok.

1. working properly, before dismiss button
![before_dismiss](https://user-images.githubusercontent.com/51432966/60642180-ea3afd80-9e69-11e9-8fae-aeb1e51bf93f.png)

2. nothing on the widget after re-open fitwidget
![after_dismiss](https://user-images.githubusercontent.com/51432966/60642187-ec9d5780-9e69-11e9-83e6-1549c68e230e.png)

PR #2665, opened 2019-07-04
Fix issue reopening FitWidget after dismiss (issue #2664)

steveksw:
Dear silx team,

Please refer to issue #2664

If FitWidget is closed with the dismiss button, nothing is displayed on the screen the next time you open it again. The problem seems to be solved by using the method of deleting and re-creating the window.


t20100:
Thanks for the PR!

By fixing it this way, you do not keep the state of the `FitWidget` when it is hidden (as it is destroyed), and as it changes the `PlotToolAction` base class, it modifies the behavior of multiple actions.

I made an alternative PR #2667 where the fix is in `FitAction` and keeps the state when hidden.

Closing as superseded  by #2667

Issue #1272, opened 2017-10-08
[debian8 package] app issue ?


payno:
Someone at icalepcs try to install silx from the debian8 package with root rights and he the executable (silx view, six convert...) where not correctly installed.

I made the installation using a clone of project.
Didn't had enough time to confirm the issue but a check on the installation of the executable would be nice to be sure.

kif:
On Sun, 08 Oct 2017 13:33:20 +0000 (UTC)
payno <notifications@github.com> wrote:

> Someone at icalepcs try to install silx from the debian8 package with root rights and he the executable (silx view, six convert...) where not correctly installed.
> 
> I made the installation using a clone of project.
> Didn't had enough time to confirm the issue but a check on the installation of the executable would be nice to be sure.

I just checked the packages and the look OK.
Did you install silx or python3-silx ? 
The former contains the scripts for installation which are used by the
later. python2-silx is not used by silx executable.

Cheers,
-- 
Jérôme Kieffer
tel +33 476 882 445

vasole:
Remember we can start the program without the scripts. From the code camp slides:

![image](https://user-images.githubusercontent.com/5636251/31327560-647829c0-acd0-11e7-865d-4561f2eb9e1a.png)

t20100:
I just installed silx (not only python*-silx) on my debian 8 machine and it works fine.

vasole:
As far as I have understood, the script is tied to a particular version of the silx python library.

If one installs the library and not the script, one has to use the python(3) -m silx approach.

t20100:
The scripts are in a `silx` package which uses `python3-silx`.
If only `python*-silx` is installed, yes, the corresponding `python* -m silx` should be used.

payno:
Ok that it.
He followed the instructions from the documentation,
which goes for 
```
sudo apt-get install python-silx python3-silx
```
Maybe since we have executable a warning/note should be added ?

vasole:
Why? If one installs silx as apt-get install silx the dependencies should be there.

vasole:
I think if the guy followed the documentation, then the documentation should follow the silx pacakge and the ways to start the application written above.

PR #2669, opened 2019-07-05
Testing if pyopencl can be used on the machine after importation

alexmarie78:
Hello,

I am trying to deal with the issue number #1272 of silx-kit/pyFAI.
We get this : 

```LogicError: clGetPlatformIDs failed: PLATFORM_NOT_FOUND_KHR```

when using pyopencl even if  it is not possible to use it.
Thus I tried to fix this issue with a test being ran just after the importation of pyopencl.
Tell me what you think about it.

Cheers,

Alex

### DIFF
--- a/silx/opencl/common.py
+++ b/silx/opencl/common.py
@@ -56,6 +56,15 @@ if os.environ.get("SILX_OPENCL") in ["0", "False"]:
 else:
     try:
         import pyopencl
+        if pyopencl:
+            try:
+                pyopencl.get_platforms()
+            except pyopen._cl.LogicError:
+                logger.warning("The module pyopenCL has been imported but can't be used here")
+                pyopencl = None
+            else:
+                pass
+
     except ImportError:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by t20100** [comment]
Hi,

Thanks for the PR.

I would be more readable if you add the `get_platforms` test in a `else` after the `try` `except` block.
Also is it best to use `pyopencl.LogicError` rather than the private package `_cl`.

Something like:
```
     try:
         import pyopencl
     except ImportError:
         logger.warning("Unable to import pyOpenCl. Please install it from: http://pypi.python.org/pypi/pyopencl")
         pyopencl = None
     else:
         try:
             pyopencl.get_platforms()
         except pyopencl.LogicError:
             logger.warning("The module pyopenCL has been imported but can't be used here")
             pyopencl = None
```

alexmarie78:
Thanks for your answer, 
I'm on it, I'll push it when it is done.
Cheers

PR #2670, opened 2019-07-05
Add build of Debian 10 packages.

t20100:
This PR adds configuration for building Debian10 packages based on current Debian packaging of 0.10.1 in experimental (http://deb.debian.org/debian/pool/main/s/silx/silx_0.10.1+dfsg-1~exp1.debian.tar.xz).

Note that the configuration provided in this PR builds packages only for Python3.


pierrepaleo:
This looks as cryptic as debian packaging can be, let us see if Bob is happy with this.

PR #2671, opened 2019-07-05
[Opencl] CSR: add support for data types other than float32

pierrepaleo:
This PR enables to use the GPU CSR utilities on other data types (float64, (u)int8,16,32,64).
Complex and half float are not supported as data types.  
Due to a pyopencl limitation in `GenericScanKernel`, indices data type must have its size divisible by 4 bytes.


kif:
Looks good !

PR #2674, opened 2019-07-09
[plot] Defer addition of dock widgets to main window

### LINKED ISSUE #1629: [plot] Docked widgets always shown at bottom when instanstiated
We seem to have lost the functionality of showing docked widgets to the right of the plot when the plot was much wider than tall.

t20100:
This PR adds `PlotWindow` dock widgets to dock area when they become visible rather than when they are created.
This allows to take the window size into account to decide where to put them

closes #1629


payno:
LGTM

PR #2678, opened 2019-07-11
Add a trilinear interpolator

t20100:
This PR adds a `silx.math.interpolate` module with a `interp3d` function for trilinear interpolation in a regular equally spaced grid.

This functions is similar to [`scipy.interpolate.interpn`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interpn.html#scipy.interpolate.interpn) except that the grid is equally spaced and it only handles 3D dataset.
It provides almost a x10 speed-up over `interpn` and almost x50 with OpenMP * (with `method=linear_omp`).

I need it for #2323

*: On my machine

### DIFF
--- a/silx/math/interpolate.pyx
+++ b/silx/math/interpolate.pyx
@@ -1,0 +1,165 @@
+# coding: utf-8
+# /*##########################################################################
+#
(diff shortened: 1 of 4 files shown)


pierrepaleo:
LGTM

PR #2636, opened 2019-06-17
Stats roi

payno:
#2684 should be merged before
The goal of this PR is to provide a convenient way to compute statistics on specific Region of interest.

Steps:
- [x] adapt the stats handler to manage roi
- [x] add a mask system to filter values in a specific roi
   - [x] test on CurveContext
   - [x] test on Imagecontext
   - [x] test on 2D ScatterContext
   - [x] test on Histogram context
   - [ ] ~test on 3D ScatterContext~ -> not managed, would need a 3d roi or an axis at leasr
   - [ ] ~test on 3D _plot3DArrayContext~ -> same as ScatterContext
- [x] add a widget to display the result for a roi and a set of stats.
   - how to present this to the user ? Make roi selection possible for a plot and display all the stats or present only one stat and all the roi, or create a filter with a set of roi and a set of stats ? Is the roi editable box available ? -> this is PR #2754
- [x] add isIn function in _RegionOfInterestBase
   - add unit test
- [x] add an example of usage
- [x] doc
- [x] speed up statistic calculation (especially calls to isIn and neighbouring functions)
   - [x] better filtering when event is 'DataChanged'
   - [x] keep contexts for (item, plot, roi, on_limits) to keep mask and avoid there avoid it recalculation when possible.

This feature requires:

- [x] minimalistic common API between ROI class and RegionOfInterestClass (see PR: #2684)
- [x] a new signal 'RegionEditionFinished' for RegionOfInterestClass ? Overwise will always update the stats. Might also be managed as plot and the '_dirty' flag ? See issue #2685 , but might be complicated.

#2724 and #2754 need to be merged before

:warning: The goal of this PR is to provide the feature of roi and stats as a first steps.
Here we are looking for high performances. If this feature is commonly use then we should look to speed up performances. Some ideas:
* we can add some filter in the isIn function to first look if a bounding box contains the pixel... before calling more complex operations.
* `isIn` functions could be done in Cython

:warning:
`isIn` functions for image context: for now have basically a binary approach, the pixel is in a roi or not. To have a fully 'correct behaviour' we should add a weight on stats calculation to moderate the pixel value.

:warning:
behaviour of scale images: now the statistics are computed on the entire data array (there is no sampling in the array or interpolation regarding the scale). This also mean that the result can differ from what is displayed. But I guess there is no perfect behaviour.

PR #2684, opened 2019-07-24
add a common base class for ROI and regionOfInterest

payno:
needed by #2636


payno:
Maybe we can have both signals, one 'generic' `sigItemChanged` and one `sigItemNameChanged` ?
Or over-engineered ?

t20100:
I remind we considered at some point to have ROI being plot items, so this point to  `sigItemChanged`.
I'm merging as it is as it's been standing opened for long, QObject inheritance and signal can be done in another PR.

PR #2692, opened 2019-07-30
h5todict asarray=True argument

### LINKED ISSUE #2689: h5todict always returns arrays even for scalar
For single values, `h5todict` returns `numpy.ndarray`.

To reproduce:
```
from silx.io.dictdump import dicttoh5, h5todict
dicttoh5({'test': 1.}, 'test.h5')

In [19]: h5todict('test2.h5')                                                   
Out[19]: {'test': array(1.)}
```

The reason is the use of `[...]` rather than `[()]` here: https://github.com/silx-kit/silx/blob/master/silx/io/dictdump.py#L315

```
In [16]: f2 = h5py.File('test2.h5', 'r')    
In [23]: f2['test'][()]                                                         
Out[23]: 1.0
In [24]: f2['test'][...]                                                        
Out[24]: array(1.)
```

To me, it is best to returns scalar in this case... yet this will break the behaviour of `h5todict`.
Adding an `asarrays=True` argument to `h5todict` would allow to keep the legacy behavior and enable the "correct" way of handling it.

Do we want to break the behaviour?

t20100:
This PR adds an `asarray` argument to `h5todict` which is `True` by default and keeps the legacy behavior, but when set to `False` provides the behavior that for me should have been the only one: scalars are converted to scalars in the output `dict` rather than arrays.

closes #2689

### DIFF
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 # /*##########################################################################
-# Copyright (C) 2016-2018 European Synchrotron Radiation Facility
+# Copyright (C) 2016-2019 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


payno:
LGTM, Thanks for documentation of the related issue by the way

PR #2697, opened 2019-08-01
silx.gui.utils.blockSignals helper

t20100:
This PR adds a `blockSignals` context manager that allows to block signals for one or many `QObject`s.

This is not much code, but it allows to replace:
```
previous = obj.blockSignals(True)
# do stuff
obj.blockSignals(previous)
```
with:
```
with blockSignals(obj):
    # do stuff
```
which makes it more readable.

### DIFF
--- a/silx/gui/qt/_utils.py
+++ b/silx/gui/qt/_utils.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2004-2017 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2019 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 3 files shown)


vallsv:
Sounds nice.

BTW, what about a module `silx.gui.qtutils` for that?

It would make the import easier.

```
from silx.gui import qt
from silx.gui import qtutils
```

t20100:
Initially I put it in `silx.gui.qt`, there is already a few helpers there in `silx.gui.qt._utils`, what do you think about that? import is even simpler... but it mixes helpers with Qt bindings.

t20100:
skip tests with Qt4 (where `QObject.objectNameChanged` is not available)

PR #2698, opened 2019-08-01
Replace pyopencl.array.zeros with empty + fill

### LINKED ISSUE #2696: [OpenCL] Avoid the usage of pyopencl.array.zeros
This command triggers the usage of enqueue_fill_buffer in opencl which is not (yet) properly implemented in the cuda backend of PortableCL:
https://github.com/pocl/pocl/issues/725

kif:
close #2696 

All OpenCL tests are passing on x86-64+PoCL but the one requiring textures ... which is known unsupported.


pierrepaleo:
Looks good, tests are passing on my machine as well.

Issue #2683, opened 2019-07-18
Image parameters maintained upon changing dataset


sjleake:
Hi,
When loading an h5 file with many datasets, when inspecting a 2D slice and then changing to a another dataset to compare, if the datasets are the same dimension, it would be useful if the image or stackview maintain the chosen slice.
Thanks

t20100:
PR #2701 provides part of it (i.e., it works for all views but the `StackView` which needs specific handling)

Obi-Wan:
Any news on this? It seems that PR #2701 was merged, but this was not resolved. Viewing stacks of images is quite a useful application.

t20100:
For image stacks, you can use the "Image" tab which keeps the selected slice when switching datasets.
The "Image stack" tab which is also available does not supports this.
Unless I missed something, the "Image stack" tab provides the same features as the "Image" one, so we'd better remove it (2 tabs providing the same feature is 1 too much).

Obi-Wan:
I agree that the redundant tab should be removed as it can be confusing.
On the other hand, I have noticed that for 3D volumes, the labels of the axes differ a bit between the two tabs. In the image tab you have `{"", "x", "y"}` while in the stack tab `{"depth", "x", "y"}`. The empty label in the image tab can be confusing as well. So, before you do away with the stack tab, please give the dimension a label different from `""`. May I suggest `"z"`? :)

PR #2701, opened 2019-08-01
silx view: keep the same axes selection whenever possible

t20100:
This PR makes the data viewer (and so `silx view`) keeping the same selection of axes when switching between datasets which have the same shape. It also keeps the selection between views with similar dimensions for the same dataset (e.g., between image view and array view).

This is changing quite a bit the internals which are not that simple, so it needs a serious look at it.

related to #2683, but not closing it: (NX)StackView need specific care and this PR is not addressing it.

### DIFF
--- a/silx/gui/data/DataViewer.py
+++ b/silx/gui/data/DataViewer.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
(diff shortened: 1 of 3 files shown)


t20100:
Tests should pass, ready for review

vallsv:
LGTM

vallsv:
As an improvement, I think it would be good to display somewhere on the numpy axes selector, that the axes was costumed, and an easy way to reset to default (a kind of warn). Just a proposal.

PR #2715, opened 2019-08-09
[Plot] add "baseline" option in addCurve

### LINKED ISSUE #2714: [plot] baseline parameter require
It would be nice to have a baseline parameter to display for example for a stack of 1D value the mean value and the standard deviation

something like:

Screenshot_2019-08-09_16-46-58

payno:
This is the PR related to #2714 to add a baseline parameter.

this include
- [x] matlab backend
- [x] opengl backend
- [x] unittest
- [x] example of usage -> not sure this is mandatory

If such a parameter is added then we could reconsider the usage of the 'fill parameter'.
/close #2714 

small script to test, for now fails in the addHistogram, with the current values:
``` python
from silx.gui import qt
import numpy

qapp = qt.QApplication([])

x = numpy.arange(0, 10, step=0.1)
my_sin = numpy.sin(x)
y = numpy.arange(-4, 6, step=0.1) + my_sin
mean = numpy.arange(-5, 5, step=0.1) + my_sin
baseline = numpy.arange(-6, 4, step=0.1) + my_sin
edges = x[y>=3.2]
histo = mean[y>=3.2] - 1.8

# opengl plot
plot_ogl = Plot1D(backend='opengl')
plot_ogl.addCurve(x=x, y=y, baseline=baseline, color='grey', legend='std-curve', fill=True)
plot_ogl.addCurve(x=x, y=mean, color='red', legend='mean')
plot_ogl.addHistogram(histogram=histo, edges=edges, color='red', legend='mean2', fill=True)

# matplotlib plot
plot_mtl = Plot1D()
plot_mtl.addCurve(x=x, y=y, baseline=baseline, color='grey', legend='std-curve')
plot_mtl.addCurve(x=x, y=mean, color='red', legend='mean')
plot_mtl.addHistogram(histogram=histo, edges=edges, color='red', legend='mean2', fill=True)

plot_mtl.show()
plot_ogl.show()

qapp.exec_()
```

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -848,7 +848,7 @@ class PlotWidget(qt.QMainWindow):
             curve.setColor(default_color)
             curve.setLineStyle(default_linestyle)
             curve.setSymbol(self._defaultPlotPoints)
-            curve.setBaseline(baseline=baseline)
+            curve._setBaseline(baseline=baseline)
 
         # Do not emit sigActiveCurveChanged,
         # it will be sent once with _setActiveItem
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -987,10 +987,8 @@ class PlotWidget(qt.QMainWindow):
             histo.setZValue(z=z)
 
         # Set histogram data
-        histo.setData(histogram, edges, align=align, copy=copy)
-        # set baseline
-        if baseline is not None:
-            histo.setBaseline(baseline=baseline)
+        histo.setData(histogram=histogram, edges=edges, baseline=baseline,
+                      align=align, copy=copy)
 
         if mustBeAdded:
             self._add(histo)
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -710,7 +710,8 @@ class PlotWidget(qt.QMainWindow):
                  xlabel=None, ylabel=None, yaxis=None,
                  xerror=None, yerror=None, z=None, selectable=None,
                  fill=None, resetzoom=True,
-                 histogram=None, copy=True):
+                 histogram=None, copy=True,
+                 baseline=None):
         """Add a 1D curve given by x an y to the graph.
 
         Curves are uniquely identified by their legend.
@@ -845,6 +846,7 @@ class PlotWidget(qt.QMainWindow):
             curve.setColor(default_color)
             curve.setLineStyle(default_linestyle)
             curve.setSymbol(self._defaultPlotPoints)
+            curve.setBaseline(baseline=baseline)
 
         # Do not emit sigActiveCurveChanged,
         # it will be sent once with _setActiveItem
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -927,6 +927,7 @@ class PlotWidget(qt.QMainWindow):
                      align='center',
                      resetzoom=True,
                      copy=True,
+                     z=None,
                      baseline=None):
         """Add an histogram to the graph.
 
@@ -958,6 +959,7 @@ class PlotWidget(qt.QMainWindow):
         :param bool resetzoom: True (the default) to reset the zoom.
         :param bool copy: True make a copy of the data (default),
                           False to use provided arrays.
+        :param int z: Layer on which to draw the histogram
         :param int baseline: baseline value
         :returns: The key string identify this histogram
         """
@@ -978,6 +980,8 @@ class PlotWidget(qt.QMainWindow):
             histo.setColor(color)
         if fill is not None:
             histo.setFill(fill)
+        if z is not None:
+            histo.setZValue(z=z)
 
         # Set histogram data
         histo.setData(histogram, edges, align=align, copy=copy)
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -891,7 +891,7 @@ class PlotWidget(qt.QMainWindow):
             if len(x) > 0 and isinstance(x[0], dt.datetime):
                 x = [timestamp(d) for d in x]
 
-            curve.setData(x, y, xerror, yerror, copy=copy)
+            curve.setData(x, y, xerror, yerror, baseline=baseline, copy=copy)
 
         if replace:  # Then remove all other curves
             for c in self.getAllCurves(withhidden=True):
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -926,7 +926,8 @@ class PlotWidget(qt.QMainWindow):
                      fill=None,
                      align='center',
                      resetzoom=True,
-                     copy=True):
+                     copy=True,
+                     baseline=None):
         """Add an histogram to the graph.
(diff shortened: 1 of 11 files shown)


silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -402,9 +402,13 @@ def addCurve(self, x, y, legend,
                                    s=symbolsize**2)
             artists.append(scatter)
 
-            if fill:
+            if fill or baseline is not None:
```
t20100: I would keep `fill` and `baseline` independent, i.e., if `fill` is `False`, then it is not filled even if there is a `baseline`.

silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -417,9 +421,13 @@ def addCurve(self, x, y, legend,
                                   markersize=symbolsize)
             artists += list(curveList)
 
-            if fill:
+            if fill or baseline is not None:
```
t20100: Same as above

silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -943,6 +943,9 @@ def addCurve(self, x, y, legend,
         if selectable:
             behaviors.add('selectable')
 
+        fillColor = None
+        if fill is True or baseline is not None:
```
t20100: Same as above

silx/gui/plot/backends/glutils/GLPlotCurve.py
```diff
@@ -124,16 +124,20 @@ def __init__(self, xData=None, yData=None,
         self.offset = offset
 
         # Offset baseline
-        self.baseline = baseline - self.offset[1]
+        if not isinstance(baseline, numpy.ndarray):
+            self._base_is_a_single_pt = True
```
t20100: This attribute can be removed, it is not used

silx/gui/plot/items/curve.py
```diff
@@ -169,6 +169,8 @@ class Curve(PointsBase, ColorMixIn, YAxisMixIn, FillMixIn, LabelsMixIn, LineMixI
     _DEFAULT_HIGHLIGHT_STYLE = CurveStyle(color='black')
     """Default highlight style of the item"""
 
+    _DEFAULT_BASELINE = None
```
t20100: This is not used.

t20100:
For the OpenGL implementation, did you make it work?
The approach needs to be completely changed when the baseline is no longer a single value.

vallsv:
If this features is a kind of way to create stacked histogram?

payno:
> If this features is a kind of way to create stacked histogram?

was not the original idea, but I guess we could have something like that from it. But would probably require a dedicated api or some thinking around the existing one.
Is it a need ?

payno:
> For the OpenGL implementation, did you make it work?
> The approach needs to be completely changed when the baseline is no longer a single value.

I make it work on my case, but there is some code for duplicating the last point (connecting degenerated triangle) that I don't fully get and is currently removed.
Maybe you also have some 'corner cases' to test and that can be interesting.

t20100:
As now the order of rendering is that of addition of items to the plot, I don't think you need to use the baseline to render stacked histograms.

Stacked histogram is simple to do (a cum sum of the values of the histograms) and you provide it in the examples, so to me it does not need to be added to `silx`.

silx/gui/plot/PlotWidget.py
```diff
@@ -955,6 +959,8 @@ def addHistogram(self,
         :param bool resetzoom: True (the default) to reset the zoom.
         :param bool copy: True make a copy of the data (default),
                           False to use provided arrays.
+        :param int z: Layer on which to draw the histogram
+        :param int baseline: baseline value
```
t20100: `baseline` can also be a numpy.ndarray

silx/gui/plot/items/histogram.py
```diff
@@ -297,6 +301,20 @@ def setData(self, histogram, edges, align='center', copy=True):
...
 
+    def setBaseline(self, baseline):
+        if isinstance(baseline, numpy.ndarray):
+            self._baseline = numpy.empty(baseline.shape[0] * 2)
+            for i_value, value in enumerate(baseline):
+                self._baseline[i_value*2:i_value*2+2] = value
+        else:
+            self._baseline = baseline
+
+    def getBaseline(self, copy=True):
```
t20100: The docstring is missing.

Defining a `BaselineMixIn` class to share the baseline implementation would allow to share the implementation, provide a fixed way of checking if the item has a baseline with `isinstance` and be consistent with the other items.

Also it doesn't cost much and is safer to always return a new array: `return numpy.array(self._baseline, copy=copy)

silx/gui/plot/items/histogram.py
```diff
@@ -297,6 +301,20 @@ def setData(self, histogram, edges, align='center', copy=True):
 
         self._updated(ItemChangedType.DATA)
 
+    def setBaseline(self, baseline):
```
t20100: Same as for `Curve`, this should be merged with set data, or you have to deal with curve and baseline of different length when updating...

silx/gui/plot/items/histogram.py
```diff
@@ -297,6 +301,20 @@ def setData(self, histogram, edges, align='center', copy=True):
...
+        if isinstance(baseline, numpy.ndarray):
+            self._baseline = numpy.empty(baseline.shape[0] * 2)
+            for i_value, value in enumerate(baseline):
+                self._baseline[i_value*2:i_value*2+2] = value
```
t20100: Not sure I get this.
To me the user should input as many baseline points as bins (and retrieve the same from `getBaseline`).
That you have to twice the number of point for rendering should not be visible from the user.

silx/gui/plot/items/curve.py
```diff
@@ -321,6 +326,15 @@ def setHighlightedColor(self, color):
         """
         self.setHighlightedStyle(CurveStyle(color))
 
+    def setBaseline(self, baseline):
```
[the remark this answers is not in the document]
t20100: The baseline needs to be converted to a `numpy` array, you don't want it to be a list or tuple afterwards.

t20100:
I put a few comments in the review

t20100:
There is still a small issue with `setBaseline`, one can provide an array with a different length as the data. Maybe this method can be private.
Also, the `Curve` does not have a `baseline` argument in `setData`.

vallsv:
BTW, what means a curve with a baseline? As a curve is supposed to be a line, there is maybe an inconsistency problem with the picking API. Then that's maybe not a curve, but another object which have a surface.

payno:
From my point of view a baseline is more a 'property' of a curve. Having a specific object would complexity the code and the usage for no reason. (in the case we want to first have a curve, then have an associated baseline). It is also how most 'ploting' libraries work I guess.

Anyway it would make sense to keep consistency. We can have a `setData` with the baseline and the `getData` which does not return the baseline. In this case the user would have to specifically call the `getBaseline` function. But the `setData` and the `getData` would not be 'homogeneous'... :confused:

payno:
After a discussion with @t20100, second option is the best one. We won't modify getData, only setData with a baseline parameter and a default value set to None.

t20100:
LGTM, @payno is there anything else to do on it?

PR #2716, opened 2019-08-09
Display title in the type field.

### LINKED ISSUE #2713: [silx view] show NXentry title
Could Silx show the NXentry title in the viewer? NXentry names are typically chosen for easy access by software (e.g. just a number) but are not very informative for humans.

vasole:
closes #2713

image

### DIFF
--- a/silx/gui/hdf5/Hdf5TreeModel.py
+++ b/silx/gui/hdf5/Hdf5TreeModel.py
@@ -511,6 +511,22 @@ class Hdf5TreeModel(qt.QAbstractItemModel):
         if index.column() == self.NAME_COLUMN:
             return node.dataName(role)
         elif index.column() == self.TYPE_COLUMN:
+            if node.dataType(role) == "NXentry":
+                if node.hasChildren():
+                    titleFound = False
+                    idx = 0                    
+                    while idx < node.childCount():
+                        item = node.child(idx)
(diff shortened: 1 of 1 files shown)


vallsv:
What about changing the description name, instead of the node name, it is also part of the tree? I think it would be really better.

I mean, it's like having custom magic filename and directory in your file system... Hmmm somebody just told me that Windows is doing that...

vasole:
The problem is the description filed is too far to the right.

vasole:
As it is now, the node name is not changed, only the type displayed.

I considered that to change the node name would be too risky

vasole:
Sorry, the value is what is too far to the right.

image

vallsv:
Also cause your change grow up the type-column ;-)

Value could be a better place, shape could be moved after value.

BTW, in case, you can reorder columns with the mouse, and hide/display things with the right mouse.

vasole:
It seems there is a problem with PySide2.

t20100:
Ideas from a brief discussion with @vasole:
- What about using the `Value` column for `NXEntry` title and moving the `Value` column to be the 2nd one.
- Alternative: Add a `Description` column as 2nd column which content varies according to the dataset/group: title for `NXEntry`, NXClass for other nexus groups, value for scalars, dtype+shape for other datasets...

t20100:
closing as superseded by PR #2804

PR #2719, opened 2019-08-14
[Scatter Mask] Prevent unhandled exception

vasole:
Attempting to disconnect a non-connected method raises a type error.

According to this old post, there is no way to know it the slot was connected or not

https://forum.qt.io/topic/6667/i-strongly-need-to-know-slot-where-signal-is-connected-to/11


t20100:
Yes.

For similar case, at other place in the code, we use:
`except (RuntimeError, TypeError):`

Example:
https://github.com/silx-kit/silx/blob/689ba07e734e8c5d8bbfbc449155f72f665f6c84/silx/gui/plot/MaskToolsWidget.py#L369-L373

But I don't remember which binding raises `RuntimeError`.

t20100:
Checking for that, `PySide2` raises `RuntimeError`, so it also needs to be catched.

t20100:
Fixed for PySide2.

LGTM

PR #2724, opened 2019-08-23
[PlotWindow] fix call to getStatsWidget.

payno:
When we try to call the getStatsWidget() the _statsDockWidget.widget() is None. This came from the merge of PR #2674 because `self._statsDockWidget.setWidget(statsWidget)` need to be set before any other call to `getStatsWidget()` which is made by getStatsAction() and bring some incoherence.

this PR fix it


vallsv:
LGTM

PR #2726, opened 2019-09-03
Clean-up some TODOs about plot event creation

t20100:
This PR remove/fix TODOs that where introduced in PR #2602 about plot events content.


vallsv:
LGTM

PR #2728, opened 2019-09-03
PlotWidget.pickItems returns objects

t20100:
This PR makes `PlotWidget.pickItems` and items' `pick` method returns a `PickingResult` object rather than a tuple (item, indices) or indices. This enables to provide richer picking information in the future without breaking the API and is consistent with the picking API of `plot3d`.

The `PickingResult` class of `plot3d` inherits from the new one of `plot` and extends it as more features are available for now in `plot3d` picking.


vallsv:
LGTM

PR #2729, opened 2019-09-03
Point roi symbol

t20100:
This PR changes the default symbol used for `PointROI` in `plot` back to '+', and updates the ROI example script.

Some changes made the default symbol used for `PointROI` a 'o' as a side effect.
'+' sounds better and this PR restores it as a default.

### DIFF
--- a/examples/plotInteractiveImageROI.py
+++ b/examples/plotInteractiveImageROI.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2018 European Synchrotron Radiation Facility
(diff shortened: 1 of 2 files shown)


vallsv:
What about using `_DEFAULT_SYMBOL` if you define it?

t20100:
`_DEFAULT_SYMBOL` is used by the `SymbolMixIn` class (which `PointROI` inherits from) to define its default.

t20100:
But yes, that's implicit from the inherited class point of view.

vallsv:
LGTM, i just add a comment relative to it. That's maybe point less. Anyway, it is merged.

PR #2730, opened 2019-09-04
Rework PlotWidget rendering/picking order

### LINKED ISSUE #2682: [plot] Rendering order of plot items changes depending on updates
Order of items at a same `z` level is in not maintained in the backend when items are updated.

An example, where updating overlapping curves not all at the same time makes them blink:
```
from silx import sx

plot = sx.Plot1D()
plot.show()

for index in range(3):
    plot.addCurve((1, 2, 3), (0, 0, 0), legend=str(index))

index = 0

def run():
    global index
    index = (index + 1) % 3
    plot.addCurve((1, 2, 3), (0, 0, 0), legend=str(index))

timer = sx.qt.QTimer()
timer.start(100)
timer.timeout.connect(run)
sx.qt.QApplication.instance().exec_()
```

There is a workaround, which is to set the `z` level when adding items to the plot the first time:
`plot.addCurve((1, 2, 3), (0, 0, 0), legend=str(index), z=index)`

But that would be better to always keep the order of insertion in the plot as the rendering order.
This can be achieved by decoupling the z order in `PlotWidget` from that of the backend, and specifying an explicit `z` order (i.e., each item having a different `z`) in the backend.

t20100:
This PR reworks completely the order in which the plot items are rendered and picked:
- Picking order was not consistent with rendering order after merging PR #2602 which didn't care about the item ordering on screen.
- Rendering order was not ideal because:
  -  for a given z value, the rendering order was defined by the order of updates of items rather than the order of addition to the plot.
  - for overlaid items (i.e., markers and draggable shape items), the z value was not taken into account.

Now the rendering order is defined by:
- first, whether or not the item is an overlay,
- then by the z value,
- and finally by the order of addition of items in the plot.

Rendering order of items is no longer managed by the backend, instead, the backend are asking the order to the PlotWIdget.

Picking order is the reversed order of rendering.

The adaptation is done for the matplotlib backend.
TODO: Make it for the OpenGL backend.

closes  #2682

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -2902,18 +2902,28 @@ class PlotWidget(qt.QMainWindow):
         """
         return self._getItem(kind='marker', legend=legend)
 
-    def _itemsFromBackToFront(self):
+    def _itemsFromBackToFront(self, condition=None):
         """Iterator of plot items ordered from back to front.
 
         This is the order used for rendering.
         It takes into account overlays, z value and order of addition of items
+
+        :param callable condition:
+           Callable taking an item as input and returning False for items
+           to skip.
+           If None (default), no item is skipped.
+        :rtpye: List[Item]
         """
         # Sort items: Overlays and markers first, then others
         # and in each category ordered by z and by order of addition
         # as _content keeps this order.
-        for item in sorted(self._content.values(),
-                key=lambda i: ((1 if i.isOverlay() else 0), i.getZValue())):
-            yield item
+        content = self._content.values()
+        if condition is not None:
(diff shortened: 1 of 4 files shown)


t20100:
Implemented for OpenGL as well.

t20100:
Ready for review.

The backends code can most probably be refactored/simplified, but the feature is there.

vallsv:
Looks fine.

I just notice a glitch around the rect of the view:

master:
image


your pr:
image

You can see that the image is displayed over the view rect. While the view rect have to be visible.

t20100:
Indeed... I missed that, thanks!

It's fixed now (I'm not fully happy of the fix though).

PR #2731, opened 2019-09-05
Refactoring plot backends

t20100:
Merge PR #2730 first!

Following the refactoring of picking and z order management in the plot, this PR makes some clean-up in the backends.
Indeed, some information is no longer need in the backend.
Also for the OpenGL backend, some of the code that was initially duplicated between the backends and `PlotWidget` was not removed. This PR removes some of this (e.g., checks in `dataToPixel` and `pixelToData` methods).

More rework/clean-up can be done, but that will be for another time...


vallsv:
LGTM

PR #2732, opened 2019-09-05
Plot: remove legend argument in backends

t20100:
Merge PR #2730 and #2731 first!

This PR removes the now useless `legend` argument in the plot backend.
This does NOT change the `PlotWidget` and its items, it just changes the backends.

### DIFF
--- a/silx/gui/plot/backends/BackendBase.py
+++ b/silx/gui/plot/backends/BackendBase.py
@@ -97,7 +97,7 @@ class BackendBase(object):
 
     # Add methods
 
-    def addCurve(self, x, y, legend,
+    def addCurve(self, x, y,
                  color, symbol, linewidth, linestyle,
                  yaxis,
                  xerror, yerror, z, selectable,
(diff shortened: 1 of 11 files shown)


silx/gui/plot/backends/BackendBase.py
```diff
@@ -263,7 +258,7 @@ def addMarker(self, x, y, legend, text, color,
                           and that returns the filtered coordinates.
         :return: Handle used by the backend to univocally access the marker
         """
-        return legend
+        return object()
```
vallsv: What about `raise NotImplementedError()`, `BaseBackend` is an abstract class no? It have to be implemented.

vallsv:
LGTM. Looks to be a good idea.

silx/gui/plot/backends/BackendBase.py
t20100: The `BackendBase` can be use to run the `PlotWidget` headless. This is done in the tests at some point.
Basically all methods of `BackendBase` should raise `NotImplementedError`.
Doing it this way provides a mock and the definition of the interface at once.

t20100:
Moving features from the backends to `PlotWidget` allows to simplify the backends.

silx/gui/plot/backends/BackendBase.py
vallsv: Ok.

PR #2733, opened 2019-09-05
Fixes bootstrap.py starting IPython the wrong way

t20100:
This PR fixes issues with `bootstrap.py` and IPython (e.g., defined variables not available in `globals()`).
It was using `embed` while `start_ipython` is preferable.


kif:
Works

PR #2734, opened 2019-09-06
Avoid some synchronization replot in matplotlib backend

t20100:
This PR uses the matplotlib `Axes.apply_aspect` method to synchronize plot limits rather than doing a full replot.
This avoids to replot when getting the limits of a plot with keep aspect ratio.


vallsv:
Sounds very good. As it was a big problem on plot blinking.

This could have an impact. I think it is a good idea to check pyFAI.

PR #2735, opened 2019-09-09
Fix float32

### LINKED ISSUE #2736: Precision lost when casting to float16
When calling the function ``min_numerical_convertible_type`` from ``silx.utils.number`` there is a use case where the precision gets lost and a float32 is wrongly cast to float16.
```
from silx.utils.number import min_numerical_convertible_type
c = "-23.172"
type = min_numerical_convertible_type(c)
print(type(c))
>>> -23.17
```

### LINKED ISSUE #2737: [fabioh5] Currently casting twice when converting from scalar to number
In FabioReader class, when obtaining a value from the header in ``get_value``, the value is converted to a numpy object:
```
def _convert_scalar_value(self, value):
        """Convert a string into a numpy int or float.

        If it is not possible it returns a numpy string.
        """
        try:
            numpy_type = silx.utils.number.min_numerical_convertible_type(value)
            converted = numpy_type(value)
        except ValueError:
            converted = numpy.string_(value)
        return converted
```

But this dtype is later modified in ``__normalize_vector_type``. This value should not be casted twice, because it leads to a change of the original scalar value.

The next code is a summary of what is done to convert the type in fabioh5:
```
import numpy
from silx.utils.number import min_numerical_convertible_type

c = "-23.17"
_type = min_numerical_convertible_type(c)
converted = _type(c)
if numpy.issubdtype(_type, numpy.floating):
    result_type = numpy.result_type(_type, numpy.float32)
    second_conversion = numpy.array(converted, dtype=result_type)
    print(second_conversion)
```
Result is ``-23.171875`` because of how -23.17 is saved in numpy.float16.

vallsv:
- Tune magic coef to allow to discrete float precision expected more than 16bit but less than 64bit bits
- Fix precision for scalar in many frames (1)
- Fix precision for list of numbers (2)
- Fix precision for list of numbers in many frames (3)

Add regression test.

Spotted by @juliagarriga 

It would be good to rework this code to allow to take care of the three use cases (1) (2) (3) in specialized functions. This could create a more readable code.

Closes #2736
Closes #2737

### DIFF
--- a/silx/io/fabioh5.py
+++ b/silx/io/fabioh5.py
@@ -614,12 +614,16 @@ class FabioReader(object):
         converted = []
         types = set([])
         has_none = False
+        is_array = False
+
         for v in values:
             if v is None:
                 converted.append(None)
                 has_none = True
             else:
                 c = self._convert_value(v)
+                if c.shape != tuple():
+                    is_array = True
                 converted.append(c)
                 types.add(c.dtype)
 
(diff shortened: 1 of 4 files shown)


juliagarriga:
continuation in issue #2756

PR #2750, opened 2019-09-17
Allows URL to go through filename expansion

### LINKED ISSUE #2749: [silx view] cannot open URLs
`silx.utils.files.expand_filenames` tries to match files for urls because of the `?`.
URLs should be excluded from the expansion

t20100:
This PR makes URL goes through the filename expansing function by adding filenames with magic that cannot extends.
Not sure it is the best (it would be better to check that it is a URL), but at least it keeps it simple.

closes #2749

### DIFF
--- a/silx/utils/files.py
+++ b/silx/utils/files.py
@@ -46,7 +46,11 @@ def expand_filenames(filenames):
         if os.path.exists(filename):
             result.append(filename)
         elif glob.has_magic(filename):
-            result += glob.glob(filename)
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM

vallsv:
I did not check before merging, but i expect to have an error message in case `silx view foo*` found no files. Then it's also better.

vallsv:
Thanks a lot

PR #2751, opened 2019-09-18
[math] FFTW: bugfix + avoid copies when possible

pierrepaleo:
This PR fixes a bug when using `silx.math.fft` with the FFTW backend, when we try to avoid memory copies.

### DIFF
--- a/silx/math/fft/fftw.py
+++ b/silx/math/fft/fftw.py
@@ -82,6 +82,10 @@ class FFTW(BaseFFT):
         self.set_fftw_flags()
         self.compute_forward_plan()
         self.compute_inverse_plan()
+        self.refs = {
+            "data_in": self.data_in,
+            "data_out": self.data_out,
+        }
 
 
     def set_fftw_flags(self):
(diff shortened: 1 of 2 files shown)


silx/math/fft/fftw.py
```diff
@@ -104,13 +108,6 @@ def _allocate(self, shape, dtype):
 
 
     def check_array(self, array, shape, dtype, copy=True):
-        """
```
payno: did you voluntarily remove the documentation of the function ?

pierrepaleo: The documentation was incorrect - the function does not return a new array. So I preferred to remove it.

pierrepaleo:
[From](https://github.com/pierrepaleo/silx/pull/7):
> - due to modifications there is incoherence between signature BaseFFT.set_data and FFTW.set_data, BaseFFT.fft and FFTW.fft, same for ifft
> - I removed the DEBUG lines but you can skip the specific commit if those are mandatory.
> - BaseFFT functions should be documented and FFTW functions documentation can be defined using the @Docstring decorator

The signatures are the same - only the parameters names are different. I agree it would be best if they had consistent parameters names across the classes, but these are internal functions not supposed to be called by the user.
Maybe `set_data` should be renamed as `_set_data`.

PR #2754, opened 2019-09-24
Add feedback on item darg & drop

### LINKED ISSUE #2685: [regionOfInterest] missing signal 'editingFinished' ?
It would be interesting to have an 'editingFinished' signal for RegionOfInterest.

For example in the case of computing statistics on specific plot item and region of interest, computation can be time consuming. So if we compute it for each new anchor position it will be a nightmare.
We should only recompute those when the edition is finished.
Or maybe the emission of `regionChanged` signal can be changed.

payno:
/close #2685

### DIFF
--- a/silx/gui/plot/PlotInteraction.py
+++ b/silx/gui/plot/PlotInteraction.py
@@ -1263,8 +1263,7 @@ class ItemsInteraction(ClickOrDrag, _PlotInteraction):
 
         if isinstance(item, items.MarkerBase):
             self._signalMarkerMovingEvent('markerMoving', item, x, y)
-            if item is not None:
-                item._startDrag()
+            item._startDrag()
 
         return True
 
@@ -1283,7 +1282,7 @@ class ItemsInteraction(ClickOrDrag, _PlotInteraction):
 
     def endDrag(self, startPos, endPos):
         item = None if self.draggedItemRef is None else self.draggedItemRef()
-        if item is not None and isinstance(item, items.MarkerBase):
+        if isinstance(item, items.MarkerBase):
             posData = list(item.getPosition())
             if posData[0] is None:
                 posData[0] = 1.
@@ -1435,7 +1434,7 @@ class ZoomAndSelect(ItemsInteraction):
         :return: True if click is catched by an item, False otherwise
         """
         eventDict = self._handleClick(x, y, btn)
-        
+
         if eventDict is not None:
             # Signal mouse clicked event
             dataPos = self.plot.pixelToData(x, y)
--- a/silx/gui/plot/PlotInteraction.py
+++ b/silx/gui/plot/PlotInteraction.py
@@ -1264,7 +1264,7 @@ class ItemsInteraction(ClickOrDrag, _PlotInteraction):
         if isinstance(item, items.MarkerBase):
             self._signalMarkerMovingEvent('markerMoving', item, x, y)
             if item is not None:
-                item.startDrag()
+                item._startDrag()
 
         return True
 
--- a/silx/gui/plot/PlotInteraction.py
+++ b/silx/gui/plot/PlotInteraction.py
@@ -1435,6 +1435,7 @@ class ZoomAndSelect(ItemsInteraction):
         :return: True if click is catched by an item, False otherwise
         """
         eventDict = self._handleClick(x, y, btn)
+        
         if eventDict is not None:
             # Signal mouse clicked event
             dataPos = self.plot.pixelToData(x, y)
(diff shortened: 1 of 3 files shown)


vallsv:
Is it part of the release?

payno:
nope

silx/gui/plot/items/roi.py
```diff
@@ -113,6 +120,7 @@ def __init__(self, parent=None):
         self._editable = False
         self._visible = True
         self.sigItemChanged.connect(self.__itemChanged)
+        self._editing = False
```
t20100: `_editing` can be removed, it is not used.

silx/gui/plot/items/marker.py
```diff
@@ -167,6 +172,12 @@ def _defaultConstraint(*args):
         """Default constraint not doing anything"""
         return args
 
+    def startDrag(self):
```
t20100: Maybe rename `start|endDrag` to `_start|_endDrag`, we don't want someone to call them directly.

silx/gui/plot/items/marker.py
```diff
@@ -35,14 +35,19 @@
...
-
+from silx.gui import qt
 
 _logger = logging.getLogger(__name__)
 
 
 class MarkerBase(Item, DraggableMixIn, ColorMixIn, YAxisMixIn):
     """Base class for markers"""
 
+    sigItemPressed = qt.Signal()
```
t20100: I'd prefer `sigDragStarted|Finished`.

silx/gui/plot/items/roi.py
```diff
@@ -100,8 +100,15 @@ class RegionOfInterest(_RegionOfInterestBase):
     sigRegionChanged = qt.Signal()
     """Signal emitted everytime the shape or position of the ROI changes"""
 
+    sigRegionEditionStarted = qt.Signal()
```
t20100: I'd prefer `sigDragStarted|Finished`

silx/gui/plot/items/roi.py
```diff
@@ -505,6 +519,15 @@ def __str__(self):
         params = '; '.join('(%f; %f)' % (pt[0], pt[1]) for pt in points)
         return "%s(%s)" % (self.__class__.__name__, params)
 
+    def _startEditing(self, *args, **kwargs):
```
t20100: Can be renamed `_start|_endDrag`

silx/gui/plot/items/roi.py
t20100: ....Though it's not exactly a drag of the whole ROI.... to see.

t20100:
Sounds like missing signals, indeed.

I put a few comments in the review, main one is: I'd rename `MarkerBase.sigItemPressed|Released` and `RegionOfInterest.sigRegionEditionStarted|Finished` to `MarkerBase.sigDragStarted|Finished` as it is more meaningful and makes the API consistent between markers and ROIs (and similar to QtQuick API https://doc.qt.io/qt-5/qml-qtquick-drag.html#attached-signals).

silx/gui/plot/PlotInteraction.py
```diff
@@ -1263,6 +1263,8 @@ def beginDrag(self, x, y):
 
         if isinstance(item, items.MarkerBase):
             self._signalMarkerMovingEvent('markerMoving', item, x, y)
+            if item is not None:
```
t20100: BTW, `item` cannot be `None`, it is an instance of `MarkerBase`

silx/gui/plot/items/roi.py
```diff
@@ -392,8 +399,14 @@ def _createPlotItems(self):
                 item.setColor(color)
                 item.setVisible(self.isVisible())
                 plot._add(item)
+                # connect item changed
                 item.sigItemChanged.connect(functools.partial(
                     self._controlPointAnchorChanged, index))
+                # connect items pressed and released signals
+                item.sigDragStarted.connect(functools.partial(
```
t20100: One final comment, the `functools.partial` is actually not used as the parameters of `_editingStarted|Finished` are not used, probably best to remove unless you want to give the index in `sigEditingStarted|Finished` which can be a good idea if it is meaningful

silx/gui/plot/items/roi.py
```diff
@@ -505,6 +518,13 @@ def __str__(self):
         params = '; '.join('(%f; %f)' % (pt[0], pt[1]) for pt in points)
         return "%s(%s)" % (self.__class__.__name__, params)
 
+    def _editingStarted(self, *args, **kwargs):
```
t20100: `*args, **kwargs` is not used.

t20100:
LGTM

PR #2758, opened 2019-09-25
ENH: add double click slot on hdf5 table to copy value to clipboard

tacaswell:
This is to make it easy to copy things like the path and name out.

### DIFF
--- a/silx/gui/data/Hdf5TableView.py
+++ b/silx/gui/data/Hdf5TableView.py
@@ -524,6 +524,13 @@ class Hdf5TableView(HierarchicalTableView.HierarchicalTableView):
         super(Hdf5TableView, self).__init__(parent)
         self.setModel(Hdf5TableModel(self))
 
+        self.doubleClicked.connect(self._dbl_click)
+
+    def _dbl_click(self, index):
+        m = self.model().data(index)
+        qapp = qt.QApplication.instance()
+        qapp.clipboard().setText(m)
+
     def isSupportedData(self, data):
         """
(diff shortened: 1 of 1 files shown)


vallsv:
Hi.

I don't think it is a good idea. It's a kind of magic shortcut. It would be better to use a context menu + a copy action.

BTW, do you know you can use ctrl-c?

What do  you think?

tacaswell:
I am used to the linux-style "copy on high-light" which is what I started trying to make this work (I could figure out how to make the text editable, but not selectable...).

right-click -> copy would also be fine.  Can you point me to an example of setting that up?

vallsv:
- From `QTableView`, you could intersept mouse events.
- From the event you can reach the mouse position
- `QTableView` provides a getter `indexAt(QPoint) -> QModelIndex`
- Using the model and the index you can reach the content

That's an example.

- You also can use the mouse event to drop a `QMenu`, with a copy action, and then providing the same next sequence.

- Finally there is a shortcut for context menu. You can reach an example here https://github.com/silx-kit/silx/blob/6cfaaf05d87e2f1d978fcc0dce06b74111e5f270/silx/gui/hdf5/Hdf5TreeView.py#L101 (`customContextMenuRequested`, `setMenuPolicy`)

vallsv:
Hi Thomas,

We have just merged into the master #2903 if you like to check.
Now you can select the text you like to copy, as a normal text.
Let us know if it is fine for you.
I close your PR.

tacaswell:
Thanks!

PR #2772, opened 2019-10-03
Create an example for the ScatterView widget

vallsv:
Hi,

I notice there was no example for the scatter view widget.

Here is one, if you like.

Screenshot (93)


t20100:
Missing, indeed.
Could you add it to the doc as well?
The list of examples is there: https://github.com/silx-kit/silx/blob/master/doc/source/sample_code/index.rst

PR #2778, opened 2019-10-05
Fix dicttoh5 for dictionaries that have lists of dictionaries as values.

### LINKED ISSUE #2768: Loading json file and dumping hdf5
The snippet below returns `TypeError: list indices must be integers or slices, not collections.OrderedDict`:

```code: python
from silx.io import dictdump

data = dictdump.load("model.json")
dictdump.dump(data, "model.h5")
```

mretegan:
Fixes #2768. This brakes however the original structure as lists become dictionaries with the indexes as keys.


mretegan:
This brakes however the original structure as lists become dictionaries with the indexes as keys.

silx/io/dictdump.py
```diff
@@ -207,35 +207,53 @@ def dicttoh5(treedict, h5file, h5path='/',
         h5path += "/"
 
     with _SafeH5FileWrite(h5file, mode=mode) as h5f:
+
+        # The root of the tree can be a list or tuple of dictionaries.
+        if (isinstance(treedict, (list, tuple))
+                and all(isinstance(item, dict) for item in treedict)):
+            for i, item in enumerate(treedict):
+                dicttoh5(item, h5f, h5path + '/%d' % i,
```
t20100: To make it sorted, maybe you want to use 0 prefixed indices with something like: `ndigits = int(numpy.log10(len(treedict))) + 1`

silx/io/dictdump.py
```diff
@@ -207,35 +207,53 @@ def dicttoh5(treedict, h5file, h5path='/',
...
+            return
+
         for key in treedict:
-            if isinstance(treedict[key], dict) and len(treedict[key]):
+            value = treedict[key]
+
+            if (isinstance(value, (list, tuple))
+                    and all(isinstance(item, dict) for item in value)):
+                for i, item in enumerate(value):
+                    dicttoh5(item, h5f, h5path + key + '/%d' % i,
```
t20100: same as above

payno:
works well here.

Maybe an interesting improvement for the future would be to get Information from gui to know the active visualization mode (irregular grid, binned statistics / mean ...)

We could change selected QAction background color or name (adding a tick) or the QMenu tooltip (not a big fan of this one but can be an idea)

PR #2783, opened 2019-10-14
Improve legend widget

### LINKED ISSUE #2774: [widget] Improve LegendIcon widget
Here is few proposal to improve the `LegendIcon` widget.

- Support symbol size
- Support colormap lut
- Move it to an own module (``silx.gui.widgets.something`)

Right now the widget provides display for lines and symbols:
Screenshot (100)

Here is my current result, created by inheritance:
Screenshot (101)

My aim is to be able to display the style of images and scatters. And i agree the current result for scatter symbol with colormap is not very nice.

If you like i can create a PR for that.

vallsv:
- Create a dedicated module for the `LegendIcon` -> `silx.gui.widgets.LegendIconWidget`
    - It was created to have no dependency with plot objects
- Support `colormap` (including grayed when disabled)
- Support `symbolColormap` to display style for scatters (including grayed when disabled)
- `LegendIcon` still exists, it provides the link between plot items and `LegendIconWidget`.
- Update of the `simplewidget` example
- Module attributes was set private
- Accesses (from `LegendModel`) like `linestyle in LegendIconWidget.NoLineStyle` was replaced by `isEmptySymbol` and `isEmptyLineStyle`
- Automatically call `update` when an attribute was changed, instead of forcing the user to do so

### Todo
- Some unittest or code coverage?

### Preview

Screenshot (115)

- line
- symbol
- line + symbol
- colormap
- symbol using colormap
- line-based symbol using colormap
- colormap + line + symbol using colormap

### Could be a problem:
- `NoLineStyle`, `Symbols`... are not anymore part of `silx.gui.plot.LegendSelector`. It maybe breaks external code

Closes #2774

### DIFF
--- a/examples/simplewidget.py
+++ b/examples/simplewidget.py
@@ -165,13 +165,23 @@ class SimpleWidgetExample(qt.QMainWindow):
         # Symbol and colormap
         legend = LegendIconWidget(panel)
         legend.setSymbol("o")
-        legend.setColormap("viridis")
+        legend.setSymbolColormap("viridis")
         layout.addWidget(legend)
 
         # Symbol (without surface) and colormap
         legend = LegendIconWidget(panel)
         legend.setSymbol("+")
-        legend.setColormap("plasma")
+        legend.setSymbolColormap("plasma")
+        layout.addWidget(legend)
+
+        # Colormap + Line + Symbol
+        legend = LegendIconWidget(panel)
+        legend.setColormap("gray")
+        legend.setLineStyle("-")
+        legend.setLineColor("white")
+        legend.setLineWidth(3)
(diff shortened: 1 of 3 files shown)


t20100:
I would consider `LegendIcon`, `NoLineStyle` and the like as not currently being part of the public API, they are not in the documentation. So to me you are free to change the API if you have to.

t20100:
So +1 for your proposal

vallsv:
I fixed the proposals.

All the module attributes are now private, and i created helper `isEmptyLineStyle` and `isEmptySymbol` to replace (from `LegendModel`) `linestyle in LegendIconWidget.NoLineStyle`.

PR #2794, opened 2019-11-06
Make ROI inherit from QObject + provide sigItemChanged

t20100:
This PR is a follow-up of PR #2684 to make ROI base class a QObject and provide a `sigItemChanged` signal as plot items.


payno:
Lgtm

Issue #2797, opened 2019-11-20
OpenGL invalid enumerant


vallsv:
Here is a report from ID21.

The user tried to display a scatter plot using PyMca, from a Debian 6, connected to a Debian 8 machine.

As i know some exceptions are already catched, that's maybe something which can be also catch to provide instead the matplotlib version?

```
<class 'OpenGL.error.GLError'> GLError(
        err = 1280,
        description = b'invalid enumerant',
        baseOperation = glGetString,
        cArguments = (GL_VERSION,),
        result = b'3.0.0 NVIDIA 410.78'
)   File "/users/blissadm/conda/miniconda/envs/bliss/lib/python3.7/site-packages/silx/gui/_glutils/OpenGLWidget.py", line 195, in initializeGL
    version = gl.glGetString(gl.GL_VERSION)
  File "/users/blissadm/conda/miniconda/envs/bliss/lib/python3.7/site-packages/OpenGL/platform/baseplatform.py", line 402, in __call__
    return self( *args, **named )
  File "/users/blissadm/conda/miniconda/envs/bliss/lib/python3.7/site-packages/OpenGL/error.py", line 232, in glCheckError
    baseOperation = baseOperation,
```

t20100:
`silx` never supported debian 6 and no longer supports debian 8 (as debian also no longer supports it)...

t20100:
That being said, this looks like an invalid OpenGL call, more info on the config would be needed to figure out what's going on (nvidia drivers or not?, any other OpenGL-based application working in this config?, version of python, pyopengl, etc...).

vasole:
@t20100 

What I understand is that the software is running on Debian 8 but seen remotely from Debian 6.

@vallsv 

What version of PyMca? The latest version performs a test to see if OpenGL can be used.

t20100:
@vasole the error is issued from `silx` not from pymca directly.

It is in the test where I check if I can get the OpenGL version to see if OpenGL actually works.
What's funny here is that it seems to get something correct `b'3.0.0 NVIDIA 410.78'`, but also raises an error...
Anyway a `try: except:` block as it's place here, as it is all about knowing whether OpenGL works or not.

vasole:
I know the error is from silx, but if OpenGL is not supported one should not reach there.

vallsv:
I am sorry, i don't have much information. But I can try to ask.

I think they have uninstalled pyopengl to fix the problem :-)

t20100:
;)

[...]

PR #2799, opened 2019-11-22
[OpenGL] Clear errors prior to initialize widget

vasole:
Related to #2797 

The problem appears with PyQt 5.9.2 on the anaconda installation of lid21nano.
This patch allows to initialize the widget by clearing the previous errors.


vallsv:
Wow, strange. Thanks for the check.

PR #2800, opened 2019-11-22
Rework matplotlib backend rendering (and picking)

### LINKED ISSUE #2785: Adding artists directly in the matplotlib backend no longer work
Quite a few users are relying on (ab)using the direct access to the matplotlib backend to benefit from all matplotlib artists.
PR #2730, which refactored the rendering had broken this.
I think even if it is not part of the public API , as it is used and there is no alternatives, this should be restored.

Thanks @mretegan for reporting it.

t20100:
This PR reworks the `matplotlib` rendering.
The way used in PR #2730 had some short-comings (e.g., see #2785 + when saving as image, the content was different from what the widget displayed...)

The idea here is to decouple the z order in `PlotWidget` from that in `matplotlib`.
Each matplotlib item is given a different z order (which order is set according to this order in the `PlotWidget`).
The drawback of this method it that items associated with left and right axes are no longer interleaved regarding z as matplotlib `Axes` are one on top of the other... but trying to avoid that (as in #2730) is quite tricky and (as shown by the issues it implies) not really robust.

closes #2785

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -2933,28 +2933,6 @@ class PlotWidget(qt.QMainWindow):
         """
         return self._getItem(kind='marker', legend=legend)
 
-    def _itemsFromBackToFront(self, condition=None):
-        """Iterator of plot items ordered from back to front.
-
-        This is the order used for rendering.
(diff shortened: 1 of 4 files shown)


payno:
LGTM

PR #2804, opened 2019-11-25
silx view display NeXus title, name, short_name, etc...

### LINKED ISSUE #2713: [silx view] show NXentry title

t20100:
This PR proposes an alternative to PR #2716 by implementing one of the options proposed in this PR (https://github.com/silx-kit/silx/pull/2716#issuecomment-525628487)

This PR:
- Moves the `Value` column in `silx view` right after the `Name` column
- Populates the `Value` column of NeXus groups with the value of a child dataset which name depends on the kind of `NX_class` (see `_NEXUS_CLASS_TO_VALUE_CHILDREN`).

By default, it looks for a "title" dataset as most (but not all...) application definitions are defining one.
This would be simpler if NeXus was using consistent names for the same thing across classes (see `_NEXUS_CLASS_TO_VALUE_CHILDREN`)...

closes #2713


t20100:
Here is a screenshot:

silx-view-nx-as-Value

t20100:
BTW, if we don't care about strict compliance to the NeXus format, the mapping from `NX_class` to title names can more or less be replaced by looking for a children by name in the following order: `short_title`, `title`, `short_name`, `name`.

vallsv:
What do you think about creating a new column? That's not a big deal.

Cause it would be better to rename it `Description`/`Info`/``Detail`, or something like that. That's not a "value".

The column id should not be edited, but the header to be fixed to use another order. But it's also fine this way. I don't think it will break things. And then you can hide the `value` column (should should only contains dataset values).

For your command about NeXus  format, i think it is better to be strict.

vallsv:
Except that it's fine.

t20100:
Updated to use the `Description` column rather than the `Value` column to display title/name/description, and to reorder the columns in `silx view` so that column ids in `silx.gui.hdf5` is left unchanged.

As for following NeXus, to me a reader should be permissive while a writter should be strict... but NeXus is already complicated so I guess it can become cumbersome to also be permissive.

vallsv:
There is a small issue relative to a test. looks easy to fix.
```
FAIL: testDatasetData (silx.gui.hdf5.test.test_hdf5.TestHdf5TreeModel)
  File "C:\projects\silx\venv_test\lib\site-packages\silx\gui\hdf5\test\test_hdf5.py", line 348, in testDatasetData
    self.assertEqual(displayed[hdf5.Hdf5TreeModel.DESCRIPTION_COLUMN, qt.Qt.DisplayRole], "")
AssertionError: '[1 2 3]' != ''
- [1 2 3]
+ 
```

vallsv:
I add a PR to your PR.
- Fix tests
- Categorize kind of descriptions
- Add icons
- Add tooltip

Screenshot (134)

This should help to distinguish different values. That could be much better, but the result is not so bad.

Feel free to merge, it's fine for me.

t20100:
The icons add visual complexity.
Giving the kind of input used to populate the description in the tooltip is probably sufficient. 
Anyway, I merged your PR, let's see how it goes.

One option could also have been to make the text bold for titles/names.

t20100:
It's ready for me, we can always rework the description icons later on if needs be (and time available).

PR #2805, opened 2019-11-25
Make reading NX_class from silx.gui.hdf5 a bit more robust to errors

t20100:
This PR adds makes `silx.gui.hdf5` and so `silx view` a bit more robust against malformed `NX_class` by making it case-insensitive.

No sure we want to go into this direction... but I encountered this case.
What do you think?

I'm fine to drop this.


vasole:
If we go that way, the user should be notified. Logging the situation as en error seems convenient.

t20100:
Added an error message when `NX_class` is not correct.

PR #2806, opened 2019-11-28
2695 tesla pocl

kif:
allows using PortableCL driver with Nvidia GPU (like on the Power9)


kif:
> The default chosen device is still CPU on our powerpc platform, despite the estimated flops of GPU being 4x higher. This can be solved in a future PR.

This is another bug then ... surprizing. I thought the selection was based on the flops value

kif:
After investigation, the selected device was "nvidia/tesla" which turns out to be unusable. Then it falls back to pocl/power9. I deactivated the nvidia driver on the power9 and now the selected device is "pocl/tesla" by default.

Issue #2743, opened 2019-09-12
[plot] Add a solid regular display mode for scatter


vallsv:
The main idea is to render a scatter in a "pixel based" way. But not to render a image using the same pixel size.

Here is constraints which i think are useful for the readability.

- The polygons should have mostly the same surface
- The polygons can be quadrilateral, i don't think we need a real voronoi rendering
- The polygons have to be mostly centered around the location of each points

![scatter-solid](https://user-images.githubusercontent.com/7579321/64764124-b8b57100-d541-11e9-854a-1fefd24401d1.png)

I don't need to display the |symbols at the location points. I can do it with another scatter item.

Relative to #2709

vallsv:
@woutdenolf send to me a dataset from real data. 3MB, about 200K points.

We can notice huge displacement around the image. Sometimes about 5x the pixel size. But globally it looks aligned.

I am curious about the result with a pixel based rendering.

vallsv:
![Screenshot (87)](https://user-images.githubusercontent.com/7579321/64958561-c5aec900-d88f-11e9-85e2-225bc54279a7.png)

PR #2810, opened 2019-12-05
Scatter plot regular grid visualization mode

t20100:
This PR provides an extra visualization of scatter plots as image for cases where the points are on a regular grid.

related to #2743

TODOs: Better guessing of the image dimension, including handling points on a single line, backward direction for both axes, y as fast moving dimension

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -922,7 +922,7 @@ class ScatterVisualizationMixIn(ItemMixInBase):
         This is based on Delaunay triangulation
         """
 
-        REGULAR_GRID = 'regular grid'
+        REGULAR_GRID = 'regular_grid'
         """Display scatter plot as an image.
 
         It expects the points to be the intersection of a regular grid,
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -922,6 +922,15 @@ class ScatterVisualizationMixIn(ItemMixInBase):
         This is based on Delaunay triangulation
         """
 
+        REGULAR_GRID = 'regular grid'
+        """Display scatter plot as an image.
+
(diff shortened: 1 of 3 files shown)


vasole:
The profile cannot be extracted unless we start from outside the image. The same happens in pure scatter view if we click on a point and not in the space between the points,

t20100:
PR #2811 is solving the scatter profile issue.

vasole:
LGTM

Should we wait for @vallsv?

vallsv:
I would like to do some tests yes. At worst tomorrow, if it is good for you.

vallsv:
Just a small comment about:
`        REGULAR_GRID = 'regular grid'`
As it is not a label, I would prefer no space here.

vallsv:
Oh, sorry, you was talking about #2811? It's fine for me.

t20100:
- `REGULAR_GRID` enum value changed
- Added support for: backward direction for both axes, y as fast moving dimension, single value in scatter.
- handling points on a single line was already implemented.

The weak point is the guessing, but I guess that's a never ending story...

I still want to test it a bit, but it should be almost final for now so it is worth testing/providing feedback now.

vallsv:
Another typo: `get_Z_line_length`. i don't see why you put upper case? It just make the thing difficult to find. But it's up to you.

I can merge it if you like. Or it is still in WIP?

t20100:
Still WIP, I'm reworking the guessing part.
I'll rename `get_Z_line_length`, it should be private.

t20100:
Ready for review.
Let's keep the rework of the guess for another PR.

t20100:
Ready for review.
Let's keep the rework of the guess for another PR.

vallsv:
LGTM

PR #2811, opened 2019-12-09
Fix scatterview profile interaction

t20100:
This PR fixes an issue in the `ScatterView` with the profile tool.
Initially raised here: https://github.com/silx-kit/silx/pull/2810#issuecomment-562220610

It was not possible to start a profile while hovering a point of the scatter plot (this was due to the scatter item being selectable and so taking focus of the mouse press event).

The scatter item is no longer selectable and so no longer takes the focus of mouse events

### DIFF
--- a/silx/gui/plot/ScatterView.py
+++ b/silx/gui/plot/ScatterView.py
@@ -79,7 +79,7 @@ class ScatterView(qt.QMainWindow):
         self._plot = weakref.ref(plot)
 
         # Add an empty scatter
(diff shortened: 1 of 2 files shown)


vallsv:
BTW what does this change means?
```
-        roi.setLabel('Profile')
+        roi.setName('Profile')
```

payno:
It comes from the fact that `ROI` and `RegionOfInterest` classes are now sharing some API.
And setLabel is now deprecated.

PR #2813, opened 2019-12-10
Add action to save stack of images

### LINKED ISSUE #2812: StackViewMainWindow - save the stack in a file
The 'save image' action available from the stack will apply on the current image.
But at one point we would like to dump all the image into a file (hdf5 in our cases).
It would be good to have this at least as an action available from the `StackViewMainWindow`.
Maybe it should be available from default ?

juliagarriga:
This PR adds a filter in the `SaveAction` of `StackView` which makes possible to save all the stack in an hdf5 file.
We need to specify that the 'IMAGE_STACK_FILTER_NXDATA' is an 'Append' action to the file.
If 'IMAGE_STACK_FILTER_NXDATA' should be stored in StackView then we need to add an 'addAppendFilter' function.

/close #2812

### DIFF
--- a/silx/gui/plot/actions/io.py
+++ b/silx/gui/plot/actions/io.py
@@ -509,7 +509,7 @@ class SaveAction(PlotAction):
                 axes_errors=[xerror, yerror],
                 title=plot.getGraphTitle())
 
-    def setFileFilter(self, dataKind, nameFilter, func, index=None):
+    def setFileFilter(self, dataKind, nameFilter, func, index=None, appendToFile=False):
         """Set a name filter to add/replace a file format support
 
         :param str dataKind:
@@ -520,9 +520,14 @@ class SaveAction(PlotAction):
         :param callable func: The function to call to perform saving.
             Expected signature is:
             bool func(PlotWidget plot, str filename, str nameFilter)
+        :param bool appendToFile: True to append the data into the selected
+            file.
         :param integer index: Index of the filter in the final list (or None)
         """
         assert dataKind in ('all', 'curve', 'curves', 'image', 'scatter')
+
+        if appendToFile:
+            self._appendFilters.append(nameFilter)
 
         # first append or replace the new filter to prevent colissions
         self._filters[dataKind][nameFilter] = func
--- a/silx/gui/plot/actions/io.py
+++ b/silx/gui/plot/actions/io.py
@@ -237,7 +237,7 @@ class SaveAction(PlotAction):
             # create new entry in new file
             return "/entry"
         else:
-            self._errorMessage('Save failed (file access issue)\n', parent=parent)
+            SaveAction._errorMessage('Save failed (file access issue)\n', parent=parent)
             return None
 
     def _saveCurveAsNXdata(self, curve, filename):
--- a/silx/gui/plot/actions/io.py
+++ b/silx/gui/plot/actions/io.py
@@ -185,10 +185,11 @@ class SaveAction(PlotAction):
         self.setShortcut(qt.QKeySequence.Save)
         self.setShortcutContext(qt.Qt.WidgetShortcut)
 
-    def _errorMessage(self, informativeText=''):
+    @staticmethod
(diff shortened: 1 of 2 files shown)


silx/gui/plot/actions/io.py
```diff
@@ -185,10 +188,11 @@ def __init__(self, plot, parent=None):
         self.setShortcut(qt.QKeySequence.Save)
         self.setShortcutContext(qt.Qt.WidgetShortcut)
 
-    def _errorMessage(self, informativeText=''):
+    @staticmethod
+    def _errorMessage(informativeText='', parent=None):
         """Display an error message."""
         # TODO issue with QMessageBox size fixed and too small
-        msg = qt.QMessageBox(self.plot)
```
t20100: What about:
```
if parent is None:
    parent = self.plot
```
so you keep the previous behavior by default and avoid changing all the calls.

silx/gui/plot/actions/io.py
```diff
@@ -220,7 +224,8 @@ def _getAxesLabels(self, item):
         ylabel = item.getYLabel() or self.plot.getYAxis().getLabel()
         return xlabel, ylabel
 
-    def _selectWriteableOutputGroup(self, filename):
+    @staticmethod
```
t20100: It cannot be static, it's using `self`

silx/gui/plot/actions/io.py
```diff
@@ -220,7 +224,8 @@ def _getAxesLabels(self, item):
...
+    def _selectWriteableOutputGroup(filename, parent):
```
t20100: Same as about, adding `parent=None` avoids a few changes and allow only to set it when it matters.

silx/gui/plot/actions/io.py
```diff
@@ -144,9 +145,11 @@ class SaveAction(PlotAction):
     SCATTER_FILTER_NXDATA = 'Scatter as NXdata (%s)' % _NEXUS_HDF5_EXT_STR
     DEFAULT_SCATTER_FILTERS = (SCATTER_FILTER_NXDATA,)
 
+    IMAGE_STACK_FILTER_NXDATA = 'Stack of images as NXdata (%s)' % _NEXUS_HDF5_EXT_STR
+
     # filters for which we don't want an "overwrite existing file" warning
     DEFAULT_APPEND_FILTERS = (CURVE_FILTER_NXDATA, IMAGE_FILTER_NXDATA,
-                              SCATTER_FILTER_NXDATA)
+                              SCATTER_FILTER_NXDATA, IMAGE_STACK_FILTER_NXDATA)
```
t20100: It would be best to define `IMAGE_STACK_FILTER_NXDATA` in `ScatterView` and provide a way to extend the `DEFAULT_APPEND_FILTERS` from outside `SaveAction`.

One way is to add an argument to `setFileFilter` (e.g., `overwriteWarning=True`) and to use a list attribute (initialised with the content of `DEFAULT_APPEND_FILTERS` to check for dialog that needs overwrite warning.

silx/gui/plot/actions/io.py
t20100: OK, I got why you want it static, then self needs to be replaced by SaveAction.

juliagarriga: Yes sorry! Forgot to change it

silx/gui/plot/actions/io.py
juliagarriga: Yes perfect, how about calling the argument `append` ?

t20100: `append` is not meaningful to me, without ready the doc, I'd expect it to "append" the filter to the end of the list of filter....
`allowOverwite` or `withOverwrite`?

t20100: or `appendToFile`?

t20100: Any other options?

juliagarriga: `appendToFile` is good

PR #2814, opened 2019-12-10
Expose ColormapNameComboBox in silx.gui.widgets

vallsv:
Move `ColormapNameComboBox` from `ColormapDialog` to the public API `silx.gui.widgets`.


vallsv:
Oups

vallsv:
Here is an update

PR #2815, opened 2019-12-10
Scatter regular grid visualization parameters

t20100:
This PR allows to provide parameters for the regular grid visualization, that when provided replace the guessing.
Those parameters are:
- `GRID_ORDER`: 'C' or 'F' as in numpy for row- or column- major order
- `GRID_BOUNDS`: ((x, y) of first point, (x, y) of last point)
- `GRID_SIZE`: (width, height) number of grid cells in each dimension

`GRID_BOUNDS` and `GRID_SIZE` can be wider than the current number of points, so one can provide the final values of those and only update the data (e.g., for live monitoring...)

Still to do: Use `GRID_BOUNDS` as the bounds of the item when provided?

related to #2743

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -931,8 +931,36 @@ class ScatterVisualizationMixIn(ItemMixInBase):
         (either all lines from left to right or all from right to left).
         """
 
+    @enum.unique
+    class VisualizationParameter(_Enum):
+        """Different parameter names for scatter plot visualizations"""
+
+        GRID_ORDER = 'grid_order'
+        """The order of points in the regular grid (row or colunm major).
+
+        Either 'C' (row-major) or 'F' (column-major) as for numpy array contiguity.
+        """
+
+        GRID_BOUNDS = 'grid_bounds'
+        """The expected range in data coordinates of the regular grid.
+
+        A 2-tuple of 2-tuple: (begin (x, y), end (x, y)).
+        This provides the data coordinates of the first point and the expected
+        last on.
+        As for `GRID_SIZE`, this can be wider than the current data.
+        """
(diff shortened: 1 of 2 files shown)


t20100:
Answer to: Use GRID_BOUNDS as the bounds of the item when provided?

No, use data bounds, so ready for review.

vallsv:
Its fine for me. But just a comment about the params:

- `GRID_ORDER`, if it is about the fast axes, i think it could be a little more understandable to use the name of this axes. You named it X and Y. -> `FAST_AXES = 'X'`. or if you expect complex things at one point `GRID_ORDER = "FAST_X" or "XY"`...
- I will receive metadata per axes, in case it is convinient for you, an API per axes fit better `AXES_X = begin,end,nbpoint`, `AXES_Y = begin,end,nbpoint`. But it is up to you.
- What if i have nbpoints for one and not for the other? Just to say that it will happen.

vasole:
@vallsv 

To have the nb of points (or intervals) is not enough. A mesh scan can be incomplete. Therefore one has to figure out the dimensions of the expected regular grid based on the available points. The nice side of the approach is that is the users asks for a regular mesh without relying on any additional data.

t20100:
@vasole, the idea of this PR was to provide some information about the grid structure in case on has it so as to avoid guessing it. If it is not provided, then the grid is guessed.

@vallsv, having `begin,end,nbpoint` for each axis might be a good, I didn't consider the fact of having info for only one axis (in fact, the info from the fast axis only is really needed).
As for the `GRID_ORDER` naming, I reused that of numpy. Calling it fast axis is somewhat specific to scans, I could use `row_major` and `column_major` which is the generic concept underneath.

PR #2818, opened 2019-12-11
Robustness on testutils

### LINKED ISSUE #2816: testutils.py failure
While executing BLISS test suite, which partly rely on silx test suite for Flint, it happened that
`_outcome` was `None`:

```
self = <test_image_plot.TestImagePlot testMethod=test_display_image>

    def _currentTestSucceeded(self):
        if hasattr(self, '_outcome'):
            # For Python >= 3.4
            result = self.defaultTestResult()  # these 2 methods have no side effects
>           self._feedErrorsToResult(result, self._outcome.errors)
E           AttributeError: 'NoneType' object has no attribute 'errors'

../../miniconda3/envs/bliss/lib/python3.7/site-packages/silx/gui/utils/testutils.py:158: AttributeError
```
(maybe more for @vallsv ?)

vallsv:
I guess it will not work for python<3.4
Let see the CI

Closes  #2816

### DIFF
--- a/silx/gui/utils/test/__init__.py
+++ b/silx/gui/utils/test/__init__.py
@@ -35,6 +35,7 @@ import unittest
 from . import test_async
 from . import test_image
 from . import test_qtutils
+from . import test_testutils
 from . import test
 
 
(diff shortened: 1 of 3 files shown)


vallsv:
Then the test is failing with python 2.7 (cause of the test itself).

I just want to skip the test for python 2.7, is that fine with you? It's not a good idea to spend time on 2.7.

t20100:
yes, skip on python 2 if failing.

vallsv:
Let's try that

PR #2822, opened 2019-12-12
[plot] Make code compatible with numpy 1.12

t20100:
This PR makes the plot picking code compatible with numpy 1.12 where the `axis` argument of `numpy.flip` has no default value.


vallsv:
LGTM

PR #2823, opened 2019-12-12
Create BoundingRect item

vallsv:
This PR introduce a `BoundingRect` item.

It is a kind of shape, which is not displayed but considered as part of the data.

This can be used to custom the expected visible region of the plot.

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -603,7 +603,7 @@ class PlotWidget(qt.QMainWindow):
         elif isinstance(item, (items.Marker,
                                items.XMarker, items.YMarker)):
             kind = 'marker'
-        elif isinstance(item, items.Shape):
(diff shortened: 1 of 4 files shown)


t20100:
Looks like there are a few things missing for the signal and clearing data bounds cache.
I'll make a PR on your branch for it.

t20100:
I made a PR on your branch with the handling of updates to refresh the data bounds whenever visibility, y axis or bounds are changed.
We can discuss it.
The whole data bounds refresh needs to be done in a better way (e.g., inheriting from a `DataItem`) but that is not for this PR.

t20100:
BTW, did you check how it behaves when the rectangle is partly negative and axes are log scale?

Also probably worth making sure the mins are below the maxs...

vallsv:
Here is an update to:
- Check argument validity
- Have an expected behaviour with log scale axes

PR #2824, opened 2019-12-12
Add an irregular grid rendering mode for scatter plots

### LINKED ISSUE #2743: [plot] Add a solid regular display mode for scatter

t20100:
This PR adds an irregular grid mode to scatter plot visualization.
It provides a way to visualize the data as an irregular grid.

I would consider it an experimental feature to see what it can provide.

Todo: testing, picking.. optimization probably

closes #2743

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -931,6 +931,15 @@ class ScatterVisualizationMixIn(ItemMixInBase):
         (either all lines from left to right or all from right to left).
         """
 
+        IRREGULAR_GRID = 'irregular_grid'
+        """Display scatter plot as contiguous quadrilaterals.
+
+        It expects the points to be the intersection of an irregular grid,
+        and the order of points following that of an image.
+        First line, then second one, and always in the same direction
+        (either all lines from left to right or all from right to left).
+        """
+
     @enum.unique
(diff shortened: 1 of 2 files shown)


t20100:
Few fixes: picking (with OpenGL at least), dimension < 2 check.
Main remaining issue is with column-major grid.

t20100:
column-major grid working.

ready for review.

vallsv:
LGTM except few things.

- The very first line is not rendered (that is not a big problem as far as i can know nothing is rendered)
- The last incomplit line is not rendered (here it is more problematic), if it is not too difficult to do it i really would like to have that.

I will merge it anyway.

vallsv:
Anyway, don't worry, that's already very nice to show what we can do.

t20100:
Yes, as it is, it only works with more than 2 lines and only displays complete lines.
This can be supported but not in this release.

PR #2829, opened 2019-12-13
Rework plot backend picking especially for matplotlib

### LINKED ISSUE #2821: picking issue with matplotlib backend
They are regression in the plot picking with the matplotlib backend with triangles and images.
The OpenGL backend looks to work fine.

t20100:
This PR reworks the plot backend picking to simplify it and solve issues with matplotlib backend

closes #2821


vallsv:
Does your PR is compatible, in case i want a curve not to be selectable?

t20100:
Yes, the selectable/draggable properties are handled in `PlotWidget`, the backend does not have to handle it.
This removes some left over in the backend that created issues.
The OpenGL backend was already not using `selectable` and `draggable` argument while the matplolib backend was using them to make matplotlib items pickable.
This creates issue because in some situations you want to be able to pick an item no matter if it is selectable or not.

t20100:
matplotlib picking fixed.

ready for review.

t20100:
Added a few tests for picking with scatter grid visualizations (which use the picking underneath)

PR #2832, opened 2019-12-13
[gui][StackView] fix autoscale=True option

### LINKED ISSUE #2809: [plot] StackView issue with colormap
There seems to be an issue with `StackView` colormap handling.

To reproduce:

```
import numpy
from silx import sx                                                     
w = sx.StackView()                                                      
w.show()                                                                
w.setColormap(colormap='viridis', autoscale=True)                       
w.setStack(numpy.arange(10**3).reshape(10, 10, 10))
```

and the colormap used is switching to the grey one....

Setting the colormap with: `w.setColormap(Colormap('viridis'))` or `w.setColormap(colormap='viridis')` works fine.

This looks to be due to setting `self.__autoscaleCmap` (https://github.com/silx-kit/silx/blob/master/silx/gui/plot/StackView.py#L834), which looks like not needed any longer and causing trouble.

Thanks to Alessandro for spotting it.

payno:
Should fix #2809 but there is probably plenty of cases where the stack colormap might fail.

from @t20100 : For now the best option seems to be to add a button to update vmin and vmax from the stack values.

### DIFF
--- a/silx/gui/plot/StackView.py
+++ b/silx/gui/plot/StackView.py
@@ -542,7 +542,12 @@ class StackView(qt.QMainWindow):
         # This call to setColormap redefines the meaning of autoscale
         # for 3D volume: take global min/max rather than frame min/max
         if self.__autoscaleCmap:
-            self.setColormap(autoscale=True)
+            # note: there is no real autoscale in the stack widget, it is more
+            # like a hack computing stack min and max
+            colormap = self.getColormap()
+            _vmin, _vmax = colormap.getColormapRange(data=self._stack)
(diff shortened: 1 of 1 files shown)


vallsv:
Yes, autoscale with the full stack was rejected cause of the efficiency.

The argue was, if you want that you still can disable the autoscale. A button for that could be helpful but it is more than what we was providing, as far as i remember.

t20100:
The problem is that as it is today, there are some inconsistencies due to partial handling of stack autoscale. So depending what you do when you get autoscale on the image or the stack.

PR #2839, opened 2019-12-17
Fix none and log bound

vallsv:
Here is a fix to avoid to fail when `BoundRect` is None and the plot is in log.


vallsv:
That's fine. Rebased to 0.12.

PR #2843, opened 2019-12-18
[mpl backend] Fix mouse-y location to integer

vallsv:
I notice that the `ypixel` of moue event is not an integer.
```
{'x': 81.35577372604021, 'y': 93.55140186915887, 'xpixel': 542, 'ypixel': 127.0, 'button': None, 'event': 'mouseMoved'}
```

Here is a fix.

Please take a look at the consequences, cause this method is not only called by mouse event call back.


t20100:
closing as supersede by #2849

PR #2849, opened 2019-12-20
[mpl backend] Fix mouse-y location to integer

t20100:
This PR includes and supersedes PR #2843 by taking care of casting to int or not depending on the usage.
It casts to int for mouse events, and leaves it as float for `dataToPixel` and `pixelToData`.
It also fixes types for `getPlotBoundsInPixels` as this is documented to be `int`.


vallsv:
LGTM

PR #2851, opened 2019-12-20
[plot] ColormapDialog: Add z order to plot items to ensure rendering order

### LINKED ISSUE #2850: Colormap dialog min/max line behind histogram
The line between min/max is under the histogram in the colormap dialog.
Check if this is a regression in 0.12 due to rewokr of rendering order.

Initially raised in #2847

t20100:
This PR makes sure plot items from the `ColormapDialog` are render in the right order by giving them a `z`.
Beforehand the rendering order was not guaranteed.. and it turned out to break with plot rendering/picking refactoring.

closes  #2850

### DIFF
--- a/silx/gui/dialog/ColormapDialog.py
+++ b/silx/gui/dialog/ColormapDialog.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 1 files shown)


vallsv:
LGTM

PR #2853, opened 2019-12-20
[plot] Fix matplotlib crosshair rendering

### LINKED ISSUE #2841: [plot] The lines displayed by the crossair option blinking a lot
Sometimes it is displayed, sometimes not depending on the mouse location.

As result sometimes the mouse stop and there is no crossair anymore.

t20100:
This PR fixes the rendering of the crosshair with the matplotlib backend which was due to the refactoring of the plot rendering and the fact that the right axes are on top of the left ones.

closes #2841


vallsv:
Oh... weird.
LGTM, thanks a lot.

PR #2860, opened 2020-01-06
Fix renaming of  get|setLabel to get|setName

t20100:
This PR fixes issues with renaming ROI's `get|setLabel` to `get|setName`:

- Renaming was not done in `sx.ginput` while it is needed.
- Renaming has been done accidentally in `plot3d` where it is not done (and can wait another release).


vallsv:
Nice catch.

t20100:
Well, nice exceptions....

PR #2861, opened 2020-01-06
Disable hdf5 file locking by default, add option to enable it

### LINKED ISSUE #2859: Do not lock the HDF5 file with silx view
This would allow to open a file while another process (e.g., acquisition) is writing it.

t20100:
This PR makes `silx view` no longer locking opened HDF5 files.

It also adds a  `--hdf5-file-locking` option to re-enable the file locking (which is the current behaviour).

What happens when someone removes stuff from the HDF5 file from another process is undefined, though quick tests shows that it does not crash.

See discussion and details here: https://gitlab.esrf.fr/bliss/bliss/issues/1241

closes #2859

### DIFF
--- a/silx/app/view/main.py
+++ b/silx/app/view/main.py
@@ -80,10 +80,10 @@ def mainQt(options):
     # Import most of the things here to be sure to use the right logging level
     #
 
-    if not options.hdf5_file_locking:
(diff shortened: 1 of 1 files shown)


t20100:
Minor update to always set the environment variable (so that behavior does not depend on this variable being set).

PR #2864, opened 2020-01-08
Fix 2858

### LINKED ISSUE #2858: [plot] stats not displayed as expected when update mode is manual
There is an issue with the displayed stats when stats widget is in manual mode: when only the active items should be displayed, all items are still displayed.

To reproduce, start `python plotStats.py --update-mode manual`,
then set to display only active items in stats widget and press the refresh stats button.
This should then only display the scatter plot stats, but stats for the 2 curves are also displayed.

payno:
close #2858


vallsv:
I updated the rebase to 0.12

PR #2868, opened 2020-01-09
Add support of not finite values in the fit

t20100:
This PR adds support of not finite values in fit by ignoring them.

It also fixes some calls to not existing `warnings.warning` function in `FitWidget`, and add top and bottom margins in the profile 1D plot.


t20100:
Added some tests with NaNs

t20100:
For the very last time: added support of python 2....

PR #2870, opened 2020-01-09
[CI] only run CI on master and x.y branches

t20100:
This PR avoids running CI on branches other than master and x.y.
This prevents to run CI on branches (so running twice) when pull requests are created from the github web interface.


vallsv:
Whle there is still CI on the PR it's fine.

PR #2874, opened 2020-01-13
Improve colormap dialog

### LINKED ISSUE #1742: [plot] Colormap dialog range selector
I would be nice to improve the colormap dialog range selection vertical bars on the histogram plot.
As is, the user does not know whether they are interactive or not.
An option is to change the color of those vertical markers depending whether the colormap bounds are in autoscale or not.
Another option proposed by @vallsv is to have vertical marker always interactive and leave autoscale mode if the user is moving the markers.

To me second option is best.

### LINKED ISSUE #2847: [colormap dialog] Few small improvements
Here is the current state of the colormap dialog
Screenshot from 2019-12-19 12-45-22

- the line between min/max is under the histogram (could be moved on top)
- vertical min/max markers could be displayed blue only if editable
- a scatter/image could be used to display the colormap within the graph (the histogram itself could be colorized, in case it works)

vallsv:
Closes  #2847

PR #2877 have to be merged first

Here is few improvement on the Colormap dialog.

- Display the LUT in the plot
- Markers can can always move, if it was in auto range, the auto range is removed
- I removed the `None` data display. I think it is never used, and plotUpdate uses data min/max anyway, there the aim of this option is pointless.
- I removed the curve displayed in the plot. It also do not have meaning (it's not a transfer function). 
- Rework the layout
- Add auto scale mode
- Rework the auto scale buttons (to also provide a single button for both auto)

Todo:
- [ ] The LUT disappear in log scale (cause it can't do the projection). But here the image have to be displayed in linear scale. That's a very special case. It could be nice and easy to find a workaround in this case. I tried things, help is welcome

Screenshot from 2020-01-17 16-01-13

Also closes #1742


vallsv:
Screenshot from 2020-01-15 09-22-33

Here is a proposal for layout improvement
- It includes the autoscale mode
- Move the min/max under the plot
- Try to rework the `auto` to provide a single button to set both min/max in auto

vallsv:
I am happy with the result.

If we can fine a solution for the LUT display in the plot in log, it's fine, else it can still be merged.

vallsv:
I changed the combobox using "stddev" instead of SD. There is enough space, and it was also used by Oxidis.

vallsv:
The code is rebased.

I still notice an assertion with min>max few days ago, when switching from log to linear. It maybe have to be taken into account.

t20100:
For displaying the colormap in log scale, maybe a set of colored triangles can do the job.

t20100:
I've fixed the tests with `PySide2`.

t20100:
@vallsv did you push all your changes? 

> I changed the combobox using "stddev" instead of SD. 

It's still `SD` in the PR.

vallsv:
@t20100 now SD is fixed. But I did not found your PR for the tests.

vallsv:
As reminded after a talk with @t20100
- [ ] Enable/Disable label of the autoscale mode
- [ ] Use Autoscale instead of Auto
- [ ] Checkbox instead of empty buttons (for min and max only autoscale)

vallsv:
- [ ] Nice to ignore
```
/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py:769: RuntimeWarning: divide by zero encountered in log10
  _data = numpy.log10(_data)
```

vallsv:
@t20100 this can be reviewed

If it is fine i can create issues for:
- Display the lut in the plot in log scale
- Limit histogram computation (only one every a delay)

vallsv:
- [ ] Data range can not be displayed when switching to log
- [ ] This following issue can occurred 

```
ERROR:silx.gui.qt._qt:<class 'AssertionError'>    File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py", line 1227, in _updateNormalization
    self._updateMinMaxData()
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py", line 1028, in _updateMinMaxData
    self._plotUpdate()
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py", line 663, in _plotUpdate
    self._updateLutItem((posMin, posMax))
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py", line 680, in _updateLutItem
    self._bound.setBounds((pos, pos, 0, -0.1))
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/items/shape.py", line 192, in setBounds
    assert rect[2] <= rect[3]
```

vallsv:
Looks like the tests segfaults

vallsv:
I did more rework than i was expected.
But the code is cleaner, and if an item is used, the min/max cache is reused.
But it have to be tested, i do not expect it to work

vallsv:
Was updated by @payno. Can be reviewed.

payno:
Nice rework :+1:

Issue #2739, opened 2019-09-12
[colormap] Add 3*std autoscale mode


vallsv:
The idea is to design it simple (without over-engineering). Then just another specific autoscale.

Basically an function `setAutoScaleMode` enum with `MIN_MAX` or `PLUS_MINUS_3_TIME_STANDARD_DEVIATION`.

If a real need is expressed we can grow the API with objects to allow to custom the values  (3 or something else), or the computation.

t20100:
+1 to keep it simple (and extensible).

PR #2876, opened 2020-01-14
Cached colormap minmax

### LINKED ISSUE #2855: [plot] Cache colormap min/max in scatter/image
This usually have to be computed many times.

It would be good to store the result in the item.

It could became critical with auto min/max based on statistics.

### LINKED ISSUE #2862: [plot] Provide colormap API to reach color from data
There is no obvious way to reach a color from few data from the `Colormap` object.

My main target is to display the color of a single picked data.

`applyToData` is only applied to the full data.

This method could take an optional min/max range.

- This should provide a consistent way with #2855 in order to reduce code overhead.
- And with #2739

vallsv:
This PR allow to cache min/max data for autoscape

This is the first step to proved advanced autoscale.

- Expand the API of `ColormapMixIn`
- Cache colormap min/max based on data and normalization
- The item store the cache, the colormap compute the min/max
- Update `ColorBar` to use items (then the min max is cached)
- Speed up `getNColors` to avoid to compute min/max
- Add a param on `applyToData` to allow to specify another object to compute the autorange
- [x] Unittest

Closes #2862
Closes #2855
Related to #2739


t20100:
I'd keep the compatibility of `ColorBar` with data and keep the `data` argument name, that doesn't sounds complicated and would avoid to break the API.

t20100:
I made a PR on your branch (https://github.com/vallsv/silx/pull/25) with some rework.

t20100:
@vallsv and I each did half of this PR, maybe someone else can do the review, @payno up for it?

vallsv:
There is still a minmax computation here:
```
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/backends/BackendMatplotlib.py", line 629, in addImage
    data = colormap.applyToData(data)
```

Easy to fix, but we have to pass the image item to the backend.
What do you think about?

vallsv:
@t20100, few comments

- I added test for the `applyToData` API
- which BTW returns an image, as it would be better to return the dimension of the `data`, which could be done without breaking the API if a `reference` is set (but it is kind of tricky...) what do you think?
- I have passed the item to the backend (e6a7afc549d30c969bd8f56e0e867075e400c06f) to reuse the cache. It is also kind of tricky, it could be better to just pass the range

What do you think?

vallsv:
@t20100 could you check the unit test and fix it if it is fine?

Maybe due to the call of `_colormapChanged` from `_setColormappedData`, there is unexpected events from the items. But it is maybe normal. But if `_setColormappedData` is called it means something else from the item have changed, and the colormap itself did not change, then it is maybe not needed.

vallsv:
Finally we decided that both extra colormap events, and item argument to the backend was good enough.

t20100:
Giving the item to the backend is not so nice, I'll change that in another PR.

PR #2878, opened 2020-01-14
Check Opengl availability

### LINKED ISSUE #2866: How do I check if I can use the gl backend?
Before requesting the a plot backend (gl or matplotlib), how should I check if the machine has opengl 2.1 ? There are a couple of cases of lower GL version that are causing problems (e.g. rnice -> X-Win32 and windows remote desktop).

t20100:
This PR adds a `isOpenGLAvailable` function in a `silx.gui.utils.glutils` module.
This function checks whether a specific version of OpenGL (2.1 by default as it is used by `silx`) is really available.
It does different checks and end-up running a small Qt OpenGL widget in a subprocess...

TODOs:

- [x] Tested in different conditions
- [x] Add some automated tests? Not sure it is such a good idea....

closes #2866

### DIFF
--- a/silx/gui/_glutils/OpenGLWidget.py
+++ b/silx/gui/_glutils/OpenGLWidget.py
@@ -44,14 +44,8 @@ from .._glutils import gl
 _logger = logging.getLogger(__name__)
 
 
-# Probe OpenGL availability and widget
-ERROR = ''  # Error message from probing Qt OpenGL support
-
-_check = isOpenGLAvailable(version=(2, 1), runtimeCheck=False)
-if not _check:
-    ERROR = _check.error
-    _logger.error('OpenGL-based widget disabled: %s', ERROR)
-    _OpenGLWidget = None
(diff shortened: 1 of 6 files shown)


t20100:
This `isOpenGLAvailable` is used in `PlotWidget` and `silx.gui._glutils.OpenGLWidget` (base class for all OpenGL widgets in silx.
For `PlotWidget`, this makes the fact that one can provide a list of backends truly effective: If one pass `backend=('gl', 'mpl')`, the OpenGL backend will be used only if available.

I decided not to use the subprocess test by default from `PlotWidget` and `silx.gui._glutils.OpenGLWidget`, so, by default, we do not fully test that OpenGL is available...

But `isOpenGLAvailable` caches the runtime checks, so if one calls `isOpenGLAvailable()` before creating any `PlotWidget` and `OpenGLWidget`, the result of this check will be used to probe the availability of OpenGL.

PR #2881, opened 2020-01-15
Fix typo in depth of collapse/expand

### LINKED ISSUE #2879: [silx view] tuple object has no attribute column
```
ERROR:silx.gui.qt._qt:<class 'AttributeError'> 'tuple' object has no attribute 'column'   File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/app/view/Viewer.py", line 382, in __expandAllSelected
    if index.column() != 0:
```

vallsv:
Closes  #2879

Without this fix the action to `expand all/collapse all` is not working.

This could be part of fix for 0.12

### DIFF
--- a/silx/app/view/Viewer.py
+++ b/silx/app/view/Viewer.py
@@ -379,12 +379,12 @@ class Viewer(qt.QMainWindow):
         model = self.__treeview.model()
         while len(indexes) > 0:
             index = indexes.pop(0)
-            if index.column() != 0:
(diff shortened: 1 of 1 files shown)


t20100:
From a look at git blame, this bug is there since v0.9.0, so I would not make a 0.12 minor release just for it, but if we do make a 0.12.1 for other reasons, this should get into it.

PR #2885, opened 2020-01-16
Pan with middle click

vallsv:
This PR allow to pan with middle-click

While the middle button is pressed to can drag the axes of the plot.


vallsv:
I tried to use the `FocusManager` without success. In case help is welcome.

PR #2886, opened 2020-01-20
Update backend following colormap min/max cache

t20100:
! Merge PR #2876 first !

This PR provides some rework of PR #2876 in order to avoid passing the item to the backend.
Instead, for autoscale colormaps, provides a copy of the item colormap with the range set to what the autoscale has computed.
This leverages the item's range cache without changing the backend.

### DIFF
--- a/silx/gui/plot/backends/BackendBase.py
+++ b/silx/gui/plot/backends/BackendBase.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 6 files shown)


vallsv:
LGTM

vallsv:
Smart solution, plus i guess it is the same behaviour for the 3D plot.

t20100:
Yes, in `plot3d` I use another colormap class in the backend, but the idea is the same.

PR #2887, opened 2020-01-20
plot backend clean-up

t20100:
This PR clean-up the `z` argument from plot backend `add*` method which is no longer used.

### DIFF
--- a/silx/gui/plot/backends/BackendBase.py
+++ b/silx/gui/plot/backends/BackendBase.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2004-2019 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2020 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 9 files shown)


vasole:
May I ask how you make sure who is on top?

vallsv:
The backend API is supposed to be private. BTW it is often changed, basically on every releases.

vasole:
The call with z=XXX was not private and it was used to decide what curve was on top of the others for instance.

vallsv:
`z=XXX` from `BackendBase`?

vasole:
OK. If it is kept at the high level interface (PlotWidget) it is fine.

vallsv:
I have fixed the conflict with the master, waiting for CI

vallsv:
There is problem with the installation of setuptools in the Linux CI.
As the tests pass on Windows and the changes are not platform dependant, i merge it.

vallsv:
THe problem is about pip 20.0 -> https://github.com/pypa/pip/issues/7620

t20100:
@vasole , the handling of `z` order is now done by the `PlotWidget`, no longer by the backend, this PR just removed arguments that were no longer used.

PR #2889, opened 2020-01-21
support uint16 RGB images with the OpenGL backend

PiRK:
uint16 RGB seems to be already supported in the OpenGL backend, except for these 2 lines.

I have been testing it for a few days, now.

To test it: 
```
from silx.gui import qt
import numpy

img_rgb = numpy.empty((300, 300, 3), dtype=numpy.uint16)
img_rgba = numpy.empty((300, 300, 4), dtype=numpy.uint16)

max_uint16 = numpy.iinfo(numpy.uint16).max

r, g = numpy.meshgrid(numpy.arange(0, max_uint16, max_uint16 / 300),
                      numpy.arange(max_uint16, 0, - max_uint16 / 300))

b = (r + g) / 2
a = b.T

img_rgb[..., 0] = r
img_rgb[..., 1] = g
img_rgb[..., 2] = b

img_rgba[..., 0:3] = img_rgb
img_rgba[..., 3] = a

app = qt.QApplication([])

pw = {}
for be in "gl", "mpl":
	pw[(be, "rgb")] = PlotWidget(backend=be)
	pw[(be, "rgba")] = PlotWidget(backend=be)
	
	pw[(be, "rgb")].setGraphTitle(be + "uint16 RGB")
	pw[(be, "rgba")].setGraphTitle(be + "uint16 RGBA")

	pw[(be, "rgb")].addImage(img_rgb)
	pw[(be, "rgba")].addImage(img_rgba)

	pw[(be, "rgb")].show()
	pw[(be, "rgba")].show()

app.exec_()
```

### DIFF
--- a/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/silx/gui/plot/backends/BackendMatplotlib.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2004-2019 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2020 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -632,6 +632,9 @@ class BackendMatplotlib(BackendBase.BackendBase):
 
         if data.ndim == 2:  # Data image, convert to RGBA image
             data = colormap.applyToData(data)
+        elif data.dtype == numpy.uint16:
+            # Normalize uint16 data to have a similar behavior as opengl backend
+            data = data.astype(numpy.float32) / 65535
 
         image.set_data(data)
         self.ax.add_artist(image)
--- a/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/silx/gui/plot/backends/BackendMatplotlib.py
@@ -634,8 +634,9 @@ class BackendMatplotlib(BackendBase.BackendBase):
             data = colormap.applyToData(data)
         elif data.dtype == numpy.uint16:
             # Normalize uint16 data to have a similar behavior as opengl backend
-            data = data.astype(numpy.float32) / 65535
-
+            data = data.astype(numpy.float32)
+            data /= 65535
+        
         image.set_data(data)
         self.ax.add_artist(image)
         return image
--- a/silx/gui/plot/backends/BackendOpenGL.py
+++ b/silx/gui/plot/backends/BackendOpenGL.py
@@ -878,6 +878,8 @@ class BackendOpenGL(BackendBase.BackendBase, glu.OpenGLWidget):
 
             if numpy.issubdtype(data.dtype, numpy.floating):
                 data = numpy.array(data, dtype=numpy.float32, copy=False)
+            elif data.dtype in [numpy.uint8, numpy.uint16]:
+                pass
             elif numpy.issubdtype(data.dtype, numpy.integer):
                 data = numpy.array(data, dtype=numpy.uint8, copy=False)
             else:
--- a/silx/gui/plot/backends/glutils/GLPlotImage.py
+++ b/silx/gui/plot/backends/glutils/GLPlotImage.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2014-2019 European Synchrotron Radiation Facility
+# Copyright (c) 2014-2020 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -583,9 +583,14 @@ class GLPlotRGBAImage(_GLPlotData2D):
 
     def prepare(self):
         if self._texture is None:
-            format_ = gl.GL_RGBA if self.data.shape[2] == 4 else gl.GL_RGB
+            formatName = 'GL_RGBA' if self.data.shape[2] == 4 else 'GL_RGB'
+            format_ = getattr(gl, formatName)
 
-            self._texture = Image(format_,
+            if self.data.dtype == numpy.uint16:
+                formatName += '16'  # Use sized internal format for uint16
+            internalFormat = getattr(gl, formatName)
+
+            self._texture = Image(internalFormat,
                                   self.data,
                                   format_=format_,
                                   texUnit=self._DATA_TEX_UNIT)
--- a/silx/gui/plot/backends/glutils/GLPlotImage.py
+++ b/silx/gui/plot/backends/glutils/GLPlotImage.py
@@ -530,7 +530,8 @@ class GLPlotRGBAImage(_GLPlotData2D):
     _DATA_TEX_UNIT = 0
 
     _SUPPORTED_DTYPES = (numpy.dtype(numpy.float32),
-                         numpy.dtype(numpy.uint8))
+                         numpy.dtype(numpy.uint8),
+                         numpy.dtype(numpy.uint16))
 
     _linearProgram = Program(_SHADERS['linear']['vertex'],
                              _SHADERS['linear']['fragment'],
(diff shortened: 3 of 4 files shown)


t20100:
It's fine for OpenGL (though the specification does not guarantee that the image will actually be stored with 16bits per channel...) but it makes a difference with the matplotlib backend that should be solved.

t20100:
I've added a patch to make both backends behave the same.
I also used a size texture format (still no guarantee from the spec, but at least the request is made).

If we integrate this, the behaviour will change for `uint16` RGB(A) images: before they were clipped to [0, 255] and now the full range [0, 65535] will be used.
Other integer types will still be clipped to [0, 255].
That sounds a bit odd.... but I would be OK with it (RGBA image is not our main target and they should be `uint8`).
Any other opinion on that?

An alternative would be for you to convert images to `float32`... but that twice the amount of memory it uses...

t20100:
Except for the question, LGTM

PiRK:
I confirm that the use of float32 for the image sizes I have to deal with is not an option. Keeping `uint16` instead of casting to `uint8` is already a problem for some of my largest images. 

The OpenGl backend is performing very well even with the largest images, once the image is loaded. The bottleneck seems to be loading the image to the GPU. `addImage` can take up to 20 seconds.

vasole:
I do not know the size of the arrays but it can be interesting to change the line:

            data = data.astype(numpy.float32) / 65535

by the lines

            data = data.astype(numpy.float32)
            data /= 65535

PiRK:
Here is an example of array size that caused issues. My initial naïve way of casting unint16 to uint8 caused a MemoryError because of the intermediate float32 array.
```
img = numpy.ones((29632, 60810, 3), dtype=numpy.uint16) 
multiplier = numpy.iinfo(numpy.uint8).max / numpy.iinfo(numpy.uint16).max
img = (img * multiplier).astype(numpy.uint8, order="C")
```

I initially fixed it with:
```
multiplier = numpy.iinfo(numpy.uint8).max / numpy.iinfo(numpy.uint16).max
numpy.multiply(img, multiplier, out=img, casting="unsafe")
img = img.astype(numpy.uint8, order="C")
```

And then later I gained a x4 factor with numba:
```
@numba.njit(parallel=True)
def uint16_to_uint8_3d(img):
    """Downscale RGB image array to uint8"""
    multiplier = numpy.iinfo(numpy.uint8).max / numpy.iinfo(numpy.uint16).max
    img_out = numpy.empty(img.shape, dtype=numpy.uint8)
    for i in numba.prange(img.shape[0]):
        for j in range(img.shape[1]):
            for k in range(img.shape[2]):
                img_out[i, j, k] = img[i, j, k] * multiplier
    return img_out
```

But I eventually ran into an issue with an image in uint16 format but with most of the values < 1000. 


> It's fine for OpenGL (though the specification does not guarantee that the image will actually be stored with 16bits per channel...)
That would be a problem...  I need to do some testing on this.

vallsv:
Sorry for my naive question but I don't understand very well why you use any arithmetic operation. Is that normalization using max instead of max+1 makes a big change? Cause basically it's just about reading a single char from a short (use hi-char values, and ignoring low-char).

t20100:
@vasole, I added your patch.

Loading everything in the GPU will reach a wall sooner or later...

BTW, any opinion on `uint16` RGB(A) image support and behavior change?

vallsv:
Here is what i mean to convert from uint16 to uint8

There is no computation:
```
import numpy                                                                                                                   
numpy.set_printoptions(formatter={'int':hex})                                                                                 

a = (numpy.random.rand(10) * 10000).astype(int)                                                                                

array([0x1c91, 0xf25, 0xb9f, 0x1077, 0x559, 0x219a, 0x15e, 0xe60, 0x12dd,
       0x2346])

a.astype(dtype=('u2',[('lo','u1'), ('hi','u1')]))["hi"]                                                                       

array([0x1c, 0xf, 0xb, 0x10, 0x5, 0x21, 0x1, 0xe, 0x12, 0x23], dtype=uint8)
```

PiRK:
Thanks for that solution. It's interesting, I didn't know that it was possible.

It is clever, and is faster than using an arithmetic operation. But I haven't figured out yet how to wrap it with numba. It's still slower than the arithmetic operation with numba:


```
In [13]: import numpy 
In [14]: img = numpy.ones((29632, 30000, 3), dtype=numpy.uint16) 
In [15]: @numba.njit(parallel=True) 
...: def uint16_to_uint8_3d(img):
...:     multiplier = numpy.iinfo(numpy.uint8).max / numpy.iinfo(numpy.uint16).max
...:     img_out = numpy.empty(img.shape, dtype=numpy.uint8)
...:     for i in numba.prange(img.shape[0]):
...:         for j in range(img.shape[1]):
...:             for k in range(img.shape[2]):
...:                 img_out[i, j, k] = img[i, j, k] * multiplier
...:     return img_out
...: 
In [16]: %timeit img2 = uint16_to_uint8_3d(img) 
750 ms ± 4.29 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [17]: %timeit img2 = img.astype(dtype=('u2', [('lo', 'u1'), ('hi', 'u1')]))["hi"]
2.28 s ± 36.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) 


In [18]: def uint16_to_uint8_ufunc(img):
...:     multiplier = numpy.iinfo(numpy.uint8).max / numpy.iinfo(numpy.uint16).max
...:     numpy.multiply(img, multiplier, out=img, casting="unsafe")
...:     return img.astype(numpy.uint8, order="C")
...:  
In [19]: %timeit img2 = uint16_to_uint8_ufunc(img)
5.6 s ± 95.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

PiRK:
Similar solution using a view (more readable, imo):
```


In [26]: def uint16_to_uint8_view(img):
...:     return numpy.ascontiguousarray(img.view(numpy.uint8)[..., 1::2])
...: 

In [41]: %timeit img2 = uint16_to_uint8_view(img)
2.58 s ± 165 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
--
```

vallsv:
Sounds weird, the bottleneck should be the memory access. @kif could have an idea?

vallsv:
BTW i still don't understand why you use a multiplier in numba? Your destination bins will not have the same size. You could use `numpy.right_shift(img, 8)`.

PiRK:
`numpy.right_shift` is not supported by numba, but your idea works, and it's slightly faster than the arithmetic solution:

```

In [11]: @numba.njit(parallel=True)
...: def uint16_to_uint8_shift(img):
...:     img_out = numpy.empty(img.shape, dtype=numpy.uint8)
...:     for i in numba.prange(img.shape[0]):
...:         for j in range(img.shape[1]):
...:             for k in range(img.shape[2]):
...:                 img_out[i, j, k] = img[i, j, k] >> 8
...:     return img_out
...:
In [12]: %timeit img2 = uint16_to_uint8_shift(img)
650 ms ± 43.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
--
```

PiRK:
Unfortunately I had to stop converting uint16 to uint8, because it seems most of my  uint16 data files have most of their values in range 0 -- ~1000, except for a few hot pixels at >50 000.
Converting to uint8 causes my "useful" dynamic range to be encoded on 4 or 5 values.

t20100:
What about log normalization in this case?
So you consider support of `uint16` to be useful?

I had a question in this comment: https://github.com/silx-kit/silx/pull/2889#issuecomment-577768068

PiRK:
I don't really understand the question. To me, clipping the int types to 0--255 is equivalent to saying we do not support them at all. So support of uint16 (no more clipping) can only been considered a progress.

But obviously you never know if some weird user is not using what i consider to be a bug as a feature :)

PiRK:
Btw, i searched on the internet, and I found a few PNG files with a RGBA 4x16 bits. So it is not completely exotic.

t20100:
It's just I had no answer, it's fine for me.

t20100:
Ready for review

vallsv:
PNG 16bits channels o_O weird

t20100:
I'll add a test.

t20100:
Test added.

Issue #573, opened 2017-01-23
hdf5widget: allow copying the data path in the clipboard


vincefn:
It would be very useful to be able to copy the path string to the clipboard, e.g. by adding "copy path to clipboard" in the contextual (right-click) menu of entries in the tree view.
This can be used to copy the path in a python script.

vallsv:
That's a good idea. We can do it when we will implement default close/synchronize options

vallsv:
For the 0.4 release, (a pull request is in process) it will be possible to copy the path (with ctrl-c) from the HDF5 view. It can already be useful.

![screenshot from 2017-01-27 13 52 59](https://cloud.githubusercontent.com/assets/7579321/22371641/7b79140a-e498-11e6-809a-b1fa78b2f31e.png)

vallsv:
Let us know it it is enough.

vincefn:
Perfect !

PR #2898, opened 2020-01-22
Add copy-to-clipboard icon to HDF5 view

t20100:
This PR adds a copy to clipboard icon on the right side of each field in the HDF5 view `Hdf5TableView` (used in `silx view`).

There are recurrent questions about how to copy-paste those fields (e.g., PR #2758, issue #573 ...).
This PR proposes an explicit way (through a button) to do so similar to what's currently done on web pages.

Screenshot (see icons on the right side):
Screenshot from 2020-01-22 16-57-54

### DIFF
--- a/silx/gui/data/Hdf5TableView.py
+++ b/silx/gui/data/Hdf5TableView.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2020 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


vallsv:
I really don't think it is a good idea, cause there is too much icons for a very basic feature. And it is non consistent, or you will also add that to the compression cells?

A tooltip make more sense. Or allow to select the text with the mouse (which should not be a big deal).

t20100:
I'd say the key things to copy are in "Path Info". I was even thinking of putting the copy icons only there. For me, an icon for a basic thing that is often used is fine.

People does not find how to copy with `Ctrl-C` by themselves, the icon is an explicit way to provide it, any other way is not.

The selectable text is a good option though, it makes more explicit that you can do `Ctrl-C`.

vallsv:
But if there is no time for better solution, it;s fine. It can be merged.

t20100:
Let's see what we can do with selectable text rather than selectable cells.

vallsv:
AppVeyor looks broken

vallsv:
Superseded by #2903

PR #2901, opened 2020-01-23
[plot] Fix crosshair offset issue with right axis

### LINKED ISSUE #2899: [plot] y-crossair can be displayed at the wrong place
I did not found example how to use time zome.

Then i enabled the datetime axis using
```
plot.getXAxis().setTimeZone("UTC")
```
I used epoch timestamps (which looks to not work properly)

As result the cross-air mouse was not displayed at the proper location (x was right, y was wrong).

Can it be reproducible? Any idea?

t20100:
This PR fixes an issue with the crosshair being displayed with an offset when the right y axis is on.
The issue was due to matplotlib events being received in right Y axes, while crosshair lines are displayed in the left Y axes.

closes #2899

If we make a 0.12.1, this should be included in it.


vallsv:
Thanks for the fix

PR #2902, opened 2020-01-24
fix display of chunks in silx view

PiRK:
The chunks cell did not display anything in silx view. A conversion from tuple to string seems to be necessary.

Capture


vallsv:
Thanks, i will just patch the result to use the shape style

vallsv:
If someone else want to review, i did not put `u""` for strings, i expect to be in Python 3.

PR #2903, opened 2020-01-24
Make text selectable in Hdf5TableView

t20100:
This PR aims at making the text of the `HDF5TableView` selectable, so that it is clearer what should be done to copy it.

It is an alternative to #2898 (as proposed by @vallsv in #2898).

The way it is done sounds like doing arm-twisting to Qt: Using read-only editor and playing around with the way `QlineEdit` automatically set selection to overcome it.

### DIFF
--- a/silx/gui/data/Hdf5TableView.py
+++ b/silx/gui/data/Hdf5TableView.py
@@ -266,7 +266,7 @@ class Hdf5TableModel(HierarchicalTableView.HierarchicalTableModel):
             return cell.span()
         elif role == self.IsHeaderRole:
             return cell.isHeader()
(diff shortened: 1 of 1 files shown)


vallsv:
Looks really nice!
- There is a small glitch with the first click, but frankly it is much better that before.
- Except numbers, which are edited using spinbox :-)

Ill try to convert everything to string

vallsv:
No, that's really nice :-)

vallsv:
Oups, you was still working on...?

t20100:
No, it's OK for. What I'm concerned with is the implementation, because some of it is working around Qt behaviors...

Issue #2830, opened 2019-12-13
[plot] Allow to add custom item in the plot


vallsv:
This proposal is there to improve OO programming with the plot.

- The current way to manage "legend" creates a lot of overhead in the code
   - `addThing` sometimes updates an existing item, sometimes not
   - Then a legend always have to be provided, which have to be unique
   - The kind is also always needed to remove an item
   - Some parameters are not provided by the functional API, then the item anyway have to be created with the functional API, then the object have to be reached using the legend, then the item can be updated with the desired arguments
- The "kind" limiting custom third party objects (exception if nothing fit)
- There is no public API to dead with plot items (`add` and `remove` are private)

Few changes could make the life easier.
- [x] Create `addItem(item)` and `removeItem(item)`. This 2 functions will not deal with legend/kind. An object is unique.
- But `addItem` is already used for shapes, then...
- [x] Copy `addItem` as `addShape`
- [x] Update `addItem` to support both single item or shape arguments
   - The shape arguments can be deprecated in the next release
   - And can be removed the release after
- [ ] Create public `setActiveItem` which takes an item

This should be enough to provide a cleaner API, without breaking the legend for the functional API.

Here what i expect to have
```
item = MyShape()
item.setThing()
plot.addItem(item)
self._item = item  # store the item somewhere for further use
...
plot.removeItem(self._item)
```

vasole:
Well, the uniqueness of the legend is not a problem. The id of the item is unique so, it is enough to set the legend to the item's id to solve most of your caveats.

t20100:
This deserves a reference to #2

vallsv:
For me the last missing thing is a global signal for the active item. I will create a dedicated issue for that.

PR #2904, opened 2020-01-24
Plot item objet API

t20100:
This PR completes the objet orient plot item API by making `add|removeItem` the public version of `_add` and `_remove`, so one can instanciate and item and add it to the plot.

With `sigItemAdded`, `sigItemAboutToBeRemoved` and `getItems` to me, the API is completed.
Compatibility of `add|removeItem` is maintained (with a deprecation).

TODO:  handle item legends: init with a unique default legend.

### DIFF
--- a/examples/fftPlotAction.py
+++ b/examples/fftPlotAction.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2020 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -110,7 +110,7 @@ class FftAction(PlotAction):
         for curve in allCurves:
             x = curve.getXData()
             y = curve.getYData()
-            legend = curve.getLegend()
+            legend = curve.getName()
             info = curve.getInfo()
             if info is None:
                 info = {}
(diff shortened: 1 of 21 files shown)


vallsv:
Just a note, i was expecting `_add` (using by `addImage`, `addCurve`...) to enforce legend, while `addItem` could be able to handle object without legend. Is that still doable with this rework?

vallsv:
FYI i am evilly using this private API:
```
plot._add
plot._setActiveItem
item._setLegend
```

t20100:
@vallsv, I kept the private API methods (`plot._add`, `plot._remove`, `item._setLegend) with deprecation warnings.
This PR is NOT changing the need for a unique legend by kind of item in a `PlotWidget`, it just changes the API (while keeping compatibility with the previous one).
This should be doable for `addItem`, but it probably tricky, so it's better to check it and do it separately.

For consistency with ROIs API, I added the `Item.getName` and `Item.setName` methods that are respectively equivalent to `Item.getLegend` (also kept) and `Item._setLegend` (deprecated).
`Item.setName` can only be called on items that are not yet in a `PlotWidget`, else they raise an exception.

t20100:
This is ready for review.

vallsv:
`setName` is nice idea, while the legend can be updated with `id(item)` as said by Armando. I can check that in a few days. But i can't update my project to use that until the integration of the next release.

t20100:
`Item._setLegend` remains available.

vallsv:
COuld you please update the conflict?

vallsv:
Looks good to me. I mean i tested flint with, without problems.

t20100:
Conflict fixed.

vallsv:
Let's try it

t20100:
related to #2830

PR #2909, opened 2020-01-28
Pan with middle button

t20100:
This PR adds pan with middle button for all available interaction modes.
This is achieved by extending the `ClickOrDrag` class to configure which button supports interaction.
The structure of the interaction code can be improved (it's a bit of a mix of things now), but the feature is there.

superseed PR #2885


t20100:
Fixed state machine tests.

vallsv:
LGTM

PR #2912, opened 2020-01-30
Plot scatter histogram visualization mode

### LINKED ISSUE #1934: [scatter view] Add a density map mode
Would be nice to be able to display large scatter plots as density map.

t20100:
This PR adds a histogram visualization mode to the plot `Scatter` item.
It displays the histogram of the scatter points.
The `ColorBar` and `ColormapDialog` uses the info from the histogram, while other features (mask, profile, position info) are based on the points.
For the position info in `ScatterView`, the point closest to the mouse in the histogram bin under the mouse is used.

closes  #1934

TODO in other PRs:
- Compute automatically shape of histogram (now (100, 100) by default)
- Provide more aggregation method (now weighted histogram, but could be other binned statistics).


vallsv:
I checked it using `./bootstrap.py examples/scatterview.py`.

But i dont know if it is related
```
ERROR:silx.gui.plot.tools.PositionInfo:Error while converting coordinates (192.000000, 178.000000)with converter 'X'
ERROR:silx.gui.plot.tools.PositionInfo:Traceback (most recent call last):
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/tools/PositionInfo.py", line 272, in _updateStatusBar
    value = func(xData, yData)
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/utils/weakref.py", line 178, in __call__
    return fn(*args, **kwargs)
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/ScatterView.py", line 220, in _getPickedX
    picking = self._pickScatterData(x, y)
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/ScatterView.py", line 182, in _pickScatterData
    lambda item: isinstance(item, items.Scatter))
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/PlotWidget.py", line 2975, in _pickTopMost
    for result in self.pickItems(x, y, condition):
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/PlotWidget.py", line 2957, in pickItems
    result = item.pick(x, y)
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/plot/items/scatter.py", line 643, in pick
    row, col = result.getIndices()[0]
TypeError: 'NoneType' object is not subscriptable
```

vasole:
@t20100 Perhaps that happens on empty bins?

t20100:
Thanks for the tests: issue is with matplotlib backend, the OpenGL one works fine.
I did develop while testing with the OpenGL backend because I was concerned by handling a lot of points and I didn't checked with matplotlib in the end.

t20100:
There was a consistency issue in the backends concerning picking of the images.
PR #2913 solves this and should be merged before this PR.

t20100:
Ready for review. It's better to merge PR #2913 first to get a cleaner diff.

t20100:
BTW, this PR just changes the display and the colorbar/colormap dialog. Profile, picking, masks are all performed on the original scatter. This can be missleading.

In the end, it's probably best to compute the histogram and display it as an image with the image tools.
Let's see.

t20100:
Rebased to master.

vallsv:
There is again a conflict

vallsv:
BTW, i don't have time to review it.

t20100:
Fixed conflict.

vasole:
Unless I am missing something, I do not find an entry in the test suite or something in the examples or an action to access the new functionality.,

vallsv:
I think it is in only in the scatterview widget, you can select it in the display mode action (from the toolbar).

vasole:
It is what I expected, but I cannot find the action there. I am running the scatterview example.

vasole:
Am I correct if I interpret that the implementation bins the amount of signal falling into the histogram bins and not the amount of points? Two points with signal z1 and z2 falling on the same bin will become a point with associated value z1+z2 and not a value of 2?

t20100:
There is a second PR following this one to add more options: see PR #2923

PR #2913, opened 2020-01-31
Consistent plot backend image picking

t20100:
This PR improves the implementation of image picking in the `PlotWidget`.
Beforehand, getting the picked array indices was done in the `Image` item itself, and the backends had different behaviours: the OpenGL was also doing it, but not the matplotilb one.

This PR makes both backends returning the picked indices in the array and removes the implementation in the `Image` and `Scatter` items. This avoids duplicating the image picking in each item that needs it.

### DIFF
--- a/silx/gui/plot/PlotWindow.py
+++ b/silx/gui/plot/PlotWindow.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2004-2019 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2020 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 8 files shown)


vallsv:
Does it mean it could break the picking? Or it is just a refactoring?

t20100:
It's a refactoring, nothing should change at `PlotWidget`, `items` level.

vallsv:
There is few errors in the CI. It looks related to the indices.

t20100:
CI issues was related to an issue with picking in matplotlib backend.

t20100:
Ready for review, needed by #2912

vallsv:
Tested again with Flint without problem.

PR #2917, opened 2020-02-06
Remove build of python2 packages for debian

t20100:
This PR removes the build of python2 packages for debian 9 and 10.
It was causing issues with pyFAI nightly builds.


kif:
LGTM

PR #2923, opened 2020-02-06
Add option to select the reduction function for "histogram" of scatter plot

t20100:
Merge PR #2912 first!

This PR adds the option to select the reduction function (mean, count, sum for now) for the histogramming visualization of the scatter plot.

This PR also renames the "histogram" visualization mode to "binned_statistic" which is more generic.

### DIFF
--- a/silx/gui/plot/PlotToolButtons.py
+++ b/silx/gui/plot/PlotToolButtons.py
@@ -482,6 +482,18 @@ class ScatterVisualizationToolButton(_SymbolToolButtonBase):
                         {Scatter.VisualizationParameter.BINNED_STATISTIC_FUNCTION: reduction}))
                     submenu.addAction(action)
 
+                submenu.addSeparator()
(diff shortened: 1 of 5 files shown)


t20100:
Fixed conflict.

t20100:
Rebased to master after merge of PR #2912 and to fix conflicts, ready for review.

payno:
works well here.

Maybe an interesting improvement for the future would be to get Information from gui to know the active visualization mode (irregular grid, binned statistics / mean ...)

We could change selected QAction background color or name (adding a tick) or the QMenu tooltip (not a big fan of this one but can be an idea)

t20100:
Thanks for the review/merge!

@payno, I wanted to do that, but it breaks the "genericity" of the actions, because there can be more than one scatter in the plot and they can have different visualization mode... So the action is setting all scatter to a particular mode but cannot provide the information...

That being said, it should be possible to do so for the `ScatterView` widget where there should not be many scatters...

PR #2925, opened 2020-02-12
[plot] Replace usage of PlotWidget's _getAllMarkers and _getItems by getItems

t20100:
This PR replace the use of `PlotWidget._getAllMarkers` and `PlotWidget._getItems` by `PlotWidget.getItems` and deprecates the former private methods.

Following PR #2904, this reduces the amount of methods used and moves towards using the item-based API internally.

### DIFF
--- a/silx/gui/plot/ItemsSelectionDialog.py
+++ b/silx/gui/plot/ItemsSelectionDialog.py
@@ -141,20 +141,22 @@ class PlotItemsSelector(qt.QTableWidget):
     def updatePlotItems(self):
         self._clear()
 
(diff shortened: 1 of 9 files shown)


t20100:
Added removing `_itemKey` (replaced by `_itemKind` and `Item.getName`) + avoid to directly use `PlotWidget._content`, use `PlotWidget.getItems` instead.

t20100:
Fixed conflict.

vallsv:
LGTM, hairy PR.

t20100:
;) thanks for the review!

PR #2930, opened 2020-02-19
Fix fabio deprecated API

vallsv:


kif:
LGTM

t20100:
Does that implies updating the minimal required version of `fabio`?
Currently `"fabio>=0.7"` in `setup.py`.

vallsv:
It's better to do it anyway yes.

t20100:
Done in PR #2937

PR #2932, opened 2020-02-20
PlotWidget XAxisExtent and YAxisExtent items

### LINKED ISSUE #2926: [plot] Create BoundRect item only for X or Y
This item will help me a lot.

Right now i have to delay computation in really sad way to provide something similar to Flint.

t20100:
This PR adds 2 items to the `PlotWidget` items: `XAxisExtent` and `YAxisExtent` which allows to define data extent along each axis to be taken into account by reset zoom. Those items do not display anything.
Those are similar to the `BoundingRect` item except that they only interfere with the data range on a single axis.

closes #2926

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -598,9 +598,13 @@ class PlotWidget(qt.QMainWindow):
         elif isinstance(item, items.Scatter):
             kind = 'scatter'
         elif isinstance(item, (items.Marker,
-                               items.XMarker, items.YMarker)):
+                               items.XMarker,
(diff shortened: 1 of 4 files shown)


vallsv:
I ask myself, cause i check the `testBoundingRectWithLog` to have in log, a valid x (cause it's linear), and an invalid y (cause it is negative). Then maybe it is better to have a single `BoundingRect` with a stronger rules (especially to go back to the right linear values, if the axis was switch to log in between).

Anyway, it's just a question. I merge it.

t20100:
I don't get what you mean, we can discuss that if you want.

PR #2933, opened 2020-02-20
Avoid to move profile after first display

### LINKED ISSUE #2931: Profile window location is reset every image update
In Flint, with a slow frame rate, 1 second, if i display the profile window, the window location is reset after each image update.

I am able to move the window, but the location is set back at the next image update.

This could be a behaviour introduced by Flint it-self.

This could be reproducible with the threaded image example, else the problem is on my side.

t20100:
This PR avoids to move the profile window after the first display.
This is to make it friendly with live update of the plot.

closes #2931

### DIFF
--- a/silx/gui/plot/Profile.py
+++ b/silx/gui/plot/Profile.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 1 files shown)


vallsv:
Wooow, this code looks really complex. I trust you.

Issue #2929, opened 2020-02-18
[Py38] warning about silx


kif:
I started to work on Ubuntu 20.04 and got those warning
```
/usr/lib/python3/dist-packages/silx/examples/plotCurveLegendWidget.py:89: SyntaxWarning: "is" with a literal. Did you mean "=="?
  curve.setYAxis('left' if yaxis is 'right' else 'right')
/usr/lib/python3/dist-packages/silx/gui/plot/ColorBar.py:856: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif self._forcedDisplayType is 'std':
/usr/lib/python3/dist-packages/silx/gui/plot/ColorBar.py:858: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif self._forcedDisplayType is 'e':
/usr/lib/python3/dist-packages/silx/gui/plot/CurvesROIWidget.py:413: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if self.roiTable.rowCount() is 0:
/usr/lib/python3/dist-packages/silx/gui/plot/CurvesROIWidget.py:706: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if len(activeItems) is 0:
/usr/lib/python3/dist-packages/silx/gui/plot/CurvesROIWidget.py:1258: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if x.size is 0:
/usr/lib/python3/dist-packages/silx/gui/plot/PlotWidget.py:1580: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if kind is 'all':  # Replace all by tuple of all kinds
/usr/lib/python3/dist-packages/silx/gui/plot/PlotWidget.py:2118: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if kind is 'all':  # Replace all by tuple of all kinds
/usr/lib/python3/dist-packages/silx/gui/plot/StatsWidget.py:1374: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if len(items) is 1:
/usr/lib/python3/dist-packages/silx/gui/plot/StatsWidget.py:1416: SyntaxWarning: "is" with a literal. Did you mean "=="?
  _item = items[0] if len(items) is 1 else None
/usr/lib/python3/dist-packages/silx/gui/plot/backends/BackendOpenGL.py:974: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if linestyle is not '-':
[...]

PR #2934, opened 2020-02-20
fix py3.8 DeprecationWarning: invalide escape character

t20100:
This PR removes a few more deprecation warning related to python 3.8

related to #2929


vallsv:
LGTM

PR #2946, opened 2020-02-28
Added StackViewPLugin for QtDesigner

alexmarie78:
Hello, 

I am currently working with Qt Designer on a project, still I use a lot of silx widgets also.

Thus I tried to create a plugin to have one of silx widget in Qt Designer.
I made up this branch (in my fork) because I may create other plugins for widgets that I need.

I think it could be nice to append this to the silx repository.

However I don't know how really it works, I just created it by using the already existing plugins.
So if anything seems wrong or not understand, I would like to have some help, or advices.

Thanks in advance

Alexandre

### DIFF
--- a/qtdesigner_plugins/stackviewplugin.py
+++ b/qtdesigner_plugins/stackviewplugin.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python
+# coding: utf-8
+# /*##########################################################################
+#
+# Copyright (c) 2015-2020 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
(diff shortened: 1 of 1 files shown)


vallsv:
Hi,

Few comments:
- You can remove `PyQt4`, we do not support it anymore
- In case, you also can use "promote widget" from Qt designer, which no not require any plugins.

alexmarie78:
Thanks for your answer,

I think it could be nice to have these as plugins so anyone that uses Qt Designer could use it too.

If you want I can clean the already existing plugins from PyQt4 and refresh the dates.

I will try to "promote widget" as you adviced me, but I am still up to create new plugins for enhance silx if you want to.

t20100:
Thanks for the contribution.

A few comments:
```
    def icon(self):
        return icons.getQIcon('stack-view')
```
This is likely to fail as it is looking for an icon file in `silx/resources/gui/icons/` and there is no `stack-view.svg` or `stack-view.png` file there.
You can use `'plot-window-image'` instead.

There is a little bit of doc on Qt designer plugins here: https://www.riverbankcomputing.com/static/Docs/PyQt5/designer.html#writing-qt-designer-plugins

Also, feel free to use the name of your institution as the copyright of things you add.

t20100:
Hi,
I'm closing this PR because development on it is stalled and it has conflicts since PR #3991.
Do not hesitate to rebase the branch and re-open the PR if needed.

Issue #2911, opened 2020-01-29
[plot] Visualize compound data


jln-ho:
I have a ton of HDF5 files containing compound arrays where the first field is a timestamp and the second field/column is a sensor value sampled at that time.

It would be really nice if two-dimensional, numeric compound arrays could be visualized as a time series line graph in the "Curve" tab. Or basically any combination of two columns of a compound array. [HDFView](https://www.hdfgroup.org/downloads/hdfview/) already does this, albeit poorly (no zooming, takes forever with larger datasets, no grid lines ... the list goes on).

I could try to add this functionality and submit a PR...
Any pointers as to where the best place to start could be?

vallsv:
Hi Julian,

We have a set of `DataView` to deal with different kind of visualization (you can check `silx.gui.data.DataViews`). It receives the raw dataset and check if it can be displayed with, then provides a factory method to create the widget and to feed the data.

You can try to copy-past  `_Plot1dView` with your own name. You also can check `examples/customDataView`.

I dont know if it is generic enough to be merge in silx, but you could be able to create your own viewer based on this.

For the time, How plot X axis provides a time based.
```
plot.getXAxis().setTickMode(TickMode.TIME_SERIES)
```

Hope it can help.

jln-ho:
I'll look into it, thanks!

jln-ho:
So I got a first version that can detect compound arrays. It creates a custom ``PlotWindow`` with a small toolbar where the user can select the field for the X and Y axis respectively (only numeric fields will be selectable).

To me it seems like this could be generic enough to be merged into ``master``. Would you be willing to review a PR later this week? Just asking because, frankly, if you're not interested at all, I'll just keep using my quick and dirty solution for now without refactoring, comments, etc.

vallsv:
Nice :-) You could at least provide a screenshot, and push it somewhere so that it can be tested. Just think about creating an example or recipe to generate a dataset. With git and github, you can clean up the stuff later, there is no problem.

I will not have time to review anything this week, but there is other people around.

jln-ho:
My colleague just submitted a PR incl. some test data. I'd be glad to hear your (or anyone else's) thoughts on this.

t20100:
closed by #2948

PR #2948, opened 2020-03-04
#2911 Add support for visualizing compound data as 2D plots

JuliusHarald:
See #2911 for details. 

Attached file [testdata.h5.zip](https://github.com/silx-kit/silx/files/4286686/testdata.h5.zip) contains three datasets covering the following scenarios:
* ``1_numeric_columns``: Compund array with only one numeric column cannot be visualized
* ``2_numeric_1_object_column``: Compund array with two numeric columns: only the numeric columns will be selectable for visualization
* ``3_numeric_columns``: Compund array with three numeric colums to showcase how different columns/signals can be selected for the x/y axes on the fly

### DIFF
--- a/silx/gui/data/DataViewer.py
+++ b/silx/gui/data/DataViewer.py
@@ -150,6 +150,7 @@ class DataViewer(qt.QFrame):
             DataViews._Plot3dView,
             DataViews._RawView,
             DataViews._StackView,
+            DataViews._Plot2dRecordView,
         ]
         views = []
         for viewClass in viewClasses:
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2019 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2020 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -844,6 +844,7 @@ class _Plot2dRecordView(DataView):
         self._data = None
         self._xAxisDropDown = None
         self._yAxisDropDown = None
+        self.__fields = None
 
     def createWidget(self, parent):
         from silx.gui import plot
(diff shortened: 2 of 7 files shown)


t20100:
Great, thanks for the contribution!

I'll look at that more in details next week and provide feedbacks.

One quick question: Do you plan on using the `RecordPlot` widget outside of `silx view`?

If not, it's probably better to keep it private for now and move it to the public API later.
If you do plan on using it as a standalone widget, then we'll check the API to be as consistent as possible to existing widgets and it would need some docstrings and a basic test.

JuliusHarald:
The plan is to just use it for silx view.

t20100:
OK, then it is best to move everything you added to `silx.gui.plot` into a module in `silx.gui.data` (e.g., `silx/gui/data/_RecordPlot.py`) for now.

I made a small PR on your branch (https://github.com/JuliusHarald/silx/pull/1) with a few minor updates. I added an option to not use data from a field for X. Does that sounds a good idea?

Finally, the selection for X and Y would be better placed at the bottom of the plot since "Axis selection" is located just below when needed.

t20100:
Also a question though for later eventually, wouldn't that be useful to be able to select multiple fields for `Y`?

JuliusHarald:
> Add an option to select no field for "X" and use it as default if no time field. What do you think of making it the default?

For our datasets it is the best way to take time as the default, because we almost always have such a field and it makes the most sense to set time for the X axes. Having the option to select no field for "X" is a good idea though.

t20100:
OK great, I left "time" as the default X if available.

JuliusHarald:
> Also a question though for later eventually, wouldn't that be useful to be able to select multiple fields for Y?

Our idea was it to have multiple tabs with different graphs

t20100:
OK, we considered having multiple tabs to open multiple datasets at the same time, but we never did it. There's probably quite a bit of work, let's discuss that in another issue/PR if need be.

jln-ho:
> > Also a question though for later eventually, wouldn't that be useful to be able to select multiple fields for Y?
> 
> Our idea was it to have multiple tabs with different graphs

IMO these are two independent features. It would be nice to have both, actually. But I agree, it's for another issue/PR.

t20100:
I removed `RecordPlot` exposition in `silx.gui.plot`, the idea is to keep it private.

LGTM, thanks for the contribution!

PR #2960, opened 2020-03-18
FitAction: Added support of auto update of fitted data and x range

t20100:
This PR provides a refactoring and extension of the plot `FitAction` in order to add:
- A mode where fitted data is updated when the active curve changes (`FitAction.set|isFittedItemUpdatedFromActiveCurve(bool)`).
- A mode where fitted range is updated when the X limits of the plot changes (`FitAction.set|isXRangeUpdatedOnZoom`).

The default is consistent with the previous behavior (i.e., no update).

What this PR does NOT change:
- The `FitWidget` is never modal
- The `FitWidget` is never reset
- The way the displayed result in the plot is managed
- Exception when estimate/fit goes wrong 

closes #2953

PR #2961, opened 2020-03-19
Some fixes and minor improvements for FitAction

### LINKED ISSUE #2952: [fit] Exception while fitting curves
Some feedback from Flint.

The following one could be catched and  displayed with a convenient message.
I really think the backtrace is pointless.
Screenshot from 2020-03-12 11-52-09

The following one should be investigated. It is maybe due to my integration.
Screenshot from 2020-03-12 11-44-01

t20100:
!!! Merge PR #2960 first !!!

This PR provides:
- Simpler message in pop-up when estimate/fit fails.
- Fix issue when plot is cleared while the user is choosing the curve to fit.
- Deprecates `FitAction`'s `x`, `y`, `xlabel` and `ylabel` attributes.
- Adds `FitAction.get[X|Y]Data` methods.
- Hides the `FitWidget` when the user `Cancel` the curve selection.
- Makes `handle_signal` resilient to being called while it should not.

closes #2952

### DIFF
--- a/silx/gui/fit/FitWidget.py
+++ b/silx/gui/fit/FitWidget.py
@@ -512,10 +512,12 @@ class FitWidget(qt.QWidget):
                 msg.setWindowTitle('FitWidget Message')
                 msg.exec_()
                 return
-        except:    # noqa (we want to catch and report all errors)
+        except Exception as e:    # noqa (we want to catch and report all errors)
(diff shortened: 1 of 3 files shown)


vallsv:
Do you think you can rebase it? To make the review easier?

t20100:
Yes, done, ready for review.

vallsv:
LGTM

PR #2964, opened 2020-03-19
[plot] Use set_frame_on to keep grid while hiding axes

### LINKED ISSUE #2943: [plot] If axes are not displayed, grid config is ignored
It would be good to provide one or the other independently.

t20100:
This PR uses matplotlib `Axes.set_frame_on` rather than `set_axis_on|off` to show/hide the axes as it keeps the grid.

closes #2943


vallsv:
LGTM

PR #2965, opened 2020-03-19
Fix PlotWindow tabbed dock widget management

### LINKED ISSUE #2940: [plot] addTabbedDockWidget do not manage removed docks
This API stores docks inside `_dockWidgets`, but i see nowhere a function to clean up this list.

As result we can add docks, but we can't remove them.

I use to create ROI docks that i destroy. At the second call, the layout is broken, cause the code tries to tabbify the new one, on the previous one which not anymore own by the plot, or whatever.

t20100:
This PR fixes the handling of tabbed dock widget in the `PlotWindow` by overriding `removeDockWidget` and adds a small test  for it.

closes #2940

### DIFF
--- a/silx/gui/plot/PlotWindow.py
+++ b/silx/gui/plot/PlotWindow.py
@@ -464,7 +464,7 @@ class PlotWindow(PlotWidget):
         """Add a dock widget as a new tab if there are already dock widgets
         in the plot. When the first tab is added, the area is chosen
         depending on the plot geometry:
-        it the window is much wider than it is high, the right dock area
+        if the window is much wider than it is high, the right dock area
         is used, else the bottom dock area is used.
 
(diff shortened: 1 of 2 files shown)


t20100:
Random segfaults in the test with `PySide2` apparently in `gc.collect()` (https://github.com/silx-kit/silx/blob/master/silx/gui/utils/testutils.py#L171).

vallsv:
I have restarted the job. There was a segfault in travis Python 3.5 PySide2

vallsv:
Oups, dup.

t20100:
Actually due to calling `self.plot.layout()`.... weird

t20100:
should be fixed now.

vallsv:
Tested with Flint

PR #2969, opened 2020-03-20
Fix FitAction support of Y right axis 

### LINKED ISSUE #2968: [plot] Fit action do work on y2
It looks like curves from y2 can't by fit.

It would be good to allow that and to create the fitted curve inside the y2 axis.

t20100:
This PR makes the `FitAction` work with curves on the right Y axis.
It also deprecates the `legend` attribute.

closes #2968

### DIFF
--- a/silx/gui/plot/actions/fit.py
+++ b/silx/gui/plot/actions/fit.py
@@ -47,7 +47,6 @@ from .. import items
 from ....utils.deprecation import deprecated
 from silx.gui import qt
 from silx.gui.plot.ItemsSelectionDialog import ItemsSelectionDialog
-from silx.gui.plot.items import Curve, Histogram
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


vallsv:
Tested, approved. I have just restarted the CI.

vallsv:
Fail again. Looks to be a segfault on PySide2 again.

t20100:
Yes and not reproduced locally....

t20100:
No way to reproduce locally so far, that is going to be painful...

t20100:
Now tests pass with `PySide2` on Windows but fail with `PyQt5` everywhere...

t20100:
Looks to be fixed! Ready for review.
It was a funny weird bug....

vallsv:
Good time to switch to PySide2 then?

vallsv:
LGTM

PR #2970, opened 2020-03-20
Fix handling of PlotWidget background color

### LINKED ISSUE #2967: [plot] Data background is not used when plot do not display the axes
I think it is a regression.

- Set both background with a different color
- One is displayed outside the data view (beyond the axes)
- Another one is dislayed under the data (inside the axes)
- Switch the plot mode to hide the axes
- The data background is not anymore used (i think we expect it to be used)

t20100:
This PR fixes a regression on the handling of plot background colors following PR #2964.

closes #2967


vallsv:
Thanks a lot. It works.

PR #2973, opened 2020-03-23
[io][url] fix path() when data_slice is not an iterable.

### LINKED ISSUE #2972: Data.path() fails if data_slice is not an iterable
For now `DataUrl` allows to define data slice as a not iterable object (like an int).
But in this case, call to DataUrl.path() raise an error.

source code to reproduce:
``` python
url = DataUrl(scheme="silx", file_path="/foo.h5", data_slice=1)
url.path()
```
the call of url.path() raises the error:

``` bash
data_slice = ",".join([slice_to_string(s) for s in self.__data_slice])
TypeError: 'int' object is not iterable
```

payno:
if data_slice is not an iterable simply convert it to str instead of calling '",".join([slice_to_string(s) for s in self.__data_slice])' which fails if data_slice is not an iterable

closes #2972

### DIFF
--- a/silx/io/url.py
+++ b/silx/io/url.py
@@ -30,6 +30,7 @@ __date__ = "29/01/2018"
 
 import logging
 import six
+from collections import Iterable
 
 parse = six.moves.urllib.parse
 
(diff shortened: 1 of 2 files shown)


silx/io/url.py
```diff
@@ -297,7 +298,10 @@ def slice_to_string(data_slice):
             if self.__data_path is not None:
                 queries.append("path=" + self.__data_path)
             if self.__data_slice is not None:
-                data_slice = ",".join([slice_to_string(s) for s in self.__data_slice])
+                if isinstance(self.__data_slice, Iterable):
+                    data_slice = ",".join([slice_to_string(s) for s in self.__data_slice])
+                else:
+                    data_slice = str(self.__data_slice)
```
t20100: Probably best to use `slice_to_string` rather than `str`?

t20100:
Sounds good to roundtrip, maybe best to use `slice_to_string` rather than `str`.

silx/io/url.py
payno: Yes, totally miss that one thanks

PR #2978, opened 2020-03-28
Create an example to custom silx view

vallsv:
This example allow to add custom behavior to silx view.

- Create a hook on the `DataViewer` to allow to custom the widgets in the application side
- Create a silx view factiory for `Viewer` (the main window) and `ApplicationContext` (a single instance object which basically share the context of the application)
- Create an example to add an action to custom the `Plot1D` background color

Capture d’écran de 2020-03-28 12-12-49

It can be tested this way:
```
./bootstrap.py examples/customSilxView.py foo.h5
```
Or this way (i guess)
```
python -m silx.examples.customSilxView foo.h5
```

### DIFF
--- a/examples/customSilxView.py
+++ b/examples/customSilxView.py
@@ -1,0 +1,60 @@
+import sys
+import numpy
+
+
+def createWindow(parent, settings):
+    # Local import to avoid early import (like h5py)
+    # SOme libraries have to be configured first properly
+    from silx.gui.plot.actions import PlotAction
+    from silx.app.view.Viewer import Viewer
+    from silx.app.view.ApplicationContext import ApplicationContext
+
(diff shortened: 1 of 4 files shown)


t20100:
This PR also includes the setting of the title for the `Plot1D` view. This would be nice for have for all views! (could be done in another PR).

t20100:
LGTM.
To me it provides the necessary hooks for making a custom `silx view`and could be extended through the already existing `DataViewHooks` class.

vallsv:
I have removed from this PR the custom title, so it can be cleaned up inside another PR.

t20100:
LGTM

t20100:
related to PR #2974

PR #2979, opened 2020-03-30
[silx view] add title to plots

t20100:
This PR adds a title to plots in silx view taking the dataset path and slicing information into account.
This paves the way to have a configurable way of deciding what the title is built from.

Initially proposed in PR #2978 by @vallsv 
related to PR #2974


vallsv:
LGTM

PR #2980, opened 2020-03-30
[plot] Keep profile window always on top

### LINKED ISSUE #2935: [plot] Switch profile windows as dialog
I think it would be better to set our side windows (like the profile window) always over there main parent window.

This can be achieved by using `windowFlags` with `qt.Qt.Dialog` + hint for minimize/maximize.

I guess there is not only the profile window to patch.

Could we do it in silx?

t20100:
This PR ensures the profile window remains on top of the plot window.

closes #2935

### DIFF
--- a/silx/gui/plot/ProfileMainWindow.py
+++ b/silx/gui/plot/ProfileMainWindow.py
@@ -59,7 +59,10 @@ class ProfileMainWindow(qt.QMainWindow):
     sum or mean)"""
 
     def __init__(self, parent=None, backend=None):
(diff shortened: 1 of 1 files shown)


vallsv:
Yet i did not try, but i think this PR will not behave as expected.

I mean it should not be displayed on the top of all the applications. But only on the top of the main window.

t20100:
Yes, you are right, my bad.

t20100:
This is using the `Dialog` flag instead.

vallsv:
Ahahah i was thinking about doing that on my PR, i do not remember it was here ^_^

PR #2981, opened 2020-03-30
Change FitAction behavior in silx view and Plot1D

t20100:
This PR proposes to change the behavior of the `FitAction` in `Plot1D` and `silx view` to use the behaviour of live update introduced in PR #2960.

It also proposes to keep the `FitWidget` always on top (only when started from the `FitAction`), related to #2935.


vallsv:
I think it is better. But `Dialog` should be used instead of `WindowStaysOnTopHint`.

t20100:
You're right, it changed it.

vallsv:
I think this behavior is better

PR #2986, opened 2020-04-02
Use icon for mask tools

vallsv:
This PR rework the tool buttons as an action toolbar.

This allows to reduce a lot the version size.

Capture d’écran de 2020-04-02 14-11-09


t20100:
LGTM.
I made 2 minor improvements:
- Make the tooltip of load and save a bit more explicit.
- Add a stretch area to "Threshold" and "Other tools" groups

If it's fine for you, I'll merge it.

vallsv:
I think the second one can be problematic for me (cause i update the group on flint), but it's fine.
I guess to set it for the horizontal layout? Cause for the vertical one it should not be needed.
You can merge, no problem.

vallsv:
Oh, be careful you did not changed the tooltip, but the text, i think you should revert, and add the tooltip. I mean both text and tooltip should be there.

t20100:
You're right, I fix the text/tooltip now.

Issue #2865, opened 2020-01-08
[plot] Split profile into manager and toolbar


vallsv:
It would be much better to have a stateless toolbar, with all the logic inside another class for flexibility.

That's something which can be done to avoid duplication code between flint and silx.

vallsv:
Closed by  #3011

Issue #2046, opened 2018-07-24
Rework profile toolbar


vallsv:
The Plot2D profile toolbar could be reworked like that
- Display profile kind in a checkable choice drop box
- Hide tools when there is no profile on the image

![rework-plot2d-profile](https://user-images.githubusercontent.com/7579321/43144367-7b56f3d8-8f4c-11e8-8620-a30652cffd3f.png)

Another proposal could be to rework the plot2d as the plot1d toolbar, and only display the profile toolbar if it is requested by a "tools" toolbar.

vallsv:
Closed by  #3011

Issue #1947, opened 2018-06-20
[plot] Make profile tools more flexible


t20100:
It would be nice to make the profile tools more modular, so one can remove some of the buttons and add others while reusing underlying profile computation.

Use case: having vertical, horizontal and profile at 45 degrees instead of free line profile.

vallsv:
Closed by  #3011

Issue #709, opened 2017-04-06
[plot] Profile toolbar refactoring


t20100:
ProfileToolBar._createProfile includes code that is only needed for Profile3DToolBar.
Would be good to refactor so that Profile3DToolBar override _createProfile to add its stuff.

payno:
I don't know how far we should go on this factorization.
First we should create two ProfileToolBar because we have the 2D and the 3D case.

But from what I recall we said we should redo some stuff on the ProfileWindow to group those.

woutdenolf:
Can be closed?

vallsv:
Yes, it was fixed.

PR #2988, opened 2020-04-06
Rework profile tool

### LINKED ISSUE #2955: [plot] Improve profile architecture
Flint integrate profile from silx.

From user feedback, some feature could be implemented, and could also benefit to silx.

But it would be good improve the current implementation first:

After a talk to @t20100 here few comments:

- The scatter profile and the image profile do not use the same architecture
    - Scatter profile uses ROIs (not the image profile)
    - It would be good to use ROIs for both
        - The limitation here is the weight of the line (it does not exist right now for ROIs)
- The main object for both is a toolbar
   - It is a strong limitation on the reusability
   - It would be better to create a dedicated manager as the main object

vallsv:
Closes #2955, #2865

Merge #3018 first

This PR provides a generic way to manage the profiles.

Everything is working, but actually nothing is used it the default silx plots.

So what an example is provided to test the feature.

## Example
```
./bootstrap.py examples/plotProfile.py
```

## Features

It provides the same features as before, but with another architecture.

Here is the changes.

- Use ROIs
- Same way to manage image/scatter/image stack
- Profile properties (line width for example) are managed per ROI, and not globally
- Use thread pool for computation
- New scatter profile to display data without interpolation (for regular scatters)
- Provide a mode to allow many profiles at the same time
- Provide a cross ROI
- Provide a specific computation of RGB images

## Coming next

This could be done.

- Use it in silx (#84, #709, #1947)
    - An integrated toolbar helper (#2046, #3001)
- The profile ROIs uses a specific signal for there properties, it could be nice to reuse the `sigItemChanged` instead
- Move the previous profile code into the new package
    - Move `silx.gui.plot.Profile.createProfile` to `silx.gui.plot.tools.profile.core`?
    - Move `silx.gui.plot.MainProfileWindow` to `silx.gui.plot.tools.profile`?
    - Move `silx.gui.plot.ProfileToolBar` to `silx.gui.plot.tools.profile.toolbar`?
- (Flint) Provide a line widget with the 0 in the first point, and the pixel distance to the last point
- Add an option to deactive the ROI action after a single use
- (Flint) Be sure a Flint integration (using docks is doable)
- A Profile ROI should refer to the plot item (the manager should manage the update to provide the current feature)
- Create single ROI per kinds (hline/vline/line/cross) which can handle any kind of items
    - First merge the math code inside a single `createProfile(roi, item)`
    - Better to wait for groups (the big deal is the specific shape we have to display on the plot)
    - This impose default ROIs to provide all the kind of properties (line width...) for configuration
- Create different profile computation according to the `ScatterItem` visualization mode

## Issues
- A single click on the cross anchor will create a new ROI (while i expect just nothing), i guess that's an issue in the plot interaction (that's cause of the shapes/markers, it have to be fixed but into another PR)
- Updating the style of a marker during an interaction stops this interaction

Capture d’écran de 2020-04-06 08-58-59


vallsv:
I have removed a test cause of that:
```
  File "C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\gui\plot\tools\test\testProfile.py", line 41, in test_init
    self.assertEqual(roi.receivers(roi.sigRegionChanged), 1)
TypeError: 'PySide2.QtCore.QObject.receivers' called with wrong argument types:
  PySide2.QtCore.QObject.receivers(SignalInstance)
Supported signatures:
  PySide2.QtCore.QObject.receivers(bytes)
```

vallsv:
Here is the slicing tool on scatter data

Capture d’écran de 2020-04-07 13-12-50

vallsv:
Ok. I am happy with the result. This PR only create features without using it in the default plots.

Let me know if it is enough.

I can create another PR to fully replace the previous toolbars, or try to be a little more conservative. Let me know.

I can merge few last commits, and move on top commits which changes other aspect of silx (and create another PR with it if you prefer). I also can remove the example if you like (in this end it should not be needed anymore).

t20100:
Let's go step by step it will be already quite huge to review, I'd say polish what you want but don't add features in this PR, create new ones.

vallsv:
So this PR is ready.

vallsv:
The very first commit is de80367, the previous ones can be checked one by one and prepare silx to the feature. I can split it into several other PR if you prefer.

vallsv:
By the way, do not trust the order list of commits displayed here (or in the commits tab) by github, it is sorted by date, but it's pointless for git. The real order of commits is not this one.

vallsv:
This PR is ready for review

vallsv:
This PR contains an example to be able to test it.

It became useless after the next related PR. When this PR became fine, i prefer to remove the commit related to the example. Or let me know.

vallsv:
This PR is again ok for a review. #3018 have to be reviewed first.

silx/gui/plot/Profile.py
```diff
@@ -180,6 +184,7 @@ def createProfile(roiInfo, currentData, origin, scale, lineWidth, method):
     :type scale: 2-tuple of float
     :param int lineWidth: width of the profile line
     :param str method: method to compute the profile. Can be 'mean' or 'sum'
+        or 'dry-run', to compute everything but the profile
```
t20100, line 187: I'd replace `dry-run` by 'none'

silx/gui/plot/ProfileMainWindow.py
```diff
@@ -56,7 +58,10 @@ class ProfileMainWindow(qt.QMainWindow):
 
     sigProfileMethodChanged = qt.Signal(str)
     """Emitted when the method to compute the profile changed (for now can be
-    sum or mean)"""
+    sum or mean)
+
+    Note: This signal should be removed.
```
t20100, line 63: Then we can first remove the docstring and put a comment

t20100: unless we consider `ProfileMainWindow` not being part of the public API.

silx/gui/plot/ProfileMainWindow.py
```diff
@@ -68,41 +73,68 @@ def __init__(self, parent=None, backend=None):
...
+        return self._plot2D
+
+    def _showPlot2D(self):
+        plot = self._getPlot2D()
+        self._layout.setCurrentWidget(plot)
+
     def setProfileType(self, profileType):
         """Set which profile plot widget (1D or 2D) is to be used
 
+        Note: This method should be removed.
```
t20100, line 119: deprecate?

vallsv: Maybe we can talk about it in #3011 cause i reworked everything there

vallsv: I just checked, i put a deprecation in the next PR.

Let me k now, if you prefer, i can remove this comment from this PR.

silx/gui/plot/ProfileMainWindow.py
vallsv: We could create a deprecated signals, with a deprecating property method returning the signal.

vallsv: But i would prefer to talk about it in the next PR.

silx/gui/plot/Profile.py
vallsv: No problem. I will fix it.

vallsv: It's done

silx/gui/plot/ProfileMainWindow.py
t20100: If you've done it in the next PR it's fine.

silx/gui/plot/ProfileMainWindow.py
t20100: OK

t20100:
I made a small PR on your branch.
Otherwise, it looks good to me, we can always adjust in up coming PRs where this will be used.

PR #2991, opened 2020-04-06
Fix out-of-bound row/column indices

### LINKED ISSUE #2990: Issue with position info picking
This issue happens from time to time when displaying an image occupying the whole plot area with a `Plot2D`:
```
ERROR:silx.gui.plot.tools.PositionInfo:Error while converting coordinates (514.000000, 48.000000)with converter 'Data'
ERROR:silx.gui.plot.tools.PositionInfo:Traceback (most recent call last):
  File "silx/gui/plot/tools/PositionInfo.py", line 273, in _updateStatusBar
    value = func(xData, yData)
  File "silx/utils/weakref.py", line 178, in __call__
    return fn(*args, **kwargs)
  File "silx/gui/plot/PlotWindow.py", line 938, in _getImageValue
    value = image.getData(copy=False)[row, col]
IndexError: index 200 is out of bounds for axis 0 with size 200
```
The picking returns an index outside the image range (here [0, 199])....
It seems to only happen with the matplotlib backend.

t20100:
This PR fixes an issue with matplotlib backend returning indices outside of the bounds of the image when on the edge.

closes #2990

### DIFF
--- a/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/silx/gui/plot/backends/BackendMatplotlib.py
@@ -375,8 +375,9 @@ class Image(AxesImage):
             x, y = mouseevent.xdata, mouseevent.ydata
             ox, oy = self.__silx_origin
             sx, sy = self.__silx_scale
-            column = int((x - ox) / sx)
-            row = int((y - oy) / sy)
+            height, width = self.get_size()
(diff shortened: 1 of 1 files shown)


vallsv:
Nice catch. I also noticed it twice recently.

Issue #2835, opened 2019-12-16
[plot] Rework StackView colormap


t20100:
There are inconsitencies in the `StackView` colormap handling (see #2809, PR #2832).
The problem is linked with the need for autoscale on current image and autoscale on the whole stack.
An idea is to have autoscale on the image so colormap handling is as usual, and to add a button to autoscale on the stack on demand (which removes autoscale and sets the colormap range to that of the stack).

vallsv:
Now we can implement that really easily.

- Create a boolean attribute `stackMode`/`fullMode` (something like that) in colormap (that's really ad hoc, but if it is useful)
    - This commit could be used as template 8abb23cf2eca3b49ecfb7a42ef6c78b6b5cdc05d
- Update the autoscale min/max computation
    - If the item is `ImageStack`, add `stackMode` is enabled, the result but be computed using `item.getStackDate` 
- Update the colormap dialog to allow to custom this attribute
    - It could can be displayed in the bottom on the scale --> [x] [auto scale] [3D] [x]
    - It should only be displayed when a `ImageStack` item is selected

vallsv:
Oh, maybe `getColormappedData` itself can be patched, and all the computation is done

vallsv:
Another approach, which looks much easier, could be to create a `ImageStack` property, and to not change at all the `Colormap`.
- `getColormappedData` could be customed for this item
- In this case switching from one one to another could be done with a `StackView` action, and could not be done with a colormap dialog button.

vallsv:
So, sorry, difficult to brainstorm remotely...

Now i think i prefer the second approach, cause it is not an ad hoc solution.

But it would be better to displayed the button in the colormap dialog. Like in the first approach. It is doable, cause the colormap dialog knows the item (in this case). That's kind of border line, in this case, it makes sens to edit in the dialog this specific property of the `ImageStack`.

What do you think?

t20100:
I will need to set it from the dialog.
An item mix-in for this makes the dialog only care about a single kind of class to take into account (be it an `ImageStack` item or something else).

t20100:
Can be fixed by dropping full stack autoscale

PR #2994, opened 2020-04-07
Rework the StackView with a stack image plot item

vallsv:
This is a small refactoring to allow to clean up other part of Silx, basically the profiles (#2988).

This PR creates an `ImageStack` in order to store in the plot a model of the stack.

An `ImageStack` is only an extends of `ImageData`. That's a 3D stack, in which only a single frame is displayed.

Storing the stack in the plot allows to rework other part of the GUI with this extra information. The profile do not need anymore a dependency to this `StackView` view in order to display a profile in the 3D depth. Other tools can also benefit to it, like a generic statusbar, for example.

### DIFF
--- a/silx/gui/plot/items/image.py
+++ b/silx/gui/plot/items/image.py
@@ -42,7 +42,6 @@ import numpy
 from ....utils.proxy import docstring
 from .core import (Item, LabelsMixIn, DraggableMixIn, ColormapMixIn,
                    AlphaMixIn, ItemChangedType)
-from ._pick import PickingResult
 
 
 _logger = logging.getLogger(__name__)
--- a/silx/gui/plot/items/image.py
+++ b/silx/gui/plot/items/image.py
@@ -458,3 +458,82 @@ class MaskImageData(ImageData):
     internal silx widgets.
     """
     pass
+
+
+class ImageStack(ImageData):
+    """Item to store a stack of images and to show it in the plot as one
+    of the images of the stack.
+
+    The stack is a 3D array ordered this way: `frame id, y, x`.
+    So the first image of the stack can be reached this way: `stack[0, :, :]`
+    """
+
+    def __init__(self):
+        ImageData.__init__(self)
+        self.__stack = None
+        """A 3D numpy array (or a mimic one, see ListOfImages)"""
+        self.__stackPosition = None
+        """Displayed position in the cube"""
+
+    def setStackData(self, stack, position=None, copy=True):
+        """Set the stack data
+
+        :param stack: A 3D numpy array like
+        :param int position: The position of the displayed image in the stack
+        :param bool copy: True (Default) to get a copy,
+                          False to use internal representation (do not modify!)
+        """
+        if self.__stack is stack:
+            return
+        if copy:
+            stack = numpy.array(stack)
+        assert stack.ndim == 3
+        self.__stack = stack
+        if position is not None:
+            self.__stackPosition = position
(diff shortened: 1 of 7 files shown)


vallsv:
@vasole any idea how the StackView is used in PyMca? It is used in a very basic way, or you create items inside?

vallsv:
@t20100 i think about creating `getPlotWidget` in order to be consistent with the `ScatterView`? What do you think?

t20100:
+1 for `getPlotWidget`, to me that should be used everywhere (which is by far not the case, `getPlot` is often used)

t20100:
That would also allow to handle colormap autoscale either on the displayed slice or on the whole stack (see #2835).

silx/gui/plot/items/image.py
```diff
@@ -459,3 +458,73 @@ class MaskImageData(ImageData):
     internal silx widgets.
     """
     pass
+
+
+class ImageStack(ImageData):
+    """Item to store a stack of images and to show it in the plot as one
+    of the images of the stack.
+
+    The stack is a 3D array ordered this way: `frame id, x, y`.
```
t20100, line 467: I think there is a typo here: `frame id, y, x`

silx/gui/plot/items/image.py
```diff
@@ -459,3 +458,73 @@ class MaskImageData(ImageData):
...
+    """
+
+    def __init__(self):
+        ImageData.__init__(self)
+        self.__stack = None
+        """A 3D numpy array (or a mimic one, see ListOfImages)"""
+        self.__stackPosition = None
+        """Displayed position in the cube"""
+
+    def setStack(self, stack, position=None):
```
t20100, line 478: I'd call it `setStackData`, in the items most if not all method taking numpy.array of data have names ends with `Data`.
Also, I'd put a `copy=True` argument with a pass-through if False and a copy if True

silx/gui/plot/items/image.py
```diff
@@ -459,3 +458,73 @@ class MaskImageData(ImageData):
...
+            return
+        assert stack.ndim == 3
+        self.__stack = stack
+        if position is not None:
+            self.__stackPosition = position
+        if self.__stackPosition is None:
+            self.__stackPosition = 0
+        self.__updateDisplayedData()
+
+    def getStack(self):
```
t20100, line 494: Same: `getStackData(self, copy=True)`

vallsv:
`getProfileWindow1D` and `getProfileWindow2D` can be removed, the content pointing to nothing, if i am not wrong.

silx/gui/plot/items/image.py
vallsv: Thanks a lot, i was not sure about that (and did not checked). As it is images inside, but in fact y,x is the normal order for image too.

silx/gui/plot/items/image.py
vallsv: Ok. I am not sure for the copy=True, cause we talk about a lot of data, which can still be in files. But it's fine, it's much better to be consistent.

t20100: Yes, you always want to put `copy=False`, but I prefer it to be explicit.

t20100:
Indeed `getProfileWindow1D` and `getProfileWindow2D` should be removed they are broken!

t20100:
macos CI seems to use python2 and fails.... It does not seem to install miniconda.

t20100:
LGTM

PR #2999, opened 2020-04-14
[silx view] When NXdata title is defined, do not append dataset name

### LINKED ISSUE #2998: [NXdata] User supplied title not fully respected
It seems the signal dataset name is appended to the user supplied title. If the user wanted that, (s)he would just add that to the title. 

image

t20100:
This PR avoid appending the dataset name to the plot title when the `title` attribute is defined.

closes #2998

### DIFF
--- a/silx/gui/data/NXdataWidgets.py
+++ b/silx/gui/data/NXdataWidgets.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017-2019 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2020 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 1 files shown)


vasole:
I am hesitating about the behavior to have for the auxiliary signals. My caveat is the data producer has no way to provide the appropriate title for them.

vasole:
Please forget my previous comment. The auxiliary signals should also be displayed.

vasole:
- For 1D data, we can display all the curves simultaneously, therefore only the user specified title is to be provided.
- For 2D data, we are unable to display all the images simultaneously and we use a slider. In that case the previous behavior is appropriate (2D + Aux. signals)

t20100:
Updated to append the dataset signal name when there is more than one.

PR #3000, opened 2020-04-14
Add a isDragged method to plot marker items

t20100:
This PR proposes to add a `isDragged` method to plot marker items.
Indeed  there is 2 signals for start and end of dragging, but no method to retrieve the current dragging state.


vallsv:
Maybe better to use `isDragging`?

t20100:
I would put `isDragged`, but I'm no native English speaker....

t20100:
Renamed to `isBeingDragged`

vallsv:
LGTM

PR #3004, opened 2020-04-15
[profile] Add free line profile

vallsv:
This PR provides a new line ROI to allow to change the behaviour of the current line profile.

- The profile ROI have an origin
- The X profile start from 0 and is displayed in distance
- This can be used easily in diffraction imaging to check the diffraction pattern

This PR is ready, but its better to:
- [ ] Add test coverage

Capture d’écran de 2020-04-15 12-57-16


silx/gui/plot/tools/profile/rois.py
```diff
@@ -258,7 +262,66 @@ def __init__(self, parent=None):
         _DefaultImageProfileRoiMixIn.__init__(self, parent=parent)
 
 
+class ProfileImageLineOriginROI(roi_items.LineROI,
```
t20100, line 265: Naming proposition: `ProfileImageDirectedLineROI`

PR #3006, opened 2020-04-15
[profile] Manager helper to only create a single profile ROI with an action

vallsv:
Merge #2988 first.

This PR provides an helper which allow to unactivate the current ROI action when a ROI was created.

Another approach, which is IMO better, would be create that by the `QAction` themself. For that the ROI manager should create specific actions, and `PlotWidget` must provide an API to allow to select the previous default mode used (`pan` or `zoom`). This can also be useful for other cases like that. If you like the idea i rework this PR.

This PR is ready, but its better to:
- [ ] Rebase to the master
- [ ] What about the second approach?


vallsv:
Now i think it have to be done in the ROI `QAction`. In fact that's a ROI feature, it have nothing to do with profiles. So it would be better to do it this way.

vallsv:
Superseded by #3014 + #3018

PR #3008, opened 2020-04-16
[roi] Allow to select a ROI

vallsv:
This PR allows to select a ROI from the plot.

- A ROI provides `setSelectable` and `setSelectionProxy`
- When an anchor is clicked, or a ROI created, the manager update the current selected ROI
- Plus other small things relative to ROIs
- Create `sigInteractiveRoiCreationFinished` to know when the manager create a new ROI (and is not just added)
- Create `sigAboutToRemove` to allow the ROIs to clean up things

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -374,7 +374,7 @@ class Item(qt.QObject):
 
 # Mix-in classes ##############################################################
 
-class ItemMixInBase(qt.QObject):
+class ItemMixInBase(object):
     """Base class for Item mix-in"""
 
     def _updated(self, event=None, checkVisibility=True):
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -144,6 +144,9 @@ class ItemChangedType(enum.Enum):
     EDITABLE = 'editableChanged'
     """Item's editable state changed flags."""
 
+    SELECTABLE = 'selectableChanged'
+    """Item's selectable state changed flags."""
+
 
 class Item(qt.QObject):
     """Description of an item of the plot"""
--- a/silx/gui/plot/tools/roi.py
+++ b/silx/gui/plot/tools/roi.py
@@ -322,6 +322,7 @@ class RegionOfInterestManager(qt.QObject):
             raise ValueError(
                 'RegionOfInterest does not belong to this instance')
 
+        roi.sigAboutToBeRemoved.emit()
         self.sigRoiAboutToBeRemoved.emit(roi)
 
         self._rois.remove(roi)
--- a/silx/gui/plot/tools/roi.py
+++ b/silx/gui/plot/tools/roi.py
@@ -306,7 +306,7 @@ class RegionOfInterestManager(qt.QObject):
         """Handle ROI object changed"""
         self.sigRoiChanged.emit()
 
-    def createRoi(self, roiClass, points, label='', index=None):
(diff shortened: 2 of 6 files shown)


vallsv:
This PR can be reviewed.

It was extracted from the other PR relative to the profile, in order to make it smaller, and easier to review.

silx/gui/plot/items/core.py
```diff
@@ -374,7 +377,7 @@ def pick(self, x, y):
 
 # Mix-in classes ##############################################################
 
-class ItemMixInBase(qt.QObject):
+class ItemMixInBase(object):
```
t20100, line 380: That's because of some issues with diamond inheritance, right?

silx/gui/plot/tools/roi.py
```diff
@@ -132,7 +146,7 @@ def getSupportedRoiClasses(cls):
 
     # Associated QActions
 
-    def getInteractionModeAction(self, roiClass):
+    def getInteractionModeAction(self, roiClass, parent=None):
```
t20100, line 149: Why do you need to set the parent of the QActions? It can be as risky as using `self` since they can be shared among multiple widgets.

silx/gui/plot/tools/roi.py
```diff
@@ -272,12 +339,23 @@ def createRoi(self, roiClass, points, label='', index=None):
...
+        if label is not None:
+            roi.setName(str(label))
         roi.setFirstShapePoints(points)
 
         self.addRoi(roi, index)
+        if roi.isSelectable():
+            self.setSelectedRoi(roi)
         return roi
 
+    def containsRoi(self, roi):
```
t20100, line 351: Is this really needed? `roi in manager.getRois()` would do the job.

t20100:
I made a small PR on your branch and I have a few questions, but overall it looks good to me!

silx/gui/plot/items/core.py
vallsv: Yes. It looks to work, but in the end you have double execution of some `__init__`, or Qt signals of instancies are not well initialized, that's tricky.

silx/gui/plot/tools/roi.py
vallsv: This function is in fact an action factory, the parent should be provided. It least you should be able to chose. When i create an action i want it to be well parented (to a menu or whatever).

I don't know how Qt deal with actions, i guess it can trig destroy to update the menus, but right now I did not find any problems to do so.

silx/gui/plot/tools/roi.py
vallsv: Yes. That avoid a copy, and as it is explicit, when you use it you don't have to think about.

t20100: Creating a tuple from a list doesn't count much, I doubt we'll have millions of rois...
To me `in` is explicit.

silx/gui/plot/tools/roi.py
t20100: Yes... but the `QAction` are cached and share a `QActionGroup` which parent is the manager.
As it is now, it is not a factory.
It was initially done with `QDockWidget.toggleViewAction` in mind.

t20100: I just checked for `QDockWidget.toggleViewAction`: it always returns the same `QAction` and its parent is the `QDockWidget`.

Issue #3001, opened 2020-04-15
[plot] Profile tools as a drop down list


vallsv:
Hi,

I am using this action in Flint in order to reduce the amount of actions in the main toolbar.
Would you like to have the same in Silx?

Doing that could break thinks, cause for example there will be no more dedicated toolbar for profile, so if you custom it on your side, you will have problems.

![Capture d’écran de 2020-04-15 09-17-32](https://user-images.githubusercontent.com/7579321/79309364-777fa800-7efa-11ea-9b1b-ab604c21aa31.png)

We also can find a mix in between, drop down for interaction only, main actions to edit and remove profile ROIs.

vallsv:
WHen i talk about a mix i talk about #2046

vallsv:
Closed by  #3011

PR #3011, opened 2020-04-17
Rework previous profile

### LINKED ISSUE #2955: [plot] Improve profile architecture

vallsv:
PR #2988 and #3008 have to be merged before

Closes #2955, #2865, #2046, #1947, #84, #2046, #3001

This PR rework the previous profile mechanism with the new one.

Then with this PR, Plot1D, Plot2D, StackView, ScatterView, ImageView... uses the the new profile based on ROIs.

- Ceate a toolbar helper
- Rework the 1D 2D scatter toolbars 
- Merge the ROI actions together
- Rewrite the tests

The old toolbar classes are still there. It should not be used by external application, but in case there is a backward compatibility for most of the previous API.

If you use attributes or methods from the previous profile toolbars, this PR can break your code.


vallsv:
This PR can be used to see the result on the default plots.

vallsv:
The free line is not working well anymore, i suspect the `setSingleShot` mode to break the interaction. I have to investigate.

t20100:
I have to small issues:
- The profile window is very small vertically when first shown.
- The horizontal or vertical scatter profile "grid" tool, if I create the profile line, release the mouse button and then drag it outside the scatter data, I can no longer move it. It does not occur if I create it and drag it without release the mouse button.

vallsv:
- I dont understand the small height size, it should use the legacy code here, but also noticed it
- For the grid it was the problem with the update of the marker when the style is updated. Now it should be fixed

vallsv:
I have fixed the profile window size, add it supports back a custom profile window. I have added a unittest about that.

So the PR is fine.

vallsv:
I can reproduce it here. So easier to fix.

vallsv:
The problem is solved.

In fact an unexpected profile window was on top of the plot, so the second subtest was not able to create the ROI with the mouse.
- small screen so no space on side for the profile window
- now the profile window is bigger (it was incorrectly small before)
- now the profile window is a dialog (so it stay on top of the main window)

Tricky.

PR #3013, opened 2020-04-20
dicttoh5: support h5 attributes

### LINKED ISSUE #3005: extend dicttoh5 to handle h5 attributes
How about extending `silx.io.dictdump.dicttoh5` to be able to handle h5 attributes?

Here is a small example:
```python
import h5py
import numpy
 
with h5py.File("myfile.h5",mode='a') as f:
    f.create_group("my_dataset")
    f["my_dataset/x"]=numpy.arange(20)
    f["my_dataset/y"]=numpy.random.rand(20)
    f["my_dataset"].attrs["NX_class"] = "NXdata"
    f["my_dataset"].attrs["signal"] = "y"
    f["my_dataset"].attrs["axes"] = "x
```

generates a file that works well with `silx view` as it displays a default plot. Couldn't  `dicttoh5` be extended to be able to achieve something similar? I imagine something in this direction:

```python
from silx.io.dictdump import dicttoh5

mydataset = {"x":numpy.arange(20),"y":numpy.random.rand(20)}

mydict=dict()
mydict["my_dataset"]=mydataset
mydict["my_dataset@NX_class"]="NXdata"
mydict["my_dataset@signal"]="y"
mydict["my_dataset@axes"]="x"

dicttoh5(mydict,"myfile.h5",mode='a')
```
where `dicttoh5` would e.g. interpret an `@` character in the keys to create the corresponding `attrs`. What do you think?

linupi:
closes #3005 

here is a first draft for h5 attribute support in `dicttoh5` and a new `dicttonx`, it is still missing doc and tests. How do feel about this?

Here is the test dict that I used for `dicttonx`:
```python
d={'ds1':1,"ds2":2,"ds1@att1":'a1',"@att0":'a0',"ds3@att3-1":'a3-1',"ds3":{'ds4':4,'ds4@attr4':'a4', '@att3-2':'a3-2'}}
```
which transforms with `dicttonx` to
```
$ h5dump /tmp/test1.h5
HDF5 "/tmp/test1.h5" {
GROUP "/" {
   ATTRIBUTE "att0" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "a0"
      }
   }
   DATASET "ds1" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SCALAR
      DATA {
      (0): 1
      }
      ATTRIBUTE "att1" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "a1"
         }
      }
   }
   DATASET "ds2" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SCALAR
      DATA {
      (0): 2
      }
   }
   GROUP "ds3" {
      ATTRIBUTE "att3-1" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "a3-1"
         }
      }
      ATTRIBUTE "att3-2" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "a3-2"
         }
      }
      DATASET "ds4" {
         DATATYPE  H5T_STD_I64LE
         DATASPACE  SCALAR
         DATA {
         (0): 4
         }
         ATTRIBUTE "attr4" {
            DATATYPE  H5T_STRING {
               STRSIZE H5T_VARIABLE;
               STRPAD H5T_STR_NULLTERM;
               CSET H5T_CSET_UTF8;
               CTYPE H5T_C_S1;
            }
            DATASPACE  SCALAR
            DATA {
            (0): "a4"
            }
         }
      }
   }
}
}
```

@vallsv agreed to add the tests


t20100:
Great, thanks!
For the case where there are some attributes defined but no group or dataset (https://github.com/silx-kit/silx/pull/3013/files#diff-896ad154ae2a8b99bc0f744ca43e0337R264), maybe you can create a group (as you can't create a dataset without information), else I would raise an exception.

vasole:
Attributes defined but no dataset can be considered as root attributes or we want to impose the "/"?

If we want to impose the "/", then I would rise an exception.

WARNING: To have "" or "/" as dataset is not the same as having None as dataset.

vasole:
For instance, a root attribute can be written as `` {"@NX_class":u"NXroot"} `` instead of the (to me stranger) `` {"/@NX_class":u"NXroot"} ``

t20100:
Then it makes sense to create a group if attributes are defined for an otherwise undefined HDF5 path.

vasole:
BTW, the code in dictdump.py is very inefficient. Instead of using require_dataset it systematically uses create_dataset after having deleted the previous one creating file leaks.

Personally I try to use require_dataset and there will be an exception if the target and the source datasets are incompatible but there are no file leaks.

linupi:
Thanks for your comments. I pushed the corresponding changes and added some doc. I leave it to @vasole to check if the example in the docstring is valid Nexus and to @vallsv to add some tests.

vallsv:
Nice, i will take a look.
Is the key order matter?

vasole:
Only NXentry stuff can be at top level:

- You have to create an NXentry-type group containing your NXdata-type group.
- You could add the @default attribute to the NXentry group and to the NXroot.
- I know today we are in python 3 world, but I would be explicit and write ``u"NXdata", u"NXroot"`` and so on for the case somebody wants to port that module to python 2.

vallsv:
> I know today we are in python 3 world, but I would be explicit and write 

There is no Python2 CI. So we can't assume that. People who wants to port that will have much more difficult problems.

woutdenolf:
Maybe add `nxtodict` as well?

woutdenolf:
FYI I've added something like this in the Bliss project for the [nexus writer](https://gitlab.esrf.fr/bliss/bliss/-/blob/master/nexus_writer_service/io/nexus.py#L1754). Was is missing here are arguments like `update` and `overwrite``which control whether existing datasets/groups/attributes are going to be updated, overwritten or untouched.

linupi:
> Only NXentry stuff can be at top level:
> 
>     * You have to create an NXentry-type group containing your NXdata-type group.
> 
>     * You could add the `@default` attribute to the NXentry group and to the NXroot.

@vasole : is this something for the example or something that should always happen? 
Also: Should sub-dicts without NX_class automatically be "NX_collections"?

t20100:
@woutdenolf good idea, but let's keep that for another PR.

t20100:
@linupi, @vasole comments about NeXus are about the sample code, I don't think we want to write a NeXus validator here.

woutdenolf:
> > Maybe add `nxtodict` as well?
> 
> we haven't touched h5todict. Do we want it to import attributes as well? maybe it should be optional.

Why would you make importing attributes optional? You could use h5todict for that.

linupi:
> @linupi, @vasole comments about NeXus are about the sample code, I don't think we want to write a NeXus validator here.

Ok, so what should the example look like in this case? I just have to add "@default":"plot"?

vasole:
@t20100  We are not writing a NeXus validator but if you want to put a nexus example put it properly particularly if the person submitting the PR asks me for validation.
@vallsv We are talking about a doc string it does not need any python 2 validation

t20100:
@vasole that was exactly what I was meaning.

vasole:
If you are not going to write a valid NeXus example write anything that has nothing to do with NeXus.

I thought we intended to write a library for experimentalists and not for engineers. For the latter h5py is enough.

I find funny that the people in this thread interested on the potential use of this PR to handle NeXus files have a scientific background and those that do not see the interest do not.

vasole:
Certainly I can do it, although it is against the philosophy  of code review.

vasole:
Are we going to allow creation of groups at different levels? That conditions the example as I could use a shortcut using "entry/plot" as dictionary key instead of plot and just add "entry@NXclass"

vasole:
It would be simpler, more compact and close to what we exchange by e-mail to allow something like below to avoid the explicit creation of intermediate dictionaries. A single level dictionary would do the job:

```
{ 
   "entry/plot/x": [1, 2, 3, 4],
   "entry/plot/y": [2, 4, 6, 8],                         
   "entry/plot@NX_class": u"NXdata",
   "entry@default": "plot",
   "entry@NX_class":u"NXentry",
   "@NX_class": u"NXroot",
   "@default":  "entry",
}
```

vasole:
Perhaps what I wrote above is already supported.

vallsv:
@vasole I don't know about `"entry/plot/x": [1, 2, 3, 4],`, but there was nothing related to it on the existing unittests, but it should work, i will take a look, and add tests into another PR.

@woutdenolf there is an `overwrite_data`, false by default. In this case attributes are skipped when there is a collision.

vasole:
@vallsv The overwrite option is inefficient because it does not allow to reuse the disk space therefore the suggestion of @woutdenolf to have an update option or mine of trying to use the existing space and only in case of troubles to revert to the previous behavior.

linupi:
thanks @vallsv for opening the other issues. From my side I am done here. Thanks to all of you for the help.

vasole:
From my side I need to know if we allow to have  the slash on the keys (h5py allows it). If that is the case, I would update the doc string to make the example easier to digest.

If we cannot decide now, we can open another PR just for the doc string.

t20100:
PR #3017 proposes to change  to allow "/" (docstring update + test).
It looks like this PR is done!

Any more comments, or can it be merged?

vasole:
We can merge. I can open a PR to provide a couple of alternative ways to write the dictionary in the doc string.

I prefer what I wrote as comment in this thread to what I wrote in the code.

PR #3014, opened 2020-04-21
[plot] Helper to set back the previous mode used

vallsv:
Here is a proposal to allow to easily set back the previously used default interactive mode.

It is useful to know what to do if we want to abort or deselect a custom interactive mode.

Now profile uses ROI, i would like to desactive the tool at the end of the interaction (in order to only deal with existing markers). And this method simplifies it.

I also did it twice on pyFAI, so i think it can be useful to share a common API for that.

You can check #3006, it can be implemented outside of the plot, but when it is used few times, i think it's better to provide it as part of the plot. 

What do you think?


vallsv:
I have renamed the method `resetInteractiveMode` as proposed by @t20100

PR #3017, opened 2020-04-21
Test dicttoh5 flat dict

vallsv:
Merge #3013 first

This PR add test to be sure there is no regression with subtree as key in `dicttonx` and `dicttoh5`.

As the documentation said it was forbidden, but the implementation was allowing to use it, the other solution is to add assertion in case of use of `/` in the keys. Let me know if you prefer this way.

Related to @vasole  comment on #3013

### DIFF
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -163,9 +163,9 @@ def dicttoh5(treedict, h5file, h5path='/',
         This function requires `h5py <http://www.h5py.org/>`_ to be installed.
 
     :param treedict: Nested dictionary/tree structure with strings or tuples as
-         keys and array-like objects as leafs. The ``"/"`` character is not
-         allowed in keys. If tuples are used as keys they should have the format
-        (dataset_name,attr_name) and will add a 5h attribute with the
+        keys and array-like objects as leafs. The ``"/"`` character can be used
+        to define sub trees. If tuples are used as keys they should have the
+        format (dataset_name,attr_name) and will add a 5h attribute with the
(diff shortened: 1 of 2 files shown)


vasole:
Since h5py allows to create the intermediate groups when there are ``/`` in the keys, I would allow them as keys in our dictionary.

vasole:
Of course, when using the keys as input. When we reverse we create the actual dictionary.

t20100:
+1 for creating intermediate groups.

t20100:
This can be rebased since PR #3013 is merged and it looks good to me.

vallsv:
I cleaned up the PR

kif:
Debian nightly build now fails ... with this error message:
```
ERROR: testAttributeValues (silx.io.test.test_dictdump.TestDictToH5)
Any NX data types can be used
  File "/home/tester/silx/build/lib.linux-x86_64-3.7/silx/io/test/test_dictdump.py", line 162, in testAttributeValues
    with h5py.File(mem, "w") as h5file:
  File "/usr/lib/python3/dist-packages/h5py/_hl/files.py", line 309, in __init__
    name = filename_encode(name)
  File "/usr/lib/python3/dist-packages/h5py/_hl/compat.py", line 111, in filename_encode
    filename = fspath(filename)
TypeError: expected str, bytes or os.PathLike object, not _io.BytesIO
```

vallsv:
Please open an issue.
Which version of h5py? As it is part of there documentation, i guess it's a problem on there side.

PR #3018, opened 2020-04-22
[roi] Provide an action to create new ROIs

vallsv:
This PR rework the QAction of the ROI manager in order to have independent actions.

As result:
- 2 actions using the same ROI class will not be triggered simultaneously
- It is possible custom the way each actions behave
- Resolve the issue with parent and `getInteractionModeAction`

### DIFF
--- a/silx/gui/plot/tools/roi.py
+++ b/silx/gui/plot/tools/roi.py
@@ -39,7 +39,6 @@ import weakref
 
 import numpy
 
-from ....utils.weakref import WeakMethodProxy
 from ... import qt, icons
 from .. import PlotWidget
 from ..items import roi as roi_items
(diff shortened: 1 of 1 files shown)


vallsv:
- I created `sigInteractiveRoiCreated̀` and `sigInteractiveRoiFinalized` to catch the live cycle of a ROI from the manager. There is maybe better names.
- We maybe could move the new action with the plot actions (maybe not nice for dependancies)

vallsv:
Also, `_MODE_ACTIONS_PARAMS` could be dispatched into the ROI classes. It is what i did for the profile PR. It's just convenient.

I think it was the case before, and i change it.

t20100:
Fixed minor typos in docstrings.
LGTM, this fixes inconsistency introduced by PR  #3008 and improves over previous state.

PR #3022, opened 2020-04-24
[roi/profile] Improve internal ROI objects

vallsv:
This PR aim to rework the ROIs object in order to help working with composition.

This should reduce the complexity of the code, by allowing split and of the code, and by removing complex design needed to provide magic helpers.

- `RegionOfInterest` is now abstract
- Add `_connectToPlot`/`_disconnectFromPlot` to help to manage composition
- ROIs based on simple markers became really simple classes (point, hrange, vline, hline)
- Create `_HandleBasedROI` for complex anchor management (rect, polygon...)
- The `ArcROI` was cleaned up

### DIFF
--- a/silx/gui/dialog/ColormapDialog.py
+++ b/silx/gui/dialog/ColormapDialog.py
@@ -828,7 +828,7 @@ class ColormapDialog(qt.QDialog):
         self._item = None
         """Weak ref to an external item"""
 
-        self._colormapChange = _ColormapChanged()
+        self._colormapChange = utils.LockReentrant()
         """Used as a semaphore to avoid editing the colormap object when we are
         only attempt to display it.
         Used instead of n connect and disconnect of the sigChanged. The
@@ -1576,19 +1576,3 @@ class ColormapDialog(qt.QDialog):
                 nextFocus.setFocus(qt.Qt.OtherFocusReason)
         else:
             super(ColormapDialog, self).keyPressEvent(event)
-
-
-class _ColormapChanged():
-    """Simple context manager to know if we should ignore or not colormap
-    modifications"""
-    def __init__(self):
-        self._ignoreColormapChange = False
-
-    def __enter__(self):
-        self._ignoreColormapChange = True
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        self._ignoreColormapChange = False
-
-    def locked(self):
-        return self._ignoreColormapChange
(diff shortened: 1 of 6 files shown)


silx/gui/dialog/ColormapDialog.py
```diff
@@ -828,7 +828,7 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         self._item = None
         """Weak ref to an external item"""
 
-        self._colormapChange = _ColormapChanged()
+        self._colormapChange = utils.LockReentrant()
```
t20100, line 831: Do you need to use LockReentrant elsewhere?

silx/gui/plot/items/roi.py
```diff
@@ -395,229 +363,283 @@ def setFirstShapePoints(self, points):
...
+    """This inheritance is needed to avoid wrong __init__ order in Qt"""
+    pass
 
-    def _updateShape(self):
-        """Called when shape must be updated.
 
-        Must be reimplemented if a shape item have to be updated.
-        """
-        return
+class _HandleBasedROI(RegionOfInterest, _Foo):
```
t20100, line 439: Why do you get the _Foo class? This is weird.

t20100:
I renamed a method.
CI is broken because of latest pre-release of cython.

t20100:
Except for the 2 questions, this LGTM.

silx/gui/plot/items/roi.py
vallsv: It was fixes the way the inheritance was behaving, removing a weird exception. I dont have solution here.

silx/gui/dialog/ColormapDialog.py
vallsv: Yes, in this PR or the next one i think, in the ROIs.

silx/gui/plot/items/roi.py
vallsv: I could try to rebase and take a look again.

t20100: You can rebase as I merged PR #3034 which should fix CI.
I'll have a look at this _Foo now.

t20100: Only alternative so far is to constraint the order of inheritance further down which is also kind of awkward.... so let's see that later and keep the ugly workaround at a single place rather than forcing inheritance order at mainy places.

PR #3026, opened 2020-04-29
[roi] Create an highlighted style for ROIs

vallsv:
This PR reuse concept from `Curve` and the `CurveStyle` object to custom the highlighted state of ROIs.

I wanted to make obvious which ROI was selected. So now it's the case.

If it's fine, i still have to update the profile with it.

Possible improvements:
- Make the shapes selectable
- Make the shapes draggable

Capture d’écran de 2020-04-29 15-08-12

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -1510,3 +1510,54 @@ class BaselineMixIn(object):
             return numpy.array(self._baseline, copy=True)
         else:
             return self._baseline
+
+
+class _Style:
+    """Object which store styles"""
+
+
+class HilightedMixIn(ItemMixInBase):
+
+    def __init__(self):
+        self._highlightStyle = self._DEFAULT_HIGHLIGHT_STYLE
+        self._highlighted = False
+
+    def isHighlighted(self):
+        """Returns True if curve is highlighted.
+
+        :rtype: bool
+        """
+        return self._highlighted
+
+    def setHighlighted(self, highlighted):
+        """Set the highlight state of the curve
+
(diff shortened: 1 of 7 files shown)


vallsv:
Tests fail... I feel like the inheritance from PyQt5 or PySide2 are not done the same...

t20100:
Yes, that can be expected....

vallsv:
I have tested on my side with PySide without problem.
It's maybe related to python version. It fails on 3.5 and 3.6.

vallsv:
`core.HilightedMixIn.__init__` was not called... which make me feel i just don't understand anything to python...

vallsv:
I will rework the commits. And it will be fine.

vallsv:
It fine now

vallsv:
I think this behavior is better

silx/gui/plot/items/core.py
```diff
@@ -1510,3 +1510,54 @@ def getBaseline(self, copy=True):
             return numpy.array(self._baseline, copy=True)
         else:
             return self._baseline
+
+
+class _Style:
+    """Object which store styles"""
+
+
+class HilightedMixIn(ItemMixInBase):
```
t20100, line 1519: There's a typo: `HilightedMixIn` -> `HighlightedMixIn`

t20100:
This LGTM but for the typo in the class name `HilightedMixIn` -> `HighlightedMixIn`.
The `getCurrentStyle` could even move to `HighlightedMixIn`, but let's keep that for later.

vallsv:
The class name have been fixed.

Issue #3028, opened 2020-05-01
unfinished polygon ROI causes error


PiRK:
When drawing a polygon ROI in a plot widget, if I accidentaly right click before finishing the polygon, I end up with an unfinished shape. The ROI appears in the ROIManager table, and it appears like it can be deleted. However, when I leave the ROI drawing mode after deleting it, I get this error :

```
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages
\silx\gui\plot\tools\roi.py", line 190, in _plotInteractiveModeChanged
    self.__roiInteractiveModeEnded()
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages
\silx\gui\plot\tools\roi.py", line 420, in __roiInteractiveModeEnded
    self.removeRoi(self._drawnROI)
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages
\silx\gui\plot\tools\roi.py", line 322, in removeRoi
    'RegionOfInterest does not belong to this instance')
ValueError: RegionOfInterest does not belong to this instance

```

And when I then call `RoiManager.getRois()`, the deleted ROI is present in the returned list.

I tried disabling the context menu in my plot widget with `self.getWidgetHandle().setContextMenuPolicy(qt.Qt.NoContextMenu)`, but the bug happens even without context menu.

PiRK:
I also see  the same error message if I delete the polygon ROI in the RoiManager table before finishing the polygon (last point not clicked). After that, if I change drawing mode, or click on polygon mode again, each new ROI that I draw is added twice to the RoiManager table.

These errors can be reproduced with `examples/plotInteractiveImageROI`

vallsv:
Hi PiRK, I was working around that, i will take a look.

I have tested this behavior without problem recently, i will try again.

vallsv:
BTW, did you try with the master?

vallsv:
With my branch, i am not able to make it fail. And if i remove the ROI, it is recreated automatically. But no exception.

PiRK:
I will try the master, thank you.

vallsv:
THat master have another pbm btw. If you have an exception with an `abort` function it will be fixed.

vallsv:
If you still have rights you could take a look at #3029 and merge it. Else @t20100 is maybe around.

t20100:
Can this be closed with PR #3029 being merged?

[...]

PR #3033, opened 2020-05-04
Improve polygon closure

vallsv:
This PR improve the behavior of the polygon draw interaction.

- Provide another action to the context menu to close the polygon while there is the draw interaction
- The context menu is now provided by the widget, not by the backend (then it can be overwrited)

It fixes 3 issues:
- During polygon creation, using the mouse right click to open the context menu was also triggering the PlotInteraction focus class, as result it was not anymore possible to interact with the plot markers
- During polygon creation, closing the context menu was not updating the polygon while the mouse was not again moved
- Restart the plot interaction if the current drawn ROI was removed (this avoid to have a strange behavior with a recreated ROI while it was explicitly removed)

There is remained small issue:
- It is possible to create a polygon ROI with 2 points, there is no real way to cancel the polygon interaction, the plot is not aware of that.

### DIFF
--- a/silx/gui/plot/Interaction.py
+++ b/silx/gui/plot/Interaction.py
@@ -137,6 +137,11 @@ class State(object):
         """
         pass
 
+    def validate(self):
+        """Called externally to validate the current interaction in case of a
+        creation.
+        """
+        pass
 
 class StateMachine(object):
     """State machine controller.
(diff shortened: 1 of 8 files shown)


PiRK:
This closes #3028 . It solves the remaining issue with the right click. I did not notice any regression, but I will keep testing.

vallsv:
Nice :-)

Issue #3027, opened 2020-04-30
Make cython files compatible with Cython3


kif:
Cython3, which is currently in alpha phase  does enforce the Python3 syntax.

Because the string management changed so much, there is a big difference and one should be treated manually. I quickly tested and beside a few warnings, there is not much to be done.

kif:
This is declared in the setup.py, not inside the files. my bad.

kif:
Apparently I was not completely drunk when I noticed this ... 
those warning are actually present:
```cythoning silx/io/specfile.pyx to silx/io/specfile.c
/opt/VirtualEnv/py38/lib/python3.8/site-packages/Cython/Compiler/Main.py:344: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /home/jerome/workspace/silx/silx/io/specfile.pyx
```
This does not mean the code is not cython3-compatible but there are a couple of bugs in the build procedure:

* [ ] the cython compiler directives and compile-time options are not taken into account (hence the warning about language_level not set, while it is declared in the setup.py
* [ ] the --force-cython option to `python setup.py build --force-cython` is not working. Nor the environment variable.

vallsv:
Ohhh, bad news. As we did a lot of custom code for the setup, it will be painful. Except if the defaut cython builder provides more features now.

t20100:
BTW, do we still need the `--force-cython`?

kif:
I have a solution ... some part of the cython code is not up to level

kif:
Cython offers now parallel compilation. 

I believe `--force-cython` is a nice feature as the "rm silx/*/*.c" is kind-of dangerous and complicated. If there is a cache, there should always be a way to bypass the cache.

t20100:
CI is broken by cython

kif:
it passed: https://travis-ci.org/github/silx-kit/silx/builds/683740401

t20100:
Sorry, I was talking of CI from other PRs.

Anyway, this is closed by  PR #3034.

PR #3034, opened 2020-05-06
Setup.py modifications to make the code Cython3 compatble

kif:
There are issues within the qhull code which is not cython3-ready.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,9 +25,8 @@
 # ###########################################################################*/
 
 __authors__ = ["Jérôme Kieffer", "Thomas Vincent"]
-__date__ = "12/02/2019"
+__date__ = "06/05/2020"
 __license__ = "MIT"
-
 
 import sys
 import os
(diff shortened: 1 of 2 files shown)


kif:
Sorry, my eclipse is rather picky about blank lines.

kif:
We are still using the old_build_ext from cython, the new one provides parallel builds. Is it a good idea to have it or will it break compatibility with elder version of cython ?

kif:
close #3027

vallsv:
AFAIR the default cython builder was not providing enough flexibility for flags, like openmp. But that's a long time ago.

vallsv:
I think it should be merged as it is cause it fixes the CI.

But it would be good to check next that openmp is still working. And clean up our custom code if it not needed anymore.

t20100:
qhull code is already fixed upstream in scipy... to me we should get rid of our copy and disable the features based on it if scipy is not available.

PR #3036, opened 2020-05-06
correct destructor

### LINKED ISSUE #3035: Error with python-3.8
Apparently the destrurctor is not making a copy of the keys to destroy the dict:
```
  File "/home/jerome/workspace/silx/build/lib.linux-x86_64-3.8/silx/opencl/sift/alignment.py", line 149, in __del__
    self._free_buffers()
  File "/home/jerome/workspace/silx/build/lib.linux-x86_64-3.8/silx/opencl/sift/alignment.py", line 171, in _free_buffers
    for buffer_name in self.cl_mem:
RuntimeError: dictionary keys changed during iteration
```

kif:
close #3035

### DIFF
--- a/silx/opencl/sift/alignment.py
+++ b/silx/opencl/sift/alignment.py
@@ -37,7 +37,7 @@ __authors__ = ["Jérôme Kieffer", "Pierre Paleo"]
 __contact__ = "jerome.kieffer@esrf.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "09/02/2018"
+__date__ = "06/05/2020"
 __status__ = "production"
 
 import os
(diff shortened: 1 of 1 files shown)


silx/opencl/sift/alignment.py
```diff
@@ -168,11 +168,11 @@ def _free_buffers(self):
...
         free all memory allocated on the device
         """
-        for buffer_name in self.cl_mem:
+        for buffer_name in list(self.cl_mem.keys()):
             if self.cl_mem[buffer_name] is not None:
                 try:
-                    del self.cl_mem[buffer_name]
-                    self.cl_mem[buffer_name] = None
+                    buffer = self.cl_mem.pop(buffer_name)
+                    del buffer
```
pierrepaleo, line 175: Strangely the error `RuntimeError: dictionary keys changed during iteration` should have happened in every previous versions of Python, but didn't. 
This correction should indeed fix the problem.

PR #3041, opened 2020-05-07
[WIP]add a CircleRoi

PiRK:
This PR adds a CircleRoi object.


vallsv:
Hi PiRK, could you please rebase it on top of the master?

PiRK:
done

vallsv:
Anyway, if you can debug it with PyQt, the center handle is not behaving as expected, it also changes the radius. I suspect there is a missing blockSignal, but i do not have checked.

PiRK:
I can reproduce the issue. I’m still looking for a solution.

PiRK:
I was not able to find a more elegant solution.  The issue has something to do with the fact that I was computing the radius based on the center position. I tried to update the calculation by using the "origin" point instead the current handle position, but it made the problem worse.

So I added an attribute to store the ROI’s radius explicitly.

vallsv:
It's anyway better. The geometry should be stored on it's own, and not on handles.
That is not the case for other ROIs ( except the arc) cause i am lazy.

vallsv:
BTW anybody knows why the CI fail? it looks like everything is fine, except the status of the CI.

PiRK:
I will work on an ellispe ROI in another PR. The interaction will be in two steps, and there are two ways I can imagine to achieve it:
- first click for the center, second click for a corner of the bounding box (like the ellipse mask). This will draw an ellipse oriented parallel to the axes. Then, after this initial interaction, the user should edit the ROI to move a handle controlling the orientation 
- first click for the center, second click for the radius and orientation. After this initial interaction, there will be a circle with two oriented handles that can be edited to control the small radius and the large radius, to adjust the eccentricity. 

I will test the two solutions, but I have the feeling that the second solution will be more convenient to use. It allows to define the center, one of the diameters and the orientation in the initial interaction. With the first solution, maybe it will be difficult to get the shape right in the first interaction, and after setting the orientation the user will need to correct the center handle again. 

Any other ideas for the ellipse? 

Is there a way to set all parameters (center, orientation, eccentricity, diameters....) in one go, without having to come back to editing a handle?

vallsv:
Here i think it really depend of your use case.
But I feel like the first iteration should be fast. A single click or a click and drag. So i don't think your second proposal is a good idea, except for specific use case. And in this case a fit from a scatter of points as input is maybe more convenient.

For generic use case i think you should do the way paint software works. Create an ellipse as you create an orthogonal rectangle. And then you can move the handles.

For the arc, i have used a predefined constraints, and the first interaction was just scaling the shape between 2 points. It also could be done the same, which allow to preset the orientation.

Anyway, circle and ellipse should be the same ROI, as we have a rect (which is not really true, cause our rect is always orthogonal).

t20100:
> BTW anybody knows why the CI fail? it looks like everything is fine, except the status of the CI.

@vallsv, there is a seg fault at the end: https://travis-ci.org/github/silx-kit/silx/jobs/684330767#L10019

vallsv:
You can rebase to the master, so what it can be reviewed and merged.

PiRK:
I did a mess with this PR (pushing to upstream instead of origin), so I will close it in favor of #3045.

PR #3044, opened 2020-05-07
[roi] Fix HandleBasedROI with PySide2

### LINKED ISSUE #3043: problem with handles on ROIs
I get an error when I try to move a handle / anchor on any type of editable ROI. Tested with circle ROI (#3041) and RetangleRoi.

```
  File "D:\git\silx\build\lib.win-amd64-3.7\silx\gui\plot\items\roi.py", line 590, in __editingStarted
    self._posOrigin = numpy.array(handle.getPosition())
AttributeError: 'NoneType' object has no attribute 'getPosition'
  File "D:\git\silx\build\lib.win-amd64-3.7\silx\gui\plot\items\roi.py", line 605, in __editingFinished
    current = numpy.array(handle.getPosition())
AttributeError: 'NoneType' object has no attribute 'getPosition'
```

The error reminds me of https://github.com/silx-kit/silx/issues/2655#issuecomment-507677306

vallsv:
This PR fixes issue #3043 

In some circumstances with PySide2, the use of `self.sender()` returns None.

- It looks like it is the case for some private methods
- Maybe only for signals without values

Closes #3043

### DIFF
--- a/silx/gui/plot/items/roi.py
+++ b/silx/gui/plot/items/roi.py
@@ -658,23 +658,23 @@ class _HandleBasedROI(RegionOfInterest, _Foo):
         handle.setVisible(editable and self.isVisible())
         handle._setDraggable(editable)
         if editable:
-            handle.sigDragStarted.connect(self.__editingStarted)
-            handle.sigItemChanged.connect(self.__editingUpdated)
-            handle.sigDragFinished.connect(self.__editingFinished)
+            handle.sigDragStarted.connect(self._handleEditingStarted)
+            handle.sigItemChanged.connect(self._handleEditingUpdated)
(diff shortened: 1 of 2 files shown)


vallsv:
I tried to reproduce the problem with a small example
- with/without async signal,
- with/without private callback
- with/without signal argument

Everything is working well. I can't reproduce the issue.

vallsv:
It looks like PySide2 fail if the signal uses a private callback and this callback is part of the inherited classes.

PiRK:
Thank you for the fix. I will be able to test it tomorrow.

PiRK:
It works on my system (PySide2-5.14.2.1)

vallsv:
Does the full test suite segfault on your side?

PR #3045, opened 2020-05-11
Ellipse roi

PiRK:
This PR adds an EllipseRoi class.

It replaces #3041 (all commits included) and cherry picks a commit from #3044


PiRK:
Regarding the CircleROI, I don’t really need it and I’m OK with removing it if you want. I just did it as an exercise. But the EllipseROI interaction currently does not provide a way of ensuring that the ellipse snaps to a perfect circle, if the user starts moving one of the handles.

Do you think we should implement something like this (snapping to circle when both axes are close enough)? I’m not sure how to define a proper criterium to decide that the shape is close enough to a circle to set the major and minor radius to the exact same value.

I don’t really see a problem with  keeping both objects, in case someone needs a simple circle that can be edited by moving a single radius handle.

t20100:
I need a `circleROI`, but it must be only a circle.
To me it is best to have 2 different ROIs for circle and ellipse.

vallsv:
Let's create 2 ROIs then.

@t20100: I am just afraid you will need specific ROIs with handles at a very specific places for consistency between views? I mean, you will need a circle that can be edited like a rectangle, more than like a centre+radius, no?

For ellipse snapping, i think you can skip this feature, it's fine.

t20100:
Yes we need 2 ROIs and snapping can be added later if need be.
@vallsv, I use 2 kinds of ROI: rectangle + circle depending on the view, a circle with center+radius is fine for me.

vallsv:
It is fine, except the change on the example i think. It looks to be a good idea, but this modes are supposed to be to create new ROIs. It is really surprising. Enabling by default all the ROIs edition is enough no?

I am still surprising the handles do not behave like a rect. I tried to use another anchor style, like pyqtgraph, but that's detail, i will maybe edit that into another PR i am working on.

Except that, I will just add a small fix for the style and we can merge it. Nice to have your feed back on that tool.

vallsv:
I did small changes to remove math lib. It looks to behave the same.

Issue #3047, opened 2020-05-12
Some requests about the new profile toolbar


mretegan:
If I am not mistaken this part of the library has gone to some major changes recently. Because I do some customization on the plot widget toolbar, I wanted to see how the behavior changed. I use the widget to plot both curves and images. This is a minimal working example to illustrate my points:

```python
import sys
import numpy as np

from silx.gui.qt import QApplication, QMainWindow
from silx.gui.plot import PlotWidget
from silx.gui.plot.Profile import ProfileToolBar, manager


class BasePlotWidget(PlotWidget):
    def __init__(self, parent=None, **kwargs):
        super(BasePlotWidget, self).__init__(parent, **kwargs)


def main():
    app = QApplication([])

    window = QMainWindow()
    widget = BasePlotWidget(parent=window)
    toolBar = ProfileToolBar(plot=widget)
    toolBar.clear()

    profileManager = manager.ProfileManager(parent=widget, plot=widget)
    # profileManager.setItemType(scatter=True)
    # profileManager.setActiveItemTracking(True)

    for action in profileManager.createImageActions(toolBar):
        toolBar.addAction(action)

    cleanAction = profileManager.createClearAction(toolBar)
    toolBar.addAction(cleanAction)
    editorAction = profileManager.createEditorAction(toolBar)
    toolBar.addAction(editorAction)

    # Plot a curve.
    # widget.addToolBar(toolBar)
    # x = np.linspace(-10, 10, 200)
    # widget.addCurve(x, np.sin(x))

    x = np.outer(np.linspace(-10, 10, 200), np.linspace(-10, 5, 150))
    data = np.sin(x) / x
    widget.addImage(data)

    window.setCentralWidget(widget)
    widget.addToolBar(toolBar)
    window.show()

    sys.exit(app.exec_())


if __name__ == "__main__":
    main()
```

Here are a few comments:

1. Before, the actions would only be enabled if there were applicable to things plotted in the widget. For example if I had a 1D curve they would be disabled.
2. In the example above the ``profileWindow`` is not displayed. If I use the default toolbar, it works as expected.
3. Would it be possible to programmatically change/disable the default label of the profile? Now it is "Profile".
4. Is it possible to change the default of the last button to "Compute profile sum"?

vallsv:
Could you please take a screenshot?

[...]

PR #3048, opened 2020-05-12
[roi/profile] Polish

vallsv:
Polish things all around ROIs and profiles

- Fix marker text location
- Fix marker to avoid to use space as margin (which create wrong alignment when using multiline text)
- Action profile editor can be used at many places
- Clean up ROI metadata
- Improve compatibility with the old profile code
- Make `HandleBasedROI` public
- Robustness on tests


mretegan:
I've just tested this and having the last profile properties stored is indeed quite nice. I've noticed however that initially, the profile appears on top of the plot widget which makes things a bit weird when drawing a line profile.

In #3047 you mentioned that it is possible to change the default of the last button to be "Compute profile sum". How exactly? `ProfileOptionToolButton` seems very hard to reach. 

Finally, would it be possible to move the `setProfileWindow` method to the `ProfileManager`? This way I can set own type of profile window.

vallsv:
> the profile window appears on top of the plot widget which makes things a bit weird

It is supposed to be displayed on the side (if you have enough space). That's the exact same algorithm as before. But at home i use a very small screen, And i suspect the profile window is bigger than previously. We have also added a `Dialog` window flag to the profile, this can be custom if you inherit the profile manager, see below. 

> you mentioned that it is possible to change the default of the last button

I have also mentioned it in the issue. The way a profile is computed (sum/mean, line width) is part of the ROI itself, so when you custom the name, you also custom that properties.

If it is not working, cause of the `StoreProperty`stuff, maybe... I think we can had default values to this object (`setDefaultMethod`...). But if the first approach is working for you, i think it is enough.

> Finally, would it be possible to move the setProfileWindow method to the ProfileManager?

Definitely no. It is also a per ROI concept. To be able to display many ROIs at the same time. That is how the cross ROI is working. But you have 2 solutions

- You can use the previous `ProfileToolBar` (from `silx.gui.qt.plot.Profile`) which provide a backward compatibility (btw i don't know how it behave right now with the cross ROI). 
- You can inherit the Profile manager and create it on your own. It is what i am doing for Flint (to dock the profile windows). You can take a look at the module `silx.gui.qt.plot.Profile` which also inherit the profile to implement the backward compatibility feature (else i can show you a Flint commit).

mretegan:
> > you mentioned that it is possible to change the default of the last button
> 
> I have also mentioned it in the issue. The way a profile is computed (sum/mean, line width) is part of the ROI itself, so when you custom the name, you also custom that properties.
> 
> If it is not working, cause of the `StoreProperty`stuff, maybe... I think we can had default values to this object (`setDefaultMethod`...). But if the first approach is working for you, i think it is enough.

Maybe you did not understand what I meant before. See how the method is set when the button is instantiated. I want to change that default to sum. I don't even need the average, but I don't mind if I can set the sum as the default.

https://github.com/silx-kit/silx/blob/c25787b07b2fdeca4c4226440dbae800a242c4ae/silx/gui/plot/PlotToolButtons.py#L269-L275

> 
> > Finally, would it be possible to move the setProfileWindow method to the ProfileManager?
> 
> Definitely no. It is also a per ROI concept. To be able to display many ROIs at the same time. That is how the cross ROI is working. But you have 2 solutions
> 
> * You can use the previous `ProfileToolBar` (from `silx.gui.qt.plot.Profile`) which provide a backward compatibility (btw i don't know how it behave right now with the cross ROI).
> * You can inherit the Profile manager and create it on your own. It is what i am doing for Flint (to dock the profile windows). You can take a look at the module `silx.gui.qt.plot.Profile` which also inherit the profile to implement the backward compatibility feature (else i can show you a Flint commit).

I don't understand this. Fine, it is a ROI concept, but I should be able to change the way all those windows look, what buttons have etc. Before I was using the `ProfileTooBar`, but if I want to customize the toolbar, I first need to instantiated it, then clear it, which is convoluted.

```python
    profileWindow = PlotWidget()
    # toolBar = QToolBar("Profile", parent=widget)
    toolBar = ProfileToolBar(parent=widget, plot=widget, profileWindow=myProfileWindow)
    toolBar.clear()

    profileManager = manager.ProfileManager(parent=widget, plot=widget)
    profileManager.setItemType(image=True)
    profileManager.setActiveItemTracking(True)

    roiManager = profileManager.getRoiManager()

    horizontalLine = profileManager.createProfileAction(
        ProfileImageHorizontalLineROI, parent=toolBar
    )
    toolBar.addAction(horizontalLine)

    verticalLine = profileManager.createProfileAction(
        ProfileImageVerticalLineROI, parent=toolBar
    )
    toolBar.addAction(verticalLine)
```

vallsv:
"It's a ROI concept", means you can edit it from the ROI. So:
```
def configureMyRois(roi):
    roi.setName("I prefer writing more things here")
    roi.setProfileMethod("mean")

roiManager = profile.getRoiManager()
roiManager.sigInteractiveRoiCreated.connect(configureMyRois)
```
And that's it.

vallsv:
I don't really understand your problem with the tool bar.

- If you use the `ProfileToolBar` from `silx.gui.plot.Profile`, obviously it is already feed with content. It was the case before and it is still the case now. It is provided for retro compatibility.
- You can use the tool bar from `silx.gui.plot.tools.profile.toolbar` which is the new one.
- And if you are still not fine with it copy-paste it and create your own tool bar `qt.QToolBar`.

vallsv:
Anyway:
```
myProfileWindow = PlotWidget()
```
This can't work. The profile window have to be a `silx.gui.plot.tools.profile.manager.ProfileWindow` (or simply a `silx.gui.plot.ProfileMainWindow` (which should be deprecated)). Or you have to implement it's API to manage new profile computation result.

mretegan:
Ok, let me try differently. Below is a minimal working example. I want two things:

1. When I do a profile to have a profile window without any buttons.
2. The "sum" should be selected for the new profile, and this should be reflected in the toolbar above. 

```python
import sys
import numpy as np

from silx.gui.qt import QApplication, QMainWindow, QToolBar
from silx.gui.plot.Profile import manager
from silx.gui.plot.tools.profile.rois import (
    ProfileImageHorizontalLineROI,
    ProfileImageVerticalLineROI,
    ProfileImageCrossROI,
    ProfileImageLineROI,
)


class BasePlotWidget(PlotWidget):
    def __init__(self, parent=None, **kwargs):
        super(BasePlotWidget, self).__init__(parent, **kwargs)


def configureRois(roi):
    roi.setName("Cut")
    roi.setProfileMethod("mean")


def main():
    app = QApplication([])

    window = QMainWindow()
    widget = BasePlotWidget(parent=window)

    toolBar = QToolBar("Profile", parent=widget)

    profileManager = manager.ProfileManager(parent=widget, plot=widget)
    profileManager.setItemType(image=True)
    profileManager.setActiveItemTracking(True)

    roiManager = profileManager.getRoiManager()
    roiManager.sigInteractiveRoiCreated.connect(configureRois)

    horizontalLine = profileManager.createProfileAction(
        ProfileImageHorizontalLineROI, parent=toolBar
    )
    toolBar.addAction(horizontalLine)

    verticalLine = profileManager.createProfileAction(
        ProfileImageVerticalLineROI, parent=toolBar
    )
    toolBar.addAction(verticalLine)

    crossLine = profileManager.createProfileAction(ProfileImageCrossROI, parent=toolBar)
    toolBar.addAction(crossLine)

    freeLine = profileManager.createProfileAction(ProfileImageLineROI)
    toolBar.addAction(freeLine)

    cleanAction = profileManager.createClearAction(toolBar)
    toolBar.addAction(cleanAction)
    editorAction = profileManager.createEditorAction(toolBar)
    toolBar.addAction(editorAction)

    x = np.outer(np.linspace(-10, 10, 200), np.linspace(-10, 5, 150))
    data = np.sin(x) / x
    widget.addImage(data)

    window.setCentralWidget(widget)
    widget.addToolBar(toolBar)
    window.show()

    sys.exit(app.exec_())


if __name__ == "__main__":
    main()
```

vallsv:
If you want to show your own profile window you have to:
```
class MyProfileWindow(ProfileWindow):
    def __init__(self, parent=None, backend=None):
        super.__foo__()
        self._plot1D = here you can create the PlotWidget you want to use for 1D profiles
        # We should create a createPlot1D so what it is obvious, so let me know

class MyProfileManager(ProfileManager):

    def createProfileWindow(self, plot, roi):
        # and here you can create it
        return MyProfileWindow(plot)
```
That's the cleanest way.

If you was using `PlotWidget` with the old toolbar, we still can profile a compatibility for that. Let me know.

vallsv:
I will create now `_createPlot1D` to the `ProfileWindow` so what it is done.

mretegan:
> I will create now `_createPlot1D` to the `ProfileWindow` so what it is done.

So I should rely on a private method to implement the behavior I want. Isn't that risky? Before I used the public API to define a custom plot widget for the profiles and this is now deprecated.

> 2\. The "sum" should be selected for the new profile, and this should be reflected in the toolbar above.

Is this possible?

mretegan:
I would find it much easier to use to just allow for another argument to be passed to the initializer of the `ProfileWindow` like this: 
```python
def __init__(self, parent=None, backend=None, plotWidget=None)```

**Comment by vallsv** [comment]
> So I should rely on a private method

That's in fact a protected method. One you can inherit, but that you should not call from outside. It is explicitly documented, so i don't think there is a problem. But it could be public, sometimes i put public sometimes protected, usually i try to match the current class implementation.

> The "sum" should be selected for the new profile

I already answer there: https://github.com/silx-kit/silx/pull/3048#issuecomment-628572015
What works for the name also works for the method.

> I would find it much easier to just allow for another...

I am sorry, but argument in constructors are mostly a pain in the ass. And here concepts do not match. The new profile expect many profile windows, so anyway it must require a factory of window and not an instance of window. And as you also see it exists many kind of profile, in 1D or 2D. So this stuffs was complex.

**Comment by mretegan** [comment]
> > The "sum" should be selected for the new profile
> 
> I already answer there: [#3048 (comment)](https://github.com/silx-kit/silx/pull/3048#issuecomment-628572015)
> What works for the name also works for the method.
> 

I'm sorry but that doesn't work. Please see the code that I've posted. I don't see the QMenu changing.

> > I would find it much easier to just allow for another...
> 
> I am sorry, but argument in constructors are mostly a pain in the ass. And here concepts do not match. The new profile expect many profile windows, so anyway it must require a factory of window and not an instance of window. And as you also see it exists many kind of profile, in 1D or 2D. So this stuffs was complex.

Ok, a setter then. I will try to work a bit on this. I find the current way to customize the plot window of the profile very convoluted.

**Comment by vallsv** [comment]
> I'm sorry but that doesn't work.

Then that's a bug, I will take a look.

**Comment by vallsv** [comment]
I copy-pasted your code, everything is working. I think i don't understand your problem. Do you use this this branch?

**Comment by mretegan** [comment]
Yes, I am using this branch. Here is a recording. I want that menu to be on the "sum".


recording

**Comment by vallsv** [comment]
In your example you have used `mean`.

I have tested it with both. Both are working.

**Comment by mretegan** [comment]
> In your example you have used `mean`.
> 
> I have tested it with both. Both are working.

Well, this is embarrassing. Sorry about that. :disappointed:

**Comment by vallsv** [comment]
You'r welcome. But i really was questioning myself about the way i explain things :-?

**Comment by mretegan** [comment]
There is an issue when setting the labels of the profile window.

screenshot

Below is part of the code I'm using:

```python
class BasePlotWidget(PlotWidget):
    def __init__(self, parent=None, **kwargs):
        super(BasePlotWidget, self).__init__(parent, **kwargs)
        self.setGraphXLabel("BaseX")
        self.setGraphYLabel("BaseY")


class CustomProfileWindow(ProfileWindow):
    def _createPlot1D(self, parent, backend):
        plot = PlotWidget(parent=parent, backend=backend)
        plot.setGraphXLabel("ProfileX")
        plot.setGraphYLabel("ProfileY")
        return plot


class CustomProfileManager(ProfileManager):
    def createProfileWindow(self, plot, roi):
        return CustomProfileWindow(plot)
```

vallsv:
The axis label is part of the profile data. So if you want to enforce something, you have to set it while you receive the profile data. So you have to inherit `ProfileWindow.setProfile()` or one of the protected class rendering one of the kind of profile data.

BTW, this `setProfile` could be renamed to `setProfileData`, for clarity. What do you think @t20100?

For me it is not a bug. Except you have a generic way, using the main plot information, through the computed profile, to provide maybe a better axis by default. But if you want to custom it here, i don't think there is choices.

vallsv:
BTW, to avoid to inherit the `ProfileManager` we can create a class setter. So you can use the default one. Let me know what you think.

```
profileManager.setProfileWindowClass(CustomProfileWindow)
```
But that's mostly all we can do.

mretegan:
> The axis label is part of the profile data. So if you want to enforce something, you have to set it while you receive the profile data. So you have to inherit `ProfileWindow.setProfile()` or one of the protected class rendering one of the kind of profile data.
> 
> BTW, this `setProfile` could be renamed to `setProfileData`, for clarity. What do you think @t20100?
> 
> For me it is not a bug. Except you have a generic way, using the main plot information, through the computed profile, to provide maybe a better axis by default. But if you want to custom it here, i don't think there is choices.

So, in my case, I would have to reimplement all the logic in `_setImageProfile` just to set the label of the plot? :anguished:


> BTW, to avoid to inherit the `ProfileManager` we can create a class setter. So you can use the default one. Let me know what you think.
> 
> ```
> profileManager.setProfileWindowClass(CustomProfileWindow)
> ```
> 
> But that's mostly all we can do.

Thanks!

mretegan:
> > So I should rely on a private method
> 
> That's in fact a protected method. One you can inherit, but that you should not call from outside. It is explicitly documented, so i don't think there is a problem. But it could be public, sometimes i put public sometimes protected, usually i try to match the current class implementation.

Could you please make it public? I hope it will not be problematic if in my implementation the function does not return an instance of `Plot1D`.

vasole:
For the example shown by @mretegan somehow I recall a discussion where we said that when using horizontal or vertical profiles the X axis should take the corresponding axis label associated to the original image.

mretegan:
> For the example shown by @mretegan somehow I recall a discussion where we said that when using horizontal or vertical profiles the X axis should take the corresponding axis label associated to the original image.

This is actually what I am trying to achieve 😄. Whenever possible use the labels of the original plot, and even set the title of the cut accordingly.

vasole:
Then that is clearly a regression:

https://github.com/silx-kit/silx/pull/2395

vasole:
Perhaps not a regression but a half-way implementation in #2395. The actual coordinates are used but effectively it was not explicitly requested to use the labels.

vallsv:
> So, in my case, I would have to re-implement all the logic in _setImageProfile just to set the label of the plot? anguished

No, you just have to call super, and set the label after that.

But i agree with @vasole, it is what i tried to explain. If the information is already known, there is no reason to custom it. In short, fix the code, instead of convoluting it.

vallsv:
Looking at the code, i think it is not a regression. But we can easily display the right label from the parent plot. So i will do it.

mretegan:
Nice! Just a final idea: maybe the label of the y-axis could be the name of the method instead of simply "Profile".

vallsv:
That's a good idea. Is something have an advice here?

silx/gui/plot/tools/profile/rois.py
```diff
@@ -232,6 +233,7 @@ def createProfile2(currentData):
                 rgba = rgba.astype(numpy.float)
             currentData = 0.21 * rgba[..., 0] + 0.72 * rgba[..., 1] + 0.07 * rgba[..., 2]
 
+        yLabel = "%s, width = %s" % (str(method).capitalize(), lineWidth)
```
mretegan, line 236: I don't think is needed to have the width in the label. For example, in the case of the "free" line, it already appears in the title.

vallsv:
Here is the result. I have also specified the line width in the yLabel. I think it's a good place, but i feel like it is not convenient to read. Then maybe the method alone if better. Let me know.

Capture d’écran de 2020-05-15 16-34-15

vallsv:
My ultimate commit, if there is no more feedback.
I will switch to another PR to fix a specific issue.

silx/gui/plot/tools/profile/rois.py
vallsv: But that's not the case on the image i provided below.

silx/gui/plot/tools/profile/manager.py
```diff
@@ -173,9 +199,7 @@ def _getPlot2D(self, init=True):
         if not init:
             return self._plot1D
```
mretegan, line 200: Is this right?

vallsv:
It's fine now. This PR can be reviewed.

vallsv:
Oh i forget about this width info... I will try to make it consistent it the diagram title

vallsv:
It's fine now

vallsv:
Alleluia

mretegan:
I've just tested it and all works as expected. Very nice work Valentin, and thanks for your patience :smile:

vallsv:
Thanks. I was expecting to finish it earlier. But that's nice you spend time to test it. It is really nice to be sure everything is working now, and not in a bug fix release.

mretegan:
Just to add one last thing (maybe not very important), this is the most commended pull request in the silx repository.

vallsv:
Also the only one with a video!

vallsv:
hmmm weird error...

vallsv:
omgwtfitrulzbro

vallsv:
Finally tests look also fine with PySide2.

PR #3052, opened 2020-05-13
Change behavior of QObject mulit-inheritance

### LINKED ISSUE #3051: Issue with 3D scene items inheritance
I have the following issue with current master with `example/plot3dSceneWindow.py`:
```
File "examples/plot3dSceneWindow.py", line 136, in <module>
    clipPlane = items.ClipPlane()  # Create a new clipping plane item
  File "silx/gui/plot3d/items/clipplane.py", line 52, in __init__
    Item3D.__init__(self, parent=parent, primitive=plane)
  File "silx/gui/plot3d/items/core.py", line 92, in __init__
    qt.QObject.__init__(self, parent)
TypeError: __init__() missing 1 required positional argument: 'plane'
```
The issue is with `PyQt5` (tested with 5.12.1 and 5.14.2) but with `PySide2` it works fine.

This looks to be due to https://github.com/silx-kit/silx/commit/d9a11d543c5885346562e161ed46a19c19ff23bf
`ClipPlane` is the only item with this issue, so it seems to be related to the fact that the mix-in takes an argument (as `PlaneMixIn` does).

t20100:
This is changing the behavior of `QObject` with PyQt5 regarding multiple-inheritance to make it behave as the other bindings (i.e., PySide2 and PyQt4) by NOT isolating its call to `super().__init__(**kwargs)`
See https://www.riverbankcomputing.com/static/Docs/PyQt5/multiinheritance.html

closes #3051

### DIFF
--- a/silx/gui/plot/items/roi.py
+++ b/silx/gui/plot/items/roi.py
@@ -766,9 +766,9 @@ class PointROI(RegionOfInterest, items.SymbolMixIn):
         elif event == items.ItemChangedType.EDITABLE:
             editable = self.isEditable()
             if editable:
-                self._marker.sigItemChanged.connect(self.__positionChanged)
+                self._marker.sigItemChanged.connect(self._pointPositionChanged)
             else:
-                self._marker.sigItemChanged.disconnect(self.__positionChanged)
+                self._marker.sigItemChanged.disconnect(self._pointPositionChanged)
             self._marker._setDraggable(editable)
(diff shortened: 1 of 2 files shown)


vallsv:
Take a look at 4dbc477fdfb1993d4d04a41293bc766212bfd2a4, it could fix your failing unittest

t20100:
I cherry-picked your commit, let's see if it fixes it.

PR #3055, opened 2020-05-18
[roi] add ROI.contains function

payno:


vallsv:
Few comments:
- `isIn`:
    - you definitely have to rename it into `contains`
    - remove "pixel" from the documentation. There is nothing about screen, so if think it would be much better to talk about "point".
    - `:param Union[float,tuple] value:` as the function is inherited, arguments must always be the same. The input argument must always be `Union[Tuple[float,float]]`.
- `BoundingBox`, maybe you could move it into `silx.math`?

t20100:
A few comments:

- +1 for `contains` (as in `QRect`) and a unique kind of argument.
- For the docstrings, you probably want to use a generic docstring and the `docstring` decorator (https://github.com/silx-kit/silx/blob/658dd96021ecaa3d94d5b3434cb0eaff0333fc5c/silx/utils/proxy.py#L226)
- I would avoid definition of functions within function and put `lines_intersection` and `segments_intersection` somewhere else. BTW, in the OpenGL backend I'm using: https://en.wikipedia.org/wiki/Cohen%E2%80%93Sutherland_algorithm which does the same unless I missunderstood (https://github.com/silx-kit/silx/blob/master/silx/gui/plot/backends/glutils/GLPlotCurve.py#L1299).
- `BoundingBox` could be moved to `silx.image` since it is 2D.. or kept private for now.

PR #3056, opened 2020-05-18
Stats roi 2

### LINKED ISSUE #2430: Stats and measurements tools based on ROIs
The idea is to allow to use editable ROIs to provide some statistics on a part of an image (or scatter plot) as well as measurement information (e.g., for rectangle width, height, area, perimeter in pixels and data units).

This would also be useful as part of silx view.

payno:
Goal
====

Add the feature to compute statistics inside a specific region of interest.

close #2430 

Information
=========

#3055 need to be merge first.
replace #2636 (split and merge commits.)


kif:
LGTM.

PR #3058, opened 2020-05-19
[profile] Split area shape from profile ROIs

vallsv:
This PR fix small issue relative to profile area update at the update of the active item.

- It extracts areas from the profile ROIs into an independent shape
- It creates a link from the profile ROI to the item

As result it is now possible to design this profiles by composition. But i have burned more than the time i wanted to spend here.

Most of ROIs could be merged together, reducing by 3 the amount of classes; by 3 the amount of actions. Reducing by far the complexity of the profile ROIs.

### DIFF
--- a/silx/gui/plot/tools/profile/core.py
+++ b/silx/gui/plot/tools/profile/core.py
@@ -113,8 +113,8 @@ class ProfileRoiMixIn:
         previousPlotItem = self.getPlotItem()
         if previousPlotItem is plotItem:
             return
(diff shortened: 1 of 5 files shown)


vallsv:
It's ok now.

vallsv:
;__;

vallsv:
I found i can have unreleased widgets on `testCustomProfileWindow` time to time ( 50% of the time), on my side. I will try to take a look.

PR #3060, opened 2020-05-20
Test remaining widgets with pyside2

t20100:
This PR enables testing of remaining widgets after each test with PySide2 as it is already the case with PyQt5.
It avoids remaining widgets at test level by using `QWidget.close` rather than `TestCaseQt.qWaitForDestroy` where it seems to be an issue, it makes sure widgets have parents when it is needed.

### DIFF
--- a/silx/app/view/Viewer.py
+++ b/silx/app/view/Viewer.py
@@ -605,7 +605,7 @@ class Viewer(qt.QMainWindow):
         action.setStatusTip("Select plot rendering backend")
         self._plotBackendSelection = action
 
-        menu = qt.QMenu()
+        menu = qt.QMenu(self)
         action.setMenu(menu)
         group = qt.QActionGroup(self)
         group.setExclusive(True)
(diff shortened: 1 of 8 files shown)


vallsv:
Looking at the result, only `TestDeprecatedProfileToolBar` is failing

t20100:
This looks to fix all remaining widgets with PySide2 spotted by the tests.

It looks best to use `qapp.processEvents` rather than `qWait` since the later might not process all pending events, thus probably leading to remaining widgets.

t20100:
If CI is happy, it's ready for review.

vallsv:
Did you have rebased it on top of the master? `testCustomProfileWindow`, is the last one i have fixed.

t20100:
Yes it was rebased, and I just re-rebased it after merging #3064.

vallsv:
So i think a861ae5a051ea685fc7812ab64bd1f9d8724f01f should be removed. This object is owned by the plot.

The implicit deletion was needed in order to make it work on both windows and linux pyside2. Else one or the other was always failing.

t20100:
Let's try it.

t20100:
Closing as outdated

PR #3074, opened 2020-05-26
Clean up roi manager

### LINKED ISSUE #3065: error when clearing RegionOfInterestManager after clearing parent plot
If I try to clear a RegionOfInterestManager after clearing its parent plot, I get the folowing error:

```
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages\silx\
gui\plot\tools\roi.py", line 415, in clear
    roi.setParent(None)
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages\silx\
gui\plot\items\roi.py", line 173, in setParent
    self._disconnectFromPlot(previousPlot)
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages\silx\
gui\plot\items\roi.py", line 152, in _disconnectFromPlot
    plot.removeItem(item)
  File "C:\Users\Pierre\AppData\Local\Programs\Python\Python37\lib\site-packages\silx\
gui\plot\PlotWidget.py", line 696, in removeItem
    raise ValueError('Item not in the plot')
ValueError: Item not in the plot
```

I’m not sure what the right fix is.

Should the RegionOfInterestManager implicitely delete ROIs when they are cleared from the parent plot, or should we just check that a ROI still belongs to the plot before trying to disconnect it?

vallsv:
This PR clean up ROI when one of its item was removed from the plot.

This allows to use `plot.clear()` to also remove the ROIs, and not only the shape of the ROIs.

Closes #3065

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -192,6 +192,12 @@ class PlotWidget(qt.QMainWindow):
     It provides the item that will be removed.
     """
 
+    sigItemRemoved = qt.Signal(items.Item)
+    """Signal emitted right after an item was removed from the plot.
+
+    It provides the item that was removed.
+    """
+
     sigVisibilityChanged = qt.Signal(bool)
     """Signal emitted when the widget becomes visible (or invisible).
     This happens when the widget is hidden or shown.
@@ -718,6 +724,8 @@ class PlotWidget(qt.QMainWindow):
         if (kind == 'curve' and not self.getAllCurves(just_legend=True,
                                                       withhidden=True)):
             self._resetColorAndStyle()
+
+        self.sigItemRemoved.emit(item)
 
         self.notify('contentChanged', action='remove',
                     kind=kind, legend=item.getName())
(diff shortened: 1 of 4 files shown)


t20100:
There is a `Item.getPlot` that can be used instead of `parent` which provides the `PlotWidget` the item belong to. I think it would be best not to use `parent` until we introduce some group items.

Also I think it would be best to avoid multiplication of signals and avoid `sigItemRemoved` if possible (and add it if no other option), so we can introduce first as a private signal first.

t20100:
I'm looking at it.

vallsv:
For the signal, i don't really agree, cause providing every steps of the life cycle of the items is really useful, depending on the use case. And there is no cost. But it's up to you anyway.

I was starting to fix the parent, but if you do it, it's fine.

vallsv:
Ill take a look of the parent and add a test.

vallsv:
I have removed the use of parent, cause it's better anyway, but this implies to use another way to flag the item. So i have created a magic attribute.

That's not really nice, but for now, while there is no groups i think is it enough.

vallsv:
Anyway i trust much more this version, than the first one i did.
In fact there is less magic now :-D

t20100:
I added an issue to create a `Group` item in `PlotWidget` which would avoid such issues and workaround.
I agree it's not nice, but I doubt there is any better solution until a `Group` is available.

PR #3078, opened 2020-05-28
Rework PlotWidget ROI items handling of events/updates

### LINKED ISSUE #3071: Cannot synchronize ROIs between themself
I have an issue with synchronizing ROIs ending in infinite recursion.
Checking that the value doesn't change when being set should fix it.

In my case it is between different plots, but here is a simple example to reproduce it with a single one:
```
from silx.gui import qt
from silx.gui.plot.tools.roi import RegionOfInterestManager
from silx.gui.plot.items.roi import RectangleROI

import numpy

app = qt.QApplication([])

roi1 = RectangleROI()
roi2 = RectangleROI()

def roi1Updated():
    roi2.setCenter(roi1.getCenter())
roi1.sigRegionChanged.connect(roi1Updated)

def roi2Updated():
    roi1.setCenter(roi2.getCenter())
roi2.sigRegionChanged.connect(roi2Updated)

roi1.setCenter((3, 3))

w = Plot2D()
w.addImage(numpy.arange(100).reshape(10, 10))
w.show()

manager = RegionOfInterestManager(w)
manager.setColor('pink')
manager.addRoi(roi1)
manager.addRoi(roi2)

app.exec_()
```

Leading to:
```
  File "silx/test_roi.py", line 17, in roi1Updated
    roi2.setCenter(roi1.getCenter())
  File "silx/gui/plot/items/roi.py", line 1303, in setCenter
    self.setGeometry(center=position, size=size)
  File "silx/gui/plot/items/roi.py", line 1320, in setGeometry
    with utils.blockSignals(self._handleBottomLeft):
  File "/usr/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "silx/gui/utils/__init__.py", line 43, in blockSignals
    blocked = [(obj, obj.blockSignals(True)) for obj in objs]
RecursionError: maximum recursion depth exceeded
```

t20100:
This PR reworks the `PlotWidget` ROI items to avoid updating and emitting the `sigRegionChanged` signal when a setter is called with the same parameters as the current ones.
This is done for all ROIs but `ArcROI` which is far more complicated than the others.
This is useful e.g., if one wants to synchronize ROI on different plots.

For `EllipsisROI`, I changed the API of `EllipsisROI.setGeometry` has changed (@PiRK is it an issue?), but now, the anchors no longer "jumps" when released after a drag.

For `HorizontalRangeROI`, I changed the order of the marker so on can enlarge the ROI when it is colapsed (before it was only possible to move it around).

closes #3071

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,10 +99,10 @@ before_script:
     # This installs PyQt and scipy if wheels are available
     - pip install ${PIP_OPTIONS} -r requirements.txt
     # Install PySide2 if needed
-    # Pin-pointing for now to avoid issues with versions [5.11.2 - 5.12.1]
+    # Pin-pointing for now to avoid issues with versions [5.15.0]
     - if [ "$QT_BINDING" == "PySide2" ];
       then
-          pip install ${PIP_OPTIONS} pyside2;
+          pip install ${PIP_OPTIONS} pyside2==5.14.2.1;
       fi
 
     # Install built package
--- a/silx/gui/plot/items/roi.py
+++ b/silx/gui/plot/items/roi.py
@@ -836,13 +836,16 @@ class CrossROI(HandleBasedROI, items.LineMixIn):
         HandleBasedROI.__init__(self, parent=parent)
         items.LineMixIn.__init__(self)
         self._handle = self.addHandle()
+        self._handle.sigItemChanged.connect(self._handlePositionChanged)
         self._handleLabel = self.addLabelHandle()
         self._vmarker = self.addUserHandle(items.YMarker())
         self._vmarker._setSelectable(False)
         self._vmarker._setDraggable(False)
+        self._vmarker.setPosition(*self.getPosition())
         self._hmarker = self.addUserHandle(items.XMarker())
         self._hmarker._setSelectable(False)
         self._hmarker._setDraggable(False)
+        self._hmarker.setPosition(*self.getPosition())
 
     def _updated(self, event=None, checkVisibility=True):
         if event in [items.ItemChangedType.VISIBLE]:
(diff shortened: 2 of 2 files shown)


PiRK:
No problem changing the setGeometry API. I have not started using it yet.

I checked that the result of __str__ remains unambiguous regarding the axis used for orientation, and it looks fine.

t20100:
Proper conversions between internal parameters and what is expected are done in the getters that are used in `__str__` so it's correct.
For the orientation I hesitated to return a value in [0, pi[ which is enough to define the ellipsis, but that would loose the  information of the exact position of the handles.

t20100:
I pinpointed pyside2 to run the tests on travis (see #3080)

vallsv:
- While you ensure that `sigRegionChanged` is also emitted when the ROI position is updated programmatically, it's fine. 
- Be careful with the changes you did with `sigItemChanged` and markers, cause my experience was, it emit signals also when the visibility change (which is not really convenient). So check the 4 combinations of stats visible(or not)+editable(or not), to check if there is no extra `sigRegionChanged` or strange behaviors.
- For `HorizontalRangeROI` i think it is better to clamp to min and max. In fact there is at least 3 different valid behaviors. But this ROI is not yet used, so if you think it is better like that, it is fine, cause that's also valid.

vallsv:
I relaunch the tests on Python3.8 PyQt5 WIndows
```
FAIL: testMethodSumLine (silx.gui.plot.tools.test.testProfile.TestProfile3DToolBar)
Simple interaction test to make sure the sum is correctly computed
  File "C:\projects\silx\venv_test\lib\site-packages\silx\utils\testutils.py", line 268, in wrapper
    result = func(*args, **kwargs)
  File "C:\projects\silx\venv_test\lib\site-packages\silx\gui\plot\tools\test\testProfile.py", line 573, in testMethodSumLine
    numpy.testing.assert_almost_equal(data, expected)
  File "C:\projects\silx\venv_test\lib\site-packages\numpy\testing\_private\utils.py", line 579, in assert_almost_equal
    return assert_array_almost_equal(actual, desired, decimal, err_msg)
  File "C:\projects\silx\venv_test\lib\site-packages\numpy\testing\_private\utils.py", line 1042, in assert_array_almost_equal
    assert_array_compare(compare, x, y, err_msg=err_msg, verbose=verbose,
  File "C:\projects\silx\venv_test\lib\site-packages\numpy\testing\_private\utils.py", line 840, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 7 decimals
Mismatched elements: 6 / 6 (100%)
Max absolute difference: 32.
Max relative difference: 0.66666667
 x: array([[ 1.,  4.],
       [ 7., 10.],
       [13., 16.]], dtype=float32)
 y: array([[ 3, 12],
       [21, 30],
       [39, 48]])

```

vallsv:
Looks good.

PR #3079, opened 2020-05-28
Plot ROI: Replaces iterChild|Handles with getChild|Handles

t20100:
This PR replaces the `iterChild` and `iterHandles` methods with `getItems` and `getHandles`.
Interest of iterator here is limited and that would be more inline with the rest of the API (and with a `Group` item).


t20100:
I pinpointed pyside2 to run the tests on travis (see #3080)

vallsv:
That's fine. This API is not supposed to be used. So it can be fixed without problem.

PR #3086, opened 2020-06-02
[CI] Update OS to ubuntu18.04 + update python 3.5 to 3.6

t20100:
This PR updates the OS to ubuntu18.04 and it drops testing of python3.5 since it is not available on the test image.


vasole:
Now it works on PySide2 but not on PyQt5 ?

t20100:
PySide2 is pin-pointed to a working version...

Issue #2100, opened 2018-09-05
[viewer] Additionnal plot description fro NXdata


kif:
For SAXS data, user expect their data to be displayed "as usual". I don't want to enforce novice user to perform 5 clicks in menus to ease the acceptance of NeXus format in this community. For this I would need 4 features which are not (yet?) provided by NXdata:

* [ ] display a plot in semi-log -x or -y or in log-log. One could use the NXdata attribute  @axis_style=["lin", "log"] for semi-log-y, @axis_style=["log", "lin"] for semi-log-x, ...
* [ ] Display all curves from a 2D dataset tagged with @interpretation="spectrum" using @quotity="all" instead of "one" which would be the current default
* [ ] Display an image using log-scale colormap, for example using the @color_normalization="log"
* [ ] Enforce the colormap when displaying images. Imaging beamlines tend to use greyscale color-maps, while diffraction beamlines prefer coloured images. Being able to enforce the colormap would allor to respect the different communities. The attribute NXdata@colormap="magma" would do the job.

This is more a request for comment ... There are little implementation difficulties.

vasole:
I am in favor of a more complete description of the representation to be used for each axis and the signal(s). I have foreseen to discuss it at the NIAC but I can also send it to mailing list to fire up the discussion.

vasole:
I would wait a couple of weeks to see if we get some feedback from https://github.com/nexusformat/NIAC/issues/29

vasole:
For the time being, reasonable candidates seem to be axes_scale_type and signal_scale_type.

On the point of showing all or one curve, that should be a decision of the viewer. Our viewer could offer to show one curve or all.

vasole:
@t20100 

One nice idea that appeared in the discussion was to use just one attribute (something in the lines of plot_hint) so that the application trying to perform the visualization would not need an ever growing set of attributes and it could just be reactive to the qualifiers it understands ignoring those that cannot.

t20100:
+1 for a prefix for extension stuff (could be `ext_` or `addon_` if we don't want to mark it as silx only).

@vasole, what you propose looks like a CSS style sheet string as for Qt widgets. Good point is indeed to avoid having too many attributes, but the down side is to add a parsing of this string to the visualization tool. And anyway, it will have to adapt to new fields.

vasole:
The attribute could be a variable length string array so that you are free to store a list of strings of variable size therefore allowing to write basically anything.

[...]

PR #3092, opened 2020-06-04
Add support of SILX_style attribute to tune default NX plot

### LINKED ISSUE #2928: [view] Hints to plot/display data in linear or logarithmic scale
Can we agree within the *silx* project to define an attribute to tell the plotting tool to set certain axis in log/linear scale when read from a nexus file ? What about 
@scale="log" or "linear" 

The nexus group officially stated this is not their business:
https://github.com/nexusformat/definitions/issues/755#issuecomment-587019174

t20100:
This PR adds support of `SILX_style` hdf5 attribute storing a json string describing the plot.
For now it only supports a `axis_scale_types` key with a list of 'linear' or 'log' value:

`@SILX_style = {'axis_scale_types': ['linear', 'log']}`
Are you OK with this syntax (inspired by https://github.com/nexusformat/NIAC/issues/29)?

This is supported for curves and 2D scatters.
Display of json in silx view is ugly because of escape `\` escape characters.

related to  #2100
closes #2928

### DIFF
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1690,11 +1690,20 @@ class _NXdataXYVScatterView(DataView):
         else:
             y_errors = None
 
+        xscale, yscale = None, None
+        axisTypes = nxd.plot_style.axis_scale_types
+        if axisTypes is not None:
+            if len(axisTypes) >= 1:
+                yscale = axisTypes[-1]
+            if len(axisTypes) >= 2:
+                xscale = axisTypes[-2]
+
         self.getWidget().setScattersData(y_axis, x_axis, values=[nxd.signal] + nxd.auxiliary_signals,
                                          yerror=y_errors, xerror=x_errors,
                                          ylabel=y_label, xlabel=x_label,
                                          title=nxd.title,
-                                         scatter_titles=[nxd.signal_name] + nxd.auxiliary_signals_names)
+                                         scatter_titles=[nxd.signal_name] + nxd.auxiliary_signals_names,
+                                         xscale=xscale, yscale=yscale)
 
     def getDataPriority(self, data, info):
         data = self.normalizeData(data)
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1667,7 +1667,7 @@ class _NXdataXYVScatterView(_NXdataBaseDataView):
     def createWidget(self, parent):
         from silx.gui.data.NXdataWidgets import XYVScatterPlot
         widget = XYVScatterPlot(parent)
-        widget.getPlot().setDefaultColormap(self.defaultColormap())
+        widget.getScatterView().setColormap(self.defaultColormap())
         widget.getScatterView().getScatterToolBar().getColormapAction().setColorDialog(
             self.defaultColorDialog())
         return widget
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1752,6 +1752,7 @@ class _NXdataImageView(_NXdataBaseDataView):
         img_slicing = slice(-2, None) if not isRgba else slice(-3, -1)
         y_axis, x_axis = nxd.axes[img_slicing]
         y_label, x_label = nxd.axes_names[img_slicing]
+        y_scale, x_scale = nxd.plot_style.axes_scale_types[img_slicing]
 
         self.getWidget().setImageData(
             [nxd.signal] + nxd.auxiliary_signals,
@@ -1759,8 +1760,7 @@ class _NXdataImageView(_NXdataBaseDataView):
             signals_names=[nxd.signal_name] + nxd.auxiliary_signals_names,
             xlabel=x_label, ylabel=y_label,
             title=nxd.title, isRgba=isRgba,
-            xscale=nxd.plot_style.axes_scale_types[-1],
-            yscale=nxd.plot_style.axes_scale_types[-2])
+            xscale=x_scale, yscale=y_scale)
 
     def getDataPriority(self, data, info):
         data = self.normalizeData(data)
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1758,7 +1758,9 @@ class _NXdataImageView(_NXdataBaseDataView):
             x_axis=x_axis, y_axis=y_axis,
             signals_names=[nxd.signal_name] + nxd.auxiliary_signals_names,
             xlabel=x_label, ylabel=y_label,
-            title=nxd.title, isRgba=isRgba)
+            title=nxd.title, isRgba=isRgba,
+            xscale=nxd.plot_style.axes_scale_types[-1],
+            yscale=nxd.plot_style.axes_scale_types[-2])
 
     def getDataPriority(self, data, info):
         data = self.normalizeData(data)
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1628,10 +1628,19 @@ class _NXdataCurveView(DataView):
                                           title=nxd.title or nxd.signal_name)
             return
 
+        xscale, yscale = None, None
+        axisTypes = nxd.plot_style.axis_scale_types
+        if axisTypes is not None:
+            if len(axisTypes) >= 1:
+                yscale = axisTypes[-1]
+            if len(axisTypes) >= 2:
+                xscale = axisTypes[-2]
+
         self.getWidget().setCurvesData([nxd.signal] + nxd.auxiliary_signals, nxd.axes[-1],
                                        yerror=nxd.errors, xerror=x_errors,
                                        ylabels=signals_names, xlabel=nxd.axes_names[-1],
-                                       title=nxd.title or signals_names[0])
+                                       title=nxd.title or signals_names[0],
+                                       xscale=xscale, yscale=yscale)
 
     def getDataPriority(self, data, info):
         data = self.normalizeData(data)
--- a/silx/gui/data/DataViews.py
+++ b/silx/gui/data/DataViews.py
@@ -1546,12 +1546,26 @@ class _InvalidNXdataView(DataView):
         return 100
 
 
-class _NXdataScalarView(DataView):
+class _NXdataBaseDataView(DataView):
+    """Base class for NXdata DataView"""
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


vasole:
Is this solution expected to be adopted to other dimension plots? 

I am asking to make clear if first item means x and then y and then z and then ... or if we have to assume that for image plots that becomes dim0, dim1, dim2

vasole:
No problem with the syntax.

A bit off-topic and just for the record: A foreseeable (future) alternative solution would be to have agreed-upon enums. For instance, following MatLab way of handling 1D plots

SILX_style = { 'plot_hint':'linear'}
SILX_style = { 'plot_hint':'semilogx'}
SILX_style = { 'plot_hint':'semilogy'}
SILX_style = { 'plot_hint':'loglog'}

vallsv:
It could be useful to add a specific documentation page for this attribute content?

vallsv:
You can remove the space from the json. it is not needed.

Anyway an HDF5 viewer is not designed to display this kind of stuff, so the result will be ugly anyway.
See Jive and json for years.

t20100:
@vallsv I'd wait a bit to truly advertise and document this feature, but yes we should clearly document the syntax.

@vasole, how it is now is: x axis and then y axis is there is 2 values, y axis if there is only one value.
The rationale I took was NXdata axes first, then signal... but we can change that to dim0, dim1, etc... I really don't know what is best: For defining a plot I would expect x then y (then z) and from a data point of view dim0, dim1... the usual (x, y) or (y, x) convention choice.

What would be good would be to be consistent with NeXus order, i.e., to use the same order as NXdata axes + [signal] (if signal is mapped to a plot axis) (or [signal] + axes depending on the order).
But, what is the order of axes in NXdata? From some data examples, silx view interprets axes in different order: 
- for a scatter axes are in [x, y] order
- for a stack of images, axes are in [slice, row, col] order.
Is it an issue in silx view or a not consistent order in NeXus?

I initially tried to define the `scale_type` associated to NXdata's `signal` and `axes` (i.e., a `signal_scale_type` value and an `axis_scale_types` list), but for curves `signal_scale_type` defines an axis, while for images it defines the colormap normalization... I found that more complicated than needed to in the end say Y is log or colormap normalization is log.

vasole:
The inconsistency is between images and everything else. For images people provides rows and columns but those correspond to y and x in the representation. If we avoid to talk about x and y and we talk about dim0, dim1 there is no ambiguity. If we want to talk about x and y than we have to make clear the image case.

The solomonic decision I can propose is to say in the documentation that the qualifiers in those 'axis_scale_types' are to be directly mapped to the NXdata@axes without any additional re-ordering or interpretation. Perhaps axes_scale_types (instead of axis_scale_types) would serve to make clear the direct mapping.

t20100:
I would also prefer to follow how NeXus orders `axes`, but it raises some questions:
- Is the inconsistency between scatter interpreting `axes` as (x, y) and image interpreting `axes` as (row, col) part of NeXus or an issue in silx?
- For the case of curves, the scale_type is not only needed for the `axes` (X axis), but also for the `signal` (Y axis). How would you deal with that? Prepend or append (or store separately) the signal's scale_type?

vasole:
In our code we have been writing the axes coordinates for images as y an x thus effectively mapping what is to be assigned to the dimensions. The inconsistency is at NeXus side where repeatedly they mix i and j or x and y when referring to indices:

https://manual.nexusformat.org/classes/base_classes/NXdetector.html#nxdetector

I do not think for a 2D detector people transposes the data to match what is written there. Furthermore, who knows how the detector was initially positioned! They should forget about x, y and z and talk about dim0, dim1, dim2.

vasole:
If we deal with axes and signal separated that already solves the linear or logarithmic case for the colormap normalization.

t20100:
For images we use (dim0, dim1) and I hope people do not transpose their data indeed!
But for scatter we use (x axis, y axis) to provide coordinates...

Separating axes/signal `scale_types` provides a way to hint the linear/log colormap normalization, but it uses the same name for setting an axis scale (for curves) OR the colormap normalization (for images, scatters)... The available scales will not be the same and I'm not sure it is easy to get compared to separating both, but that can also work.

t20100:
I see 2 ways to see it:
- As plot hints: so colormap_normalization, plot axes hints (where I would use (x, y, z) order).
- As data description: so `axes_scale_types` (in same order as NeXus) and a `signal_scale_type` that is then interpreted either as colormap_normalization or Y axis scale type depending on the case.

I see the latter more complex to understand, but there is no discussion about order of the axes.

vasole:
To me it uses a representation a signal (linear or log). I do not know if some software will display a curve as a one raw image but it is possible :-)

vasole:
I would prefer the second solution. It does not have any ambiguity.

t20100:
OK, I'll update it.
In the meantime, I implemented the `colormap_normalization`, so I can switch to option 2.

vallsv:
On side note, about log and images, silx do not support something else than linear axes for images. So it should be skipped (no problem of order ^_^).

vasole:
As far as I have understood non-equidistant axes and logarithmic axes are supported (perhaps only in OpenGL backend?)

They are not exposed in our viewers but they are there.

t20100:
The OpenGL backend supported log scale axes for images but it is not possible to use it through the `PlotWidget` which is removing all images from the backend when switching to log scale.

Yet, in `silx view`, when displaying an image as NXdata with irregular axes, it is displayed as a scatter so it supports log scale axes.

vallsv:
Oh, interesting

t20100:
Definitely not final, but with `colormap_normalization` support (way simpler than what I first did yesterday). So current state is plot style oriented: `axis_scale_type` + `colormap_normalization`.

vasole:
I was expecting axis_scale_type and signal_scale_type with the latter covering the colormap normalization.

t20100:
Yes that's an intermediate implementation, I'm converting it to `axes/signal_scale_type`

t20100:
Here is the version with `axes_scale_types` and `signal_scale_type`:
Still to do: handle log axes with 2d image displayed as scatter plot.

I really think doing it this way is obfuscated and does not solve the axes order convention since the issue is a NeXus level between scatter and images.

In the end, I would prefer to use `colormap_normalization` and `axis_scale_types` with an order that matches that of NeXus (so is inconsistent between scatter and images)...

Also in the code, it supports any number of axes for NXdata, is it something we want to handle and in which way?

vallsv:
I feel like there is no use cases for log axes for 2d images. So for an hidden feature for the release it is fine as it is (which there is log to say it is not supported).

vasole:
Well, I suppose we are going to save HKL data at a certain point, therefore we need to go beyond just 2 axes.

The inconsistency images versus other things is also present on higher dimensions but at the end what one wants is to be able to associate axes values and signals and the visualization code has some freedom there.

Assuming a 1 to 1 mapping of our axes_scale_types to the nexus attribute axes, the clear interpretation is 

1 - we have as many values on each axis as items in the signal. The natural interpretation is that the signal has been collected at those "positions/points" and the natural choice is (X, Y, Z) rather than (Z, Y, X).

2 -  we number of values on each axis map each of the dimensions to be visualized: the order clearly corresponds to the signal dimensions and how to show them is a choice of the visualization program. Silx translates that into visualizing as (dim0=Z, dim1=Y, dim2=X) but **any program choice** would be valid.

Common sense tells us than when providing axis coordinates for each point, one does it in the order (x, y, z) while when providing coordinates for each dimension one does it in the order (dim0, dim1, dim2). As far as we do not try to be too smart and we blindly map each scale_type to each axis in the same order everything will work.

t20100:
Ok, makes it clear to me, thanks!

t20100:
Ready for review.

You can generate test data with `./tools/create_h5_sample.py` which generates an `all_types.h5` file.

It seems to work fine for the current cases.

I have some serious doubt on the implementation of the XYV scatter parsing of NXdata in general not just for `SILX_style` if there is more than 2 axes.
It takes the last 2 axes in this order: (x, y), where I would expect to take the first 2 if order is (x, y).
Anyway the scale_type are taken in the same way.

vasole:
In the multidimensional case, you have to distinguish.

You are providing as many axes as dimensions and the product of the axes sizes matches the number of signals. Then clearly you are providing the coordinates along each dimension and how you visualize them is your choice. In silx we are following the Z, Y, X (for instance, stack of images) If you have a single image you have the last dimensions as Y and X and if you have n frames, you still have Z=n_frame, Y, X. If you would have more dimensions, you would still visualize the last dimensions as what they are and for the others you would provide sliders. Remember that we are going to have always (n_points, WHATEVER) where whatever is spectra, images,...).

If you have as many values for each axis as signal items (images, spectra, points), then you visualize in the order X, Y, Z because it is logic to supply coordinates in that way. At the end, what is important is to be able to associate the signal to the corresponding axes coordinates irrespective of what you choose to show as X, as Y and as Z.

vallsv:
The implementation is fine. My only concern is about the previous user colormap configuration (before activating an NXdata which uses an imposed colormap). Maybe it would be better in this case to create and use a dedicated colormap instead of editing the default one. But for now i think it is not a huge problem, and it's maybe safer to not touch too deep inside.

vasole:
To help you decide. If you are using the dimension case and you have an image, the last, and only, two dimensions are the rows and columns and we **choose** to map them as y and x. If we would have an additional dimension (n_frames), we would be showing them as (z, y, x) and it is still the same thing. Your last two coordinates represent the rows and columns in both cases.

In the non dimension mapping, you are always providing as many "point" coordinates as signals and they are always in the order X, Y, Z. One axis -> X, two axes ->X, Y, three axes -> X, Y, Z In that way your first coordinate is always assumed to be X, the second always assumed to be Y, and so on.

How you display things is at the end irrelevant. You have an unique association of each signal to a set of "axes coordinates" and how you present it to the user is your choice.

t20100:
@vallsv, regarding the colormap instance to use, it's surely possible to use a different one (with the constraint of handling the colormap dialog properly). Yet using different instance makes it harder to understand than now where colormap is a unique global state.

As for coordinates, I think there is an issue in the scatter even before this PR about more than 2 dimensions.

t20100:
What I have doubts with is 2 to 3 years-old code... It can't be that bad!
It's ready for me, let me know if you want me to change something.

vallsv:
If there is no other remarks, i will merge it just before lunch. Problems still can be fixed later anyway.

PR #3104, opened 2020-06-05
Fix deprecation warnings

woutdenolf:
Deprecation warnings found when preparing for release 0.13

### DIFF
--- a/examples/findContours.py
+++ b/examples/findContours.py
@@ -549,9 +549,10 @@ class FindContours(qt.QMainWindow):
         self.__customValue = value
         div = 12
         delta = (self.__image.max() - self.__image.min()) / div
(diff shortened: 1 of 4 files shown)


vallsv:
LGTM. Is there a reason why you round first here? `int(numpy.round(value))`

woutdenolf:
> LGTM. Is there a reason why you round first here? `int(numpy.round(value))`

API needs integers so I used floor for the min, ceil for the max and round for the value in between.

vasole:
I avoid to use int(round(x)) in favor of int(x + 0.5)

numpy.round(0.5) = 0.0
numpy.round(1.5) = 2.0

Now it is not as critical as when Python 2 was still used, but bankers rounding is not what is usually expected in science.

PR #3120, opened 2020-06-10
Fix backward compatibility of PlotWidget.addItem

t20100:
This PR ensures a smooth deprecation of `addItem` behaviour replaced by `addShape`.

Keyword only parameter issue reported by @PiRK, thanks!
Change of return value impacting application.


vallsv:
Yes i saw that too on BLISS recently, the syntax is confusing, but in fact `*args` definition is not considered as a positional argument, which makes sense.

vallsv:
But maybe, we could consume the `item` argument from `**kwargs`. So what it make it clearer.

t20100:
Up to you if you want to change the signature as long as it is compatible.
I initially added the `item` argument because it is the only one that should remain in the end (`addItem(self, item)`).

PR #3125, opened 2020-06-15
Fixes rectangle mask drawn outside image

### LINKED ISSUE #3123: Mask image issue when drawing outside the image
Drawing a rectangle mask ouside an image on the left side of the image actually draws a rectangle mask on the image....

t20100:
This PR fixes an issue with the rectangle mask tool when drawing on the left side of the image due to negative indices in slicing.

closes #3123

### DIFF
--- a/silx/gui/plot/MaskToolsWidget.py
+++ b/silx/gui/plot/MaskToolsWidget.py
@@ -187,6 +187,8 @@ class ImageMask(BaseMask):
         :param bool mask: True to mask (default), False to unmask.
         """
         assert 0 < level < 256
+        if row + height <= 0 or col + width <= 0:
+            return  # Rectangle outside image, avoid negative indices
         selection = self._mask[max(0, row):row + height + 1,
                                max(0, col):col + width + 1]
         if mask:


vallsv:
I did not check the result. But dont you think it is better to clamp the area, instead of aborting the mask?

vasole:
I would also clamp the area. It allows the user to be less "careful" when drawing the mask.

I also wonder if we should clamp the area to the visible portion of the image too. For instance, when the image extends beyond the region of the displayed axes.

t20100:
The clamp is done as part of the slicing (https://github.com/silx-kit/silx/blob/51161a6658706c792dd66ac1707c25ad5c5fbe97/silx/gui/plot/MaskToolsWidget.py#L192), except for the issue with negative `end` slice indices which the added test avoids. It can be rejected because the rectangle fully lies outside the image.

Also, you cannot draw outside of the plot area, so you cannot draw a mask on what is no visible... unless you use the arrow keys to pan the plot, but then you intend to do so.

vallsv:
LGTM

PR #3130, opened 2020-06-15
Sample code: Remove auto-edit behavior when adding ROIs

t20100:
This PR removes the toggle of editable ROI when selecting a mode to draw a ROI since this is hard to understand when running the sample code.


vallsv:
LGTM. Smaller sample code, better

PR #3131, opened 2020-06-15
[ROI] Allow to move anchors to the other side of the rect

t20100:
This PR allows to edit rectangular ROI an move the anchors freely without moving the other anchors even when "reverting" the rectangle (i.e., moving top-left anchor to be either the bottom-left or top-right one).


t20100:
This should now works fine without other anchors moving with 'inversing' the rectangle area.
Ready for review.

vallsv:
It's annoying now... it just works.

PR #3135, opened 2020-06-16
fix Plot2D position info with masks

### LINKED ISSUE #3134: [silx view] Pixel intensity with mouse over is not anymore displayed with a mask
On the status bar i always have `Data: 0, Masked`.

t20100:
This PR fixes the displayed information in `Plot2D`'s `PositionInfo` widget when there is a mask.
This is a regression from PR #2913

closes  #3134


vallsv:
Nice, thanks a lot.
Works well.

PR #3140, opened 2020-06-23
Update documentation following drop of python<=3.4

t20100:
- Update documentation to match drop of Python <=3.4.
- Fix sphinx build warnings
- Add graphviz to debian package build dependencies so documentation is built with inheritance images


vallsv:
LGTM

PR #3142, opened 2020-06-24
Added github Pull Request template

### LINKED ISSUE #3141: Improving management of changelog
It's quite tedious to prepare the changelog for the release and some features are not ideally described.
I see a few ways to improve this:

1. Updating the changelog for each important PR.
2. Writing a "news fragment" file for each PR and using a tool to concatenate them (e.g., https://regro.github.io/rever-docs/, https://github.com/twisted/towncrier as numpy) as other projects do (e.g., h5py, numpy). 
3. Adding a PR template which advertise adding a `# changelog` section in the PR description. Then for the release we can extract those to build the final changelog.

- We wanted to do `1.` but it never happened.
- `2.` basically adds one file per PR and you still need to think about doing it or while reviewing it... or add a CI service to check it (e.g., https://github.com/pypa/browntruck)
- `3.` seems to me the simplest way both when making PR and to generate the changelog for the release (and it's easy to set up).

What do you think?

t20100:
This PR proposes to add a PR template aiming at automating/easing the generation of the changelog for releases.

closes #3141

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+<!--
+You are encouraged to write a PR title that is meaningful by itself.
+It will be used to auto-generate the complete changelog.
+
+To highlight a change in the changelog, please add a line starting with `Changelog: ` in the PR description.
+It will be used to generate the changelog's summary.
+
+E.g., Changelog: Added new wonderful feature
+-->
+
+
+Changelog: 


vasole:
I guess we do not have the choice between rst and markdown.

t20100:
I don't think so... I followed github documentation

t20100:
There is no choice for the template as for PR/issue description, but it does not stop us to use rst for the changelog.

payno:
Now you have an (almost) empty template right ? 

It looks like you can have a template for issues and one for PR.
Here is a project containing some examples: https://docs.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates

payno:
ok, I didn't had this one.

PR #3143, opened 2020-06-30
silx.gui.colors: Added `Colormap.get|setNaNColor` to change color used for NaN

t20100:
NaN values in images were displayed differentely with the matplotlib backend (transparent) and the opengl one (as lowest bound of the colormap). 
This PR fixes this difference by making the OpenGL backend behave as the matplotlib one.
Furthermore, it adds `Colormap.get|setNaNColor` methods to allow to change the color used for NaN and it supports it in both `PlotWidget` and `plot3d`.

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -543,12 +543,14 @@ class Colormap(qt.QObject):
     sigChanged = qt.Signal()
     """Signal emitted when the colormap has changed."""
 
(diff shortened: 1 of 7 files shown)


vallsv:
Looks good to me, except the NaN color should be saved in saveState/restoreState. I can do it if you like.

t20100:
I updated the store/restore methods.

vallsv:
Well done. Very nice to check the v2 binary.

PR #3145, opened 2020-07-01
silx.gui: Fixed matplotlib 3.3.0rc1 deprecation warnings

### LINKED ISSUE #3144: [plot] Matplotlib 3.3.0rc1 deprecation warnings
image

t20100:
This PR avoids deprecation warnings raised by matplotlib 3.3.0rc1 during tests.

closes #3144

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -46,9 +46,11 @@ _logger = logging.getLogger(__file__)
 
 try:
     from matplotlib import cm as _matplotlib_cm
+    from matplotlib.pyplot import colormaps as _matplotlib_colormaps
 except ImportError:
     _logger.info("matplotlib not available, only embedded colormaps available")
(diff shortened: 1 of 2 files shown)


vallsv:
Is that a very long deprecation warning?
```
        if _matplotlib_colormaps is not None:
            colormaps.update(_matplotlib_colormaps())
        elif _matplotlib_cm is not None:
            colormaps.update(_matplotlib_cm.cmap_d.keys())
```
Else it is maybe better to do that?

t20100:
Even without the deprecation, `matplotlib.pyplot.colormaps` is what should be used, it's been there for a very long while, see https://github.com/matplotlib/matplotlib/blame/master/lib/matplotlib/pyplot.py#L1874

vallsv:
SOunds good then

PR #3149, opened 2020-07-07
silx.gui.plot3d: Bug fix and minor improvement in scene rendering

t20100:
This PR:
- fixes an issue in rendering/picking of none square 2D images (due to using not reversed shape),
- allows to toggle keep aspect ratio on/off for `Orthographic` (i.e., parallel) projection.


t20100:
Added fix of indentation

PR #3150, opened 2020-07-07
silx.gui.plot.tools.profile: Improved image profile tool to support PlotWidget item extension

t20100:
This PR provides support for more Image item classes than `ImageData` and `ImageRgba`.


vallsv:
LGTM

PR #3153, opened 2020-07-08
Fix custom levels colormap dialog

### LINKED ISSUE #3148: Colormap Dialog looses focus in min/max input widget on image update in 0.13
we are using ImageView for live viewing detector images at 1Hz and noticed a bug in version 0.13
With auto scaling disabled in the Colormap Dialog the min/max input fields loose focus whenever the images updates, making it impossible to adjust the min/max values by hand.
This was working fine in version 0.12

vallsv:
This PR fixes 2 issues on the `ColormapDialog` which was forbid the user to custom the levels of the colormap when the image was often refreshed.

It was still possible to custom the values during the time of a single frame.

This PR:
- Make sure the result of the histogram range edition is used (it was not the case)
- Make sure the handle moved bu the user are to reset during the interaction

Closes #3148

### DIFF
--- a/silx/gui/dialog/ColormapDialog.py
+++ b/silx/gui/dialog/ColormapDialog.py
@@ -155,6 +155,9 @@ class _BoundaryWidget(qt.QWidget):
         if self.__textWasEdited:
             value = self._numVal.value()
             self.__realValue = value
+            with utils.blockSignals(self._numVal):
+                # Fix the formatting
+                self._numVal.setValue(self.__realValue)
             self.sigValueChanged.emit(value)
             self.__textWasEdited = False
 
--- a/silx/gui/dialog/ColormapDialog.py
+++ b/silx/gui/dialog/ColormapDialog.py
@@ -174,16 +174,6 @@ class _BoundaryWidget(qt.QWidget):
             return self.__realValue
         return self._numVal.value()
 
-    def getFiniteValue(self):
-        if not self._autoCB.isChecked():
-            if self.__realValue is not None:
-                return self.__realValue
-            return self._numVal.value()
-        else:
-            if self.__realValue is not None:
-                return self.__realValue
-            return self._numVal.value()
-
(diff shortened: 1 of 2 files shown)


vallsv:
If i could vote, I would be happy to backport this issue in a bug fix release.

t20100:
I was thinking so too.
Maybe we wait a little bit to see if more bugs like this one are showing up.

vallsv:
I think we can wait until mid/end of august without problem.

t20100:
This fixes the issue while dragging the markers, but no when trying to set the min or max in the `LineEdit`.
The latter remains impossible.

I don't get why the displayed min/max are updated when changing the dataset while the colormap is not in autoscale.

vallsv:
You are right. I did not try. I will take a look.

The widget have to know the available theoretical values, in order to abort wrong values for example. But it should just be filter on it's own.

vallsv:
Here is a fix.

The first commit is only a clean up. There was a lot of dead code.

The second filter the `setValue` update to do it only if it is needed.

t20100:
Now it works fine when setting the bounds from the `LineEdit`.

It is still a bit weird when changing the bounds with the markers on the histogram since the values in the `LineEdit` are reset to their previous value at each update of the image.

CI seems to be unhappy.

vallsv:
Ill take a look. If there is too much things to change i don't think it is a good idea for a big fix release. In this case it is better to stay with a small glitch.

vallsv:
The tests was not use the API the right way.

And there was a filter inhibiting the change of the colormap when dragging. Now it's much better. The image color is updated during the interaction, which was not the case before.

vallsv:
While i was there i have removed an unused function.
It's fine to be reviewed.

vallsv:
Which is btw the smartest if i never saw.

t20100:
LGTM, works smoothly, thanks!

PR #3166, opened 2020-07-22
[gui][utils] add signalproxy from pyqtgraph.

payno:
rework a bit to fit silx.
Use a default Qtimer and submitToQtMainThread instead of a ThreadsafeTimer
Was also distributed as MIT code.

### DIFF
--- a/copyright
+++ b/copyright
@@ -24,6 +24,10 @@ Files: silx/third_party/modest_image.py
 Copyright: 2013 Chris Beaumont
 License: MIT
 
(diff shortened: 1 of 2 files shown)


t20100:
You should add the license and copyright owner/date at the beginning of the file and an entry in https://github.com/silx-kit/silx/blob/master/copyright

t20100:
I updated the copyright owner in both the copyright file and the source code to reflect what's in the License file.

LGTM

PR #3168, opened 2020-07-28
Speed up image.tomography.get_next_power()

pierrepaleo:
The module `silx.image.tomography` uses a function `get_next_power()` which, given an integer `n`, finds the next "friendly" power of `(2, 3, 5 ,7)`. This is used for padding data before performing FFT.

The current implementation builds a sorted list of numbers in the form `(2**a)*(3**b)*(5**c)*(7**d)`, and finds the "next friendly power" by bisection. 

Building this list takes some time, which can be noticeable if `get_next_power()` is needed many times. In this PR, the list of powers is cached in a LRU structure.

### DIFF
--- a/silx/image/tomography.py
+++ b/silx/image/tomography.py
@@ -32,6 +32,7 @@ __date__ = "12/09/2017"
 
 import numpy as np
 from math import pi
+from functools import lru_cache
 from itertools import product
 from bisect import bisect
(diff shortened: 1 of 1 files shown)


pierrepaleo:
By the way I wonder if `get_next_power()` and `generate_powers()` should belong to `silx.math.fft`. Any thoughts ?

PR #3176, opened 2020-08-27
silx view: Prevent collapsing browsing panel, Added `-f` command line option

### LINKED ISSUE #3169: [silx view] The right panel can be hidden for users
The right panel can be hidden (drag the panel and hide it on the side).

The layout configuration is saved.

As result, the next user can be surprised not to found the right panel. And reach it back is kind of magic.

Feed back from scientists.

Proposal: Add an option in the `Window` menu to hide and show the right panel using the same layout feature used to hide it with the mouse. So what it became obvious to hide/show the right panel.

t20100:
This PR prevents from completely hiding the left panel and adds the `-f` option to silx view as a short-cut for `--fresh`.

I choose to solve issue #3169 by disabling collapsing of the spliter since it is a simple an clean way to solve it.
The alternative of making sure the browsing panel  is visible sounds more complicated for little improvement if any.

closes #3169

### DIFF
--- a/doc/source/applications/view.rst
+++ b/doc/source/applications/view.rst
@@ -48,7 +48,7 @@ Options
   -h, --help           Show this help message and exit
   --debug              Set logging system in debug mode
   --use-opengl-plot    Use OpenGL for plots (instead of matplotlib)
-  --fresh              Start the application using new fresh user preferences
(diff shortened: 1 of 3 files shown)


vallsv:
Sounds good to me.

PR #3179, opened 2020-08-31
Remove LEFT_BTN filter on FocusManager

### LINKED ISSUE #3178: [plot] Middle-click panning is not used for roi edition
It would be good to be consistent with this behaviour

vallsv:
This PR remove a filter with was used on `FocusManager` in order to limit the manager to the left button.
It sounds like a typo.

As result `DrawSelectMode` can be used within middle click pan.

Closes  #3178

But maybe this `FocusManager` behavior was normal and there is better way to fix the middle pan in `DrawSelectMode`.

Changelog: Fix FocusManager on plot interaction

### DIFF
--- a/silx/gui/plot/PlotInteraction.py
+++ b/silx/gui/plot/PlotInteraction.py
@@ -1350,12 +1350,11 @@ class FocusManager(StateMachine):
     """
     class Idle(State):
         def onPress(self, x, y, btn):
-            if btn == LEFT_BTN:
-                for eventHandler in self.machine.eventHandlers:
-                    requestFocus = eventHandler.handleEvent('press', x, y, btn)
-                    if requestFocus:
-                        self.goto('focus', eventHandler, btn)
-                        break
+            for eventHandler in self.machine.eventHandlers:
+                requestFocus = eventHandler.handleEvent('press', x, y, btn)
+                if requestFocus:
+                    self.goto('focus', eventHandler, btn)
+                    break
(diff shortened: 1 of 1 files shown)


t20100:
Yes, looks like a typo.

t20100:
Actually, the modification was done 4 months ago in the other direction to fix an issue with polygons, see #3033 

So, it seems it needs a bit more digging to fix #3178 without ending-up undoing PR #3033...
I'm looking into it.

t20100:
With the current change, pressing and releasing the middle button while creating a polygon ROI makes a mess.
Reproducible with `examples/plotInteractiveImageROI.py`

vallsv:
Yes sure, i was afraid of side effects.

I think during the focus, if another button is pressed, the focus should be cancelled. The focus became then kind of exclusive. That's the easiest way to fix it. It removes a lot of flexibility, which could be convenient, but i don't think we really need it.

If you think it is good enough, i can make a try.

t20100:
I am looking at refactoring the basic interaction code to have consistent features (i.e., zoom on wheel and pan on middle click) across all interaction modes and reserve only left button for custom interaction.

vallsv:
Sounds good and the easiest route.

t20100:
Easier to say than to do...

t20100:
Closing as supersede by PR  #3291 
Larger refactoring is needed: #3292

PR #3180, opened 2020-09-02
silx.gui.plot.items: Added `DATA_BOUNDS` visualization parameter for `Scatter` item histogram bounds

### LINKED ISSUE #3177: [scatter] Histo ranreding with BINNED_STATISTIC_SHAPE and GRID_BOUNDS
I would like to know if `BINNED_STATISTIC_SHAPE` uses `GRID_BOUNDS`.

I though it was the case, but with few rework of Flint i doubt about it. Or i have buggy metadata.

I feel like only the data bound is used.
It would be nice to use the grid bound too, in order to stabilize the cell size with live scatters.
I guess it should not be a huge fix.

t20100:
This PR adds `VisualizationParameter.DATA_BOUNDS` to set the histogram range for binned statistics.

As it is, it is completely decoupled from `GRID_BOUNDS`, but it could be related, e.g., using `GRID_BOUNDS` if `DATA_BOUNDS` is not defined.

It would be cleaner to replace `GRID_BOUNDS` with `DATA_BOUNDS` as in this PR and a `GRID_DIRECTION` or `GRID_ORIENTATION` providing the information about whether X and Y increase or decrease from one point to the next, i.e., `(y direction, x direction)`... but that would change the API or make it redundant.

closes #3177

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -1067,6 +1067,15 @@ class ScatterVisualizationMixIn(ItemMixInBase):
         Available reduction functions are: 'mean' (default), 'count', 'sum'.
         """
 
+        DATA_BOUNDS = 'data_bounds'
+        """The expected bounds of the data in data coordinates.
+
+        A 2-tuple of 2-tuple: ((ymin, ymax), (xmin, xmax)).
+        This provides the expected data ranges in both dimensions.
+
+        WARNING: dimension 0 i.e., Y first.
+        """
+
     _SUPPORTED_VISUALIZATION_PARAMETER_VALUES = {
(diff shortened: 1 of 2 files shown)


vallsv:
I would prefer to use the ranges as an hint.

I can vaguely know the range, but the real data could be outside. So i think it would be better to min/max the data within the provided range, and not overwrite the min/max data.

What do you think?

t20100:
OK, I'll change that.

t20100:
I changed `DATA_BOUNDS` to `DATA_BOUNDS_HINT` so that the hint is eventually enlarged with actual data bounds.

vallsv:
LGTM. Works as expected.

PR #3184, opened 2020-09-07
silx.gui.utils: Fix `glutils.isOpenGLAvailable` issue

### LINKED ISSUE #3182: ModuleNotFoundError: No module named 'concurrent.futures'; 'concurrent' is not a package
Fun issue, no idea why it happens now on my machine.

This avoid to launch Flint with OpenGL, so it is quite annoying.

```
In [1]: import silx.gui.utils.glutils              
   ...: silx.gui.utils.glutils.isOpenGLAvailable()                                                                                                                                                                                                   
['/users/valls/Software/miniconda3/envs/silxenv/bin/python', '/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/utils/glutils.py', '2', '1']
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/utils/glutils.py", line 30, in <module>
    import subprocess
  File "/users/valls/Software/miniconda3/envs/silxenv/lib/python3.7/subprocess.py", line 50, in <module>
    import signal
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/utils/signal.py", line 32, in <module>
    from silx.gui.utils import concurrent
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/utils/concurrent.py", line 35, in <module>
    from concurrent.futures import Future
  File "/users/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/utils/concurrent.py", line 35, in <module>
    from concurrent.futures import Future
ModuleNotFoundError: No module named 'concurrent.futures'; 'concurrent' is not a package
Out[1]: <_isOpenGLAvailableResult: False, "Qt OpenGL widget error: retcode=1, error=b''">
```

t20100:
This PR fixes the runtime check of OpenGL by removing `sys.path[0]` when running the subprocess since it contains the directory of the "script", here it is `silx/gui/utils/` which also contains `signal` and `concurrent` which conflicts with python modules....
There is a check in case something is already messing around with `sys.path` (e.g., through `PYTHONSTARTUP`).

closes #3182

Some alternatives for the record:
- Copying (or creating) the script in a temporary folder
- Creating a `silx.gui.glutils` package with only `__init__` and the check script.

### DIFF
--- a/silx/gui/utils/glutils.py
+++ b/silx/gui/utils/glutils.py
@@ -27,6 +27,13 @@
 
 import os
 import sys
+
+if __name__ == "__main__":
+    # When run as a script, remove directory from sys.path
+    # This avoids other script in same directory to override Python modules
+    if os.path.abspath(sys.path[0]) == os.path.abspath(os.path.dirname(__file__)):
+        sys.path.pop(0)
+
 import subprocess
 from silx.gui import qt
 


vallsv:
It's true it's kind of dirty.

Up to you for the fix. But I would prefer another script file instead of creating new one every time we trigger this function.

t20100:
From a quick look at cpython code, I didn't find a way to avoid the addition of the script directory only.

If it's OK for you, I'm fine with this fix, it's no nice but it avoids spreading this small functionality across multiple files.

vallsv:
I still feel like it would be cleaner with a side script.
But i just expect it to work, so it is fine to me.

vallsv:
BTW do we know if it is working on 3.5 now?

vallsv:
Ok i understand the patch now, so it is strong than what i expect. Should be fine.

PR #3195, opened 2020-09-21
silx.gui.plot.PlotWidget: Fixed time serie axis for range < 2.5 microseconds

### LINKED ISSUE #3192: Plot: AssertionError: Integer value or tickstep is 0
I found that on a beamline.

No idea how to reproduce it.

No idea if it is from an image or a curve widget...

Maybe you have some more ideas.

Screenshot from 2020-09-21 11-42-37

t20100:
This PR fixes a PlotWidget issue with time serie axis when the extent of the axis is small (< 2.5us).

There was 2 issues:
- range < 2.5us as in #3192
- range << 1us

closes #3192


vallsv:
Sounds good.

PR #3196, opened 2020-09-22
silx.gui.plot.PlotWidget: Added `get|setAxesMargins` methods to control margin ratios around plot area

### LINKED ISSUE #2681: PlotWidget doesn't use available screen space
I'm using the `PlotWidget`, integrated into one of our applications.

However, I find that the available screen space is not efficiently used. There is too much empty space around the data and axes, as displayed below:
silx-widget-space

The blue is the widget background color, as set via `PlotWidget.setBackgroundColor`.

Is there any way to improve this?

### LINKED ISSUE #3193: Colormap dialog: Very small histogram in OpenGL
Hi,

No idea if it is an issue, but i feel like if silx is setup to use OpenGL, the colormap dialog then provide a very small histogram which is not convenient to use. Or which could be improved.

Do you think there is easy way to optimize the space here?

Maybe disabling OpenGL for this specific plot?

Screenshot from 2020-09-21 11-46-11

t20100:
This PR adds `PlotWidget.setAxesMargins(left, top, right, bottom)` and `PlotWidget.getAxesMargins` methods for controlling the margins in which the axes and labels are displayed.
It is implemented for both matplotlib and OpenGL backends.

closes #3193 
closes #2681

### DIFF
--- a/doc/source/modules/gui/plot/plotwidget.rst
+++ b/doc/source/modules/gui/plot/plotwidget.rst
@@ -82,6 +82,7 @@ The following methods handle plot limits, aspect ratio, grid and axes display:
 .. automethod:: PlotWidget.setKeepDataAspectRatio
 .. automethod:: PlotWidget.getGraphGrid
 .. automethod:: PlotWidget.setGraphGrid
+.. automethod:: PlotWidget.isAxesDisplayed
 .. automethod:: PlotWidget.setAxesDisplayed
 .. automethod:: PlotWidget.getAxesMargins
 .. automethod:: PlotWidget.setAxesMargins
--- a/doc/source/modules/gui/plot/plotwidget.rst
+++ b/doc/source/modules/gui/plot/plotwidget.rst
@@ -83,6 +83,8 @@ The following methods handle plot limits, aspect ratio, grid and axes display:
 .. automethod:: PlotWidget.getGraphGrid
 .. automethod:: PlotWidget.setGraphGrid
 .. automethod:: PlotWidget.setAxesDisplayed
+.. automethod:: PlotWidget.getPlotMargins
+.. automethod:: PlotWidget.setPlotMargins
 
 Reset zoom
 ..........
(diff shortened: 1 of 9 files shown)


vallsv:
I don't really agree for replacing the visibility of the axis with the margin. It's not the same concept.

It's somehow like saying you can remove `button.setVisible(False)` cause you can use `button.setFixedSize(0, 0)`.

I think i want to be able to set the margin, and i want to be able to hide or display the axis. That's commonly used for Flint plot displaying images. I don't want to store on my side the previous margin size to be able to restore it. That's something you easily can deal with.

This said i have not checked the PR.

t20100:
OK, should be simple to change (only need to be implemented in `PlotWidget`).
Then, I'll rename `get|setPlotMargins` to `get|setAxesMargins` so it ressembles `setAxesDisplayed`... and also probably make `isAxesDisplayed` public.

vasole:
I'm a bit confused.

Are we talking about margins inside the plot (so that maximum does not touch the borders and so on), about margins between the plot rectangle and the axes labels or about something else?

t20100:
That's the margins between the axes and the border of the widget (i.e., where the labels and ticks are).
It also makes OpenGL and matplotlib backends consistent (it was not).

vasole:
I thought those borders could be controlled just by using Qt, so you are clearly talking about other thing.

| widget margins (Qt)|Left axis Labels|Plot Rectangle|right Axis labels|widget margins(Qt)|

So, you are talking about margins between the plot rectangle and the axes labels.

t20100:
Yes, that's the "margins" where the labels are drawn.

vallsv:
Thanks a lot. Yes the getter should be public, i guess i use it.

t20100:
Updated to restore `setAxesDisplayed` and made `isAxesDisplayed` public.

vallsv:
Very nice. Thanks for the patch.

PR #3197, opened 2020-09-23
silx.gui.plot: Fixed `RegionOfInterest` refresh when highlighted

### LINKED ISSUE #3191: [roi] setColor do not update the plot
Sounds like from my test that the `setColor` from ROI object do not trigger a repaint of the plot.

As result the ROI color is updated at the next plot refresh, like a zoom in/out.

t20100:
This PR triggers the refresh of the style on more events when highlighted to make sure the style is refreshed.

closes #3191


vallsv:
Nice!

PR #3199, opened 2020-09-23
Save error-bars when saving a plot

### LINKED ISSUE #3198: [silx view] Export error-bars when saving curves
Plots with error-bars should be saved with error-bars (needed for both SAXS and EXAFS)

kif:
close #3198 

Changelog: Save error-bars when saving a plot


kif:
Missing: the saving spec-files with multiple columns

kif:
hold on, "x y dx dy" is more common than "x dx y dy" ... I'll change the output format (as it is hard coded)

silx/io/utils.py
```diff
@@ -190,34 +187,46 @@ def save1D(fname, x, y, xlabel=None, ylabels=None, filetype=None,
...
+            specf = savespec(fname, x, y, xlabel, ylabels, fmt=fmt,
+                     scan_number=1, mode="w", write_file_header=True,
+                     close_file=False)
+        else:
+            y_array = numpy.asarray(y)
+            # make sure y_array is a 2D array even for a single curve
+            if y_array.ndim == 1:
+                y_array.shape = 1, -1
+            elif y_array.ndim not in [1, 2]:
+                raise IndexError("y must be a 1D or 2D array")
```
vallsv, line 218: This should not be an IndexError. Could be a ValueError, or just a RuntimeError

silx/io/utils.py
```diff
@@ -333,40 +344,51 @@ def savespec(specfile, x, y, xlabel="X", ylabel="Y", fmt="%.7g",
 
     x_array = numpy.asarray(x)
     y_array = numpy.asarray(y)
+    if y_array.ndim > 2:
+        raise IndexError("Y columns must have be packed as 1D")
 
-    if y_array.shape[0] != x_array.shape[0]:
+    if y_array.shape[-1] != x_array.shape[0]:
         raise IndexError("X and Y columns must have the same length")
```
vallsv, line 351: Oh, i see, same exception as this one. Both could be fixed, i think.

silx/io/utils.py
kif: The error previously raised was already this one ... I did not want to change it.

silx/io/utils.py
kif: I let Armando or Thomas decide if it needs to be fixed (or not). There are also some consistencies in the specfile part.

t20100: For both, `ValueError` would indeed be better, but it's been `IndexError` for the last 5 years....
To me it is not worth changing it.

kif:
@vasole do you agree with this ?

vasole:
Concerning the objective of the PR I have no problem.

What I am not sure about is to replace python by python3 in the scripts. To me it is up to the user to configure its environment to have python as python3 and the scripts whould have  ``#!/usr/bin/env python`` as before.

Unless there is strong consensus (and even then), I would put that change in a separate PR and not hidden into a PR to save error bars.

t20100:
`/usr/bin/env python` is not working on systems where you only have python3 installed and since we no longer support python2, to me, there is no reason not to use `python3` for dev scripts.

If we need to discuss this further, it is indeed best in a separate PR, else to avoid hiding it, adding mention of it after `Changelog:` in the first comment will make us consider it to write the next version's changelog.

vasole (approved):

t20100 (approved):

PR #3208, opened 2020-09-30
use Python3 for all developer tools

kif:
Changelog: Switch to Python3 for all development tools.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,5 +1,5 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
+#!/usr/bin/env python3
+# coding: utf8
 """
 Bootstrap helps you to test scripts without installing them
 by patching your PYTHONPATH on the fly
@@ -10,8 +10,7 @@ example: ./bootstrap.py ipython
 __authors__ = ["Frédéric-Emmanuel Picca", "Jérôme Kieffer"]
 __contact__ = "jerome.kieffer@esrf.eu"
 __license__ = "MIT"
-__date__ = "26/07/2018"
-
+__date__ = "30/09/2020"
 
 import argparse
 import distutils.util
@@ -20,7 +19,6 @@ import os
 import subprocess
 import sys
 import tempfile
-
 
 logging.basicConfig()
 logger = logging.getLogger("bootstrap")
@@ -69,6 +67,7 @@ def _get_available_scripts(path):
 
 
 if sys.version_info[0] >= 3:  # Python3
+
     def execfile(fullpath, globals=None, locals=None):
         "Python3 implementation for execfile"
         with open(fullpath) as f:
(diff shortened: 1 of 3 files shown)


t20100 (approved):
LGTM

t20100:
@vasole I'm adding you to the reviewers since you initiated the discussion around this.

vasole:
My take on this is that the proper way of doing thing is ``#!/usr/bin/env python`` and that writing python3 is a linux distribution bound thing.

Apparently I am not the only one thinking that way

https://bugs.python.org/issue28686

vallsv:
`python` target isnt supposed to be already python 3?

vasole:
Conda is doing things properly, and when you activate a environment, python refers to the python of that environment even if python3 or python2 are available.

For instance, under windows python3 does not exist. It is just python.

vasole:
@vallsv 

Apparently not all the unix distributions have caught up. With conda it is the case.

kif:
then setup.py should be reverted as well ...

vallsv:
It should not change much things anyway, no?

I mostly never use this scripts without prefixing them with the python i want to use.

t20100:
virtualenv behaves the same as conda.
For me, since I use virtualenvs for dev, this PR will not change anything.

It's for the "dev" scripts so it should not change much.

BTW both ubunut20.04 and debian11 seems to consider python as python2.

vasole:
To me the change is not needed and it is a question of time to have debian and ubuntu considering python as whatever python is first in the path.

I am -1 on this one but I am just one of many.

kif:
Ubuntu has a package: https://packages.ubuntu.com/focal/python-is-python3
but this does not exist under debian11

mretegan:
On the cluster, the python command doesn't point to any version of Python.

```
retegan@slurm-nice-access201:~> type -a python
python not found
retegan@slurm-nice-access201:~> type -a python3
python3 is /usr/bin/python3
python3 is /bin/python3
```

vasole:
@mretegan 

Since the cluster is ubuntu, then it is perfect to ask installation of the package @kif mentioned above:

https://packages.ubuntu.com/focal/python-is-python3

mretegan:
@vasole I was just trying to point out that not having the python command might be the default case for many Linux installations. Defaulting to python3 seems like the safer choice, at lest for Linux. Hi

vasole:
My point is that it is the safest choice for ubuntu/debian.

Anyone using a virtual env, conda or windows is better served with python (that sooner or later will default to the python 3 version).

In any case I gave my vote. The others can vote too and that's it.

kif:
> Anyone using a virtual env, conda or windows is better served with python (that sooner or later will default to the python 3 version).

venv & conda are working fine with python3. For windows, I have limited experience.

Moreover those are dev' only scripts.

t20100:
I also gave my vote by accepting this PR and @kif by proposing it.
Unless more people disagree on this, I'll merge this PR.

PR #3212, opened 2020-10-07
silx.gui.plot.items: Refactoring PlotWidget items by adding a DataItem base class

t20100:
This PR reworks PlotWidget items by adding a `DataItem` base class for items having a "data extent" in the plot.

It allows to mutualise some code and fix some minor inconsitencies between the previous different copies...
It also prepares for support of plot items updating their content depending on their visible extent.


vallsv:
Sounds better with `bounds` instead of `dataExtent` concept.

LGTM.

PR #3223, opened 2020-10-14
silx.gui.plot.items: Added item visible bounds feature to PlotWidget items

t20100:
This PR adds method `Item.getVisibleBounds`~~`Extent`~~ to PlotWidget items to retrieve the visible extent of the item bounding box in the plot area.
It also adds a private API to enable tracking changes of this visible extent through the `Item._sigVisibleBounds`~~`Extent`~~`Changed` signal.
This signal is only triggered when enabling "visible extent tracking" through `Item._setVisibleBounds`~~`Extent`~~`Tracking` and which can be retrieved through`Item._isVisibleBounds`~~`Extent`~~`Tracking`.

### DIFF
--- a/silx/gui/plot/items/core.py
+++ b/silx/gui/plot/items/core.py
@@ -37,6 +37,7 @@ except ImportError:  # Python2 support
 from copy import deepcopy
 import logging
 import enum
+from typing import Optional, Tuple
 import warnings
 import weakref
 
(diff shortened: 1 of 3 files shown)


vallsv:
Be careful  that `Tuple` concept is like C, fixed number of typed elements. `Tuple[float]` is a tuple of a single float. Here you want `Tuple[float,float,float,float]`.

Maybe it would be better to rename `getVisibleExtent` with `bound` something?

t20100:
Renamed `visibleExtent` to `visibleBounds` and fixed `getVisibleBounds` typing.

PR #3162, opened 2020-07-21
silx.io.dictdump: Make `h5todict` resilient to issues in the HDF5 file

t20100:
This PR makes the `h5todict` function resilient to issues in the HDF5 file such as broken soft or external links or corrupted datasets, and in those cases it prints an error message and skip the link/dataset instead of raising an exception.
It also prints an error if the `path` argument is not in the file and returns an empty dict.

Basically this turn exceptions to print an error and skip.
Do we want to do that?

closes #3160

attn @vasole

PR #3224, opened 2020-10-14
dictdump: support HDF5 link preservation

### LINKED ISSUE #3213: extend dicttoh5 to handle links
In analogy to what we did for attributes (#3005) I need a syntax to specify links for dicttoh5 and h5todict. For example when doing data processing we often want to do this:

```python
{
    "@NX_class": "NXroot",
    "1.1": {
        "@NX_class": "NXentry",
        "instrument": {
            "@NX_class": "NXinstrument",
            "fx8_det0": {"@NX_class": "NXdetector", "data": [0, 1, 2, 3]},
        },
        "my_calc": {
            "@NX_class": "NXprocess",
            "configuration": {
                "@NX_class": "NXcollection",
                ">y": "../instrument/fx8_det0/data",  # SYNTAX FOR LINK
            },
        },
    },
}
```

So attributes start with "@" and links start with ">"? The library would automatically handle relative links and soft vs. external. For example `"scans/test.h5::/1.1/instrument/fx8_det0/data"` would be a relative external link and `"/1.1/instrument/fx8_det0/data"` would be an absolute softlink.

Before spending time on the implementation I'd like to know whether this idea finds enough support and what the appropriate syntax could be. For example `">linkname"` or `"linkname>"`.

woutdenolf:
closes #3213 

HDF5 link support:

 * `dicttoh5`: values `h5py.SoftLink` and `h5py.ExternalLink` are supported
 * `h5todict`: `dereference_links` option. `True` by default, otherwise the dict will contain `h5py.SoftLink` and `h5py.ExternalLink` instances.

HDF5 attribute support:

 * `h5todict`: `include_attributes` option. `False` by default, otherwise the dict will contain attributes as tuple keys.

Nexus wrappers: support `">"` notation and refactor

* `h5_to_nexus_dict`:
  * convert tuple keys to `"@"` key notation
  * convert `h5py.SoftLink` and `h5py.ExternalLink` values to `">"` key notation (value is a URL as a string)
* `nexus_to_h5_dict`:
  * convert `"@"` key notation to tuple keys
  * convert `">"` key notation (value is a URL as a string) to  `h5py.SoftLink` and `h5py.ExternalLink` values
  * add up-link support for soft link with the `">"` key notation (resolved to absolute link)
* `nxtodict`: same as `h5_to_nexus_dict(h5todict(ddict))`
* `dicttonx`: same as `dicttoh5(nexus_to_h5_dict(ddict))`

### DIFF
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -219,66 +219,58 @@ def dicttoh5(treedict, h5file, h5path='/',
                 h5f.create_group(h5path)
 
         for key in filter(lambda k: not isinstance(k, tuple), treedict):
-            if isinstance(treedict[key], dict) and len(treedict[key]):
+            key_is_group = isinstance(treedict[key], dict)
+            h5name = h5path + key
+
+            if key_is_group and treedict[key]:
                 # non-empty group: recurse
-                dicttoh5(treedict[key], h5f, h5path + key,
+                dicttoh5(treedict[key], h5f, h5name,
                          overwrite_data=overwrite_data,
                          create_dataset_args=create_dataset_args)
+                continue
 
-            elif treedict[key] is None or (isinstance(treedict[key], dict) and
-                                           not len(treedict[key])):
-                if (h5path + key) in h5f:
-                    if overwrite_data is True:
-                        del h5f[h5path + key]
-                    else:
-                        logger.warning('key (%s) already exists. '
-                                       'Not overwriting.' % (h5path + key))
-                        continue
+            if h5name in h5f:
+                # key already exists: delete or skip
+                if overwrite_data is True:
+                    del h5f[h5name]
+                else:
+                    logger.warning('key (%s) already exists. '
+                                    'Not overwriting.' % (h5name))
+                    continue
+
+            if treedict[key] is None or key_is_group:
                 # Create empty group
-                h5f.create_group(h5path + key)
-
+                h5f.create_group(h5name)
             else:
                 ds = _prepare_hdf5_dataset(treedict[key])
                 # can't apply filters on scalars (datasets with shape == () )
                 if ds.shape == () or create_dataset_args is None:
-                    if h5path + key in h5f:
-                        if overwrite_data is True:
-                            del h5f[h5path + key]
-                        else:
-                            logger.warning('key (%s) already exists. '
-                                           'Not overwriting.' % (h5path + key))
-                            continue
-
-                    h5f.create_dataset(h5path + key,
+                    h5f.create_dataset(h5name,
                                        data=ds)
                 else:
-                    if h5path + key in h5f:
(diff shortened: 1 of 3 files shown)


woutdenolf:
@t20100 I need to do up-link support to dicttonx differently (parents should be derived and not an argument). The rest is done.

woutdenolf:
Ok I fixed the up-link support. Should be good now.

silx/io/dictdump.py
```diff
@@ -412,28 +413,124 @@ def h5todict(h5file, path="/", exclude_names=None, asarray=True):
...
+            h5name = path + "/" + key
+            # Preserve HDF5 link when requested
+            if not dereference_links:
+                lnk = h5f.get(h5name, getlink=True)
+                if is_link(lnk):
+                    ddict[key] = lnk
+                    continue
+
+            try:
+                h5obj = h5f[h5name]
```
woutdenolf, line 440: @t20100 I added this but I should remove it again? (see #3162)

woutdenolf: I removed it

silx/io/dictdump.py
```diff
@@ -291,78 +287,83 @@ def dicttoh5(treedict, h5file, h5path='/',
...
-    h5path="/",
-    mode="w",
-    overwrite_data=False,
-    create_dataset_args=None,
-):
-    """
-    Write a nested dictionary to a HDF5 file, using string keys as member names.
-    The NeXus convention is used to identify attributes with ``"@"`` character,
-    therefor the dataset_names should not contain ``"@"``.
+def nexus_to_h5_dict(treedict, parents=tuple()):
```
t20100, line 293: Shouldn't this be private?

silx/io/dictdump.py
```diff
@@ -291,78 +287,83 @@ def dicttoh5(treedict, h5file, h5path='/',
...
+            copy[key] = nexus_to_h5_dict(value, parents=parents+(key,))
+        else:
+            copy[key] = value
+    return copy
 
-    Example::
 
-        import numpy
-        from silx.io.dictdump import dicttonx
+def h5_to_nexus_dict(treedict):
```
t20100, line 339: Same, shouldn't it be private?

t20100 (approved):
LGTM.
Minor comments about making 2 internal functions private.

silx/io/dictdump.py
woutdenolf: The functions `h5_to_nexus_dict` (returns a dict without special objects) and `nexus_to_h5_dict` (returns a dict with special objects) could be useful if someone want to serialize/deserialize?

t20100: OK, if that can be useful let's keep it, from looking at the code, I though that was only for internal use.

PR #3225, opened 2020-10-15
silx.gui.hdf5.test.test_hdf5: forgot to add TestH5Item to `suite`

woutdenolf:
@vallsv  Locally I have this test failing (not sure if it is related to my changes).

```python
ERROR: testSynchonized (silx.gui.hdf5.test.test_hdf5.TestHdf5TreeModelSignals)
  File "/users/denolf/dev/silx/build/lib.linux-x86_64-3.8/silx/gui/hdf5/test/test_hdf5.py", line 440, in testSynchonized
    self.model.synchronizeH5pyObject(self.h5)
  File "/users/denolf/dev/silx/build/lib.linux-x86_64-3.8/silx/gui/hdf5/Hdf5TreeModel.py", line 613, in synchronizeH5pyObject
    self.synchronizeIndex(qindex)
  File "/users/denolf/dev/silx/build/lib.linux-x86_64-3.8/silx/gui/hdf5/Hdf5TreeModel.py", line 591, in synchronizeIndex
    self.insertFileAsync(filename, index.row(), synchronizingNode=node)
  File "/users/denolf/dev/silx/build/lib.linux-x86_64-3.8/silx/gui/hdf5/Hdf5TreeModel.py", line 679, in insertFileAsync
    qt.silxGlobalThreadPool().start(runnable)
RuntimeError: wrapped C/C++ object of type QThreadPool has been deleted
```

### DIFF
--- a/silx/gui/hdf5/test/test_hdf5.py
+++ b/silx/gui/hdf5/test/test_hdf5.py
@@ -1132,6 +1132,7 @@ def suite():
     test_suite.addTest(loadTests(TestNexusSortFilterProxyModel))
     test_suite.addTest(loadTests(TestHdf5TreeView))
     test_suite.addTest(loadTests(TestH5Node))
+    test_suite.addTest(loadTests(TestH5Item))
     return test_suite
 
 


vallsv:
`silxGlobalThreadPool` is supposed to be global, so i don't really know how it could be released.

vallsv:
This could be protected even more by using the Qt API to check if the c++ object is still alive.What do you think @vincefn?

But i feel it weird anyway.

t20100:
Looks weird indeed.

What context do you run the test in (PyQt5/PySide2 version, OS, python version)?

Only reason I can see for it to happen is if someone sets the `QThreadPool` parent and the parent is destroyed... but from a search in the code it doesn't seems to be the case (and it would fail everywhere).

This case could be prevented by using a subclass of `QThreadPool` to avoid someone to set the parent, something like:
```python
class _SilxThreadPool(qt.QThreadPool):
    def __init__(self):
        super().__init__()

    def setParent(self, parent):
        if parent is not None:
            raise RuntimeError("Cannot set parent on silx global thread pool")
        else:
            super().setParent(parent)
```

woutdenolf:
```python
python run_tests.py silx.gui.hdf5.test.test_hdf5
 ```
or
 ```python
python bootstrap.py -m silx.gui.hdf5.test.test_hdf5
 ```
Python 3.8 (inside a venv), PyQt5 on Ubuntu 20.04.1 LTS

 ```python
(py38) denolf@lindenolf:/home/denolf/dev/silx$ python ci/info_platform.py 
Python 64 bits
       maxsize: 9223372036854775807      maxunicode: 1114111
3.8.5 (default, Sep 15 2020, 11:23:56) 
[GCC 9.3.0]
 
Platform: Linux-5.4.0-48-generic-x86_64-with-glibc2.2.5
- Machine: x86_64
 
Config: ["'--prefix=/users/denolf/.pyenv/versions/3.8.5' '--enable-shared' '--libdir=/users/denolf/.pyenv/versions/3.8.5/lib' 'LDFLAGS=-L/users/denolf/.pyenv/versions/3.8.5/lib -Wl,-rpath=/users/denolf/.pyenv/versions/3.8.5/lib ' 'CPPFLAGS=-I/users/denolf/.pyenv/versions/3.8.5/include '"]

Numpy 1.19.2
      include /users/denolf/virtualenvs/silx/py38/lib/python3.8/site-packages/numpy/core/include
      options {'edgeitems': 3, 'threshold': 1000, 'floatmode': 'maxprec', 'precision': 8, 'suppress': False, 'linewidth': 75, 'nanstr': 'nan', 'infstr': 'inf', 'sign': '-', 'formatter': None, 'legacy': False}

PyOpenCL platform:
  <pyopencl.Platform 'NVIDIA CUDA' at 0x558b17898240>
    <pyopencl.Device 'Quadro P1000' on 'NVIDIA CUDA' at 0x558b17897dd0> max_workgroup_size is 1024
  <pyopencl.Platform 'Portable Computing Language' at 0x7fe7bd7a0020>
    <pyopencl.Device 'pthread-Intel(R) Xeon(R) W-2133 CPU @ 3.60GHz' on 'Portable Computing Language' at 0x558b1789c590> max_workgroup_size is 4096
Unable to import silx
Qt (from PyQt5): 5.15.1
SIP: 5.4.0
 ```

t20100:
It is not related to you change, I reproduced with current master.

woutdenolf:
Ok then this can be merged (I forgot to add a test to the `suite`).

PR #3227, opened 2020-10-15
build: Added pyproject.toml to comply to PEP 518

t20100:
This PR adds a `pyproject.toml` to the project to comply with [PEP 518](https://www.python.org/dev/peps/pep-0518/).
This enables `pip` to install `silx` in a fresh virtualenv without numpy already installed.


vasole:
I guess a working compiler is assumed at the building side.

t20100:
Yes what else.
I forgot to mention that it is only for source install, not for wheels of course.

PR #3231, opened 2020-10-16
Build: Fixed CI by temporarily using h5py2 for tests

t20100:
Only for CI to pass, does not fix anything....


vallsv:
Sounds pragmatic for now

Issue #2429, opened 2019-02-12
[ColormapDialog/PlotWidget] Add an option to set colormap range from visible data


t20100:
It would be convenient to have a button in the ColormapDialog to set the colormap range from the currently visible data when displaying an image (or a scatter plot), so one can adjust the contrast to what currently displayed.

t20100:
Feedback from beamline: +1

payno:
IT could be also generalize for other Items (curve...). But will bring some complexity. Let's focus on image / scatter plot from a RegionOfInterestManager (let's here also only focus on a rectangular ROI). In fact some of those development should / could be related with computing statistics on ROI.

PR #3234, opened 2020-10-22
WIP: Add option to compute colormap VRange from a roi (plot limits or user roi)

payno:
INFO
====
This will:
- add an optional `_roiForAutoscale` variable to the ColormapMixIn. This is for now a RectanleROI. @vallsv suggest to use a 'simpler' ROI (origin + size) in order for the variable to be 'pure' data. And not an object. Would simplify the conception for future. In this case I guess RectangleROI should inherit / or contain from this data structure
- add a `_filter_data` function to filter item data according to the roi if it is set. For now only Scatter2D and ImageData will overwrite it.
- roi should be a RectangleROI

related to #2429 and https://gitlab.esrf.fr/tomotools/tomwer/-/issues/494

- I guess there is something to be done with `_filter_data` and taking a roi as a parameter to have a common treatment with statistics.
- setting the roi to the item should be check twice because I am not sure I didn't mess origin / scale management or 'reference'
- Scatter origin is also probably miss handled
TODO
=====
- [x] update `ColormapDialog` interface
- manage
    - [x] handle `ImageData`
    - [ ] handle `Scatter`
    - [x] plot limits (using an 'invisible roi'): name `from visible data`
    - [x] user defined roi (can be edited): name `from roi`
    - the current mode would be renamed from `all data`
    - [ ] display options 'visible data' and `roi` only if the item type is compatible (`ImageData` or `Scatter`)

- [ ] test
    - [x] `ImageData`
    - [ ] `Scatter`
- [ ] doc

- ~~[ ] instead of redefining a definition of `Rectangle` it would be better to add a dependency to a library defining those geometric primitive. Don't reinvent the wheel~~ If done will be part of another PR

Screenshot
=========

An idea of the possible update at the gui side

Screenshot from 2020-10-22 17-13-52

Screenshot from 2020-10-22 17-14-02

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -41,7 +41,6 @@ from silx.math import colormap as _colormap
 from silx.utils.exceptions import NotEditableError
 from silx.utils import deprecation
 from silx.resources import resource_filename as _resource_filename
-from silx.utils.enum import Enum as _Enum
 
 
 _logger = logging.getLogger(__file__)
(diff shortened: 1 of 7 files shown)


payno:
replace by https://github.com/silx-kit/silx/pull/3349

PR #3238, opened 2020-10-23
Add test for legend icon widget

vallsv:
Here is a few test coverage for LegendIconWidget.


silx/gui/widgets/test/__init__.py
```diff
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2020 European Synchrotron Radiation Facility
```
t20100, line 4: I think for copyright it is best to keep start-end, i.e., 2016-2020 here, but I will merge it anyway.

PR #3239, opened 2020-10-23
Use widget QFontMetrics

vallsv:
This PR reduce memory allocation by using already defined `fontMetrics` instread of creating a new one.


t20100:
LGTM

PR #3240, opened 2020-10-23
HDF5 strings: handle h5py 2.x and 3.x

### LINKED ISSUE #3218: CI: tests breaks with h5py 3.0rc1
Tests fails for both Linux and Windows with the following message with h5py3.0rc1:
```

ERROR: testArrayUnicode (silx.gui.data.test.test_textformatter.TestTextFormatterWithH5py)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/test/test_textformatter.py", line 170, in testArrayUnicode

    result = self.formatter.toString(d[()], dtype=d.dtype)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 299, in toString

    text = [self.toString(d, dtype) for d in data]

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 299, in <listcomp>

    text = [self.toString(d, dtype) for d in data]

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 302, in toString

    text = self.__formatH5pyObject(data, dtype)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 270, in __formatH5pyObject

    return self.__formatText(data)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 200, in __formatText

    text = "\"%s\"" % text.replace("\\", "\\\\").replace("\"", "\\\"")

TypeError: a bytes-like object is required, not 'str'

ERROR: testUnicode (silx.gui.data.test.test_textformatter.TestTextFormatterWithH5py)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/test/test_textformatter.py", line 133, in testUnicode

    result = self.formatter.toString(d[()], dtype=d.dtype)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 302, in toString

    text = self.__formatH5pyObject(data, dtype)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 270, in __formatH5pyObject

    return self.__formatText(data)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/gui/data/TextFormatter.py", line 200, in __formatText

    text = "\"%s\"" % text.replace("\\", "\\\\").replace("\"", "\\\"")

TypeError: a bytes-like object is required, not 'str'

ERROR: test_hdf5_array_slice_out_of_range (silx.io.test.test_utils.TestGetData)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/io/test/test_utils.py", line 570, in test_hdf5_array_slice_out_of_range

    self.assertRaises(ValueError, utils.get_data, url)

  File "/opt/python/3.8.1/lib/python3.8/unittest/case.py", line 816, in assertRaises

    return context.handle('assertRaises', args, kwargs)

  File "/opt/python/3.8.1/lib/python3.8/unittest/case.py", line 202, in handle

    callable_obj(*args, **kwargs)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/io/utils.py", line 816, in get_data

    data = data[data_slice]

  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper

  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/h5py/_hl/dataset.py", line 704, in __getitem__

    return self._fast_reader.read(args)

  File "h5py/_selector.pyx", line 344, in h5py._selector.Reader.read

  File "h5py/_selector.pyx", line 148, in h5py._selector.Selector.apply_args

IndexError: Index (5) out of range for (0-1)

FAIL: testTitle (silx.io.test.test_spectoh5.TestConvertSpecHDF5)

Test the value of a dataset

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/io/test/test_spectoh5.py", line 133, in testTitle

    self.assertEqual(title12,

AssertionError: b'aaaaaa' != 'aaaaaa'

FAIL: testSpectra (silx.io.test.test_nxdata.TestNXdata)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/io/test/test_nxdata.py", line 241, in testSpectra

    self.assertEqual(nxd.title, "Title as dataset (like nexpy)")

AssertionError: "b'Title as dataset (like nexpy)'" != 'Title as dataset (like nexpy)'

- b'Title as dataset (like nexpy)'

? --                             -

+ Title as dataset (like nexpy)

FAIL: testFile (silx.app.test.test_convert.TestConvertCommand)

  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/silx/app/test/test_convert.py", line 143, in testFile

    self.assertEqual(title12,

AssertionError: b'aaaaaa' != 'aaaaaa'
```

woutdenolf:
closes #3218 

More than just accommodating the H5T_STRING decoding changes in h5py 3.0.0rc1, this MR makes H5T_STRING decoding uniform for h5py>=2.8.0. Also the decoding behavior of datasets and attributes is made uniform.

`silx.io.test.test_utils.TestH5Strings` demonstrates the behavior.

Implementation
--------------------

Inspired by the `AsStrWrapper` of h5py 3.0 this MR introduces 2 utility classes and 3 utility methods:

  * `H5pyDatasetReadWrapper`:
     * wraps any dataset, regardless of the `dtype`
     * NON-ASCII H5T_STRING: decode
     * ASCII H5T_STRING: decode when requested
     * when decoding fails, return the bytes
  * `H5pyAttributesReadWrapper`:
     * wraps an AttributeManager
     * NON-ASCII H5T_STRING: decode
     * ASCII H5T_STRING: decode when requested
     * when decoding fails, return the bytes
  * `h5py_read_dataset`: use wrapper and read
  * `h5py_read_attribute`: use wrapper and read
  * `h5py_read_attributes`: use wrapper and read

Discussion
---------------

Whether ASCII H5T_STRING should be decoded by default or not, depends on what our priority is:

 1. **return the same `dtype` as written**: when the file is written by h5py, the only way we can obtain ASCII H5T_STRING is from writing `bytes`
    * Hence do not decode ASCII H5T_STRING (`decode_ascii=False`).
    * If non-h5py code creates an ASCII H5T_STRING with the intention of actually writing a string, `decode_ascii=False` is not appropriate.
 2. **strict HDF5 type**: H5T_STRING means a string so return `str`.
    * Hence decode ASCII H5T_STRING (`decode_ascii=True`).
    * If someone abuses H5T_STRING to save an array of bytes (h5py does this), `decode_ascii=True` is not appropriate.

In this MR I made choice 1 (decoding ASCII H5T_STRING is optional).

Other changes in this MR:
-----------------------------------
  * Fix the tests that failed for h5py 3.0.0rc1
  * Use wrappers in `h5todict`
  * `h5todict` always saves variable-length strings

Future: 
---------
We will probably have to use the wrappers in more places in silx. I only did it in `h5todict` and all the places that broke CI.

### DIFF
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -37,6 +37,8 @@ from .configdict import ConfigDict
 from .utils import is_group, is_link, is_softlink, is_externallink
 from .utils import is_file as is_h5_file_like
 from .utils import open as h5open
+from .utils import h5py_read_dataset
+from .utils import H5pyAttributesReadWrapper
 
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
@@ -44,34 +46,30 @@ __date__ = "17/07/2018"
 
 logger = logging.getLogger(__name__)
 
-string_types = (basestring,) if sys.version_info[0] == 2 else (str,)    # noqa
+try:
+    vlen_utf8 = h5py.special_dtype(vlen=unicode if sys.version_info[0] == 2 else str)
+    vlen_bytes = h5py.special_dtype(vlen=bytes)
+except AttributeError:
+    # h5py does not support variable-length types
+    vlen_utf8 = vlen_bytes = None
 
 
-def _prepare_hdf5_dataset(array_like):
+def _prepare_hdf5_write_value(array_like):
     """Cast a python object into a numpy array in a HDF5 friendly format.
 
     :param array_like: Input dataset in a type that can be digested by
         ``numpy.array()`` (`str`, `list`, `numpy.ndarray`…)
     :return: ``numpy.ndarray`` ready to be written as an HDF5 dataset
     """
-    # simple strings
-    if isinstance(array_like, string_types):
-        array_like = numpy.string_(array_like)
-
-    # Ensure our data is a numpy.ndarray
-    if not isinstance(array_like, (numpy.ndarray, numpy.string_)):
-        array = numpy.array(array_like)
-    else:
-        array = array_like
-
-    # handle list of strings or numpy array of strings
-    if not isinstance(array, numpy.string_):
-        data_kind = array.dtype.kind
-        # unicode: convert to byte strings
-        # (http://docs.h5py.org/en/latest/strings.html)
-        if data_kind.lower() in ["s", "u"]:
-            array = numpy.asarray(array, dtype=numpy.string_)
-
+    array = numpy.asarray(array_like)
+    if vlen_bytes:
+        # Fix-length string type to variable-length type
+        kind = array.dtype.kind.lower()
+        if kind in ["s", "u"]:
+            if kind == "s":
+                return numpy.array(array_like, dtype=vlen_bytes)
+            else:
+                return numpy.array(array_like, dtype=vlen_utf8)
     return array
 
 
@@ -246,14 +244,14 @@ def dicttoh5(treedict, h5file, h5path='/',
             elif is_link(value):
                 h5f[h5name] = value
             else:
-                ds = _prepare_hdf5_dataset(value)
+                data = _prepare_hdf5_write_value(value)
                 # can't apply filters on scalars (datasets with shape == () )
-                if ds.shape == () or create_dataset_args is None:
+                if data.shape == () or create_dataset_args is None:
                     h5f.create_dataset(h5name,
-                                       data=ds)
+                                       data=data)
                 else:
                     h5f.create_dataset(h5name,
-                                       data=ds,
+                                       data=data,
                                        **create_dataset_args)
 
         # deal with h5 attributes which have tuples as keys in treedict
@@ -280,14 +278,8 @@ def dicttoh5(treedict, h5file, h5path='/',
 
             # Write attribute
             value = treedict[key]
-
-            # Makes list/tuple of str being encoded as vlen unicode array
-            # Workaround for h5py<2.9.0 (e.g. debian 10).
-            if (isinstance(value, (list, tuple)) and
-                    numpy.asarray(value).dtype.type == numpy.unicode_):
-                value = numpy.array(value, dtype=h5py.special_dtype(vlen=str))
-
-            h5f[h5name].attrs[attr_name] = value
+            data = _prepare_hdf5_write_value(value)
+            h5f[h5name].attrs[attr_name] = data
 
 
 def nexus_to_h5_dict(treedict, parents=tuple()):
@@ -467,7 +459,8 @@ def h5todict(h5file,
 
         # Read the attributes of the group
         if include_attributes:
-            for aname, avalue in root.attrs.items():
+            attrs = H5pyAttributesReadWrapper(root.attrs)
+            for aname, avalue in attrs.items():
                 ddict[("", aname)] = avalue
         # Read the children of the group
         for key in root:
@@ -504,19 +497,20 @@ def h5todict(h5file,
             else:
                 # Child is an HDF5 dataset
                 try:
-                    data = h5obj[()]
+                    data = h5py_read_dataset(h5obj)
                 except OSError:
                     _handle_error(errors,
                                   OSError,
                                   'Cannot retrieve dataset "%s"',
-                                  subpath)
+                                  h5name)
                 else:
                     if asarray:  # Convert HDF5 dataset to numpy array
                         data = numpy.array(data, copy=False)
                     ddict[key] = data
                     # Read the attributes of the child
                     if include_attributes:
-                        for aname, avalue in h5obj.attrs.items():
+                        attrs = H5pyAttributesReadWrapper(h5obj.attrs)
+                        for aname, avalue in attrs.items():
                             ddict[(key, aname)] = avalue
     return ddict
 
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -46,12 +46,8 @@ __date__ = "17/07/2018"
 
 logger = logging.getLogger(__name__)
 
-try:
-    vlen_utf8 = h5py.special_dtype(vlen=unicode if sys.version_info[0] == 2 else str)
-    vlen_bytes = h5py.special_dtype(vlen=bytes)
-except AttributeError:
-    # h5py does not support variable-length types
-    vlen_utf8 = vlen_bytes = None
+vlen_utf8 = h5py.special_dtype(vlen=str)
+vlen_bytes = h5py.special_dtype(vlen=bytes)
 
 
 def _prepare_hdf5_write_value(array_like):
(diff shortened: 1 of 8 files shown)


silx/io/dictdump.py
```diff
@@ -37,41 +37,39 @@
...
 
 __authors__ = ["P. Knobel"]
 __license__ = "MIT"
 __date__ = "17/07/2018"
 
 logger = logging.getLogger(__name__)
 
-string_types = (basestring,) if sys.version_info[0] == 2 else (str,)    # noqa
+try:
+    vlen_utf8 = h5py.special_dtype(vlen=unicode if sys.version_info[0] == 2 else str)
```
t20100, line 50: We've dropped python2, so no need for `sys.version_info[0] == 2` any longer

silx/io/dictdump.py
```diff
@@ -37,41 +37,39 @@
...
+    vlen_bytes = h5py.special_dtype(vlen=bytes)
+except AttributeError:
+    # h5py does not support variable-length types
+    vlen_utf8 = vlen_bytes = None
```
t20100, line 54: vlen seems supported since some time.
What about setting a minimum version of `h5py`? I opened an issue for it: https://github.com/silx-kit/silx/issues/3246

silx/io/dictdump.py
```diff
@@ -37,41 +37,39 @@
...
-
+    array = numpy.asarray(array_like)
+    if vlen_bytes:
+        # Fix-length string type to variable-length type
+        kind = array.dtype.kind.lower()
+        if kind in ["s", "u"]:
+            if kind == "s":
+                return numpy.array(array_like, dtype=vlen_bytes)
+            else:
+                return numpy.array(array_like, dtype=vlen_utf8)
```
t20100, line 72: That's a good cleaning-up!

silx/io/test/test_utils.py
```diff
@@ -716,6 +718,153 @@ def test_conflicts(self):
...
+        assert type(data) == type(result), data
+        assert data == result, data
+
+        self.file.attrs["vlen_data"] = self._make_array(value, 2)
+        data = utils.h5py_read_attribute(self.file.attrs, "vlen_data", decode_ascii=decode_ascii)
+        assert type(data[0]) == type(result), data[0]
+        assert data[0] == result, data[0]
+        numpy.testing.assert_array_equal(data, [result]*2)
+
+    def test_dataset_ascii_bytes(self):
```
t20100, line 805: You could also use [`TestCase.subTest`](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests).

silx/io/utils.py
```diff
@@ -957,3 +959,177 @@ def info_file_to_dict():
...
+        else:
+            if encoding != "ascii":
+                # Decoding is already done by the h5py library
+                return value
+            if encoding == "ascii" and not self._decode_ascii:
+                return value
+            if encoding == "ascii":
+                # ASCII can be decoded as UTF-8
+                encoding = "utf-8"
+            return h5py_decode_value(value, encoding=encoding)
```
t20100, line 1107: Can't this be simplified:
```python
if encoding == "ascii" and self._decode_ascii:
    return h5py_decode_value(value, encoding=encoding)
else:
    return value
```

silx/io/dictdump.py
t20100: BTW, this can be cleaned-up with #3246.

woutdenolf: Hmmm I wonder why I did that. I tried down to 2.8 but 2.8 already has vlen. I'll remove it.

silx/io/test/test_utils.py
woutdenolf: I could but why? I like the verbosity.

silx/io/utils.py
```diff
@@ -957,3 +959,177 @@ def info_file_to_dict():
...
+
+    :param h5py.Dataset dset:
+    :param index: slicing (all by default)
+    :param bool decode_ascii:
+    """
+    return H5pyDatasetReadWrapper(dset, decode_ascii=decode_ascii)[index]
+
+
+def h5py_read_attribute(attrs, name, decode_ascii=False):
+    """Read data from dataset object. UTF-8 strings will be
```
t20100, line 1127: Typo: `Read data from dataset object.` -> `Read data from attributes.`

t20100 (approved):
I left a few comments, but it overall LGTM, thaks for having taken care of this!!

t20100:
Without much thinking, I would have put `decode_ascii=True` by default to have a "decode early" behavior, but it's fine as long as one can choose.

I guess this is the first PR of a long list of fixes on this subject....

silx/io/test/test_utils.py
t20100: No need to change it , but I find it simpler to read a `with self.subTest` block than many methods.

silx/io/utils.py
woutdenolf: Thanks! Done.

silx/io/utils.py
woutdenolf: Thanks! Done.

silx/io/dictdump.py
woutdenolf: Removed it.

silx/io/dictdump.py
woutdenolf: Dropped python 2 support here

t20100:
Great! I'll merge it tomorrow.

woutdenolf:
> Without much thinking, I would have put `decode_ascii=True` by default to have a "decode early" behavior, but it's fine as long as one can choose.

Yes that's option 2 in the discussion. Neither option is satisfactory in all cases. Option 1 is more consistent when writing and reading is done in python (you get the same type you saved). Option 2 ignores the abuse of H5T_STRING by h5py to store `bytes` and decodes all H5T_STRING types.

t20100:
OK, if option 1 is better for round-tripping, let's keep it as default.

woutdenolf:
Sorry I was thinking about my use cases and I need another utility method: h5py_read_attributes (all attributes).

Issue #3244, opened 2020-10-28
QThreadPool.tryTake is not available in PyQt5=5.7 (used in debian9)


cpascual:
Using the [debian9 backported package](http://www.silx.org/pub/debian/dists/binary/python3-silx_0.13.1-1~bpo9+1_amd64.deb) we ran into the following problem:

```
https://github.com/silx-kit/silx/issuesTraceback (most recent call last):  
File "/usr/lib/python3/dist-packages/silx/gui/plot/tools/profile/rois.py", line 237, in __regionChanged    
    self.invalidateProfile()  
File "/usr/lib/python3/dist-packages/silx/gui/plot/tools/profile/core.py", line 102, in invalidateProfile    
    profileManager.requestUpdateProfile(self)  
File "/usr/lib/python3/dist-packages/silx/gui/plot/tools/profile/manager.py", line 818, in requestUpdateProfile    
   if threadPool.tryTake(runner):
AttributeError: 'QThreadPool' object has no attribute 'tryTake'
Aborted (core dumped)
```

According to the Qt docs, [`QThreadPool.tryTake` was introduced in Qt=5.9](https://doc.qt.io/qt-5/qthreadpool.html#tryTake), but in debian9 it is 5.7

AFAIKT (according to the requirements in setup.py), PyQt=5.7 is supported by silx, and therefore the usage of `tryTake` should be replaced or at least protected. 
Otherwise, the requirements should specify  PyQt5>=5.9 **and some patch would be needed to fix the debian9 package**

vallsv:
Hi.

Surprisingly I was working yesterday on a worker which could replace that code.

I see a way to replace that call by something else, as a work around.

Let us know if you expect a bug fix for the 0.13?

cpascual:
> Let us know if you expect a bug fix for the 0.13?

Of course it would be nice, but I do not think that would be critical for us. We can develop and test using conda until it is ready. What isimportant for us is that at some point we can use the profile feature in a debian9

cpascual:
Oh, and just re-reading my previous posts in this issue, I realized that I sound quite rude... 
It was not my intention!
It was just that I was writing in a hurry while multitasking. 
So, let me thank now @vallsv for the prompt reply and also let you know that if there is something I can do (test or work on a PR, just say so)

vallsv:
No worry. Sounds like a normal bug report to me :-)

t20100:
closed by PR #3250

PR #3250, opened 2020-11-04
Workaround to use only Qt5.7 API

vallsv:
Here is a workaround for #3244. This remove the use of `tryTake` in a lazy way.

If we want to create a bug fix release, I suggest to only apply that in the branch, and update the supported Qt version on the master branch.

I still have to setup a Qt5.7 env to check it manually.

Changelog: 

- Fix use of `QThreadPool.tryTake` to be Qt5.7 complient


vallsv:
I did check it. Sounds good.

t20100 (approved):
I added some comments to flag as Qt<5.9 support.
LGTM

vallsv:
Yes good idea.

PR #3255, opened 2020-11-09
silx.gui.plot.PlotWidget: Added `PlotWidget.[get|set]Backend` enabling switching backend

### LINKED ISSUE #3159: Provide a hot switch of backend
It would be nice to provide a hot switch of the backend during the execution.

So what OpenGL or matplotlib can be used when it's needed.

t20100:
This PR allows to switch the backend of a `PlotWidget` while using it.
It adds 2 methods: `setBackend` and `getBackend` to provide this feature.

It was a lot simpler to do than expected (and than when I first tried to do it a while ago).

It should already provide "full" support of state switching.
From what I quickly tested it looks to work fine.

There is an issue with widgets messing up with `PlotWidget`'s central widget (e.g., adding the colorbar there), that needs be to fixed somehow.
It also requires adding a few automated tests, but it is already ready for early testers.

closes #3159

### DIFF
--- a/silx/gui/plot/ColorBar.py
+++ b/silx/gui/plot/ColorBar.py
@@ -142,11 +142,9 @@ class ColorBarWidget(qt.QWidget):
             self._isConnected = True
 
     def setVisible(self, isVisible):
-        # isHidden looks to be always synchronized, while isVisible is not
-        wasHidden = self.isHidden()
-        qt.QWidget.setVisible(self, isVisible)
-        if wasHidden != self.isHidden():
-            self.sigVisibleChanged.emit(not self.isHidden())
+        if isVisible != self.isVisible():
+            qt.QWidget.setVisible(self, isVisible)
+            self.sigVisibleChanged.emit(isVisible)
 
     def showEvent(self, event):
         self._connectPlot()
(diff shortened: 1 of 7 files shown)


t20100:
- Added basic tests.
- Fixed issues with `PlotWindow` and `ImageView` which overrides `PlotWidget` central widget.
   Reusing the same central widget and replacing the plot backend in it was failing (there was some weird glitches when the colorbar was hidden during the change of backend and shown back).
   This is why for both `PlotWindow` and `ImageView`, the central widget is recreated for each backend switch. If someone 

It took actually longer to fix those issues than to implement the backend switch in `PlotWidget`....

I made a small change in `silx/gui/plot/ColorBar.py` (see commit e1b342f ), because the colorbar action was no longer synchronized when changing the backend. Previous implementation was introduced in #1627 to fix #1592. From what I checked #1592 is still fixed with this change.

t20100:
I still need to make `setBackend` resilient to rubbish.

t20100:
Done, `setBackend('rubbish')` or any other issue creating the backend (e.g., with OpenGL) will not break the currently running backend.

t20100:
BTW, that's a very long pending feature now available.

vallsv:
Cause we are smarter than before

Issue #3200, opened 2020-09-25
Plot: Micro glitch in ArcRoi rendering


vallsv:
When displaying a donut, the start and the stop of the inner circle looks to be not connected.

![Screenshot from 2020-09-25 12-02-31](https://user-images.githubusercontent.com/7579321/94254682-6a45fb00-ff27-11ea-8073-0e4f9792abd7.png)

PR #3257, opened 2020-11-09
Fix arc roi shape

vallsv:
Use `linspace` to make sure start end end points are part of the shape


vallsv:
closes #3200

vallsv:
Oups... i have merged that by mistake.

t20100:
LGTM ;)

PR #3260, opened 2020-11-10
Polar arc roi

vallsv:
This MR provide multi interaction mode for ROIs, and especially for arc ROI.

The arc ROI have to provide a way to setup the cercle with 3 points and with polar values.
That's useful to match different use cases (when the center is visible or not, for instance)

To manage that this requires too much handles, which make the result unusable.

So this PR provides a way to switch from an interaction mode to another.

As result the arc ROI provides for now
- `curvature` mode, 3-points to define a circle
- `polar`, which can move start and stop angles + the radius  
- `move` which only provides handle to move the shape (in case the center is not visible), (which could be removed one day when we will be able to dnd the shape itself)

Changelog: 
- Added multi interaction mode for ROIs (can be switched with a single click on an handle, or the context menu)
- Added polar interaction mode for arc ROI
- Fixes duplicated callback on ROIs (there was one for each ROI managed created on the plot)
- Added `PlotWidget.sigDefaultContextMenu` to allow to feed the default context menu
- Added context menu to the selected ROI to remove it

### DIFF
--- a/examples/plotInteractiveImageROI.py
+++ b/examples/plotInteractiveImageROI.py
@@ -38,6 +38,7 @@ from silx.gui import qt
 from silx.gui.plot import Plot2D
 from silx.gui.plot.tools.roi import RegionOfInterestManager
 from silx.gui.plot.tools.roi import RegionOfInterestTableWidget
+from silx.gui.plot.tools.roi import RoiModeSelectorAction
 from silx.gui.plot.items.roi import RectangleROI
 from silx.gui.plot.items import LineMixIn, SymbolMixIn
 from silx.gui.plot.actions import control as control_actions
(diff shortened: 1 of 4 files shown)


t20100:
It is good to have multiple interaction modes.

A few comments:
- I don't see the need for the drag only mode since the other modes provides dragging?
- Changing mode by clicking on the anchors is great!
- For the change with the `Ctrl` key, I was expecting the more usual behavior of enabling temporarily an alternative mode while `Ctrl` is pressed rather than switch the mode on `Ctrl` click.

With `examples/plotInteractiveImageROI.py` when creating an arc ROI, a diamond marker is created at the origin of the plot and doesn't seems to be used.

vallsv:
> I don't see the need for the drag only mode since the other modes provides dragging?

There is too much anchors, it is not really manageable. So if you have the center outside of your image you will not be able to move that shape. That is why i have created a mode here. The day we can drag and drop the shape by clicking on it, we could remove it. It's a workaround.

> Changing mode by clicking on the anchors is great!

I am afraid it is not very nice, if you have a better idea you are welcome. Same for Ctrl.
Hmm, what about an action on the active ROI? Which could be displayed by the ROI manager? In this case the ROI have to emit another event. But it could be a little more exposed.

Here it is open. We can remove the mode switch from this MR. While the mode is part of the ROI i am fine, Cause i already can use it statically by inheritance on my application (2 different kind of ROIs for examples).

t20100:
> The day we can drag and drop the shape by clicking on it, we could remove it.

Yes that would be better! I don't see blocking things but it is a bit short for this release.

For the modes, it could also be in a context menu on the shape... But I don't think we have the machinery to support this.. not even mentionning overriding the context menu.

I would be for keeping the mode change when clicking on the anchors, but remove the `Ctrl` way (or change it to a temporary change e.g., to the next mode).

vallsv:
Then i will remove the CTRL. It was designed to replace the mouse click. And i will try to add the magic in an action.

vallsv:
I was also thinking of move arcroi out of the roi module.
It's about 1000 line of code, And the ROI module is about 3000 lines of code now.
What do you think?

vallsv:
> For the modes, it could also be in a context menu on the shape... But I don't think we have the machinery to support this.. not even mentionning overriding the context menu.

Oh... yes we have. That's a stuff i have added few time ago to be able to close the polygon with the context menu... Which also make us in trouble with the mouse click :-D I have to make a try.

vallsv:
The context menu was not propagated so i did it.

It is  displayed if the ROI is editable and if the cursor is inside the shape.

And as it was useful and easy, i have added an action to remove a ROI. It could be part of another PR.

I think while we are here we could propagate the context menu to the ROI object. This makes the API even more powerful.

t20100:
I made `PlotWidget.sigDefaultContextMenu` private.

LGTM

PR #3261, opened 2020-11-10
Plot action to switch on/off OpenGL

vallsv:
Following #3255, this PR provides a plot action which can be used with a silx plot.

It have to be setup manually.

It only shows if OpenGL is enabled or not. I guess it is already complex for most people.

That's more or less the way i will use it in Flint.

It can be used with this example.

```
./bootstrap.py examples/plotInteractiveImageROI.py
```

Screenshot from 2020-11-10 15-00-38

Changelog: 
- Added an action to switch on/off OpenGL rendering on a plot


silx/gui/plot/actions/control.py
```diff
@@ -632,3 +632,73 @@ def _modeChanged(self, source):
 
     def _actionTriggered(self, checked=False):
         self.plot._eventHandler.validate()
+
+
+class OpenGlAction(PlotAction):
```
t20100, line 637: Best renamed to `OpenGLAction` to follow naming convention.

silx/gui/plot/actions/control.py
```diff
@@ -632,3 +632,73 @@ def _modeChanged(self, source):
...
+                        "Custom rendering")
+        }
+
+        name = self._getBackendName(plot)
+        self.__state = name
+        icon, tooltip = self._states[name]
+        super(OpenGlAction, self).__init__(
+            plot,
+            icon=icon,
+            text='Toggle keep aspect ratio',
```
t20100, line 663: `text` was not updated after copy-paste I guess.

silx/gui/plot/actions/control.py
```diff
@@ -632,3 +632,73 @@ def _modeChanged(self, source):
...
+        name = self._getBackendName(self.plot)
+        if self.__state != name:
+            # THere is no event to know the backend was updated
+            # So here we check if there is a mismatch between the displayed state
+            # and the real state of the widget
+            self._backendUpdated()
+            return
+        if name is not "opengl":
+            from silx.gui.utils import glutils
+            result = glutils.isOpenGLAvailable()
```
t20100, line 697: I would make the test at action creation, but I guess you want to keep it fast.
Once it failed once, you can disable the action, it will never become available.

t20100 (approved):
I left a few minor comments, else LGTM

silx/gui/plot/actions/control.py
vallsv: Yes i don't want crazy tests at creation.

I don't know about disabling anyway. I prefer not to do so.

If you click on it again you have an obvious error message, you don't need to make it safer.
If you are able to install `pyopengl` in your env, the problem could be fixed without restarting the application.

silx/gui/plot/actions/control.py
vallsv: This one is fixed

vallsv:
I have reworked the PR in order fix the text and the name.

I have also removed the disabled icon, and make it checkable. This allow to use it as it is with a menu.

silx/gui/plot/actions/control.py
t20100: It would work for install `pyopengl`, but for any other error, it would not since `isOpenGLAvailable` caches the runtime check result.

vallsv: That's right, we could add a parameter to skip the cache here.

t20100: Well, there is no reason the result of the runtime check would change...
`pyopengl` availability, `$DISPLAY` and Qt reporting an issue are checked each time, only the runtime check which is run after those first checks is cached.

vallsv: Driver update, or lib update

t20100: without restarting the app, things will already be loaded, no?

Issue #3252, opened 2020-11-06
Tests broken on debian9 (h5py v2.7.0)


t20100:
With current master, there are quite a few errors in the test when building debian9 packages:
```
ERROR: testAttributeAlreadyExists (silx.io.test.test_dictdump.TestDictToH5)
-> RuntimeError: Expected 1 WARNING (got 0)

ERROR: testAttributeValues (silx.io.test.test_dictdump.TestDictToH5)
ERROR: testAttributes (silx.io.test.test_dictdump.TestDictToH5)
ERROR: testH5CityAttrs (silx.io.test.test_dictdump.TestDictToH5)
ERROR: testKeyOrder (silx.io.test.test_dictdump.TestDictToH5)
ERROR: testPathAttributes (silx.io.test.test_dictdump.TestDictToH5)
ERROR: testAttributeValues (silx.io.test.test_dictdump.TestDictToNx)
ERROR: testAttributes (silx.io.test.test_dictdump.TestDictToNx)
ERROR: testKeyOrder (silx.io.test.test_dictdump.TestDictToNx)
ERROR: testAsArrayFalse (silx.io.test.test_dictdump.TestH5ToDict)
ERROR: testAsArrayTrue (silx.io.test.test_dictdump.TestH5ToDict)
ERROR: testDereferenceLinks (silx.io.test.test_dictdump.TestH5ToDict)
ERROR: testExcludeNames (silx.io.test.test_dictdump.TestH5ToDict)
ERROR: testPreserveLinks (silx.io.test.test_dictdump.TestH5ToDict)
ERROR: testStrings (silx.io.test.test_dictdump.TestH5ToDict)
ERROR: testAttributes (silx.io.test.test_dictdump.TestNxToDict)
-> TypeError: No conversion path for dtype: dtype('<U[...]')
```

See full log with all errors: https://gitlab.esrf.fr/silx/bob/silx/-/jobs/156681

Question: Do we still support debian9?
My opinion is yes at least for v0.14, and warn that we will drop debian9 and python3.5 in v0.15.

woutdenolf:
I cannot reproduce the issue on Debian9 with h5py 2.7

This issue was actually solved by #3251

PR #3271, opened 2020-11-18
Drop support for h5py<2.10

woutdenolf:


t20100:
By dropping h5py<2.10 we also drop debian10 (h5py 2.8)...

If having h5py<2.8 doesn't change anything, maybe it is better? Else debian10 will stay with v0.13 max or a backport of h5py (which is fine for me).

woutdenolf:
In that case I think it is better to still support 2.7

t20100:
Maybe not, debian10 comes with h5py v2.8 and matplotlib v3.0.2... so currently released version of silx is compatible.

woutdenolf:
Hmmm weird, I have not problems with h5py 2.7 on my Ubuntu 20.04 machine.

woutdenolf:
This is no longer needed. See #3252

Issue #3243, opened 2020-10-27
backprojection not usable


vallsv:
Hi,

It looks like on my platform/device IMAGES are not supported.

Any idea if it could be known before creating anything?

```
Traceback (most recent call last):
  File "/users/valls/workspace/bliss.git/bliss/flint/widgets/slice_reconstruction_widget.py", line 39, in __selectionChanged
    self.setItem(item)
  File "/users/valls/workspace/bliss.git/bliss/flint/widgets/slice_reconstruction_widget.py", line 64, in setItem
    sino_shape=sino.shape, angles=angles, devicetype="all"
  File "/users/valls/Software/miniconda3/envs/blissenv/lib/python3.7/site-packages/silx/opencl/backprojection.py", line 110, in __init__
    self._allocate_memory()
  File "/users/valls/Software/miniconda3/envs/blissenv/lib/python3.7/site-packages/silx/opencl/backprojection.py", line 184, in _allocate_memory
    self._allocate_textures()
  File "/users/valls/Software/miniconda3/envs/blissenv/lib/python3.7/site-packages/silx/opencl/backprojection.py", line 252, in _allocate_textures
    hostbuf=np.zeros(self.shape[::-1], dtype=np.float32)
  File "/users/valls/Software/miniconda3/envs/blissenv/lib/python3.7/site-packages/pyopencl/__init__.py", line 987, in image_init
    image_old_init(self, context, flags, format, desc, hostbuf)
pyopencl._cl.RuntimeError: clCreateImage failed: IMAGE_FORMAT_NOT_SUPPORTED
```

pierrepaleo:
What are the platform/devices ?

This error occurred when the OpenCL backprojection kernel was compiled for CPU which does not support "textures" (`Image`). There is a dedicated kernel (`backproj_cpu_kernel`) for it, but it does not seem to be dispatched correctly here.

vallsv:
I finally found something which is working.

Here is the one which was failing. I can try to reach more informations if you like.
```
In [2]: import pyopencl                                                                                                                                                                                                      

In [4]: pyopencl.get_platforms()                                                                                                                                                                                             
Out[4]: [<pyopencl.Platform 'Intel Gen OCL Driver' at 0x7fcdc31f4400>]

In [6]: p = pyopencl.get_platforms()[0]                                                                                                                                                                                      

In [7]: p.get_devices()                                                                                                                                                                                                      
[...]

PR #3273, opened 2020-11-19
[OpenCL] Check textures availability

### LINKED ISSUE #2700: [OpenCL] Convolution: better mechanism for testing texture availability
The `Convolution` class tries to use textures when instantiated on GPU (unless explicitly told otherwise).
However, it might be that textures are not supported on GPU (ex. with the `pocl` implementation).   
A new way to determine whether textures are available has to be implemented.

pierrepaleo:
This PR adds a mechanism for checking textures (openCL "images") availability.

Depending on the underlying openCL implementation, we can never be sure that textures are actually available, before explicitly trying to create one.  
For example, with `pocl`:  `backprojector.ctx.devices[0].get_info(cl.device_info.IMAGE_SUPPORT)`  returns `1` while `backprojector.check_texture_availability()` returns `False`.

Close #2700, #3243

Changelog: OpenCL: add textures availability check

### DIFF
--- a/silx/opencl/backprojection.py
+++ b/silx/opencl/backprojection.py
@@ -164,9 +164,7 @@ class Backprojection(OpenclProcessing):
     def _allocate_memory(self):
         # Host memory
         self.slice = np.zeros(self.dimrec_shape, dtype=np.float32)
-        self.is_cpu = False
-        if self.device.type == "CPU":
-            self.is_cpu = True
+        self._use_textures = self.check_textures_availability()
 
         # Device memory
         self.buffers = [
@@ -180,7 +178,7 @@ class Backprojection(OpenclProcessing):
         self.d_sino = self.cl_mem["d_sino"]  # shorthand
 
         # Texture memory (if relevant)
-        if not(self.is_cpu):
+        if not(self._use_textures):
             self._allocate_textures()
 
         # Local memory
@@ -209,7 +207,7 @@ class Backprojection(OpenclProcessing):
             _idivup(int(self.dimrec_shape[0]), 32) * self.wg[1]
         )
         # Prepare arguments for the kernel call
-        if self.is_cpu:
+        if self._use_textures:
             d_sino_ref = self.d_sino.data
         else:
             d_sino_ref = self.d_sino_tex
@@ -281,7 +279,7 @@ class Backprojection(OpenclProcessing):
         sino2 = sino
         if not(sino.flags["C_CONTIGUOUS"] and sino.dtype == np.float32):
             sino2 = np.ascontiguousarray(sino, dtype=np.float32)
-        if self.is_cpu:
+        if self._use_textures:
             ev = pyopencl.enqueue_copy(
                                         self.queue,
                                         self.d_sino.data,
@@ -301,7 +299,7 @@ class Backprojection(OpenclProcessing):
         return EventDescription(what, ev)
 
     def _transfer_device_to_texture(self, d_sino):
-        if self.is_cpu:
+        if self._use_textures:
             if id(self.d_sino) == id(d_sino):
                 return
             ev = pyopencl.enqueue_copy(
@@ -335,7 +333,7 @@ class Backprojection(OpenclProcessing):
         with self.sem:
             events.append(self._transfer_to_texture(sino))
             # Call the backprojection kernel
-            if self.is_cpu:
+            if self._use_textures:
                 kernel_to_call = self.kernels.backproj_cpu_kernel
             else:
                 kernel_to_call = self.kernels.backproj_kernel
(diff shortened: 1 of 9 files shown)


pierrepaleo:
Tested on the following machines.

## Machine `A`

Hardware: GeForce GTX 580, Quadro P400, Xeon Silver 4114 CPU  
Software: Debian 10, python 3.7

  - [x] Convolution: OK on all platforms. With CPU+POCL, we have warning about clean-up fails (dead contexts ?).
  - [x] Backprojection: OK on all platforms
  - [x] Projection: OK on all platforms

  ## Machine `B`

Hardware: Quadro P2000, Xeon E5-2643-0 CPU
Software: Ubuntu 20.04, python 3.8

  - [ ] Convolution: TODO
  - [ ] Backprojection: TODO
  - [ ] Projection: TODO


##  Machine `C`

Hardware: GTX 1080 Ti, Xeon CPU E5-2643 v3 
Software: Debian 8.8, python 3.7
OpenCL platforms:
```
[0] <pyopencl.Platform 'NVIDIA CUDA'>
[1] <pyopencl.Platform 'Intel(R) OpenCL'>
[2] <pyopencl.Platform 'AMD Accelerated Parallel Processing'>
```

  - [ ] Convolution: **FAIL**
  - [ ] Backprojection: **TODO** the first test fails on platforms 1 and 2
  - [x] Projection: OK on all platforms


## Machine `D`

Hardware: Tesla V100, Power9  
Software: Ubuntu 20.04, python 3.8

  - [ ] Convolution: TODO
  - [ ] Backprojection: TODO
  - [ ] Projection: TODO

pierrepaleo:
Funny thing: running
```bash
./run_tests.py silx.opencl.test.test_convolution.suite
```
works, but running
```bash
./run_tests.py silx.opencl.test.test_convolution
```
fails, because `unittest` does not seem to be able to pass the `param` argument at class initialization.

t20100:
You can't provide a module as argument, it needs to be a `TestSuite`, a `TestCase` or a `TestCase` method.

pierrepaleo:
Understood, thanks !

kif:
Do you actually win anything by using textures ? how much (beside the approximate interpolator)?

pierrepaleo:
For convolution, on a GTX 1080 Ti:

```python
import numpy as np
from silx.opencl.convolution import Convolution
import pyopencl.array as parray
from scipy.misc import ascent
from silx.image.utils import gaussian_kernel
from time import sleep

img = ascent().astype("f")
img2 = np.tile(img, (8,8))
kern = gaussian_kernel(3)
C = Convolution((4096, 4096), kern, axes=(0, 1), extra_options={"dont_use_textures": False}, profile=True)
C0 = Convolution((4096, 4096), kern, axes=(0, 1), extra_options={"dont_use_textures": True}, profile=True)

d_img = parray.to_device(C.queue, img2)
d_o = parray.zeros_like(d_img)

def print_event(proc, i):
    evt = proc.events[i]
    t_ms = (evt.event.profile.end - evt.event.profile.start)/1e6
    print("%s: %.3f" % (evt.name, t_ms))

C(d_img, output=d_o)
sleep(1)
C(d_img, output=d_o)
sleep(1)
print_event(C, -1)
print_event(C, -2)
print_event(C, -3)

C0(d_img, output=d_o)
sleep(1)
C0(d_img, output=d_o)
sleep(1)
print_event(C0, -1)
print_event(C0, -2)
```

gives (timings in ms):

```
# textures
batched convolution: 2.637
Transfer to texture: 0.442
batched convolution: 2.661
# no textures
batched convolution: 10.590
batched convolution: 10.989
```
and this is for convolution, where no interpolation is done, only a 1D stencil pattern.  
A bit more work is to be done for comparing things with `Backprojector` (as it always use texture whenever available) but since the access pattern is a lot more complicated, I suspect the gain will be significant.

pierrepaleo:
Ready for review. 

Some items worth mentioning:
  - With POCL: `backprojector.ctx.devices[0].get_info(cl.device_info.IMAGE_SUPPORT)` returns `1` while `backprojector.check_texture_availability()` returns `False`  (this is one reason for using `check_texture_availability` in the first place).
  - Replaced `self.is_cpu` with `self._use_textures`. The two are not equivalent:
    - Textures might be available on CPU, depending on the OpenCL implementation, although certainly emulated. When using them, precision is lowered (so we still have to check whether the device is a CPU for the tests)
    - Textures might be not available on GPUs depending on the OpenCL implementation (ex. POCL)
  - With POCL + nvidia hardware: compilation fails when compiling a kernel with "image2d_" functions, so I added some `#ifndef` guards

kif:
I am reviewing it ...

kif:
is this branch on the silx-kit repo?

kif:
Looks good to me. I'll destroy the branch once merged.

PR #3279, opened 2020-11-24
silx.gui.plot.Stackview: Replaced `setColormap` `autoscale` argument by `scaleColormapRangeToStack` method

### LINKED ISSUE #2835: [plot] Rework StackView colormap

t20100:
This PR deprecates the `autoscale` argument of `StackView.setColormap` which triggers the scaling of the colormap to the whole stack when `setColormap` or `setStack` is called, and so it is NOT use the whole stack when in autoscale mode as on would expect.
A `scaleColormapRangeToStack` method is added to which sets the range of the colormap to that of the stack.
This does not prevent adding a true autoscale to stack later.

If someone has a better idea to name the `scaleColormapRangeToStack` method, I'd take it.

To me it closes #2835

### DIFF
--- a/silx/gui/plot/StackView.py
+++ b/silx/gui/plot/StackView.py
@@ -778,7 +778,7 @@ class StackView(qt.QMainWindow):
         # specifying a special colormap
         return self._plot.getDefaultColormap()
 
-    def scaleColormapRangeFromStack(self):
+    def scaleColormapRangeToStack(self):
         """Scale colormap range according to current stack data.
 
         If no stack has been set through :meth:`setStack`, this has no effect.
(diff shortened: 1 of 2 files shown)


vallsv:
I think it's fine

payno:
+1

Issue #3258, opened 2020-11-10
GPU Memory Leak


weichengkai:
Hi,
After running the following code,GPU memory leak.
silx version=0.13.2

```
import numpy
import silx.opencl.sift as sift

def siftTest(image):
    gpuSiftPlan = sift.SiftPlan(template=image, devicetype="GPU")
    gpuSiftPlan.keypoints(image)

image = numpy.zeros((100, 100, 3), dtype=numpy.uint8)
for i in range(500000):
    siftTest(image)
    print(i)

```

pierrepaleo:
Hi @weichengkai 

The purpose of a `SiftPlan` is to be created once for a given image shape, and then to be used multiple times over different images. It is similar to the "FFTW plan" concept. Of course creating it each time will consume a lot of memory.

So instead, I would rather use:

```python
import numpy
import silx.opencl.sift as sift

image = numpy.zeros((100, 100, 3), dtype=numpy.uint8)
gpuSiftPlan = sift.SiftPlan(template=image, devicetype="GPU")
for i in range(500000):
    gpuSiftPlan.keypoints(image)
    print(i)
```

weichengkai:
**image = numpy.zeros((100, 100, 3), dtype=numpy.uint8)** just for testing
Actually, Images come from local files and come in different image shape.
Function **siftTest** get a image and return image's keypoint.Like following code
```
def siftTest(image):
    gpuSiftPlan = sift.SiftPlan(template=image, devicetype="GPU")
    kps = gpuSiftPlan.keypoints(image)
    return kps
```
When I called the function **siftTest** many times, I found that GPU memory could not be released.
What can I do to free memory before I return kps?

kif:
If your image size differs, you will have to destroy the plan each time ...

weichengkai:
The plan instantiated within function **siftTest**，after the function runs, the plan is automatically destroyed.  
The problem now is that GPU memory is not released after the plan is destroyed.

kif:
pyopencl keeps the context in caches but it should recycle them.
We could destroy the CommandQueue but it is not enough.

payno:
Let's consider this one is solved from the warning emission

PR #3280, opened 2020-11-24
Emit a warning when there is an issue in the Ocl destruction

kif:
Does not address #3258.

I suspect the bug is in pyopencl as our caches should be clean.

Changelog: report error if encountered in destructor.


pierrepaleo:
Looks good

Issue #3124, opened 2020-06-12
Combining mask and histogram in widgets/silx view


t20100:
I had a request for it.
Do we want to allow to combine the histogram plot action with the mask tool, from widgets? and from `silx view`?

Below is a sample code that works for images by overriding `PixelIntensitiesHistoAction.computeIntensityDistribution` to add :
```
        source = self.getSelectionMaskSource()
        if source is not None:
            mask = source.getSelectionMask()
            if mask is not None:
                array = numpy.array(array, dtype=numpy.float64, copy=True)
                array[mask != 0] = numpy.nan  # Replace masked values with NaN
```

Sample code:
```
import numpy

import fabio

from silx.gui import qt
from silx.gui.plot import Plot2D, items
from silx.math.combo import min_max
from silx.math import Histogramnd
from silx.gui.plot.actions import histogram as histogram_action


class HistoMaskAction(histogram_action.PixelIntensitiesHistoAction):
    """Histogram action with support for a mask to compute the histogram"""

    def __init__(self, *args, **kwargs):
        self.__selectionMaskSource = None
        super().__init__(*args, **kwargs)

    def setSelectionMaskSource(self, source):
        """Set the source of the selection mask.

        :param source:
            QObject with sigMaskChanged signal and getSelectionMask method.
        """
        if self.__selectionMaskSource is not None:
            self.__selectionMaskSource.sigMaskChanged.disconnect(
                self.computeIntensityDistribution)
        self.__selectionMaskSource = source
        if self.__selectionMaskSource is not None:
            self.__selectionMaskSource.sigMaskChanged.connect(
                self.computeIntensityDistribution)

    def getSelectionMaskSource(self):
        """Returns the source of selection mask.

        :rtype: QObject with sigMaskChanged signal and getSelectionMask method.
        """
        return self.__selectionMaskSource

    def computeIntensityDistribution(self):
        """Get the active image and compute the image intensity distribution
        """
        item = self._getSelectedItem()

        if item is None:
            self._cleanUp()
            return

        if isinstance(item, items.ImageBase):
            array = item.getData(copy=False)
            if array.ndim == 3:  # RGB(A) images
                _logger.info('Converting current image from RGB(A) to grayscale\
                    in order to compute the intensity distribution')
                array = (array[:, :, 0] * 0.299 +
                         array[:, :, 1] * 0.587 +
                         array[:, :, 2] * 0.114)
        elif isinstance(item, items.Scatter):
[...]

PR #3281, opened 2020-11-24
Use mask information for histograms

kif:
Changelog: Use the information from the mask when available to calculate the histogram.
                   Masked values are just not used


kif:
close #3124

kif:
It is probably better to use the mask shipped with the image.
superseded by #3289

PR #3282, opened 2020-11-24
silx.math.colormap: Improved performances for small datasets

### LINKED ISSUE #3268: Hight cpu load when using Colormap.applyToData on a single value
I am using this small code on a custom image item to display color of the hovered pixel value.

It uses a hi CPU load, maybe cause of OpenMP, but i only have a single value to convert.

Do you have any idea if this could be speeded up in this case?

Or maybe i could maybe do it on my own?

```
    def _getColoredChar(self, value):
        colormap = self.getColormap()
        data = numpy.array([float(value)])
        colors = colormap.applyToData(data, reference=self)
        cssColor = f"#{colors[0,0]:02X}{colors[0,1]:02X}{colors[0,2]:02X}"
        char = "■"
        return f"""<font color="{cssColor}">{char}</font>"""
```

t20100:
This PR makes use of 1 thread if less than 1000 items need to be converted.
Also, the number of threads used is at most `min(4, OMP_NUM_THREADS)`, so one can avoid using OpenMP here by setting `OMP_NUM_THREADS` to 1.

The 1000 items threshold is arbitrary (i.e, it is the threshold when OpenMP becomes useless/detrimental on my machine).
The benchmark was: 1000 items: 1.5x speed-up, 10k items: 3x speed-up, 100k, 1M items: 4x speed-up.

closes #3268

### DIFF
--- a/silx/math/colormap.pyx
+++ b/silx/math/colormap.pyx
@@ -56,6 +56,8 @@ else:  # Fallback
     DEFAULT_NUM_THREADS = 1
 # Number of threads to use for the computation (initialized to up to 4)
 
+cdef int USE_OPENMP_THRESHOLD = 1000
+"""OpenMP is not used for arrays with less elements than this threshold"""
 
 # Supported data types
 ctypedef fused data_types:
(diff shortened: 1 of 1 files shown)


kif:
LGTM

PR #3283, opened 2020-11-24
Split arc roi

vallsv:
This is a maintenance MR.

- Split ROI module into 3: base/common/arc_roi
- Add test for ROI interaction mode


t20100 (approved):
LGTM

PR #3284, opened 2020-11-24
Clamp stddev*3 with minmax of the data

vallsv:
This PR changes the 3*STDDEV autoscale in order to improve the contrast on range where there is no data.

The result of the 3*STDDEV is now clamped with the minmax data (the real range of the data).

As result is the data is only positive close to 0, the 3*STDDEV will not be netagtive (or smaller than the data).

I still ask if it is better to change the 3*STD or to create another algorithm (in order to be able to use or not this clamp).
If you have an opinion.

Tests still have to be fixed.

Bellow we can see that we win some contrast.
Screenshot from 2020-11-24 16-49-56

Changelog:
- Update stddev*3 autoscale algorithm
   - Clamp it with the minmax data in order to improve the contrast

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -366,7 +366,9 @@ class _NormalizationMixIn:
         if mode == Colormap.MINMAX:
             vmin, vmax = self.autoscaleMinMax(data)
         elif mode == Colormap.STDDEV3:
-            vmin, vmax = self.autoscaleMean3Std(data)
+            dmin, dmax = self.autoscaleMinMax(data)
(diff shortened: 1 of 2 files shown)


t20100:
To me it is better this way, there is no reason for the colormap autoscale range to span beyond the range of the data.

vallsv:
I just tested with Oxidis, it is clamped with the data. https://gitlab.esrf.fr/bliss/qub/-/blob/master/Qub/CTools/pixmaptools/pixmaptools_lut.cpp#L1576

So it's fine to me.
I will update the tests

PR #3289, opened 2020-11-26
Mask out regions for various operation: histograms, colormap selection and profile ... 

### LINKED ISSUE #3124: Combining mask and histogram in widgets/silx view

### LINKED ISSUE #3263: Combining mask and profile in widgets/silx view
Allowing to use current mask to compute profile.

Similar to #3124 but for profile.

Regarding how to connect mask tool to profile, any better idea than what I proposed in #3124, i.e., using:
```
def setSelectionMaskSource(self, source):
        """Set the source of the selection mask.

        :param source:
            QObject with sigMaskChanged signal and getSelectionMask method.
        """
```

An alternative I see is to attach the mask to the image at the item level, that might be more generic since it does not rely on the mask tool.

### LINKED ISSUE #3285: Profile drawing and masked regions ...
After discussion with Thomas, we agree this is a wished feature. Nevertheless taking into account the mask should be wired down at the array level and not connecting at the mask widget as it is proposed in #3281. This will simplify also the calculation of the range of the colormap with masked data (especially important with Eiger2 detectors are the dummy value is above 4 billions, 2^32-1)

This approach will require much more work at lower levels but will allow the generalization to all kind of processing very easily.

Here is a TODO list:

- [ ] Provide the associated mask in the image object like: image.getData() one could have image.getMask()
- [x] Profile calculation are performed in Cython. It is the right time to refactor the code to take mask into account
- [ ] Wire the mask in the ProfileToolBar.

The mask should be an array of same size as the image with dtype uint8 (or None if absent)
The alternative, numpy.maskedarray, has been discarded because we which to be able to manage multiple masks, either 7 overlapping masks or 254 non overlapping masks.

Sponsor: ID17

kif:
Changelog: Associate a mask to the plotting of the image

- [X] Histogram
- [x] Colormap
- [x] Profile

closes #3263
closes #3124
closes #3285


kif:
close #3263

kif:
Sorry, the message was for Valentin ... @vallsv

t20100:
@kif, I'll have a look at this.

t20100:
I'm reworking this PR (extending support to complex images, making sure it works when image size changes,...).
I'll make another PR based on this one because of issues with rebase.

t20100:
I'm closing this PR, it is continued in PR #3369

PR #3291, opened 2020-11-26
silx.gui.plot.PlotWidget: Add pan interaction to ROI authoring (`select-draw`) interaction mode

### LINKED ISSUE #3178: [plot] Middle-click panning is not used for roi edition

t20100:
This PR adds pan interaction to `PlotWidget`'s `select-draw` interaction mode.
The implementation is ugly, yet it looks to work fine and does not change the implementation of other interaction modes.

Having grown quite a bit, PlotWidget interaction needs to be refactored, but that's another story.

closes #3178

### DIFF
--- a/silx/gui/plot/PlotInteraction.py
+++ b/silx/gui/plot/PlotInteraction.py
@@ -1604,6 +1604,8 @@ class DrawSelectMode(FocusManager):
 
     def __init__(self, plot, shape, label, color, width):
         eventHandlerClass = _DRAW_MODES[shape]
+        self._pan = Pan(plot)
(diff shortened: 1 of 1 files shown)


vallsv:
Good thing is it is located at a single place.

Maybe just add a comment with the code to say how it should be implemented.

t20100:
It is not only this mode that needs refactoring, zoom on wheel and pan should be implemented once for all rather than composed in each interaction.
I opened issue #3292 for it.

t20100:
BTW, I added a comment with a ref to #3292

PR #3306, opened 2020-11-30
Fix the sift algorithm 

kif:
Tested on AMD Vega56 OpenCL


t20100:
That was merged into master, I guess you want to backport this to 0.14 as well.

Issue #3303, opened 2020-11-30
OpenCL performance tuning


kif:
Each compiled kernel contains information about CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE
https://www.khronos.org/registry/OpenCL/sdk/1.1/docs/man/xhtml/clGetKernelWorkGroupInfo.html

This provide hints on the prefered workgroup size. One can chose to adjust the group size accordingly at  runtime

PR #3307, opened 2020-11-30
[OpenCL] Expose the workgroup size min, max and memory usage

kif:
This allows ahead of time work-group optimization

Changelog: Expose function to probe kernel and optimize work repartitions

### DIFF
--- a/silx/opencl/common.py
+++ b/silx/opencl/common.py
@@ -34,7 +34,7 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "2012-2017 European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "28/11/2019"
+__date__ = "30/11/2020"
 __status__ = "stable"
 __all__ = ["ocl", "pyopencl", "mf", "release_cl_buffers", "allocate_cl_buffers",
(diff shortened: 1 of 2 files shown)


kif:
close #3303

kif:
I will need this feature in pyFAI 0.20 ...

t20100:
Is it for 0.14? Then it should be merged into branch 0.14 directly.

pierrepaleo:
Not sure `test_query()` will always pass smoothly, I see that some information like `PREFERRED_WORK_GROUP_SIZE_MULTIPLE` are either available in OpenCL 3 and/or `pyopencl >= 2020.3`. But perhaps the pyopencl documentation is incorrect and it will work... I'll test on several machines.

pierrepaleo:
I have the following error with an Intel Xeon W-2133 CPU (pyopencl 2019.1.1, Intel OpenCL ICD):

```
FAIL: test_query (silx.opencl.test.test_addition.TestAddition)
  File "/tmp/silx/build/lib.linux-x86_64-3.7/silx/opencl/test/test_addition.py", line 137, in test_query
    self.assertEqual(3, len(query_kernel_info(program=self.program, kernel="addition", what="COMPILE_WORK_GROUP_SIZE")), "3D kernel")
AssertionError: 3 != 0 : 3D kernel

```

it seems that `COMPILE_WORK_GROUP_SIZE` often gives weird values on CPU, for example I can get `[0, 0, 0]`

PR #3313, opened 2020-12-02
Relax FFT tests tolerance

pierrepaleo:
This PR relaxes the FFT tests tolerance. It seems that Cuda and FFTW do not provide consistent results from one machine to another (even using the same software versions).

Changelog: 

Fix FFT tests tolerance


kif:
For info the tests used to pass on my computer

Issue #3315, opened 2020-12-02
specfile locale tests fail on windows


woutdenolf:
```
FAIL: test_locale_de_DE (silx.io.test.test_specfile.TestSFLocale)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\users\paleo\personal\envs\38\lib\site-packages\silx\io\test\test_specfile.py", line 412, in test_locale_de_DE
    self.crunch_data()
  File "c:\users\paleo\personal\envs\38\lib\site-packages\silx\io\test\test_specfile.py", line 405, in crunch_data
    self.assertAlmostEqual(self.sf3[0].data_line(1)[2],
AssertionError: 1.0 != 1.56 within 7 places (0.56 difference)
```

pierrepaleo:
Handled by #3316

PR #3316, opened 2020-12-02
[specfile] use locale.h on Windows

woutdenolf:
Fix issue #3315

### DIFF
--- a/silx/io/setup.py
+++ b/silx/io/setup.py
@@ -51,7 +51,7 @@ else:
     SPECFILE_USE_GNU_SOURCE = int(SPECFILE_USE_GNU_SOURCE)
 
 if sys.platform == "win32":
-    define_macros = [('WIN32', None)]
+    define_macros = [('WIN32', None), ('SPECFILE_POSIX', None)]
 elif os.name.lower().startswith('posix'):
     define_macros = [('SPECFILE_POSIX', None)]
     # the best choice is to have _GNU_SOURCE defined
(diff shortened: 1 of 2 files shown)


woutdenolf:
@pierrepaleo I used the 0.14 branch as base for the merge. Is that ok?

pierrepaleo:
Yes perfect ! The fixes have to target 0.14.

vasole:
On windows the correct solution would be to use _atof_l

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/atof-atof-l-wtof-wtof-l?view=msvc-160

woutdenolf:
So the use to locale to convert strings to floats in the specfile C-code was disabled on Windows. So I enabled it again and solved what was probably the original issue that was "fixed" by disabling it: the static buffer for the locale name was too short. See here:
https://docs.microsoft.com/en-us/windows/win32/intl/locale-name-constants

On my windows laptop: `localeBuffer = "Dutch_Belgium.1252" (18+1 characters)
On appveyor: "localeBuffer = English_United States.1252" (26+1 characters)

It seems people are using `LOCALE_NAME_MAX_LENGTH=85` so I did the same. But it might be better to use `malloc` or `_atof_l`.

@vasole Is `_atof_l` platform independent?

woutdenolf:
Actually the exit code `-1073740791` (0xC0000409) is STATUS_STACK_BUFFER_OVERRUN. And this was exactly the problem. Sometimes exit codes are helpful ;-).

vasole:
_atof_l is the windows equivalent of strtod_l that is the solution used when setting SPECFILE_USE_GNU_SOURCE

vasole:
I have made a (I think) more efficient implementation, but I do not want to delay things. Since the problem is already solved, I will wait for the release to merge.

PR #3323, opened 2020-12-03
compare integers, not strings

kif:
Compare integers, not strings


woutdenolf (commented):
Good!

kif:
Nop, not good: does not work with 3.1

PR #3326, opened 2020-12-04
Expose minimum (reasonable) workgroup size

kif:
Changelog: Expose the minimum workgroup size


pierrepaleo:
Checked on two different platforms with `Backprojector` kernels, seems to be OK.

PR #3333, opened 2020-12-08
Fix 3332 - issue with silx view when NXclass attribute is not recognized

### LINKED ISSUE #3332: issue with silx view (data from another institute - UCL ?)
``` bash
ERROR:silx.gui.qt._qt:<class 'UnboundLocalError'> local variable 'formatedNX_class' referenced before assignment   File "/mntdirect/_scisoft/tomotools_env/tomwer/ubuntu20_x86_64/0.6/lib/python3.8/site-packages/silx/gui/hdf5/Hdf5TreeModel.py", line 514, in data
    return node.dataType(role)
  File "/mntdirect/_scisoft/tomotools_env/tomwer/ubuntu20_x86_64/0.6/lib/python3.8/site-packages/silx/gui/hdf5/Hdf5Item.py", line 422, in dataType
    text = self.nexusClassName
  File "/mntdirect/_scisoft/tomotools_env/tomwer/ubuntu20_x86_64/0.6/lib/python3.8/site-packages/silx/gui/hdf5/Hdf5Item.py", line 390, in nexusClassName
    if text != formatedNX_class:
```
minimasltic bug fix:

``` python
    @property
    def nexusClassName(self):
        """Returns the Nexus class name"""
        if self.__nx_class is None:
            obj = self.obj.attrs.get("NX_class", None)
            if obj is None:
                text = ""
            else:
                text = self._getFormatter().textFormatter().toString(obj)
                text = text.strip('"')
                # Check NX_class formatting
                formatedNX_class = None
                lower = text.lower()
                if lower.startswith('nx') or lower.startswith('nxs'):
                    formatedNX_class = 'NX' + lower[2:]
                if lower == 'nxcansas':
                    formatedNX_class = 'NXcanSAS'  # That's the only class with capital letters...
                if text != formatedNX_class:
                    _logger.error("NX_class: %s is malformed (should be %s)",
                                  text,
                                  formatedNX_class)
                text = formatedNX_class

            self.__nx_class = text
        return self.__nx_class
```

payno:
Fix handling of `nexusClassName` property from `silx.gui.hdf5.hdf5item.Hdf5Item` if node contains an `NX_class` unrecognized.

closes #3332

### DIFF
--- a/silx/gui/hdf5/Hdf5Item.py
+++ b/silx/gui/hdf5/Hdf5Item.py
@@ -383,12 +383,16 @@ class Hdf5Item(Hdf5Node):
                 text = text.strip('"')
                 # Check NX_class formatting
                 lower = text.lower()
+                formatedNX_class = None
                 if lower.startswith('nx'):
                     formatedNX_class = 'NX' + lower[2:]
                 if lower == 'nxcansas':
                     formatedNX_class = 'NXcanSAS'  # That's the only class with capital letters...
-                if text != formatedNX_class:
-                    _logger.error("NX_class: %s is malformed (should be %s)",
+                if formatedNX_class is None:
+                    _logger.error("NX_class: '%s' is malformed",
+                                  text)
+                elif text != formatedNX_class:
+                    _logger.error("NX_class: '%s' is malformed (should be '%s')",
                                   text,
                                   formatedNX_class)
                 text = formatedNX_class
--- a/silx/gui/hdf5/Hdf5Item.py
+++ b/silx/gui/hdf5/Hdf5Item.py
@@ -383,15 +383,12 @@ class Hdf5Item(Hdf5Node):
                 text = text.strip('"')
                 # Check NX_class formatting
                 lower = text.lower()
-                formatedNX_class = None
+                formatedNX_class = ""
                 if lower.startswith('nx'):
                     formatedNX_class = 'NX' + lower[2:]
                 if lower == 'nxcansas':
                     formatedNX_class = 'NXcanSAS'  # That's the only class with capital letters...
-                if formatedNX_class is None:
-                    _logger.error("NX_class: '%s' is malformed",
-                                  text)
-                elif text != formatedNX_class:
+                if text != formatedNX_class:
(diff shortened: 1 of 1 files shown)


payno:
Fix issue see on an nexus-hdf5 provided by an other institute.
I also update `%s` to `'%s'` on log message for readability.

silx/gui/hdf5/Hdf5Item.py
```diff
@@ -383,12 +383,16 @@ def nexusClassName(self):
                 text = text.strip('"')
                 # Check NX_class formatting
                 lower = text.lower()
+                formatedNX_class = None
```
t20100, line 386: This results in `nexusClassName` that can be `None` while it was always `str` before.

From a quick look at the code it can lead to an issue here: https://github.com/silx-kit/silx/blob/617001f8fce5d39498869bc1153c88fe8bbc0dfc/silx/gui/hdf5/Hdf5Item.py#L426 where text is otherwise returned.

Maybe safer to initialize to empty string.

payno: That was one of my interrogation because for "def dataType(self, role)" can return `None`. And my interpretation of `nexusClassName` is more or less 'should be None' if unknow/undefined.
But I don"t know much this part of the source code.

payno: but if you think this is safer I can update it and test it on a file raising the error.

t20100: I don't know much this code either and I would also be OK with `nexusClassName` to be None if no valid NX class.
Yet, when there is no a`NX_class` attribute, it is set to `''`: https://github.com/silx-kit/silx/blob/617001f8fce5d39498869bc1153c88fe8bbc0dfc/silx/gui/hdf5/Hdf5Item.py#L380
So I think it is better and safer to be consistent.

payno: yes you are right.

payno:
tested with the update and with a file raising an error previously. Ok with the modification & update

PR #3334, opened 2020-12-08
Fixed interactive tests on Windows10

t20100:
This PR fixes interactive tests that were failing under Windows10 during 0.14 testing.

This currently targets 0.14 branch. For me, it is not critical to embed but still nice to have (provided it does not break those tests on other platforms.

So far tested on Windows10 with PyQt5 and PySide2.

### DIFF
--- a/silx/gui/plot/tools/test/testROI.py
+++ b/silx/gui/plot/tools/test/testROI.py
@@ -557,8 +557,9 @@ class TestRegionOfInterestManager(TestCaseQt, ParametricTestCase):
         mx, my = self.plot.dataToPixel(*center)
         self.mouseMove(widget, pos=(mx, my))
         self.mousePress(widget, qt.Qt.LeftButton, pos=(mx, my))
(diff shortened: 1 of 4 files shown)


silx/gui/plot/tools/test/testROI.py
```diff
@@ -557,8 +557,9 @@ def testLineInteraction(self):
         mx, my = self.plot.dataToPixel(*center)
         self.mouseMove(widget, pos=(mx, my))
         self.mousePress(widget, qt.Qt.LeftButton, pos=(mx, my))
+        self.mouseMove(widget, pos=(mx, my+25))
         self.mouseMove(widget, pos=(mx, my+50))
-        self.mouseRelease(widget, qt.Qt.LeftButton, pos=(mx, my))
+        self.mouseRelease(widget, qt.Qt.LeftButton, pos=(mx, my+50))
```
vallsv, line 562: Did you try to remove the pos on mouseRelease? Sounds like kind of duplication.

PR #3336, opened 2020-12-09
silx.gui.plot: Fixed RegionOfInterest `contains`

### LINKED ISSUE #3335: Right button click on a plot with profile tools leads to an exception
With both current master and 0.14 branch, creating a horizontal line profile on an image and then pressing right click on the plot leads to this exception:
```
ERROR:silx.gui.qt._qt:<class 'TypeError'> 'float' object is not subscriptable   File "/silx/gui/plot/tools/roi.py", line 606, in _feedContextMenu
    if roi.contains(data):
  File "/silx/gui/plot/items/roi.py", line 405, in contains
    return position[1] == self.getPosition()[1]
```
It is not the case with silx v0.13.2.

Reproducible with `examples/plotInteractiveImageROI.py` or `silx view` with an image

t20100:
This PR fixes crashes on right click with profiles and ROIs by implementing `contains` methods.
Those methods are exact, so their use in interaction might not be ideal since it is almost impossible.

Still todo: ellipse and testing.

closes #3335

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Change Log
 0.14.0: 2020/12/11
 ------------------
 
+This is the first version of `silx` supporting `h5py` >= v3.0.
+
 * `silx.gui`:
 
   * Added support for HDF5 external data (virtual and raw) (PR #3222)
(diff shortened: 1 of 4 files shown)


vallsv:
Be careful you have triggered other commits with this PR.

pierrepaleo:
Thanks ! Can you target `0.14` ?

t20100:
Ready for review.
I'm doing a bit more testing but it should be OK.

t20100:
Added entry in changelog

t20100:
There's still a bug to fix... and `contains` does not always return a `bool`.

t20100:
Fixed bug in circle ROI and make sure `contains` always return a `bool` (it wasn't the case with `LineROI`).

payno (approved):
LGTM

PR #3338, opened 2020-12-11
Backport 0.14 changes to master

pierrepaleo:
This PR aims at merging all the changes made in the `0.14` branch to `master`.

PR #3341, opened 2020-12-14
Update version.py

kif:
One should always look at the code before merging :)


pierrepaleo:
No problem, but I don't understand how `MINOR` got to `15` ? It was [not changed](https://github.com/silx-kit/silx/pull/3338/files) in the `0.14` branch.

vallsv:
The minor version 15 was already part of master. And the merge did change dev to final, as it's line by line merge.

I think there is nothing weird here.

Usually it's better to remove the commit which change this stuff for the the backport. I through it was fine cause there was no conflict. My bad.

PR #3349, opened 2021-01-05
ColormapDialog: add an option to compute min / max value from visible data or a RectangleROI

payno:
TODO
=====

- [x] update interface to add a button to set min and max from a ROi or from the visible data (and with the active item)
- [x] handle ImageData
- [x] handle Scatter
- [x] create a util to compute `DataItem` mask with a ROI (silx.gui.plot.tools.roifiltering). This is only doing minimalistic operations just to avoid code duplication. Maybe this could include in the future more processing from stats.
- [x] add test
- [x] add doc

Changelog: ColormapDialog can now compute min / max value from visible data or a RectangleROI

Screenshots
==========

image

image

More
====

- closes https://gitlab.esrf.fr/tomotools/tomwer/-/issues/494

### DIFF
--- a/silx/gui/colors.py
+++ b/silx/gui/colors.py
@@ -925,15 +925,7 @@ class Colormap(qt.QObject):
             vmax = None
 
         if vmin is None or vmax is None:  # Handle autoscale
-            from .plot.items.core import ColormapMixIn  # avoid cyclic import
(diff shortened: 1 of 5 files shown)


payno:
replaced by #3365

PR #3356, opened 2021-01-22
silx.gui.utils.glutils: Fixed `isOpenGLAvailable`

t20100:
Should fix issues with import when running `isOpenGLAvailable`. PR 3184 was apparently not enough.

It still needs testing.
Alternative can be to import `subprocess` lazily in `def _runtimeOpenGLCheck(version)`

Related to #3182

### DIFF
--- a/silx/gui/setup.py
+++ b/silx/gui/setup.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2021 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -43,6 +43,7 @@ def configuration(parent_package='', top_path=None):
     config.add_subpackage('data')
     config.add_subpackage('dialog')
     config.add_subpackage('utils')
+    config.add_subpackage('utils.glutils')
     config.add_subpackage('utils.test')
 
     return config
--- a/silx/gui/utils/glutils.py
+++ b/silx/gui/utils/glutils.py
@@ -74,7 +74,7 @@ def _runtimeOpenGLCheck(version):
         script_file = shutil.copy(__file__, directory)
         try:
             error = subprocess.check_output(
-                [sys.executable, script_file, major, minor],
+                [sys.executable, '-s', '-S', script_file, major, minor],
                 env=env,
                 timeout=2)
         except subprocess.TimeoutExpired:
--- a/silx/gui/utils/glutils.py
+++ b/silx/gui/utils/glutils.py
@@ -32,7 +32,7 @@ if __name__ == "__main__":
     # When run as a script, remove directory from sys.path
     # This avoids other script in same directory to override Python modules
     sys.path = [path for path in sys.path
-        if os.path.abspath(path) == os.path.abspath(os.path.dirname(__file__))]
+        if os.path.abspath(path) != os.path.abspath(os.path.dirname(__file__))]
 
 import subprocess
 from silx.gui import qt
--- a/silx/gui/utils/glutils.py
+++ b/silx/gui/utils/glutils.py
@@ -31,8 +31,8 @@ import sys
 if __name__ == "__main__":
     # When run as a script, remove directory from sys.path
     # This avoids other script in same directory to override Python modules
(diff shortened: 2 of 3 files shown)


vallsv:
Ah ok i see.

But what is the part of the code relying on `concurrent.futures`?

vallsv:
Oh i remember now... Was the relative import or our concurrent module. No luck.

t20100:
It doesn't fix the issue, that's a weird issue.

vallsv:
To me the good way to handle it is to use `-m` with the module name.

Relative modules will just be relative, as we expect.

vallsv:
Here is traceback from a beamline
```
ERROR 2021-01-13 15:03:00,611 flint.output: Traceback (most recent call last):
ERROR 2021-01-13 15:03:00,611 flint.output:   File "/users/blissadm/conda/miniconda/envs/bliss_dev/lib/python3.7/site-packages/silx/gui/utils/glutils.py", line 37, in <module>
ERROR 2021-01-13 15:03:00,611 flint.output:     import subprocess
ERROR 2021-01-13 15:03:00,611 flint.output:   File "/users/blissadm/conda/miniconda/envs/bliss_dev/lib/python3.7/subprocess.py", line 50, in <module>
ERROR 2021-01-13 15:03:00,611 flint.output:     import signal
ERROR 2021-01-13 15:03:00,611 flint.output:   File "/opt/bliss/conda/miniconda/envs/bliss_dev/lib/python3.7/site-packages/silx/gui/utils/signal.py", line 32, in <module>
ERROR 2021-01-13 15:03:00,611 flint.output:     from silx.gui.utils import concurrent
ERROR 2021-01-13 15:03:00,611 flint.output:   File "/users/blissadm/conda/miniconda/envs/bliss_dev/lib/python3.7/site-packages/silx/gui/utils/concurrent.py", line 35, in <module>
ERROR 2021-01-13 15:03:00,611 flint.output:     from concurrent.futures import Future
ERROR 2021-01-13 15:03:00,611 flint.output:   File "/opt/bliss/conda/miniconda/envs/bliss_dev/lib/python3.7/site-packages/silx/gui/utils/concurrent.py", line 35, in <module>
ERROR 2021-01-13 15:03:00,611 flint.output:     from concurrent.futures import Future
ERROR 2021-01-13 15:03:00,611 flint.output: ModuleNotFoundError: No module named 'concurrent.futures'; 'concurrent' is not a package
```
No idea why they don't complain anymore.

t20100:
`-m` fixes the issue, thanks @vallsv for the idea.
There is still an issue when run through `bootstrap.py` though.

t20100:
Root cause is here:

- Import of `silx.gui._glutils.OpenGLWidget` in the script part (https://github.com/silx-kit/silx/blob/master/silx/gui/utils/glutils.py#L152):
```python
if __name__ == "__main__":
    from silx.gui._glutils import OpenGLWidget
``` 
- and import of `silx.gui.utils.glutils` in `OpenGLWidget` as usage of `isOpenGLAvailable`, so the test is actually using itself...

t20100:
Updated with an altenative: copy `glutils.py` in a temporary folder before running it as a script.
From what I tested, it does not seems to change the duration of the test.

This avoids:
- Running a script in the `silx.gui.utils` folder (which cause python's `signal` and `concurrent` packages to be overridden by those of `silx.gui.utils`)
- Using `-m` which fails when run through `bootstrap.py`

I also added `-s -S` option to preserve as much as possible `sys.path` from calling process.

Still need to be tested on macos and windows

vallsv:
Any idea why the bootstrap is not working with `-m`? Souds like there is something weird with, or it just about our own protections?

t20100:
From my current understanding, `bootstrap` patches `sys.path` and this is patch is not done for the subprocess so you end-up with the root of the project in `sys.path`.
One option would be in `bootstrap` to change the current working directory when it is the root of the project, but that has other side-effects.

I'll still dig a bit, I am not happy with the temporary file, but for now it is what I found that works in most conditions.

t20100:
One last proposition (the best from my point of view): move `glutils.py` to `glutils/__init__.py` and run it as a script in the subprocess, so there is no conflict with what's in the script directory.

Works fine, no need for temporary file.

I added [-s](https://docs.python.org/3/using/cmdline.html#cmdoption-s) and [-S](https://docs.python.org/3/using/cmdline.html#id3) `python` options to keep `sys.path` as it is in the calling process.

Ready for review.

vallsv:
Sounds like a cheat, but it's fine for me.

t20100:
Yes, I would have preferred to use `-m` but it fails with `bootstrap.py`... unless we move the sources in a sub-folder (#3357)

PR #3353, opened 2021-01-12
Misc.: Use github action to run CI on linux and macos

t20100:
closes #3272

PR #3359, opened 2021-01-25
Misc.: Removed Travis config, Added OpenCL on linux in gitub actions

t20100:
Some clean-up and added OpenCL installation on Linux with github actions.


t20100:
Follow-up of #3353

Updated matrix of effectively tested features:

os | GUI | OpenGL | OpenCL
-- | --- | ------ | ------
linux | ✓ | ✓ | ✓
macos | ✓ | ✓ | ✓
windows | ✗ | ✗ | ✗

payno:
those if statement are a bit confusing but seems the "github" / yaml way to do it

t20100:
Yes in gh actions, you can add some condition each the steps

PR #3360, opened 2021-01-26
Misc.CI: Run tests on macos 11.0 with github actions

t20100:
This should make it pass on macos where pyopencl needs to be compiled and latest pyoepncl version seems to require macos 11.0.


t20100:
`macos-11.0` not currently publicly available on github actions.

PR #3361, opened 2021-01-26
Misc. CI: Fixed testing on macos

t20100:
Pin-point pyopencl on macos since v2021.1 requires macos11.

This should fix CI.


t20100:
BTW, this is related to https://github.com/inducer/pyopencl/issues/422

PR #3365, opened 2021-01-28
silx.gui.dialog.ColormapDialog: Added scale to visible or selected area buttons

### LINKED ISSUE #2429: [ColormapDialog/PlotWidget] Add an option to set colormap range from visible data

t20100:
This PR proposes an alternative to #3349. Compared to it it also supports mean+/-3std and takes a smaller footprint in the `ColormapDialog`.

It provides 2 on-demand colormap range scaling:
- From the currently visible area of the item that is registered to the `ColormapDialog`
- From a rectangular selection on the registered item. When pressing the "Selection" button, its text changes to "Draw area..." with a waiting icon and waits for user input. This mode is cancelled by pressing again the button or changing the plot interaction mode (e.g., select zoom mode).

Both uses the current autoscale mode (i.e., min/max or mean+/-3std).
It works for `Scatter` and `ImageData`.

Screen Shot 2021-01-28 at 11 25 30

closes #2429

### DIFF
--- a/silx/gui/dialog/ColormapDialog.py
+++ b/silx/gui/dialog/ColormapDialog.py
@@ -1684,6 +1684,8 @@ class ColormapDialog(qt.QDialog):
         colormap = self.getColormap()
         enabled = self._item is not None and colormap is not None and colormap.isEditable()
         self._visibleAreaButton.setEnabled(enabled)
+        if not enabled:
+            self._selectedAreaButton.setChecked(False)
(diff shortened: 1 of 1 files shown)


payno:
Great. Can't wait to have it on the next release :)

payno (approved):

PR #3367, opened 2021-01-29
silx.gui.hdf5: Made humanReadableHdf5Type a static method of Hdf5Formatter

loichuder:
...to be able to reuse it more easily in other (non-Qt) projects.

### DIFF
--- a/silx/gui/hdf5/Hdf5Formatter.py
+++ b/silx/gui/hdf5/Hdf5Formatter.py
@@ -186,7 +186,8 @@ class Hdf5Formatter(qt.QObject):
         dtype = dtype.newbyteorder('N')
         return text
 
-    def humanReadableHdf5Type(self, dataset):
+    @staticmethod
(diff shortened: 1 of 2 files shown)


vallsv:
I don't think it's a good idea.

Here it is object oriented. If you want to use it outside, i think it is better to move the function outside. And use the normal function in the method if you like.

loichuder:
> Here it is object oriented. If you want to use it outside, i think it is better to move the function outside. And use the normal function in the method if you like.

You mean copying the function _outside of the silx project_ or _outside of the class_ ?

vallsv:
Outside the class. And if you don't want to use it in Qt context, that's probably better to move it out the this package anyway.

loichuder:
Yeah, in the end, I think I will copy paste the function in our project to avoid pulling `silx` as a dep just for that.

Thanks :+1:

PR #3368, opened 2021-01-29
Concurrent h5py

### LINKED ISSUE #3354: Concurrent HDF5 reading and writing without SWMR
For several reasons HDF5's SWMR is not usable:
1. No new groups/datasets can be created after the writer has enabled SWMR mode.
2. SWMR only works when the writer and the readers do things in a specific, synchronized order.
3. Non-SWMR readers (and that's basically most tools) will prevent the writer from enabling SWMR mode (or open the file to begin with). 
4. The SWMR feature does not work on network file systems (NFS, SMB/Windows file shares, and so on).

The alternative to SWMR is having a writer in normal mode (default h5py usage) and a reader that does not lock the file and captures exceptions caused by reading something that is being modified by the writer.

Proposal:`silx.io.h5py` to replace `h5py`

woutdenolf:
Closes #3354

Tested with h5py 2.8, 2.9, 2.10, 3.1

Stress tests: https://gitlab.esrf.fr/denolf/concurrent_h5py

The main API:
* `silx.io.h5py_utils.File`: replace `h5py.File` mainly to handle file locking
* `silx.io.h5py_utils.retry`: retry the method whenever you get an HDF5 IO exception
* `silx.io.h5py_utils.retry_in_subprocess`: retry in a subprocess to capture segfaults
* `silx.io.h5py_utils.retry_contextmanager`: retry entering the context manager

The retry logic can also be used for other things than HDF5 reading:
* `silx.utils.retry.retry`
* `silx.utils.retry.retry_in_subprocess`
* `silx.utils.retry.retry_contextmanager`

Common use case of `silx.io.h5py_utils` for processing a Bliss Nexus file:

```python
import silx.io.h5py_utils

@silx.io.h5py_utils.retry()
def process_scan(filename, name):
    """The method will be executed again if
    any HDF5 IO fails.
    """
    with silx.io.h5py_utils.File(filename) as h5file:
        scan = h5file[name]
        I0 = scan["measurement/I0"][()]
        It = scan["measurement/It"][()]
        return It/I0

scans = silx.io.h5py_utils.safe_top_level_names("myfile.h5")

for name in scans:
    process_scan("myfile.h5", name)
```

### DIFF
--- a/silx/io/h5py_utils.py
+++ b/silx/io/h5py_utils.py
@@ -1,0 +1,317 @@
+# coding: utf-8
+# /*##########################################################################
+# Copyright (C) 2016-2017 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ############################################################################*/
+"""
+This module provides utility methods on top of h5py, mainly to handle
+parallel writing and reading.
+"""
+
+__authors__ = ["W. de Nolf"]
+__license__ = "MIT"
+__date__ = "27/01/2020"
+
+
+import os
+import traceback
+import h5py
+
+from .._version import calc_hexversion
+from ..utils import retry as retry_mod
+
+H5PY_HEX_VERSION = calc_hexversion(*h5py.version.version_tuple[:3])
+HDF5_HEX_VERSION = calc_hexversion(*h5py.version.hdf5_version_tuple[:3])
+
+HDF5_SWMR_VERSION = calc_hexversion(*h5py.get_config().swmr_min_hdf5_version[:3])
+HDF5_TRACK_ORDER_VERSION = calc_hexversion(2, 9, 0)
+
+HAS_SWMR = HDF5_HEX_VERSION >= HDF5_SWMR_VERSION
+HAS_TRACK_ORDER = H5PY_HEX_VERSION >= HDF5_TRACK_ORDER_VERSION
+
+
+def _is_h5py_exception(e):
+    for frame in traceback.walk_tb(e.__traceback__):
+        if frame[0].f_locals.get("__package__", None) == "h5py":
+            return True
+    return False
+
+
+def _retry_h5py_error(e):
+    """
+    :param BaseException e:
+    :returns bool:
+    """
+    if _is_h5py_exception(e):
+        if isinstance(e, (OSError, RuntimeError)):
+            return True
+        elif isinstance(e, KeyError):
+            # For example this needs to be retried:
+            # KeyError: 'Unable to open object (bad object header version number)'
+            return "Unable to open object" in str(e)
+    elif isinstance(e, retry_mod.RetryError):
+        return True
+    return False
+
+
+def retry(**kw):
+    """Decorator for a method that needs to be executed until it not longer
+    fails on HDF5 IO. Mainly used for reading an HDF5 file that is being
+    written.
+
+    :param **kw: see `silx.utils.retry`
+    """
+    kw.setdefault("retry_on_error", _retry_h5py_error)
+    return retry_mod.retry(**kw)
+
+
+def retry_contextmanager(**kw):
+    """Decorator to make a context manager from a method that needs to be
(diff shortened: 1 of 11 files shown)


silx/io/h5py_utils.py
```diff
@@ -0,0 +1,328 @@
...
+        :param str or None mode: read-only by default
+        :param bool or None enable_file_locking: by default it is disabled for `mode='r'`
+                                                 and `swmr=False` and enabled for all
+                                                 other modes.
+        :param bool or None swmr: try both modes when `mode='r'` and `swmr=None`
+        :param **kwargs: see `h5py.File.__init__`
+        """
+        if mode is None:
+            mode = "r"
+        elif mode not in ("r", "w", "w-", "x", "a"):
```
payno, line 224: 'r+' is missing. By the way according to what you discovered with the 'r' mode (and probably r+ ?) maybe it is worth adding a warning if libver == 110 ?

woutdenolf: A python warning or a warning in the docs? If a python warning: a warning on the writer side saying "file cannot be read until closed" or a warning on the reader side saying "will fail to open when being written"?

payno: A python warning. My thought was more or on the reader side.

woutdenolf:
```python
# writer
import h5py
f=h5py.File("test.h5", mode="a", libver="latest")
f.flush()
```

```python
# reader
import h5py
f=h5py.File("test.h5", mode="r")

OSError: Unable to open file (file is already open for write (may use <h5clear file> to clear file consistency flags))
```

If you want a warning on the reader side, you'll have to check the libver the file is written in. This means you need to open the file which mean you already get the exception "file is already open for write". In any case, I don't think it is useful the have a warning just before the exception basically saying the same thing.

payno: :ok_hand:

t20100:
Nice to have the generic retry stuff apart from the HDF5 utility!

I hope that one day we won't need this any more, but in the (probably very long) meantime it makes it work.

A few minor "cosmetic" suggestions (up to you):
* ` silx/io/h5py_utils.py `:
  * Make `is_h5py_exception`, `retry_h5py_error`, `include_default` private?
  * I would not put a `open` function since `h5py` does not provide one
* `silx/utils/retry.py`:
  * Make `default_retry_on_error` private?
  * Remove `retry_` prefix from arguments `retry_timeout`, `retry_period` and `retry_on_error` since function are named `retry*`

woutdenolf:
Thanks for the review!

> Nice to have the generic retry stuff apart from the HDF5 utility!
> 
> I hope that one day we won't need this any more, but in the (probably very long) meantime it makes it work.

Exactly, we make it work like this and work on a better solution afterwards.

> A few minor "cosmetic" suggestions (up to you):
> 
>     * `silx/io/h5py_utils.py`:
>       
>       * Make `is_h5py_exception`, `retry_h5py_error`, `include_default` private?

Ok, except `include_default` as it may be useful for `open_item` as well (None is the default there). I will rename it to `group_is_complete` or something.

>       * I would not put a `open` function since `h5py` does not provide one

Ok.

>     * `silx/utils/retry.py`:
>       
>       * Make `default_retry_on_error` private?

OK

>       * Remove `retry_` prefix from arguments `retry_timeout`, `retry_period` and `retry_on_error` since function are named `retry*`

I added the prefix because you can override the timeout (and also the others) in each individual call while the argument names without the "retry_" prefix are rather generic and have higher probability of colliding with the method's named arguments. So for example:

```python
@retry(retry_timeout=10)
def method(..., timeout=None):
     # "timeout" has nothing to do with retrying

# I want to override 10 by 1 while the method already has
# a timeout argument for some other purpose
method(retry_timeout=1, timeout=5)
```

t20100:
> I added the prefix because you can override the timeout (and also the others) in each individual call 

OK. That's the `kw.pop`s.

payno (commented):
pretty convenient to use (regarding the context :) and nice test suite)

I was wondering if it would be beneficial to have a redefinition of `silx.io.utils.get_data(url, timeout)` function in `silx.io.h5py_utils`.
This can be done in another PR by someone else.
But from my point of view it could be convenient to have 
something simple like: 
``` python
@retry()
def get_data(url):
    return silx_get_data(url)
```
or something more advance like: 
`get_data(url, retry_timeout, retry_invalid...)` and call open_item if this is an hdf5 file otherwise call the original get_data. But maybe this is overkill.

silx/io/h5py_utils.py
```diff
@@ -0,0 +1,317 @@
...
+        """
+        if mode is None:
+            mode = "r"
+        elif mode not in ("r", "w", "w-", "x", "a", "r+"):
+            raise ValueError("invalid mode {}".format(mode))
+        if not HAS_SWMR:
+            swmr = False
+
+        if enable_file_locking is None:
+            enable_file_locking = bool(mode != "r" or swmr)
```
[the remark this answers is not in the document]
woutdenolf: "r+" is like append mode in most regards so file locking need to be enabled. If you have any case that does this, make sure the change that asap, you'll be corrupting the HDF5 file otherwise.

silx/io/h5py_utils.py
```diff
@@ -0,0 +1,317 @@
...
+                    libver = self._SWMR_LIBVER
+                super().__init__(
+                    filename, mode=mode, swmr=swmr, libver=libver, **kwargs
+                )
+            else:
+                raise
+        else:
+            self._add_nopen(1)
+            try:
+                if mode != "r" and swmr:
```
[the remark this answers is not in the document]
woutdenolf: Same comment as above: "r+" needs to be treated as a writer when dealing with SWMR and locks.

woutdenolf:
> I was wondering if it would be beneficial to have a redefinition of `silx.io.utils.get_data(url, timeout)` function in `silx.io.h5py_utils`.

I'm not quite sure a method in `silx.io.h5py_utils` should be able to handle non-HDF5 things. It should be the other way around imo: `silx.io.h5py_utils.get_data` (HDF5 only with retry) which gets called by `silx.io.utils.get_data` but then it's probably wise to overwrite the default `retry_timeout=None`. Otherwise this call will block forever when the HDF5 file is corrupt.

Lets keep that for another PR.

silx/io/h5py_utils.py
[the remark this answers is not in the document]
payno: My bad sorry, I thought it was a mode "create" file if does not exist and read only (that you couldn't do any "append" on it but I was wrong)

payno (approved):

t20100 (approved):

PR #3371, opened 2021-02-03
Consistency code/doc

vallsv:
Here is a small typo on the documentation.

The method is defined as:
```addCurve(replace=False)```


payno:
nice catch

PR #3373, opened 2021-02-10
[gui][ImageStack] add an option to avoid reseting the plot each time

payno:


t20100:
A few comments:
- Would be good to have a docstring for `ImageStack.setResetZoom`
- Would be good to have a getter as well: `isResetZoom`
- Alternative method name: `is|setAutoResetZoom`?
- When calling `set(Auto)ResetZoom(True)`, you might want to reset zoom at that time as well?

payno:
Yes you are right others are named "is | setAutoResetZoom".
I also added the auto reset when set to True

PR #3376, opened 2021-02-17
dicttoh5 overwrite behavior

### LINKED ISSUE #3226: dicttoh5: properly define update/overwrite behavior
There are several inconsistencies in the overwrite behavior of `dicttoh5`. For example this one:

```python
from pprint import pprint
from silx.io import dictdump

ddict = {"group1": {"data1": 1}, "group1@attr": "attr"}
dictdump.dicttoh5(ddict, "test.h5", overwrite_data=True, mode="w")
pprint(dictdump.h5todict("test.h5"))

ddict = {"group1": {"data2": 2}}
dictdump.dicttoh5(ddict, "test.h5", overwrite_data=True, mode="a")
pprint(dictdump.h5todict("test.h5"))  # group1 has attr, data1 and data2

ddict = {"group1": {}}
dictdump.dicttoh5(ddict, "test.h5", overwrite_data=True, mode="a")
pprint(dictdump.h5todict("test.h5"))  # group1 has only attr
```

These inconsistencies need to be solved and the `overwrite_data` behavior needs to be clearly defined. In addition I need a way to "update" an existing HDF5 file.

For both issues I propose these three dump modes:
   * __overwrite__ (existing things may be overwritten): the HDF5 tree is **exactly** the dict tree
   * __update__ (existing things may be overwritten): the dict tree is a **subtree** of the HDF5 tree
   * __add__ (existing things are untouched): nodes from the dict tree are added to the HDF5 tree when the existing structure allows it

As for the existing `overwrite_data`: unfortunately it behaves sometimes like __overwrite__ and sometimes like __update__. As this behavior is confusing I don't think we need to preserve it.

@vasole @t20100 @vallsv Comments?

woutdenolf:
Closes #3226 

Add missing features so the Bliss Nexus writer can use `dicttonx` from `silx`.

- `dicttoh5` has a new `existing` argument to specify overwrite behavior: 
   * **add**: Existing HDF5 items are untouched. Nodes from the dict tree are added to the HDF5 tree when the existing HDF5 tree allows it.
   * **update**: Existing HDF5 items may be updated or deleted explicitly by assigning them to `None`. The dict tree will be a subtree of the HDF5 tree.
   * **overwrite**: Existing HDF5 items may be overwritten or deleted. The HDF5 tree will be exactly the dict tree, except for already existing top-level items that were not deleted explicitly by assigning them to `None`.

- A `None` value in the dict of `dicttoh5` will never create anything (previously it was creating an empty HDF5 group). When `existing == "overwrite"` or `existing == "update"` a `None` will delete the associated key from the HDF5 file if it exists.

- The `overwrite_data` argument of `dicttoh5` is deprecated.
  * `True` is approx. equivalent to `existing="update"`
  * `False` is equivalent to `existing="add"`

- `dicttonx`: by default add missing "NX_class" attributes of groups when `existing in [None, "add"]`

- `dicttoh5`/`h5todict`: accept a `Group` or `Dataset` as HDF5 root (before it was only `str` and `File`)

- `nxtodict`: now includes attributes by default (as opposed to `h5todict`). Attributes are essential for the Nexus format.

Remark: The `h5file` argument should now be called `h5root` but this would break the API (it seems some calls use `dicttoh5(treedict, h5file=...)`).

### DIFF
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -206,13 +206,14 @@ def dicttoh5(treedict, h5file, h5path='/',
     :param create_dataset_args: Dictionary of args you want to pass to
         ``h5f.create_dataset``. This allows you to specify filters and
         compression parameters. Don't specify ``name`` and ``data``.
-    :param existing: can be 'add' (default), 'update' or 'overwrite'.
-        add: Existing HDF5 items are untouched. Nodes from the dict tree are
+    :param existing: Can be ``add`` (default), ``update`` or ``overwrite``.
+
+        * ``add``: Existing HDF5 items are untouched. Nodes from the dict tree are
             added to the HDF5 tree when the existing HDF5 tree allows it.
-        update: Existing HDF5 items may be updated or deleted explicitly
+        * ``update``: Existing HDF5 items may be updated or deleted explicitly
             by assigning them to `None`. The dict tree will be a subtree
             of the HDF5 tree.
-        overwrite: Existing HDF5 items may be overwritten or deleted.
+        * ``overwrite``: Existing HDF5 items may be overwritten or deleted.
             The HDF5 tree will be exactly the dict tree, except for already
             existing top-level items that were not deleted explicitly by
             assigning them to `None`.
--- a/silx/io/dictdump.py
+++ b/silx/io/dictdump.py
@@ -246,8 +246,8 @@ def dicttoh5(treedict, h5file, h5path='/',
 
     if overwrite_data is not None:
         reason = (
-            '`overwrite_data=True` becomes `existing="update"` and '
-            '`overwrite_data=False` becomes `existing="add"`'
+            "`overwrite_data=True` becomes `existing='update'` and "
+            "`overwrite_data=False` becomes `existing='add'`"
         )
         deprecated_warning(
             type_="argument",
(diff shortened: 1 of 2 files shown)


woutdenolf:
We need to decide on the argument name (currently called `existing`) and its values (currently "add", "update" and "overwrite")

t20100:
What about `update_mode`?

t20100:
Or `sync_mode`?

And some propositions for the values:
- `add` -> `append`, `extend`, `safe`
- `update` -> `update` sounds the best to me, `sync`, `backup`
- `overwrite` -> `replica` or `mirroring`

payno:
`update_mode` seems fine to me. Or `write_behavior`. I am less fan of `sync_mode`

Options names are good as they are but for :
* `add`: `append` and `extend` work as well
* for the other two I would keep the existing names

woutdenolf:
Thanks for the suggestions. How about `update_mode` with following values and docstrings
* `extend`: Extend the existing HDF5 tree when possible. Existing HDF5 items (groups, datasets and attributes) remain untouched.
* `modify`: Extend the existing HDF5 tree when possible, edit existing HDF5 items when possible (i.e. attributes and ame-sized datasets) and delete HDF5 items with a `None` value in the dict tree. The dict tree will be a subtree of the HDF5 tree.
* `replace`: Replace the existing HDF5 tree. The HDF5 tree will be exactly the dict tree, except for already existing top-level items that were not deleted explicitly by assigning them to `None`.

PR #3380, opened 2021-02-22
Refactor image view

### LINKED ISSUE #288: [Plot] ImageView: possibility to display X and Y profiles instead histograms
For some applications it will be interesting for me to customize ImageView (or to have another widget with similar aspect) in order to display the X and Y profiles when crossing the mouse on the image. 

I know it is possible to use the profile tool, but sometimes is easier and quicker like that.

### LINKED ISSUE #3347: Be able to reuse ImageView profile
For Flint i would like to have an `ImageView` with the profiles, optionally displayed.

There is probably a way to rework the image view in order to reuse each component. So what i could integrate them the way i want.

### LINKED ISSUE #3348: ImageView profile based on a cross ROI
It would be nice to be able to connect the cross profiles and the ImageView.

This could be done by designing a specific cross ROI for this use case.

vallsv:
This PR clean up the `ImageView` in order to be able to reuse few components and provide embedded profile.

Closes #3348 
Closes #3347
Closes #288

- [x] Rework the code to use up-to-date silx concepts (sync axis, block signals, data margins)
- [x] Rework `RadarView` as a reusable component
- [x] Small changes on the `RadarView` (displays data range + active items bounds (scatter, curve, image))
- [x] Split data processing and display
- [x] Provide a cross ROI to display a specific profile selection

Screenshot from 2021-02-23 13-53-49

Changelog: 

- Update `ImageView` with update-to-date silx concepts
    - Result of horizontal/vertical ROI profiles are now displayed embedded in the side plots
    - A h/v profile sum is displayed by default, except when there is ROI profile


silx/gui/plot/ImageView.py
```diff
@@ -656,13 +610,13 @@ def setRadarView(self, radarView):
         :param RadarView radarView: Widget subclassing RadarView to replace
                                     the lower right corner widget.
         """
-        self._radarView.visibleRectDragged.disconnect(self._radarViewCB)
+        if self._radarView is not None:
```
t20100, line 613: radarView can't be `None`, else `self.centralWidget().layout().addWidget(self._radarView, 1, 1)` would fail.

silx/gui/plot/tools/RadarView.py
```diff
@@ -0,0 +1,356 @@
...
+        """Update the content to the current data content"""
+        plot = self.getPlot()
+        if plot is None:
+            return
+        ranges = plot.getDataRange()
+        width = ranges.x[1] - ranges.x[0]
+        height = ranges.y[1] - ranges.y[0]
+        self.setDataRect(ranges.x[0], ranges.y[0], width, height)
+
+        def updateItem(rect, item):
```
t20100, line 340: Could be moved as a private method

silx/gui/plot/tools/RadarView.py
```diff
@@ -0,0 +1,356 @@
...
+    def __setVisibleRectFromPlot(self, plot):
+        """Update radar view visible area.
+
+        Takes care of y coordinate conversion.
+        """
+        xMin, xMax = plot.getXAxis().getLimits()
+        yMin, yMax = plot.getYAxis().getLimits()
+        self.setVisibleRect(xMin, yMin, xMax - xMin, yMax - yMin)
+
+    def getPlot(self):
```
t20100, line 256: Could be renamed `getPlotWidget`

silx/gui/plot/tools/RadarView.py
```diff
@@ -0,0 +1,356 @@
...
+        Returns the connected plot
+        """
+        if self.__plotRef is None:
+            return None
+        plot = self.__plotRef()
+        if plot is None:
+            self.__plotRef = None
+        return plot
+
+    def connectPlot(self, plot):
```
t20100, line 267: Can be combined with `disconnectPlot` to make a `setPlotWidget` method

t20100:
I had 2 issues while testing it (on macos10.12, python3.7, pyqt5.13.2, matplotlib 3.3.4):

- When panning in the radar view, the image is moved but not the integrated profiles. This is not the case with the line profiles, and it gets updated when resizing the window....
- When drawing a hline, vline or cross profile with only a click (and no drag), the profile(s) are not displayed, they only get displayed if the profile line is moved or if the window is resized.
Might be a macos only problem...

vallsv:
I did try with:
- Debian 10, Python 3.7, mpl 3.3.4
- Tested with Qt 5.9 and Qt 5.12,

I don't have such problems.

PyQt 5.13 is not available to me.

vallsv:
As tested together with Thomas, it is working well on his machine with OpenGL.

t20100:
Works fine on debian, let's merge it.

PR #3381, opened 2021-02-25
silx.gui.plot.ImageView: Enhanced using mask for computing profiles

t20100:
This PR follows PR #3380 and  #3369 and addresses https://github.com/silx-kit/silx/pull/3369#issuecomment-784212703:
> The mask is not used for the profile sum, which could be done later (to avoid ocnflict with my branch)

Item mask information is used for both default integrated profiles and line profiles.


vallsv:
LGTM

PR #3385, opened 2021-02-25
silx.gui.utils.glutils: Enhanced `isOpenGLAvailable` function: avoid blinking icon in the app bar on linux

### LINKED ISSUE #3339: isOpenGLAvailableResult create a blink on Debian 10 appbar
I guess as it setup a qt application, this test application is exposed by Debian 10 on the application bar.

There is maybe a way to avoid that.

t20100:
Changing the WindowFlags allows to avoid the blinking icon in the app bar at least on debian10.
It does not prevent it on macos.

closes #3339


vallsv:
Thats' already fine. Thanks

PR #3386, opened 2021-02-26
silx.gui.plot3d: Added `HeightMapData` and `HeightMapRGBA` items

### LINKED ISSUE #3379: Add a plot3d item for colored height map
Idea is to display an image as a height map plus some more information as the color.

Example of visualization:

t20100:
This PR adds 2 items to `plot3d` scene to provide a first display of height maps.
Both take a 2D data array that is used as the height of points based on a 2D regular grid and a second array for the colors:

- `HeightMapData` for a height map with colormapped data
- `HeightMapRGBA` for a height map with RGBA colors.

If both arrays (height and color) have different shapes, one is interpolated to match the other.

It can definitely be improved and provide more visualization modes (e.g., a display as a real surface), but it is already useable as it is.

closes #3379

TODO:
- [x] testing
- [x] doc ~~example~~

### DIFF
--- a/doc/source/modules/gui/plot3d/items.rst
+++ b/doc/source/modules/gui/plot3d/items.rst
@@ -112,6 +112,29 @@ The following classes allows to configure :class:`ScalarField3D` visualization:
              getParameters, setParameters,
              getDisplayValuesBelowMin, setDisplayValuesBelowMin
 
+Height map
+----------
+
+.. currentmodule:: silx.gui.plot3d.items.image
+
+:class:`HeightMapData`
+++++++++++++++++++++++
+
+:class:`HeightMapData` inherits from :class:`.DataItem3D` and also provides its API.
+
+.. autoclass:: HeightMapData
+   :members: getData, setData,
+             getColormappedData, setColormappedData
+
+:class:`HeightMapRGBA`
+++++++++++++++++++++++
+
+:class:`HeightMapRGBA` inherits from :class:`.DataItem3D` and also provides its API.
+
+.. autoclass:: HeightMapRGBA
+   :members: getData, setData,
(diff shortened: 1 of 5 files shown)


t20100:
Sample code:
```python
# coding: utf-8
import sys
import numpy

from silx.gui import qt
from silx.gui.plot3d.SceneWindow import SceneWindow, items


SIZE = 2048


qapp = qt.QApplication([])
window = SceneWindow()
sceneWidget = window.getSceneWidget()


coords = numpy.linspace(0, 1, SIZE)
x, y = numpy.meshgrid(coords, coords)
height = SIZE * numpy.exp(- 11. * ((x - .5) ** 2 + (y - .5) ** 2))

rgb_img = numpy.ones((SIZE, SIZE, 4), dtype=numpy.float32)
rgb_img[:, :, :3] = numpy.random.random(3 * SIZE ** 2).reshape(SIZE, SIZE, 3)
item = items.HeightMapRGBA()
item.setData(height)
item.setColorData(rgb_img)

sceneWidget.addItem(item)

window.show()
sys.excepthook = qt.exceptionHandler
qapp.exec_()
```

t20100:
Added basic tests, entry in documentation, picking support.

Changed interpolation mode in case the height map and the color/data map have different shapes to: use height map shape and generate color/data array of same shape through nearest interpolation to have a consistent behavior for both data and RGB(A) colors.

This is a first implementation, it can optimised if needed (e.g., store the height and/or colors in textures, do not provide coordinates for each point, cache some info/reduce amount of transformed data in picking); and extended (adding linear color/data interpolation, display as a solid surface with some smoothing).

juliagarriga (approved):
Looks great!

PR #3391, opened 2021-03-01
silx.gui.plot: Added some statistic indicators in `PixelIntensitiesHistoAction` action

### LINKED ISSUE #3215: Add statistics indicator to PlotWidget histogram action
Adding display of min/max, mean/stdev, sum values to histogram action ("a la" oxidis).

Feedback from beamline.

t20100:
This PR adds basic stats (min, max, mean, std, sum) at the bottom of the histogram window of PixelIntensitiesHistoAction

closes #3215

### DIFF
--- a/silx/gui/plot/actions/histogram.py
+++ b/silx/gui/plot/actions/histogram.py
@@ -39,6 +39,7 @@ __license__ = "MIT"
 
 import numpy
 import logging
+import typing
 import weakref
 
 from .PlotToolAction import PlotToolAction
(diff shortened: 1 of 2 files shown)


t20100:
Ready for review. To me it could be integrated in 0.15

I have some doubts/questions on the histogram behavior when items disappear from the plot, but I'll make a separate PR for this.

t20100:
histogram behavior fixed in PR #3396

vallsv:
I think it is not a good idea to use docks because you provide flexibility for no real need.

If you have to setup it every time you launch the appli, you don't win much.

Anyway I prefer to use it, you anyway have to handle properly the close of the dock.
If i am not wrong, for now you have to restart the application to display it back.

t20100:
I used dock for convenience in the code, but I agree it should not be moved: I'll check whether to rewrite the window or to change the dock widget properties to disable closing and moving.

This is hidden but in the context menu of the toolbar of a main window, you can open the dock widgets... except that in the `PlotWidget` we overload the context menu so it is not available... Anyway, there is no way to guess that.

vallsv:
It also could be part of the "Options" menu.

t20100:
For the added stats in the histogram window, I'd to keep them always visible, this does not take much space.

vallsv:
I agree, if people complain, you will have time to rework it.

t20100:
Updated to avoid using a dock widget and have a widget for the histogram window that is separated from the action.

t20100:
Added a `setItem` to the `HistogramWidget` and move item update tracking and histogram/stats computation there.
Also added use of `ElidedLabel` and `sizeHint` as in PR #3399

t20100:
Added used of `ElidedLabel` similar to #3399

vallsv:
Looks good to me

PR #3393, opened 2021-03-01
Returns the records together with the logging exception

vallsv:
On fails `TestLogging` returns no the logging records together with the exception.

Changelog: 
- TestLogging returns now the logging records when there is mismatch


t20100:
LGTM

t20100 (approved):

PR #3399, opened 2021-03-02
silx.gui.plot: Enhanced `PositionInfo` information labels layout

### LINKED ISSUE #2997: [plot] Width of the Plot2D widget really big
As you can see in the screenshot the smallest size of the Plot2D is much bigger than the Plot1D.

Cause of the statusbar, and the recent add of the image size.

It would be good to improve that
- Position:
    - X/Y could be displayed together
    - It's an image, we know the max size of this and so the max size it can take
    - Could be displayed in integer
- Use elided text
- On pyfai i have created cells which can grow up according to the size needed
- BTW, what about "Size" instead of "Dims"?

Capture d’écran de 2020-04-08 21-27-19

t20100:
This PR allows `PositionInfo` field to be smaller... yet keep their default size, this makes `Plot2D` default size the same, but it can be shrinked down by removing space between the fields.

To me, it is another way to close #2997

Default and min:
Plot2D

### DIFF
--- a/silx/gui/plot/tools/PositionInfo.py
+++ b/silx/gui/plot/tools/PositionInfo.py
@@ -44,17 +44,17 @@ import numpy
 from ....utils.deprecation import deprecated
 from ... import qt
 from .. import items
+from ...widgets.ElidedLabel import ElidedLabel
 
 
 _logger = logging.getLogger(__name__)
 
 
-class _PositionInfoLabel(qt.QLabel):
+class _PositionInfoLabel(ElidedLabel):
     """QLabel with a default size larger than what is displayed."""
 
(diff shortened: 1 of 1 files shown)


vallsv:
I think it is not a good idea while the label content will enforce the plot size.

Cause moving the move on the data is changing the plot width now.

Silx already provides an `ElidedLabel` with displays `...` if there is not enough space. If it could be used here, it could be better. But maybe toolbars already provides such features.

t20100:
Works nicely with `ElidedLabel`, thanks for the idea!

vallsv:
Looks good, thanks a lot.
I think it's huge improvement for people with small screen or in remote.

PR #3401, opened 2021-03-04
silx.gui.plot.Plotwidget: Added unified `selection` handler compatible with active item management

t20100:
This PR adds a `PlotWidget.selection()` method giving access to an object handling selection in the `PlotWidget` in a way that is compatible with current active item management.

The selection object provides:
- `getCurrentItem` and `sigCurrentItemChanged` to retrieve and track a single current item which is the latest item in the plot to have been "activated".
- `getSelectedItems` which provides the list of all activated items ordered from latest to oldest "activated" item.
- `setCurrentItem`which updates active items accordingly.

The selection object is instantiated on demand so it does not interfere if not used.
This is aimed to help features requiring management of active items of different kinds.

### DIFF
--- a/silx/gui/plot/PlotWidget.py
+++ b/silx/gui/plot/PlotWidget.py
@@ -103,40 +103,39 @@ class _PlotWidgetSelection(qt.QObject):
         assert isinstance(parent, PlotWidget)
         super(_PlotWidgetSelection, self).__init__(parent=parent)
 
-        self.__current = self.__findCurrentItem()
+        # Init history
+        self.__history = [  # Store active items from most recent to oldest
+            item for item in (parent.getActiveCurve(),
+                              parent.getActiveImage(),
+                              parent.getActiveScatter())
+            if item is not None]
+
+        self.__current = self.__mostRecentActiveItem()
 
         parent.sigActiveImageChanged.connect(self._activeImageChanged)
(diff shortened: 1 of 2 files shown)


t20100:
Added `sigSelectedItemsChanged` to track changes in any active item type.

vallsv:
That's nice. I have already setup Flint to use that.

But i was guessing what will happen in this case?

```
set active image a
set active scatter b
set active image None
# -> i expect active item to be b
set active scatter None
# -> i expect active item to be None
```
And
```
set active image a
set active scatter b
set active scatter None
# -> i expect active item to be None
```
I think it is consistent with what we should have if we could take the time to rework the item picking.

What do you think?
What about doing this behaviour?

Anyway, i am fine with the current state i think. I don't think we need to involve too much time here.

t20100:
The implemented behaviour is:
```
set active image a
# -> current item a
set active scatter b
# -> current item b
set active image None
# -> current item b
set active scatter None
# -> current item None
```
and
```
set active image a
# -> current item a
set active scatter b
# -> current item b
set active scatter None
# -> current item a
```

But when calling `plot.selection().setCurrentItem(None)`, all active items (i.e., image, scatter and curve) are set to `None`.

vallsv:
Thanks

PR #3405, opened 2021-03-05
silx.gui.plot: Fixed histogram picking, Enabled `PositionInfo` snapping to histogram

t20100:
This PR:
- Fixes histogram picking and returned indices
- Enables snapping to histogram items in the PositionInfo widget

There is still an issue with the OpenGL backend where the filled part is not taken into account for picking.


t20100:
Implemented picking of "filled" histogram.

This is done in the `Histogram` item since it is not implemented in the OpenGL backend and returns erroneous indices in the matplotlib backend.

It is not correct in terms of structure, but way simpler to implement and more efficient.
A generic picking of filled curves (and histograms) in both backend would be nice to have though.

PR #3411, opened 2021-03-10
Misc: Fixed `silx test` application on Windows

### LINKED ISSUE #3410: Windows fat binary multiprocessing issue
There is an issue running `silx test` with Windows fat binary, I expect it is due to `multiprocessing` being used in `silx.io.h5py_utils` which makes a fork-bomb.
Using `multiprocessing.freeze_support()` should fix it.

t20100:
Avoid fork-bomb when running `silx test` with Windows fat binary.

closes #3410

### DIFF
--- a/silx/__main__.py
+++ b/silx/__main__.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


t20100:
Funny issue with `silx.app.view.test.test_launcher.TestLauncher.testExecuteSilxViewHelp` during `import multiprocessing` which does `import math` at some point, which imports `silx.math` because running `silx/__main__.py` as a script adds `silx/` to `sys.path`.
I fixed it by copying the file to run in a temporary directory.

vallsv:
Could it be done using `-m` instead?

t20100:
It would be definitely better but there's maybe a reason it was not already the case.

t20100:
Looks tests now pass fine.

BTW, when running `silx test` in a Windows fat binary, some tests fails (see https://gitlab.esrf.fr/silx/bob/silx/-/jobs/191819) and I personnally think it is not an issue, but it is no longer a fork bomb.

PR #3412, opened 2021-03-10
Misc.: Reorganized project structure

### LINKED ISSUE #3357: Moving silx source to a src sub-folder
Since the `silx` folder is in the root of the repository, it is loaded first when running something from there.
We have the `bootstrap.py` script to mitigate that, but this is not enough under some circumstances, e.g., when using `subprocess` (see PR #3356 for example).

Keep the source in a `src` sub-directory would avoid that.
`matplotlib` is using this with a `lib` sub-directory (https://github.com/matplotlib/matplotlib/tree/master/lib)

t20100:
This PR moves the `silx` sources in `src/silx` and does the related modifications.

This simplifies (a very little bit in the end) `setup.py` and `bootstrap.py` (and makes `bootstrap.py` working on the macbook I'm using which is broken otherwise).
It would also have simplified things for PR  #3356.

In the end the gain is small, but it was also not complex to change (and I was curious about it).

closes #3357


vallsv:
+1

t20100:
Fixed debian packaging.

t20100:
Ready for review

t20100:
Rebased after port of 0.15 branch, should be ready.

PR #3418, opened 2021-03-15
silx view: Support very large 1D datasets in "Raw" table view

### LINKED ISSUE #3417: silx view segmentation fault when displaying a 1D dataset of 570 millions values
This is a limitation of Qt: The [memory limit of 2GB of QVector](https://doc.qt.io/qt-5/qvector.html#maximum-size-and-out-of-memory-conditions) limits the number of sections in `QHeaderView` and thus the number of number of items in a `TableView`.

Related Qt issues: https://bugreports.qt.io/browse/QTBUG-31215 and https://bugreports.qt.io/browse/QTBUG-18490

t20100:
This PR allows to overcome the limited number of items supported by QTableView by clipping 1D array when it is above a threshold.
This makes sure the application does not seg fault.

I set threshold to 1 million, but we can make it 10 millions.

Clipping is displayed as follow with a tooltip on the `...`:

Screenshot from 2021-03-15 12-43-45

closes #3417

### DIFF
--- a/silx/gui/data/RecordTableView.py
+++ b/silx/gui/data/RecordTableView.py
@@ -150,7 +150,10 @@ class RecordTableModel(qt.QAbstractTableModel):
     :param numpy.ndarray data: A numpy array or a h5py dataset
     """
 
-    MAX_NUMBER_OF_ROWS = 10e6 + 1
+    MAX_NUMBER_OF_ROWS = 10e6
+    """Maximum number of display values of the dataset"""
+
+    _CLIPPING_TOOLTIP = "Dataset is too large: display is clipped"
 
     def __init__(self, parent=None, data=None):
         qt.QAbstractTableModel.__init__(self, parent)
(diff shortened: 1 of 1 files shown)


t20100:
To me it is not necessary, but do you think the threshold needs to be settable?

vallsv:
I don't think so.

If you are about to view billions of rows, you probably don't use the right tool.

It is still nice cause you can see the layout of your data, and it does not segfault. But i dont think we should spend time to tweak the widget.

vallsv (approved):

t20100:
Rework to make `isClipped` private and avoid some duplicated code

t20100:
CI broken because of set-up issues:
```
E: The repository 'https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  Release' no longer has a Release file.
```

t20100:
CI fixed

PR #3423, opened 2021-03-17
silx.gui.qt: Removed support of PySide/PyQt4

t20100:
This PR clean-up code supporting PySide and PyQt4 that's anyway no longer supported.
Clean-up in CI scripts,`run_tests.py`, qtdesigner plugins and `silx.gui.qt`.

Idea is too clean-up code before adding support of PyQt6/PySide6.


t20100:
related to  #3432

PR #3424, opened 2021-03-17
silx: Removed code supporting PyQt4/PySide

t20100:
Merge PR #3423 first!

This PR clean-up code supporting PySide and PyQt4 that's anyway no longer supported.
Clean-up in silx packages

Idea is too clean-up code before adding support of PyQt6/PySide6.


t20100:
related to  #3432

PR #3431, opened 2021-03-22
Draft: Setup the project to use pytest

vallsv:
This PR is a minimal rework of the project to use pytest.

- It reworks `run_tests` to encapsulate pytest runner in the build env with compiled extensions
- It supports most of the specific options which was previously used (`no-gui`, `no-opencl`, `low-mem`...), short options was removed cause of conflict with pytest options
- Package boiler plate was removed
- Tests are autodiscovered
- To run a specific module test filename is used instead of qualified module name

Run all tests:
```
./run_tests.py
./run_tests.py src/silx
``` 

Run a specific module:
```
./run_tests.py src/silx/opencl/test/test_convolution.py
``` 

Changelog: 

- Setup the project to use pytest


kif:
My concern is that some tests are not discovered which can create a false feeling of security.

kif:
well -1000 loc is not bad ... less to maintain.

t20100:
> well -1000 loc is not bad ... less to maintain.

That's the goal.

vallsv:
Here is 2 examples of how it can be used.

- 3c4ef663e14c007a8f091f05706f0a0daa5c62a8 (`qapp`, `qapp_utils` to rework TestCaseQt + parametrization)
- 2663986e04e35b0418ba6bb76aec8c46e9f0f131 (tmp file creation)

But basically if you still use `unittest.TestCase` it is not very convenient.

vallsv:
Got that on AppVeyor ` Environment: PYTHON_DIR=C:\Python38-x64, QT_BINDING=PyQt5, WITH_GL_TEST=True, PIP_OPTIONS=-q --pre` in case you have an idea 
```
__________________________________ test_plot __________________________________
sx = <module 'silx.sx' from 'C:\\projects\\silx-7wjwv\\venv_test\\lib\\site-packages\\silx\\sx\\__init__.py'>
    def test_plot(sx, qapp_utils):
        """Test plot function"""
        y = numpy.random.random(100)
        x = numpy.arange(len(y)) * 0.5
    
        # Nothing
>       plt = sx.plot()
E       AttributeError: module 'silx.sx' has no attribute 'plot'
```

t20100:
`sx.enable_gui` was not called? Or failed because there is no display?

vallsv:
I dont know, cause the tests pass now. Maybe i did a typo somehow.

vallsv:
> My concern is that some tests are not discovered which can create a false feeling of security.

I would say, that's exactly why you have to discover the tests.

No body maintain lists. This have to be maintained close to the tests as metadata.

Actuqlly, you can create a test which is executed when you run the module manually (`unittest` will autodiscover the methods), but not when you run the whole tests (cause the tested classes as fixed, and sometimes the tested methods are also fixed).

This PR fixes few places where tests was not executed, or was disabled without comment.

vallsv:
Notice that i did a trick for GUI widget fixtures.

It sounds like the `yield` share the reference somehow, so the `close` is not working. So i have used a weakref proxy. This pattern looks to work well.

```
@pytest.fixture
def colormapDialog(qapp):
    dialog = ColormapDialog.ColormapDialog()
    dialog.setAttribute(qt.Qt.WA_DeleteOnClose)
    yield weakref.proxy(dialog)
    qapp.processEvents()
    from silx.gui.qt import inspect
    if inspect.isValid(dialog):
        dialog.close()
        qapp.processEvents()
```

t20100:
CI also works on gitlab: https://gitlab.esrf.fr/silx/bob/silx/-/pipelines/49329

Issue #3443, opened 2021-04-28
Memory leak on opengl backend plots


vallsv:
There is a lazy remove on the OpenGL backend in order to clean up OpenGL item in the right thread.

This code is executed during the `paintGL`.

If the plot is never displayed, items are never cleaned up.

But the plot can be updated with new items which will be never displayed.

As result the amount of items increase, until the plot is displayed again.

When using tabs, it's really common to have hidden plots.

----

With Flint, while the plot was on an hidden tab.

The plot is a curve plot with only 3 curves.

```
In [7]: len(w._silxPlot()._backend._glGarbageCollector)
Out[7]: 2190
```

After a display of this tab

```
In [8]: len(w._silxPlot()._backend._glGarbageCollector)
Out[8]: 0
```

----

A backport to 0.14 would be very welcome.

vallsv:
Related to https://gitlab.esrf.fr/bliss/bliss/-/issues/2733

tobiasvos:
We also encountered this issue and I can confirm that it is fixed from silx 0.14.1 onwards. However, on Debian 11 we are stuck on version 0.14.0. Is there a way to make the fix available on our operating system, without resorting to pip? For example by providing an official Debian backport, or an unofficial package for Debian 11 (right now these are only available for Debian 10)?

t20100:
Hi,

To build unofficial Debian/Ubuntu packages, you can use this script: https://github.com/silx-kit/silx/blob/main/build-deb.sh
You can check the CI we use to build the unofficial packages we provide: https://gitlab.esrf.fr/silx/bob/main/-/blob/master/job-templates.yml#L810-865

Otherwise to get an official backport, since we are not much involved in official Debian packaging, it's best to ask a service company that provide this kind of service.

tobiasvos:
Many thanks!

My apologies if this is in the wrong place, but I get the following error when trying to build a Debian package (and even with just `./build-deb.sh --help`):

```
Fatal Python error: init_sys_streams: can't initialize sys standard streams
Python runtime state: core initialized
Traceback (most recent call last):
  File "/data/tobiasv/silx/src/silx/io/__init__.py", line 36, in <module>
  File "/data/tobiasv/silx/src/silx/io/utils.py", line 34, in <module>
  File "/usr/lib/python3.9/logging/__init__.py", line 26, in <module>
  File "/usr/lib/python3.9/traceback.py", line 5, in <module>
[...]

PR #3445, opened 2021-04-29
Fixed leak of OpenGL backend items on branch 0.14

t20100:
Under some conditions (PlotWidget in a not visible tabbed DockWidget + some condition not yet clearly defined...) it happens that `PlotWidget.replot` is called (since `PlotWidget.isVisible()` is True) but `BackendOpenGL.paintGL` is not.
This can leads to new items to be created in the backend (in `PlotWidget.replot`) and never discarded (in `BackendOpenGL.paintGL`)

This PR workarounds this issue by adding the list of backend items to "discard " (which is done lazily in `paintGL`) only the ones that really needs to free some GPU resources (i.e., that has already been displayed).
It has the advantage of only modifying the OpenGL backend.

Related to #3443

### DIFF
--- a/silx/gui/plot/backends/BackendOpenGL.py
+++ b/silx/gui/plot/backends/BackendOpenGL.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2014-2020 European Synchrotron Radiation Facility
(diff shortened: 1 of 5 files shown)


vallsv:
Looks good to me.
Ill check it now.

PR #3447, opened 2021-04-30
silx.gui.plot.PlotWidget: Optimize regular grid visualisation of scatters

t20100:
This PR adds an optimisation of scatter when rendered as regular grid with the OpenGL backend.
In this case, it avoids computing the colormap on the CPU if it is possible (i.e., float data and no alpha data).


t20100:
Rebased to use `pytest`. Ready for review.

PR #3457, opened 2021-05-07
silx.gui.plot.ImageView: Fixed profile window, added `setProfileWindowBehavior` method

### LINKED ISSUE #3454: New line profiles in ImageView removed essential functionality for OASYS users
In the last release of silx, the line profile function of ImageView widget drastically changed behavior. When selecting a line profile (horizontal or vertical), the secondary windows with the 1D profile is no more visible, but the profile appears directly in the ImageView side histograms. 
I also noted that, when using other line profile options, the window with the 1D profile appears with minimal dimensions on the screen, requiring a manual resize to be visible.

This had a massive (and detrimental) impact on the OASYS users, that used that function to be able to do fits and other operations on the 1D profile. It is **very urgent** for the OASYS community the have the functionality restored as it was. 

Screen Shot 2021-05-05 at 11 18 36 AM

Screen Shot 2021-05-05 at 4 45 50 PM

t20100:
This PR:
- restores previous behavior for displaying profile windows (i.e., pop-up)
- adds `ImageView.get|setProfileWindowBehavior` method to change it to embed horizontal, vertical and cross profiles on the side widgets.
- adds `ImageView.getProfileToolBar` and deprecates `ImageView.profile` attribute.

closes  #3454

### DIFF
--- a/src/silx/gui/plot/ImageView.py
+++ b/src/silx/gui/plot/ImageView.py
@@ -420,8 +420,8 @@ class ImageView(PlotWindow):
 
         self._initWidgets(backend)
 
-        self.__profile = None
-        self.setProfileWindowBehavior(self.ProfileWindowBehavior.POPUP)
+        self.__profile = ProfileToolBar(plot=self)
+        self.addToolBar(self.__profile)
 
     def _initWidgets(self, backend):
         """Set-up layout and plots."""
@@ -846,7 +846,7 @@ class ImageViewMainWindow(ImageView):
         menu.addAction(actions.control.YAxisInvertedAction(self, self))
 
         self.__profileMenu = self.menuBar().addMenu('Profile')
-        self.__profileMenu.__updateProfileMenu()
+        self.__updateProfileMenu()
 
         # Connect to ImageView's signal
         self.valueChanged.connect(self._statusBarSlot)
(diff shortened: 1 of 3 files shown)


woutdenolf (approved):
* API looks good.
* I cannot judge what additional things need to be done when swapping toolbars.
* Perhaps you can change this for clarity
```python
class ImageView(PlotWindow):
    def __init__(self, parent=None, backend=None):
        self.__profile = ProfileToolBar(plot=self)
        self.addToolBar(self.__profile)
```
to this
```python
class ImageView(PlotWindow):
    def __init__(self, parent=None, backend=None):
        self.__profile = None
        self.setProfileWindowBehavior('popup')

class ImageViewMainWindow(ImageView):
    def __init__(self, parent=None, backend=None):
           self.__profileMenu = None
    
    def __updateProfileMenu(self):
        if self.__profileMenu is None:
            return
```
Or something like that? Up to you.

woutdenolf:
@t20100 I'm still bothered by this toolbar swapping. Would it be possible to have the behaviour option in `ProfileManager` so that we don't need a `_CustomProfileManager` and `_ProfileToolBar`?

t20100:
Should be fine now: no longer need for `silx.gui.plot.imageview._CustomProfileManager` and `silx.gui.plot.imageview._ProfileToolBar`.

woutdenolf (approved):
Thanks @t20100 it's better:
* no swapping of toolbars
* no extra toolbar and manager classes in ImageView.py
* explicit default behavior

LGTM

PR #3460, opened 2021-05-10
`silx.io.dictdump` extract update modes list to a constant global variable

jpcbertoldo:
This allows users to properly create scripts on top of `silx` without needing to manually copy those constants.


kif:
LGTM ...

PR #3463, opened 2021-05-11
silx.io.spech5: Enhanced robustness for missing positioner values

### LINKED ISSUE #3459: [silx.io.open] fails reading ESRF-BM30 spec files
`silx.io.open` fails reading this Spec file from BM30 at ESRF: [testBM30.zip](https://github.com/silx-kit/silx/files/6442920/testBM30.zip) (unzip it).

It happens because the `#P` fields in the scan header do not fully corresponds to the `#O` fields of the file header. The old `specfilewrapper` is able to ignore this misformat in the Spec file. 

Old `specfilewrapper` works:

```python
from silx.io.specfilewrapper import Specfile
sf = Specfile('./testBM30.spec')
```
`specH5` fails:

```python
from silx.io.spech5 import SpecH5
sf = SpecH5('./testBM30.spec')
```

Would it be possible to have `spech5` ignoring it too ? It should not take too much work to fix this bug.

t20100:
This PR allows to have less values in `#P` than in `#O` as it is the case in #3459.
In such case, it logs a warning and sets the value to `inf`. I choose to do so since when `#P` is completely missing `inf` is used for all values.

Disclaimer: I'm not used to spec at all!

 closes #3459

### DIFF
--- a/src/silx/io/spech5.py
+++ b/src/silx/io/spech5.py
@@ -670,9 +670,14 @@ class PositionersGroup(commonh5.Group, SpecH5Group):
     def __init__(self, parent, scan):
         commonh5.Group.__init__(self, name="positioners", parent=parent,
                                 attrs={"NX_class": to_h5py_utf8("NXcollection")})
+
+        dataset_info = []  # Store list of positioner's (name, is_data, value)
+        is_error = False   # True if error encountered
+
         for motor_name in scan.motor_names:
             safe_motor_name = motor_name.replace("/", "%")
-            if motor_name in scan.labels and scan.data.shape[0] > 0:
+            is_data = motor_name in scan.labels and scan.data.shape[0] > 0
+            if is_data:
                 # return a data column if one has the same label as the motor
                 motor_value = scan.data_column_by_name(motor_name)
             else:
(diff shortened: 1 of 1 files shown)


vasole:
I meant between setting all them to inf (your solution) or to ignore all of them.

I prefer to ignore all of them.

Perhaps @maurov or @woutdenolf have other opinion.

t20100:
OK, I have no opinion, so just let me know!

I set to "inf" because this is what is currently happening when `#P` is completely missing.

t20100:
Updated to discard all motors with a single scalar value when a value is missing.

maurov:
@t20100 @vasole would it be possible to merge this bug fix and have a minor release out? This would help us a lot on the beamline reading Spec files with Larch (see https://github.com/xraypy/xraylarch/issues/281)!

payno:
LGTM

src/silx/io/spech5.py
```diff
@@ -678,10 +682,24 @@ def __init__(self, parent, scan):
...
+                except SfErrColNotFound:
+                    is_error = True
+                    motor_value = float('inf')
+            dataset_info.append((safe_motor_name, motor_value))
+
+        if is_error:  # Filter-out scalar values
+            logger1.warning("Mismatching number of elements in #P and #O: Ignoring")
+            dataset_info = [
+                (name, value) for name, value in dataset_info
+                if not isinstance(value, float)]
```
woutdenolf, line 696: @t20100 When do you have a value that is not a float?

src/silx/io/spech5.py
```diff
@@ -678,10 +682,24 @@ def __init__(self, parent, scan):
...
                 # (may return float("inf") if #P line is missing from scan hdr)
-                motor_value = scan.motor_position_by_name(motor_name)
-            self.add_node(SpecH5NodeDataset(name=safe_motor_name,
-                                            data=motor_value,
-                                            parent=self))
+                try:
+                    motor_value = scan.motor_position_by_name(motor_name)
+                except SfErrColNotFound:
+                    is_error = True
+                    motor_value = float('inf')
```
woutdenolf, line 689: @t20100 Why inf and not nan?

woutdenolf (commented):
Added two questions.

PR #3466, opened 2021-05-17
Implementation of the double-word library

### LINKED ISSUE #3451: OpenCL expanded arithmetic library for statics
This PhD provides expanded arithmetic for `+`, `-`, `*` and `/`:
http://www.theses.fr/2017LYSEN036

In the conclusion of the chapter 2, the number of FLOPS for double-word calculation is given: from 6 to 31(depending on the operation). This is far less than the double-precision/single-precision ratio on the latest generation of Nvidia hardware (64 for Ampere RTX30xx). 

This would allow to calculate a single passe mean/std on GPU which would be much more reliable.

On CPU double-word arithmetic is useless since they have the same performances in single and double precision.

kif:
for compensated arithmetics.
* Still buggy
* Still WIP
* Tests ongoing

close #3451 

Changelog: Implement the doubleword OpenCL library (+tests +statistics update)

### DIFF
--- a/src/silx/opencl/statistics.py
+++ b/src/silx/opencl/statistics.py
@@ -29,13 +29,10 @@
 large data where numpy is not very efficient.
 """
 
-from __future__ import absolute_import, print_function, with_statement, division
-
-
 __author__ = "Jerome Kieffer"
 __license__ = "MIT"
-__date__ = "11/01/2019"
-__copyright__ = "2012-2017, ESRF, Grenoble"
+__date__ = "17/05/2021"
+__copyright__ = "2012-2019, ESRF, Grenoble"
 __contact__ = "jerome.kieffer@esrf.fr"
 
 import logging
@@ -128,7 +125,7 @@ class Statistics(OpenclProcessing):
                                          self.kernel_files,
                                          "-D NIMAGE=%i" % self.size)
         compiler_options = self.get_compiler_options(x87_volatile=True)
-        src = concatenate_cl_kernel(("kahan.cl", "statistics.cl"))
+        src = concatenate_cl_kernel(("doubleword.cl", "statistics.cl"))
         self.reduction_comp = ReductionKernel(self.ctx,
                                               dtype_out=float8,
                                               neutral=zero8,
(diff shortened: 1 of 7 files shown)


kif:
Pierre, since you are interested in this, I made you reviewer :)
That said, after cleaning up the code and fixing the bugs, I performed the profiling and compared with the former versions:
The runtime dropped from 106µs to 55µs (for a 1Mpix image). Surprisingly the number of of FLOP increased from 143 to 190 millions but with many more additions. The number of registers used dropped fro 45 to 38 too.
[former_reduce_kernel_stage1.pdf](https://github.com/silx-kit/silx/files/6501668/former_reduce_kernel_stage1.pdf)
[newer_reduce_kernel_stage1.pdf](https://github.com/silx-kit/silx/files/6501669/newer_reduce_kernel_stage1.pdf)

kif:
Maybe I'll add a full double precision version to validate the performances on different classes of GPU: Quadro vs Tesla.

kif:
For info, the "double" version takes 36µs on the Tesla V100 (vs 55µs). So not worth on this Tesla.
The same test on a Pascal (GT1030) gives:
* Single 13ms, 
* Compensated: 25ms
* Double: 170ms
Figures on the Quadro are likely to be similar to the consumer cards.

pierrepaleo:
This looks very good. I still have to test, though.

Some typos:
  - [ ] `test_doubleword.py`: `Project: Azimuthal integration`
  - [ ] `test_kahan.py:` `Copyright (C) 2015-2032 European Synchrotron Radiation Facility, Grenoble, France`

Questions:
  - The `DOUBLEDOUBLE` option does not seem to be tested, but I believe this is one compiler option away ?
  - Can you explain the tolerance set to `k*EPS**2` where `k` is 2, 3, 5, 6 in some cases ?

kif:
The numerical values can be found in Table 2.2 (page 68) from the PhD thesis referenced:
https://hal.archives-ouvertes.fr/tel-01534090

pierrepaleo (approved):
Looks very good. Tested on `scisoft14` and `linperceval`

Issue #3432, opened 2021-03-22
Add support for PySide6/PyQt6


t20100:
Related links:
- https://www.riverbankcomputing.com/static/Docs/PyQt6/pyqt5_differences.html
- https://doc.qt.io/qtforpython/porting_from2.html
- https://doc.qt.io/qt-6/portingguide.html
- https://www.learnpyqt.com/blog/pyqt6-vs-pyside6/

TODO list:

- [x] Remove code for PyQt4/PySide2 support
- [ ] Handle enums ~~Change enums to use  fully qualified names, e.g., `qt.Qt.ItemIsSelectable` -> `qt.Qt.ItemFlags.ItemIsSelectable`~~
- [ ] Testing/Handle specific cases
- [x] Handle change of naming in PyQt6 only: `exec_` -> `exec` (and `print_` -> `print`)

t20100:
Status of enums changes accross Qt bindings by examples with `ItemFlag`'s `ItemIsSelectable`

It comes in 3 flavors (!), with  `from <>.QtCore import Qt`:
- **Legacy** (current way to access in silx): `Qt.ItemIsSelectable`
- **New** standard: `Qt.ItemFlag.ItemIsSelectable`
-  **pyqt6.0**, mind the **s** in the enum name: `Qt.ItemFlags.ItemIsSelectable`

Status of bindings:

- PySide6: **Legacy** and **New**
- PySide2: **Legacy** and **New**

- PyQt6 >= 6.1: only **New**
- PyQt 6 < 6.1: only **pyqt6.0**

- PyQt5 >= 5.11: **Legacy** and **New**
- PyQt5 < 5.11: only **Legacy**

To limit supported cases, we could:
- do NOT support `PyQt6==6.0` which is an outlier
- drop `PyQt5<5.11`

and switch to the **New** syntax, otherwise we'll have to monkey-patch for `PyQt6==6.0` and `PyQt5<5.11` or make an explicit compatibility Qt enum module (but I would rather not).

vasole:
Since we are already using legacy, I do not see more work associated to adding compatibility with PyQt5 < 5.11 respect to the 100% move to the new syntax.

t20100:
Not sure I understand:

- To support PyQt6 >=6.1 either we have to move to the new syntax or to make a monkey-patching to provide the legacy syntax.
- To support PyQt5 < 5.11 either we keep the current syntax or we make a monkey-patching to provide the new syntax.
[...]

PR #3478, opened 2021-06-03
Fixed SVG icons for Qt6

t20100:
This PR cleans-up SVG from `radialGradient` with radius `r` = 0 and path with a single point.
It should not change the visible aspect but avoids a lot of warnings with Qt6:

```
qt.svg: zoom-original.svg:2:1306: Could not parse node: radialGradient
qt.svg: zoom-original.svg:3: Could not resolve property: #h
```

Related to #3432


vallsv:
Nice to know.
You fixed it manually of there is a tool (maybe the tool we use for SVG optimization have such features ?)

t20100:
I did it manually...

PR #3480, opened 2021-06-07
silx.gui: Qt4 support cleanup

t20100:
More clean-up following removal of PyQt4/PySide support (PR #3424): 
- Remove unused files, left-over after 3424
- Change use of `QPalette.ColorRole` enum obsolete values. Those values where already [obsolete in Qt4](https://doc.qt.io/archives/qt-4.8/qpalette.html#ColorRole-enum), and a removed in Qt6 (related to #3432).


vallsv:
Do you want to merge it?

Else there is still things which could still be cleaned up like a special file model for Qt4 (provided for windows, `SafeFileSystemModel`), and probably special methods used for section of qheader view.

Maybe check
- `QTVERSION < "5.0"`
- `hasattr` on the GUI package
- `setResizeMode`/`setSectionResizeMode`
-

t20100:
> Do you want to merge it?

Yes, when possible, I prefer to make small and quickly merged PR than bigger ones that are pending forever.

t20100:
> Maybe check `QTVERSION < "5.0"`

You're right, we have a `QTVERSION < '4.0.0'` ;)

vallsv:
eheh :-D

PR #3481, opened 2021-06-07
silx.gui: Prepare for Qt6 support

t20100:
This PR prepares for Qt6 support by changing implementation for obsolete/inconsistent features that does not require Qt6 specific code:

- `QAction.setMenu` is obsolete in Qt6, so when it is used, `QAction` needs to be replaced with `QMenu` or `QToolButton`.
- `QColor.getRgbF` is not exactly consistent across Qt5 and Qt6 (precision, rounding change?), so use `QColor.getRgb` instead

Related to #3432

### DIFF
--- a/src/silx/app/view/Viewer.py
+++ b/src/silx/app/view/Viewer.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 # /*##########################################################################
-# Copyright (C) 2016-2020 European Synchrotron Radiation Facility
+# Copyright (C) 2016-2021 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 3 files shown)


vallsv:
DId you found a "how to convert from Qt5 to Qt6"? The menu/action sounds weird, i would be happy to have an explanation.

t20100:
This is not in "how to convert from Qt5 to Qt6", but `QAction.setMenu` is obsolete in Qt6: https://doc.qt.io/qt-6/qaction-obsolete.html
Here is an explanation: https://doc.qt.io/qt-6/widgets-changes-qt6.html#qaction-qactiongroup

Anyway, it is not available from `PySide6`.

PR #3484, opened 2021-06-08
silx.gui: Prepare for Qt6 support (follow-up)

t20100:
This PR prepares for Qt6 support by changing implementation for obsolete/inconsistent features that does not require Qt6 specific code, completes PR #3481:

`QAction.setMenu` is obsolete in Qt6, so when it is used, `QAction` needs to be replaced with `QMenu` or `QToolButton`.

This 2 fixes were not highlighted by tests.

Related to #3432


t20100:
Rebased to use `pytest`. Ready for review.

PR #3486, opened 2021-06-09
silx.gui: Renamed `exec_` methods to `exec` for compatibility with Qt6

t20100:
This makes PyQt5/PySide2 provide PyQt6/PySide6 API regarding `exec_` -> `exec` renaming:
- It adds necessary `exec` methods to Qt objects providing `exec_` for PyQt5 and PySide2
- It makes use of `exec` throughout silx.
- `exec_` methods in silx are renamed `exec` to be consistent but `exec_` compatibility wrapper are also provided (without deprecation warnings).
- For `FitWidget`, since API allows to provide dialogs (`FitWidget.associateConfigDialog`), here it works with either `exec` (tried first) or `exec_` (as now).

This avoids warnings with PySide6.

Related to #3432


t20100:
Rebased to use`pytest`. Ready for review.

t20100:
Test fixed, ready for review

vallsv:
Sounds good.

COuld you add a test to check that `exec_` still exists in Qt5?

t20100:
- Added some tests for `exec_` and `exec` method availability.
- Also reworked the wrapper: both using lambda function or assigning `<class>.exec = <class>.exec_` failed.

Ready for review.

Note: As opposed to previous attempts, the `exec` method is only available through `silx.gui.qt` and not directly from the binding.

payno:
LGTM

PR #3487, opened 2021-06-11
Supports RGB and RGBA in ImageView

vallsv:
This PR release constraints on `setImage` to allow to set RGB and RGBA images.

Few other places were fixed, like statusbar display, but not much.

Changelog: 
- Added support of RGB and RGBA images to ImageView

### DIFF
--- a/src/silx/gui/plot/ImageView.py
+++ b/src/silx/gui/plot/ImageView.py
@@ -756,7 +756,7 @@ class ImageView(PlotWindow):
 
         data = numpy.array(image, order='C', copy=copy)
         assert data.size != 0
-        assert data.ndim in (2, 3)
+        assert data.ndim == 2 or (data.ndim == 3 and data.shape[2] in (3, 4))
 
         self.addImage(data,
                       legend=self._imageLegend,
--- a/src/silx/gui/plot/ImageView.py
+++ b/src/silx/gui/plot/ImageView.py
@@ -345,7 +345,7 @@ class ImageView(PlotWindow):
     """Minimum size in pixels of the image area."""
 
     # Qt signals
-    valueChanged = qt.Signal(float, float, float)
+    valueChanged = qt.Signal(float, float, object)
     """Signals that the data value under the cursor has changed.
 
     It provides: row, column, data value.
@@ -483,7 +483,7 @@ class ImageView(PlotWindow):
             activeImage = self.getActiveImage()
             if activeImage is not None:
                 data = activeImage.getData(copy=False)
-                height, width = data.shape
+                height, width = data.shape[0:2]
 
                 # Get corresponding coordinate in image
                 origin = activeImage.getOrigin()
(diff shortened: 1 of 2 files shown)


vallsv:
I think it is fine to back port it to 0.15, but it's up to you.

src/silx/gui/plot/ImageView.py
```diff
@@ -821,15 +821,29 @@ def __updateProfileMenu(self):
         self.__profileMenu.addAction(profile.lineAction)
         self.__profileMenu.addAction(profile.clearAction)
 
+    def _formatValueToString(self, value):
+        try:
+            if isinstance(value, numpy.ndarray):
+                if len(value) == 4:
+                    return "RGBA: %i %i %i %i" % (value[0], value[1], value[2], value[3])
```
t20100, line 828: RGBA can also be float in [0, 1], this formatting works for both int and float:

```suggestion
                    return "RGBA: %.3g, %.3g, %.3g, %.3g" % (value[0], value[1], value[2], value[3])
```

I also added `,` separator, to me it makes it easier to parse visually.

src/silx/gui/plot/ImageView.py
```diff
@@ -821,15 +821,29 @@ def __updateProfileMenu(self):
...
+                elif len(value) == 3:
+                    return "RGB: %i %i %i" % (value[0], value[1], value[2])
```
t20100, line 830: Same for RGB:

```suggestion
                    return "RGB: %.3g, %.3g, %.3g" % (value[0], value[1], value[2])
```

src/silx/gui/plot/ImageView.py
```diff
@@ -843,8 +857,8 @@ def setProfileWindowBehavior(self, behavior: str):
     @docstring(ImageView)
     def setImage(self, image, *args, **kwargs):
         if hasattr(image, 'dtype') and hasattr(image, 'shape'):
-            assert len(image.shape) == 2
-            height, width = image.shape
+            assert image.ndim in (2, 3)
```
t20100, line 860: Maybe catch early RGB(A) without the right number of channels:
```suggestion
            assert image.ndim == 2 or (image.ndim == 3 and image.shape[2] in (3, 4))
```

src/silx/gui/plot/ImageView.py
```diff
@@ -756,7 +756,7 @@ def setImage(self, image, origin=(0, 0), scale=(1., 1.),
 
         data = numpy.array(image, order='C', copy=copy)
         assert data.size != 0
-        assert len(data.shape) == 2
+        assert data.ndim in (2, 3)
```
t20100, line 759: Maybe catch early RGB(A) without the right number of channels:
```suggestion
        assert data.ndim == 2 or (data.ndim == 3 and data.shape[2] in (3, 4))
```

t20100 (commented):
I made a few suggestions to support RGBA as float in [0,1] and catch malformed RGB images early.
Else LGTM, I don't see what this PR can break in current API except for the `valueChanged` signal when you load a RGB image, but that's a new feature so it's fine.

PR #3488, opened 2021-06-11
Show/hide ImageView side histogram

vallsv:
This MR provide an API to show/hide the side histogram of the `ImageView`.

It is added as a tool in the toolbar, and a new entry in the (image view main window)

Screenshot from 2021-06-11 13-27-30

Changelog: 
- Added an action to show/hide the side histogram of the `ImageView`
- Few clean up around `ImageView` module
- Update the `imageview` example with a `--live` option
- Added `resetzoom` to `ImageView.setImage`


vallsv:
I have replaced `layout.addWidget(self._histoVPlot.getWidgetHandle(), 0, 1)` by `layout.addWidget(self._histoVPlot, 0, 1)`.
I think it only makes sens for the main widget.

t20100:
> I have replaced layout.addWidget(self._histoVPlot.getWidgetHandle(), 0, 1) by layout.addWidget(self._histoVPlot, 0, 1).

Sounds good. I think it was there to really make sure there is no additional margins in `PlotWidget` which would make the histograms not aligned with the image plot, but it should not be needed.

I'd move `ShowSideHistogramsAction` outside of `silx.gui.plot.actions` since all actions in `silx.gui.plot.actions` are compatible with `PlotWidget`.
I would put it as a private class in `ImageView.py` as it is only used there, else eventually in a dedicated module in `silx.gui.plot.actions`.

vallsv:
This can be reviewed.

I did some cleanup around the `ImageView`.

The main weird change is the `setStartusBar(None)` that i have removed. This was creating blinking cause just before it is set to something. So i dont know if it have to be displayed or not.

The other change is the use of `setImage(resetzoom)`, instead of `reset`, for consistency with the other plot API. I don't like much the `resetzoom`, but i think it is better to be share the same API. `reset` is still working.

Issue #3470, opened 2021-05-19
Backports for 0.15.2


t20100:
Lists of PR to backports from master branch for a v0.15.2:

- #3469
- #3463 

Please add any PR to backport here or flag issues/PRs with milestone 0.15.2

t20100:
Also backport #3476

t20100:
Also fixing issue #3489

t20100:
@kif, do you want to backport #3466 and #3472 in this micro release?

t20100:
Also backport #3490

maurov:
@t20100 when do you plan to release `0.15.2`?

t20100:
> @t20100 when do you plan to release 0.15.2?

ASAP: if there is no last minute issue to fix, tomorrow sounds too optimistic and next week sounds realistic.

t20100:
Closing: All backports were done.

PR #3495, opened 2021-06-18
0.15.2 doubleword

kif:
I have rebased all commits relative to the doubleword library ...


t20100:
related to #3470

t20100:
Could you also add an entry in the changelog for 0.15.2?
It's been filled as we go for 0.15.2, so that would make it complete: https://github.com/silx-kit/silx/blob/0.15/CHANGELOG.rst

PR #3503, opened 2021-07-08
silx.gui.plot: Added `ImageDataAggregated` item

t20100:
This PR adds a plot item that allows to aggregate image data before display.
By default it displays the data image as `ImageData` but the `setAggregationMode` allows to reduce the data image before display depending on the current level of details displayed.
For now the aggregation is done in a fast/approximate manner, a second more accurate mode could also be added.

Which "aggregation modes" are needed? For now it provides "none" (no aggregation), "min" (nanmin), "max" (nanmax), "mean" (nanmean)

See https://gitlab.esrf.fr/bliss/bliss/-/merge_requests/3838

### DIFF
--- a/src/silx/gui/plot/items/__init__.py
+++ b/src/silx/gui/plot/items/__init__.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017-2020 European Synchrotron Radiation Facility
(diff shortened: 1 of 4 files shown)


t20100:
This is ready for review, we can add more mode later on.
Question about naming: should I rename `min`, `max`, `mean` to `nanmin`, `nanmax`, `nanmean` or not?

PR #3504, opened 2021-07-09
silx.gui.plot: Fixed OpenGL issue with out-of-bouds axes limits

### LINKED ISSUE #3501: Assert error during resetZoom
Hi,

Here is a problem found in beamline logs.

I did not try to reproduce it.

Maybe already fixed.

Screenshot from 2021-07-07 14-43-07

t20100:
This PR removes an `assert` about axis limits in the `PlotWidget` OpenGL backend and replaces it with a change of the range in case of issue.
The code validating the limits range is the same as the `Axis` one and has been modified to ensure the limits are within the supported range even if the range is changed from the API or through `resetZoom` (which was not the case before).

It also adds some basic tests of out-of-bound limits.

closes #3501

### DIFF
--- a/src/silx/gui/plot/_utils/__init__.py
+++ b/src/silx/gui/plot/_utils/__init__.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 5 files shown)


vallsv:
Thx a lot

PR #3505, opened 2021-07-16
string conversion for key hdf5

rnwatanabe:
Changelog: string conversion for key hdf5 by casting the variable key to string in dicttoh5 function of dictdump.py file.


kif:
LGTM

PR #3507, opened 2021-07-22
silx.io.spech5: Make SpecH5 resilient to malformed #G1 and #G3 header lines

### LINKED ISSUE #3506: error reading spec file with silx.io.open()
@maurov I think this is separate from #3459, but it might be related.

We're trying to use `silx.io.open` to read Spec files in xraylarch -- we're very happy to be able to support ESRF Bliss files!.  But a user posted a spec file:
   https://millenia.cars.aps.anl.gov/gsecars/data/Larch/datafiles/break_silx0152.spec

that breaks in a way I don't completely understand:

```
Python 3.8.10 | packaged by conda-forge | (default, May 10 2021, 22:58:09)
[Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import silx
>>> silx.version
'0.15.2'
>>> from silx.io import open as silx_open
>>> f = silx_open('break_silx0152.spec')
  File "<stdin>", line 1, in <module>
  File "/Users/Newville/xraylarch/lib/python3.8/site-packages/silx/io/utils.py", line 590, in open
    h5_file = _open_local_file(url.file_path())
  File "/Users/Newville/xraylarch/lib/python3.8/site-packages/silx/io/utils.py", line 504, in _open_local_file
    return spech5.SpecH5(filename)
  File "/Users/Newville/xraylarch/lib/python3.8/site-packages/silx/io/spech5.py", line 574, in __init__
    scan_group = ScanGroup(scan_key, parent=self, scan=scan)
  File "/Users/Newville/xraylarch/lib/python3.8/site-packages/silx/io/spech5.py", line 629, in __init__
    if _unit_cell_in_scan(scan) or _ub_matrix_in_scan(scan):
  File "/Users/Newville/xraylarch/lib/python3.8/site-packages/silx/io/spech5.py", line 309, in _unit_cell_in_scan
    return numpy.any(_parse_unit_cell(scan.scan_header_dict["G1"]))
  File "/Users/Newville/xraylarch/lib/python3.8/site-packages/silx/io/spech5.py", line 298, in _parse_unit_cell
    return numpy.array(list(map(float, header_line.split()))[0:6]).reshape((1, 6))
ValueError: cannot reshape array of size 1 into shape (1,6)
>>>
```

The lines like `#G1 0` seem to be the problem. I have to admit, I don't know much about the origin of the file, or what those lines are meant to encode.  Is `#G1 0` always invalid?

Is there a way we could work around this? 

Thanks!

t20100:
This PR makes `SpecH5` resilient to `#G1` and `#G3` having less values than expected.

closes #3506

### DIFF
--- a/src/silx/io/spech5.py
+++ b/src/silx/io/spech5.py
@@ -298,6 +298,7 @@ def _ub_matrix_in_scan(scan):
     try:
         ub_matrix = _parse_UB_matrix(header_line)
     except ValueError:
+        logger1.warning("Malformed G3 header line")
         return False
     return numpy.any(ub_matrix)
 
@@ -327,6 +328,7 @@ def _unit_cell_in_scan(scan):
     try:
         unit_cell = _parse_unit_cell(header_line)
     except ValueError:
+        logger1.warning("Malformed G1 header line")
         return False
     return numpy.any(unit_cell)
 
--- a/src/silx/io/spech5.py
+++ b/src/silx/io/spech5.py
@@ -278,8 +278,12 @@ def _parse_UB_matrix(header_line):
 
     :param str header_line: G3 header line
     :return: UB matrix
+    :raises ValueError: For malformed UB matrix header line
     """
-    return numpy.array(list(map(float, header_line.split()))).reshape((1, 3, 3))
+    values = list(map(float, header_line.split()))  # Can raise ValueError
+    if len(values) < 9:
+        raise ValueError("Not enough values in UB matrix")
+    return numpy.array(values).reshape((1, 3, 3))
 
 
 def _ub_matrix_in_scan(scan):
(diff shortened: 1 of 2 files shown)


t20100:
attn @maurov  @newville this fixes reading the malformed spec file.

src/silx/io/spech5.py
```diff
@@ -278,8 +278,12 @@ def _parse_UB_matrix(header_line):
 
     :param str header_line: G3 header line
     :return: UB matrix
+    :raises ValueError: For malformed UB matrix header line
     """
-    return numpy.array(list(map(float, header_line.split()))).reshape((1, 3, 3))
+    values = list(map(float, header_line.split()))  # Can raise ValueError
+    if len(values) < 9:
```
payno, line 284: well I guess this should be if len(values) != 9 no ?

src/silx/io/spech5.py
```diff
@@ -288,13 +292,28 @@ def _ub_matrix_in_scan(scan):
...
 def _parse_unit_cell(header_line):
-    return numpy.array(list(map(float, header_line.split()))[0:6]).reshape((1, 6))
+    """Parse G1 header line and return unit cell
+
+    :param str header_line: G1 header line
+    :return: unit cell
+    :raises ValueError: For malformed unit cell header line
+    """
+    values = list(map(float, header_line.split()[0:6]))  # can raise ValueError
+    if len(values) < 6:
```
payno, line 314: same here this should be != 6

src/silx/io/spech5.py
payno: but could bring to some "errors that might be ignore" as some comments...

t20100: I agree it would be best, but the previous behavior was to provide something if `len(values) >= 9` and I have no clue how this is used/abused (and if it was on purpose or not), so I prefer to keep backward compatibility.

src/silx/io/spech5.py
t20100: same to keep backward compatibility

newville:
@t20100 @payno thanks, but I'm slightly confused.   If I'm reading the spec (https://github.com/silx-kit/silx/issues/3506#issuecomment-884682378) correctly  `#G2` contains lattice parameters and could be expected to have 6 numerical values.   But I don't see why the description of `#G1`   ('geometry mode, sector, etc') would lead one to expect a fixed number of numerical values.   Are you mixing up `#G1` and `#G2` here?

t20100:
@newville you look to be right, there is an issue with reading `#G1` instead of `#G2` for the unit cell...
I'll prepare a PR to fix it, but I'm really not speaking SPECfile format so no clue if it is really the right thing to do (but it seems so).

PR #3509, opened 2021-07-23
silx.io.spech5: Fixed unit cell header line interpretation

t20100:
This PR fixes the header line that is interpreted as `unit_cell`, it was `#G1` and is now `#G2`.
This looked like a bug, but if someone with good knowledge of SPECfile can check, that'd great.

Related to:
- https://github.com/silx-kit/silx/pull/3507#issuecomment-885576858
- https://github.com/silx-kit/silx/issues/3506#issuecomment-884682378

### DIFF
--- a/src/silx/io/spech5.py
+++ b/src/silx/io/spech5.py
@@ -304,9 +304,9 @@ def _ub_matrix_in_scan(scan):
 
 
 def _parse_unit_cell(header_line):
-    """Parse G1 header line and return unit cell
+    """Parse G2 header line and return unit cell
 
-    :param str header_line: G1 header line
+    :param str header_line: G2 header line
     :return: unit cell
     :raises ValueError: For malformed unit cell header line
     """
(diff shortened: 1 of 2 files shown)


t20100:
attn @newville @maurov

vallsv:
Any feed back here?

maurov:
Hello, according to the [Spec manual](https://certif.com/downloads/css_docs/spec_man.pdf), `#G2` is the field for the lattice constants. Personally, I would merge like this.

t20100:
In another documentation, the [scans one](https://certif.com/spec_help/scans.html), you can find:
> #G1
>     Specifies parameters associated with the diffractometer geometry. The parameters are those from the U[] array, which include lattice constants and orientation reflections. 

We've checked what the macro does, and it saves the U[] array as `#G1` and there is no `#G2`.

BTW, support of unit_cell as `#G1` was introduced in #722 and it is like that since 4 and a half years.

maurov:
@t20100 I have no strong opinion on this, as I have never used `#G*` parameters myself. Maybe @vasole knows some beamlines using them.

t20100:
> I have no strong opinion on this,
@maurov me neither!

Documentation is inconsistent, but the code of the SPEC macro that BCU just checked is clear: it is using `#G1`, so I'm closing this.

newville:
@t20100 @maurov Thanks!

PR #3510, opened 2021-07-26
Darft: [gui][ComapreImages] fix forcing vmin and vmax values

payno:
vmin and vmax values are forced when we want to display both image A and image B.
But this should not be the case for the other modes.
This commit attempt to fix this issue.

Changelog: 
CompareImages: fix colormap: avoid forcing vmin and vmax when not in HRIZONTAL_LINE or VERTICAL_LINE mode

### DIFF
--- a/src/silx/gui/plot/CompareImages.py
+++ b/src/silx/gui/plot/CompareImages.py
@@ -726,11 +726,18 @@ class CompareImages(qt.QMainWindow):
         """
         if self.__visualizationMode == mode:
             return
+        previousMode = self.getVisualizationMode()
         self.__visualizationMode = mode
         mode = self.getVisualizationMode()
         self.__vline.setVisible(mode == VisualizationMode.VERTICAL_LINE)
         self.__hline.setVisible(mode == VisualizationMode.HORIZONTAL_LINE)
-        self.__updateData()
+        visModeRawDisplay = (VisualizationMode.ONLY_A,
+                             VisualizationMode.ONLY_B,
+                             VisualizationMode.VERTICAL_LINE,
+                             VisualizationMode.HORIZONTAL_LINE)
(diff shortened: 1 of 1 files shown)


vallsv:
Sounds good to me.

But i would prefer a
```
if ..:
    do sothing
```
Instead of 
```
if ...:
    return
```
As the method as a logic inside i think it is better not to break the method in the middle.

Another solution is to refactor the end of the method into a dedicated method which only have normalize the min/max.

What do you think?

payno:
> Sounds good to me.
> 
> But i would prefer a
> 
> ```
> if ..:
>     do sothing
> ```
> 
> Instead of
> 
> ```
> if ...:
>     return
> ```
> 
> As the method as a logic inside i think it is better not to break the method in the middle.
> 
> Another solution is to refactor the end of the method into a dedicated method which only have to normalize the min/max (if needed).
> 
> What do you think?

In fact I was wondering if I shouldn't do it this way. So fine with me.

vallsv:
It think all this thing should be reworked taking care of the `Colormap` object.
It was probably designed before colormap was a persistent object.

What you really want to have here is to have an autoscape/min/max taking care of the concatenation of the 2 images at the same time ; anyway only A or B is displayed (you can't compare images if the level change in between during a switch).

If you switch from one image to another you should use the same contrast (anyway it was using autoscale or not), which is not the case anymore.

If you have time i would propose to update the following way:

- Create a dedicated colormap for this use case (raw image display), and setup the min/max only when images are updated
- When you switch the mode to `ONLY_A ONLY_B HORIZONTAL_LINE VERTICAL_LINE` you can set this colormap

The autoscale will still not work (which is the real difficult task), but you will be able to tune vmin/vmax manually (will you do not edit the images).

vallsv (approved):

PR #3511, opened 2021-07-26
`silx.io.convert`: Added `write_to_h5` supporting HDF5 file as input 

### LINKED ISSUE #3468: Seamless support of HDF5 input file in write_to_h5
Initially raised by @maurov: https://github.com/silx-kit/silx/issues/3354#issuecomment-768893312 but ignored so far:
One cannot use `write_to_h5` with a HDF5 input file since it raises errors when input is a HDF5 file

To workaround issue #1378, PR #1385 raised exceptions when the `infile` is a `h5py` object, e.g.: https://github.com/silx-kit/silx/blob/master/src/silx/io/convert.py#L300

It would be good to support it so one can open whatever file with `silx.io.open` and use seamlessly `write_to_h5`

t20100:
This PR allows to provide a HDF5 file as input to `write_to_h5`.
This was not possible before because `commonh5.visititem` had a different API than it's `h5py` equivalent and was disabled as part of #1385.

To achieve this, this PR adds a `silx.io.utils.visitall` function that provides a visitor of all items including links that works for both `commonh5` and `h5py`. This function is used by `write_to_h5` implementation.

It also provides some basic tests.

Current limitation: External links in input file are ignored (with a warning).

closes #3468

### DIFF
--- a/src/silx/io/commonh5.py
+++ b/src/silx/io/commonh5.py
@@ -24,9 +24,6 @@
 """
 This module contains generic objects, emulating *h5py* groups, datasets and
 files. They are used in :mod:`spech5` and :mod:`fabioh5`.
-
-.. note:: This module has a dependency on the `h5py <http://www.h5py.org/>`_
-    library, which is not a mandatory dependency for `silx`.
 """
 import collections
 try:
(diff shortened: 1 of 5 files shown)


t20100:
@maurov if you have time to give it a try with more complex files than the tests, that'd be great.

Also, is external link support needed? I haven't done it for now.

kif:
LGTM

PR #3514, opened 2021-07-29
silx.gui.plot.actions.histogram: Added controls to change histogram nbins and range

### LINKED ISSUE #3216: Allow user to set nbins of PlotWidget histogram action
It would be convenient to be able to change the number of bins and range in the histogram displayed from the histogram action.
Range can be changed by zooming into the plot, but having a range slider + text control can be easier to use.

Feedback from beamline.

t20100:
This PR adds 2 widgets to set the number of bins and the range of the histogram in the plot from the histogram action.

The number of bins is guessed by default (and displayed as placeholder text) and when the user inputs it, it is kept when data is updated (e.g., browsing a stack of image in silx view). To fallback to bin number guessing, the user has to delete the input text.
Updates are done on "Enter".

The absolute range is NOT persisted when changing data, instead the "relative range" is persisted (i.e., the histogram bounds in the data are changed but the positions of the thumb on the range slider are not).
I see it hard to keep the absolute range since it can fall outside of the data range when changing data.
At the same time, it might be useful to keep some similar range from one image to the next in a stack.

Screenshot from 2021-07-29 14-11-35

closes #3216

### DIFF
--- a/examples/plotUpdateImageFromThread.py
+++ b/examples/plotUpdateImageFromThread.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


t20100:
Still need to handle live data update and editing nbins at the same time

t20100:
Ready for review.

- Data update fixed, also handle corner cases (all data is the same and range is empty).
- The number of bins is not updated while typing to avoid computing the histogram all the time, it is only updated when pressing Enter of loosing focus. To show that the displayed value is not the currently used one, it is displayed in *italic* until effectively validated.

PR #3515, opened 2021-07-29
Misc.: Fixed invalid value, deprecation and test warnings

t20100:
Fixes the following warnings raised in CI on Windows when running the tests:
```
venv_test\lib\site-packages\silx\utils\testutils.py:116
  C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\utils\testutils.py:116: PytestCollectionWarning: cannot collect test class 'TestLogging' because it has a __init__ constructor (from: gui/plot/tools/test/testTools.py)
    class TestLogging(logging.Handler):

gui/plot/test/testPlotWidget.py::TestPlotCurve_Gl::testPlotCurveInfinite
  C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\gui\plot\backends\glutils\GLPlotCurve.py:1147: RuntimeWarning: invalid value encountered in subtract
    self.yData = (yData - self.offset[1]).astype(numpy.float32)

gui/plot/test/testPlotWidget.py::TestPlotCurveLog_Gl::testPlotCurveErrorLogXY
  C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\gui\plot\backends\glutils\GLPlotCurve.py:993: RuntimeWarning: invalid value encountered in subtract
    xCoords[3:endXError:4] = self._xData - xErrorMinus

gui/plot/test/testPlotWidget.py::TestPlotCurveLog_Gl::testPlotCurveErrorLogXY
  C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\gui\plot\backends\glutils\GLPlotCurve.py:1019: RuntimeWarning: invalid value encountered in subtract
    yCoords[endXError+3::4] = self._yData - yErrorMinus

gui/plot3d/test/testSceneWidgetPicking.py::TestSceneWidgetPicking::testPickMeshWithIndices
  C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\gui\_glutils\utils.py:103: RuntimeWarning: invalid value encountered in true_divide
    barycentric = subVolumes[intersect] / volume[intersect].reshape(-1, 1)

gui/plot3d/test/testSceneWidgetPicking.py::TestSceneWidgetPicking::testPickMeshWithIndices
  C:\projects\silx-7wjwv\venv_test\lib\site-packages\silx\gui\_glutils\utils.py:108: RuntimeWarning: invalid value encountered in true_divide
    t = volAlpha / volume[intersect]  # segment parameter of intersected triangles

[...]

PR #3516, opened 2021-07-30
`silx.utils.testutils`: Renamed `TestLogging` and `test_logging` to `LoggingValidator` and `validate_logging`

t20100:
This renaming is to avoid pytest test gathering possible issue with functions/classes that starts with `test`/`Test` but are not tests.

It follows https://github.com/silx-kit/silx/pull/3515#issuecomment-889282524

### DIFF
--- a/src/silx/app/test/test_convert.py
+++ b/src/silx/app/test/test_convert.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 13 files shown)


t20100:
I put some backward compatibility wrappers just in case even if I really doubt anyone is using this. :shrug:

vallsv:
SOunds good to lme

PR #3522, opened 2021-08-24
`silx.math.colormap`: Improved `cmap` error messages

### LINKED ISSUE #3520: silx.math.colormap.cmap: Obfuscated error message
The `cmap` function does not support `None` and the error message is not obvious. Either we support None and return an empty array or provide a clearer error message.
Also only `uint8` and `float32` are supported for `colors` and the same error message is triggered. In this case we could either support all types or convert (with a warning) to supported data type

First reported here: https://gitlab.esrf.fr/ui/daiquiri/-/issues/19

t20100:
This PR improves error messages raised when `data` or `colors` are not supported.
It also cast `colors` to supported types (with a warning) whenever possible.

closes  #3520

### DIFF
--- a/src/silx/math/colormap.pyx
+++ b/src/silx/math/colormap.pyx
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2018-2020 European Synchrotron Radiation Facility
+# Copyright (c) 2018-2021 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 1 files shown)


t20100:
CI fails for another issue: See #3523

t20100:
Pin-pointing `h5py` to fix tests on CI. See #3523

t20100:
Ready for review

t20100:
Rebased after PR #3526 merge to fix CI issues

Issue #3523, opened 2021-08-24
Test failure with h5py v3.4.0


t20100:
CI fails with latest version of h5py (v3.4.0):

```
___________________ TestH5pyUtils.test_modes_multi_process ____________________
self = <silx.io.test.test_h5py_utils.TestH5pyUtils testMethod=test_modes_multi_process>
    def wrapper(self):
        for _ in self._subtests():
            with self.subTest(**self._subtest_options):
>               test(self)
venv_test\lib\site-packages\silx\io\test\test_h5py_utils.py:123: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv_test\lib\site-packages\silx\io\test\test_h5py_utils.py:217: in test_modes_multi_process
    with self._open_context(filename, mode="r") as f:
C:\Python38-x64\lib\contextlib.py:113: in __enter__
    return next(self.gen)
venv_test\lib\site-packages\silx\io\test\test_h5py_utils.py:163: in _open_context
    with _open_context(filename, **kw) as f:
C:\Python38-x64\lib\contextlib.py:113: in __enter__
    return next(self.gen)
venv_test\lib\site-packages\silx\io\test\test_h5py_utils.py:78: in _open_context
    with h5py_utils.File(filename, **kw) as f:
venv_test\lib\site-packages\silx\io\h5py_utils.py:231: in __init__
    super().__init__(filename, mode=mode, swmr=swmr, libver=libver, **kwargs)
venv_test\lib\site-packages\h5py\_hl\files.py:455: in __init__
    fid = make_fid(name, mode, userblock_size,
venv_test\lib\site-packages\h5py\_hl\files.py:199: in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
h5py\_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py\_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
>   ???
E   OSError: [Errno 0] Unable to open file (unable to lock file, errno = 0, error message = 'No error', Win32 GetLastError() = 33)
h5py\h5f.pyx:100: OSError
[...]

PR #3526, opened 2021-09-01
Misc: Fixed tests

### LINKED ISSUE #3524: Test fails with matplotlib and PySide2 on macos
There is an issue in `test_sx.py` which crashes and it looks to be related with `matplotlib` v3.5.0b1

~~PR #3522 pin-points matplotlib, this will need to be removed once fixed.~~

t20100:
This PR should fix CI by updating a pytest fixture and removes a warning by fixing a typo.

closes #3524


t20100:
Also pin-point `h5py` until #3523 is fixed.

pierrepaleo:
Looks good

PR #3527, opened 2021-09-03
Misc.: Fixed `run_test.py --qt-binding` option

t20100:
This PR makes sure the chosen Qt binding is loaded by importing it as early as possible, else this can cause issues if multipe Qt bindings are available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,18 +21,18 @@ jobs:
     strategy:
       matrix:
         include:
-          - name-suffix: "PyQt5 sdist"
+          - name-suffix: "PySide2 sdist"
             os: ubuntu-20.04
             python-version: 3.6
             BUILD_COMMAND: sdist
-            QT_BINDING: PyQt5
-            RUN_TESTS_OPTIONS: --qt-binding=PyQt5 --no-opengl
-          - name-suffix: "PySide2 wheel"
+            QT_BINDING: PySide2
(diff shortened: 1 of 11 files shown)


t20100:
completes PR #3486

Related to #3432

t20100:
Issue with CI on macos with matplotlib v3.5.0b1 (see #3524), pin-pointing matplotlib on macos for now.

t20100:
-  Fixed `run_tests.py` on Windows
- Switch CI tests: PyQt5 on macos and PySide2 on Linux (There are still issues to solve with PySide2 on macos:  #3524)
- Updated tests and some code with circular reference to make tests pass with PySide2.

vallsv:
Are you sure about `_set_qt_binding` inside `pytest_addoption`?

`pytest_addoption` is supposed to be called to register available option, not to setup them, right?

t20100:
Yes, maybe using `pytest_configure` is indeed best.
Anyway the need is to import the Qt binding before the tests look-up, else PyQt5 will be imported during test look-up.

vallsv:
Let me know if you want to change it, else ill merge it as it is.

t20100:
Yes I test it and if it does the job I update the PR, else I let you know.

t20100:
`_set_qt_binding` is already called from `pytest_configure` ;)

vallsv:
Looks good. Sorry.

PR #3529, opened 2021-09-03
silx view: Keep curve legend selection with changing dimensions

### LINKED ISSUE #3518: silx view: slider resets legend selection
When moving the sliders in the NXdata view below, the legend selection gets reset (e.g. in this case I wanted to uncheck "subtracted" but it gets re-checked every time I move the sliders):

image

t20100:
For NxData with nD dataset displayed as curves, this PR keeps curves setting (visibility, Y axis) when changing the selected dimensions.
The active is curve is NOT preserved and reset what's in NxData `signal`.

closes #3518

### DIFF
--- a/src/silx/gui/data/NXdataWidgets.py
+++ b/src/silx/gui/data/NXdataWidgets.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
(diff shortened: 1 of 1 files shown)


woutdenolf (approved):
I checked with an example. LGTM

PR #3532, opened 2021-09-23
silx.gui.plot.actions.fit: FitAction: Updated behaviour of fitted item auto update

### LINKED ISSUE #3531: Line profile/fit action ease of use
Line profile does not select active curve, then fit action has nothing to fit by default, thus adding an extra step of selection of the curve.
This can be avoided by setting the active curve in the profile window or by changing the fit action to use the first curve in the plot (or maybe better, if there is only one curve, just use it for fitting).

t20100:
This PR proposes to update the behaviour of the plot's `FitAction` "fitted item auto update mode":
- Current this modes tracks the active curve.
- This PR propose that when there is no active curve, if there is only one curve or one histogram in the plot, this item is used as the fitted item.

To me it is fine to change this since it is an "auto mode", and it sounds better to add this selection fallback.

closes #3531

### DIFF
--- a/src/silx/gui/plot/actions/fit.py
+++ b/src/silx/gui/plot/actions/fit.py
@@ -39,7 +39,7 @@ __license__ = "MIT"
 __date__ = "10/10/2018"
 
 import logging
-
+import weakref
 import numpy
 
 from .PlotToolAction import PlotToolAction
@@ -64,10 +64,11 @@ def _getUniqueCurveOrHistogram(plot):
     if curve is not None:
         return curve
 
-    histograms = [item for item in plot.getItems()
-                  if isinstance(item, items.Histogram) and item.isVisible()]
-    curves = [item for item in plot.getItems()
-              if isinstance(item, items.Curve) and item.isVisible()]
+    visibleItems = [item for item in plot.getItems() if item.isVisible()]
+    histograms = [item for item in visibleItems
+                  if isinstance(item, items.Histogram)]
+    curves = [item for item in visibleItems
+              if isinstance(item, items.Curve)]
 
     if len(histograms) == 1 and len(curves) == 0:
         return histograms[0]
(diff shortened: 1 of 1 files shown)


t20100:
Closing as the complete implementation is more complex: it needs to track both the content of the plot and the currently selected item, which makes it not easy to follow.

t20100:
Reopened this PR with a more involved fitted item selection machinery.
In "fitted item auto update mode", the content of the plot is monitored and fitted item is updated according to `_getUniqueCurveOrHistogram` result.

I had an issue with disconnecting a signal at Python exit during the call to `__plotDeleted` if used through `weakref.ref(plotWidget, callback=self.__plotDeleted)`, here:
https://github.com/silx-kit/silx/blob/5a7855161483fe11f61183b2810cca8c68b1a3b1/src/silx/gui/plot/actions/fit.py#L164
Error:
```
QObject::disconnect: No such signal QObject::sigItemChanged(PyQt_PyObject)
Exception ignored in: <bound method _FitItemSelector.__plotDeleted of <silx.gui.plot.actions.fit._FitItemSelector object at 0x7f0d048753a8>>
  File "silx/gui/plot/actions/fit.py", line 150, in __plotDeleted
    self.__setCurrentItem(None)
  File "silx/gui/plot/actions/fit.py", line 164, in __setCurrentItem
    previousItem.sigItemChanged.disconnect(self.__itemUpdated)
TypeError: disconnect() failed between 'sigItemChanged' and 'unislot'
```
To avoid this I used [weakref.finalize](https://docs.python.org/3/library/weakref.html#weakref.finalize) which has a `atexit` property.

t20100:
This avoids useless curve selection in the profile window, but also more generally when only one curve or histogram is present in the plot.
Once a second curve is present, the FitWidget request a selection of one of the curves.

t20100:
I'll try to change the use of `weakref.finalize` with `sys.is_finalizing`, should be simpler.

t20100:
ready for review

PR #3535, opened 2021-09-29
Provide with the compute capabilities of new GPUs

kif:
Support for latest GPU from Nvidia

Changelog: 
- describe compute capabilities for Ampere generation GPU from Nvidia


pierrepaleo:
Looks good. So, the GA102 chip have `256` flops/core.

PR #3536, opened 2021-10-01
Add aggregation mode to imageview

vallsv:
This MR adds aggregation mode to `ImageView`

This was designed as a patch of `ImageView` to avoid side effects on other plots.

Screenshot from 2021-10-01 18-18-40

Changelog: 
- Added aggregation mode action to `ImageView`


src/silx/gui/plot/ImageView.py
```diff
@@ -338,6 +339,83 @@ def _actionTriggered(self, checked=False):
...
+    def __init__(self, parent):
+        qt.QWidgetAction.__init__(self, parent)
+
+        toolButton = qt.QToolButton(parent)
+
+        filterAction = qt.QAction(self)
+        filterAction.setText("No filter")
+        filterAction.setCheckable(True)
+        filterAction.setChecked(True)
+        filterAction.setProperty("aggregation", None)
```
t20100, line 356: What about using the same type every where?
```suggestion
        filterAction.setProperty("aggregation", items.ImageDataAggregated.Aggregation.NONE)
```

src/silx/gui/plot/ImageView.py
```diff
@@ -338,6 +339,83 @@ def _actionTriggered(self, checked=False):
...
+        """Set an Aggregated enum from ImageDataAggregated"""
+        for a in self.__densityGroup.actions():
+            if a.property("aggregation") is mode:
+                a.setChecked(True)
+
+    def getAggregationMode(self):
+        """Returns an Aggregated enum from ImageDataAggregated"""
+        densityAction = self.__densityGroup.checkedAction()
+        if densityAction is None:
+            return None
```
t20100, line 415:
```suggestion
            return items.ImageDataAggregated.Aggregation.NONE
```

src/silx/gui/plot/ImageView.py
```diff
@@ -338,6 +339,83 @@ def _actionTriggered(self, checked=False):
...
+
+        self.setDefaultWidget(toolButton)
+
+    def _aggregationModeChanged(self):
+        self.sigAggregationModeChanged.emit()
+
+    def setAggregationMode(self, mode):
+        """Set an Aggregated enum from ImageDataAggregated"""
+        for a in self.__densityGroup.actions():
+            if a.property("aggregation") is mode:
```
t20100, line 408: As it is setAggregationMode(items.ImageDataAggregated.Aggregation.NONE) has no effect.

src/silx/gui/plot/ImageView.py
```diff
@@ -810,11 +912,41 @@ def setImage(self, image, origin=(0, 0), scale=(1., 1.),
...
 
-        self.addImage(data,
-                      legend=self._imageLegend,
-                      origin=origin, scale=scale,
-                      colormap=self.getColormap(),
-                      resetzoom=False)
+        aggregation = self.getAggregationModeAction().getAggregationMode()
+        if data.ndim != 2 and aggregation is not None:
+            # RGB/A with aggregation is not supported
+            aggregation = None
```
t20100, line 918:
```suggestion
            aggregation = items.ImageDataAggregated.Aggregation.NONE
```

src/silx/gui/plot/ImageView.py
```diff
@@ -810,11 +912,41 @@ def setImage(self, image, origin=(0, 0), scale=(1., 1.),
...
+
+        if aggregation is None:
```
t20100, line 920:
```suggestion
        if aggregation is items.ImageDataAggregated.Aggregation.NONE:
```

src/silx/gui/plot/test/testImageView.py
```diff
@@ -165,3 +165,30 @@ def testRGBAImage(self):
...
+        self.qWait(100)
+
+    def testImageAggregationModeBackToNormalMode(self):
+        """Test setImage"""
+        image = numpy.arange(100).reshape(10, 10)
+        self.plot.setImage(image, reset=True)
+        self.qWait(100)
+        self.plot.getAggregationModeAction().setAggregationMode(items.ImageDataAggregated.Aggregation.MAX)
+        self.qWait(100)
+        self.plot.getAggregationModeAction().setAggregationMode(None)
```
t20100, line 184:
```suggestion
        self.plot.getAggregationModeAction().setAggregationMode(items.ImageDataAggregated.Aggregation.NONE)
```

t20100 (approved):
It would be simpler and neater to always use an `ImageDataAggregated` for data images... but I think the way you did it is safer since it keeps the previous behavior until the user changes the aggregation mode.

vallsv:
I have designed such way to touch as less as possible of stuffs. But I don't think what you think about will work. Because of intensity and RGB data, which uses 2 different items. But i can be wrong.

A cleaner way is IMO to handle items only (and remove the use of `addImage`) inside the `ImageView.setImage`. The use of the object API (instead of the functional API) in views makes sense. But i am fine with the actual implementation anyway.

t20100:
I am also fine with the actual implementation, it is safer.
And without changing the behavior, it would be complicated to change the way it is done.

LGTM.

PR #3540, opened 2021-10-21
Switch python version used for pyside2 test

t20100:
`pywin32` v302 wheel for python3.6 on Windows seems to have an issue, this PR just exchange the version of python used for pyqt5/pyside2 tests.

PR #3542, opened 2021-11-03
Fixes PySide6/Qt6 support

t20100:
~~Merge PR #3540 first!~~

This PR fixes a couple of Qt6/PySide6 compatibility issues spotted while testing `silx view`:
- Adapt to Qt6 API changes
- PySide6 bug workaround with private methods used as signals
- Avoid to raise exception with scatter profiles expecting a regular grid while a grid cannot be guessed.

related to #3432


t20100:
CI fails after tests pass (see #3553)

payno (approved):

PR #3547, opened 2021-11-04
silx view: Updated About dialog

t20100:
This PR removes `h5py` and `fabio` from advertised optional dependencies.

It also remove `import resource` from `run_tests.py` since it is not used.


t20100:
CI fails after tests pass (see #3553)

payno (approved):

PR #3549, opened 2021-11-05
silx view: Fixed settings when switching between PyQt5/PySide6

t20100:
When changing from PyQt5 to PySide6 (or the other way around), the settings are not correctly recovered and the backend is set to `None`... This PR ignores the value in this case.

Related to #3542

### DIFF
--- a/src/silx/app/view/ApplicationContext.py
+++ b/src/silx/app/view/ApplicationContext.py
@@ -74,8 +74,9 @@ class ApplicationContext(DataViewHooks):
         plotImageYAxisOrientation = settings.value("plot-image.y-axis-orientation", "")
         settings.endGroup()
 
-        if plotBackend != "":
-            silx.config.DEFAULT_PLOT_BACKEND = plotBackend
+        # Use matplotlib backend by default
+        silx.config.DEFAULT_PLOT_BACKEND = \
+            "opengl" if plotBackend == "opengl" else "matplotlib"
         if plotImageYAxisOrientation != "":
             silx.config.DEFAULT_PLOT_IMAGE_Y_AXIS_ORIENTATION = plotImageYAxisOrientation
 


payno:
Not sure but wouldn't it make more sense to have also something like (a bit upper in the source code):

``` python
  plotBackend = settings.value("plot.backend", None)
  plotImageYAxisOrientation = settings.value("plot-image.y-axis-orientation", None)
```

instead of 

``` python
    plotBackend = settings.value("plot.backend", "")
    plotImageYAxisOrientation = settings.value("plot-image.y-axis-orientation", "")
```

vallsv:
Be careful that settings are stored in the client and are typed.
It's not because you request `None` that you will not retrieve an empty string.

It would be interesting to check the type of the value here, maybe the returned type is really not what you think, like a `b""`

t20100:
Looking back into this, the issue is when saving the default backend which has 2 values (`('matplotlib', 'opengl')`) into the settings. In this case `PySide` and `PyQt5` are not storing it the same way since it is not a basic type.

There is errors like `QVariant::load: unknown user type with name PyQt_PyObject.` or `QVariant::load: unknown user type with name PySide::PyObjectWrapper.`.

vallsv:
Did you try to store a list? I guess it's also a structure supported by the settings?

payno (approved):

t20100:
Using `list` indeed works, but the code in the `Options` menu -> `Plot rendering backend` is anyway not expecting a string, and displays:
`Plot rendering backend: ('matplotlib', 'opengl')`
 And both submenu displays:
- `matplotlib (applied after application restart)`
- `opengl (applied after application restart)`

So this PR avoids this.

An alternative is to set `silx.config.DEFAULT_PLOT_BACKEND = 'matplotlib`` by default rather than a list.

vallsv:
I think the idea here was to support both. I can't tell if it is important or not.

The problem with PySide/PyQt is probably because in case of Python type they have to mangle the data, and they have there own incompatible way to do it. But if a list of string is a raw type supported by `QSettings` you should not have the problem.

The GUI issue is another story, this can be fixed to properly display both cases anyway.

t20100:
> I think the idea here was to support both, list or string. I can't tell if it is important or not.
Internally in `silx` it was the idea, but for `silx view`, I am not sure it makes much sense.

The default value: `('matplotlib', 'opengl')` is basically equivalent to `'matplotlib'` since the OpenGL backend cannot run without `matplotlib` as it is.
The only case the list of backends has an interest for me is for: `('opengl', 'matplotlib')` so it provides a fallback in case there is an issue with the OpenGL backend like when running through ssh.

PR #3562, opened 2021-11-19
Add silx-view.gif to documentation

payno:
# TODO
- [x] add it as a 'small' image in index.html
- [x] add it larget in the silx view doc index page
- [x] add a link on the README.rst (should be active once in master only)

# Notes
I am not a big fan of the layout but with rst hard to do something more 'fancy'.

# Extra information

## screenshot of the generated documentation:
Screenshot from 2021-11-19 14-05-01
Screenshot from 2021-11-19 14-04-47

## raw videos

if needs to be reused at one point

https://user-images.githubusercontent.com/12907796/142628021-b554f620-ede9-438f-85c0-34953d7345d7.mp4

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,9 @@
 silx toolkit
 ============
 
+.. |silxView| image:: http://www.silx.org/doc/silx/img/silx-view-v1-0.gif
+   :height: 480px
+
 The purpose of the *silx* project is to provide a collection of Python packages to support the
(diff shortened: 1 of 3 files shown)


t20100:
The GIF is maybe not needed to be stored in the repo?

It could be stored on silx.org with a version/date in the file name to keep the older ones when updating, else as an image uploaded to github (as you did for the video).

t20100 (approved):

PR #3572, opened 2021-12-06
HDF5 table view bug for virtual datasets in the same file

### LINKED ISSUE #3571: silx view: exception for virtual source
When clicking on a virtual dataset I get the exception `<class 'AttributeError'> 'str' object has no attribute 'pop'`

```
ERROR:silx.gui.qt._qt:<class 'AttributeError'> 'str' object has no attribute 'pop'   File "/usr/lib/python3/dist-packages/silx/gui/data/Hdf5TableView.py", line 273, in data
    value = value(self.__obj)
  File "/usr/lib/python3/dist-packages/silx/gui/data/Hdf5TableView.py", line 457, in _first_source
    firstExtSource.pop(0)
```

woutdenolf:
Closes #3571

image

The first virtual source is
```
/data/id21/inhouse/eduardo/out/fitted_EP_701_1_859_12158.h5::/2.1/blisspymca.1/results/parameters/Al_K
```

Alternatively this could also be

```
.::/2.1/blisspymca.1/results/parameters/Al_K
```

I prefer the first (absolute path) because it might not be clear that a "dot" refers to the current file and not to the current directory.

### DIFF
--- a/src/silx/gui/data/Hdf5TableView.py
+++ b/src/silx/gui/data/Hdf5TableView.py
@@ -450,8 +450,9 @@ class Hdf5TableModel(HierarchicalTableView.HierarchicalTableModel):
                     return firstExtSource
 
                 if firstExtSource[0] == ".":
-                    firstExtSource.pop(0)
-                return os.path.join(os.path.dirname(filename), firstExtSource)
+                    return filename + firstExtSource[1:]
+                else:
+                    return os.path.join(os.path.dirname(filename), firstExtSource)
 
             self.__data.addHeaderRow(headerLabel="External sources")
             self.__data.addHeaderValueRow("Type", extType)


vasole:
I would use:

`os.path.abspath(filename) + firstExtSource[1:]`

and

`return os.path.join(os.path.dirname(os.path.abspath(filename)), firstExtSource)`

if you want absolute paths everywhere.

payno:
> I would use:
> 
> `os.path.abspath(filename) + firstExtSource[1:]`
> 
> and
> 
> `return os.path.join(os.path.dirname(os.path.abspath(filename)), firstExtSource)`
> 
> if you want absolute paths everywhere.

this is a good question. Do we want absolute path everywhere. If we go for it I guess we should add somewhere information regarding if the virtual dataset contains some relative / absolute path. Because this is very convenient to have this information (even if for now it is here only partially because only the first virtual source is checked)

woutdenolf:
@vasole `filename` is an already absolute path and it is the file that is open (i.e. the one containing the link, not the destination file). Then `firstExtSource` is the URI of the destination which is either `/abs/path/to/file.h5::/dataset/in/file` or `.::/dataset/in/file` where the "dot" refers to `filename`.

woutdenolf:
I missed one case

* `./path/to/file.h5::/dataset/in/file`

woutdenolf:
I missed one case

* `./path/to/file.h5::/dataset/in/file`

t20100:
@woutdenolf does that cause trouble with `silx view`? Need a bug fix release?

woutdenolf:
the displayed URL is wrong but it doesn't affect anything else

woutdenolf:
I'll make a bug fix as soon as I have time.

t20100:
OK, great! I opened issue #3716 for it for next release, so we don't forget it.

PR #3573, opened 2021-12-08
CI: Updated build with setuptools's embedded distutils

t20100:
Trying to build with upcoming `setuptools` changes related to [PEP 632](https://www.python.org/dev/peps/pep-0632/) and checking `numpy.distutils` compatibility.


t20100:
#3581 superseeds this PR

Issue #3578, opened 2021-12-20
Checksum downloaded data from ExternalRessources


kif:
It looks like some test-unreliability in pyFAI are linked to corrupted test-data, or data which got modified on the disk (for a reason or another).
I suggest to calculate the checksum of downloaded data and store it in the ExternalResources.all_data as a dict with the filename as key.
When re-requested, such data should be checksummed to ensure integrity of the served file.

PR #3580, opened 2021-12-21
store the checksum of downloaded data

kif:
Prevent any inplace modification/corruption

Changelog: Store checksum of downloaded data to prevent unexpected failure of tests due to corrupted data.

This commonly occurs in pyFAI when one calibrates test-data, the poni-file gets modified and this leads to failing tests later on.
The checksum function is SHA256 but it can easily be changed (defined at a single place). Maybe Pierre has an opinion on this ?


kif:
close #3578

pierrepaleo:
Looks good. `sha256` should be more than safe for this use case. The CI fails are unrelated to this PR.

PR #3582, opened 2022-01-05
Enhanced `peak_search` documentation

### LINKED ISSUE #3579: [silx.math.fit] peak_search returns peak indices as float
Not sure if that is intended.

The [doc of the peak_search function](http://www.silx.org/doc/silx/latest/modules/math/fit/peaksearch.html#silx.math.fit.peaks.peak_search) says it will return *peak indices*. 

I would therefore expect them to be `int` to be reused directly like `data[peak_index]`.

t20100:
This PR clarifiy the returned type of peak indices (`float`, not `int` as one can expect for indices).

closes #3579

### DIFF
--- a/src/silx/math/fit/peaks.pyx
+++ b/src/silx/math/fit/peaks.pyx
@@ -65,6 +65,7 @@ def peak_search(y, fwhm, sensitivity=3.5,
         if ``relevance_info`` is ``False``.
         Else, sequence of  ``(peak_index, peak_relevance)`` tuples (one tuple
         per peak).
+        WARNING: Peak indices are returned as float64.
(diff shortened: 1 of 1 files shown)


loichuder (approved):
Thanks !

vasole:
No time to provide feedback...

I would have said ``it returns the closest index as a float. To get an actual index the use of int(returned_value + 0.5) is recommended.``

t20100:
From the look at the code I had, even if it is floats, values are always integers so there is no need to take care of rounding.
If I missed something, a PR is welcomed!

vasole:
It is for the case we decide to return the weighted average of three channels. Now it is correct.

t20100:
OK, then we can update the documentation when this is changed.

PR #3583, opened 2022-01-06
Fixed build by NOT using setuptools>=60.0.0

t20100:
This PR constraints `setuptools` to versions < 60.0.0 in `pyproject.toml`, `requirements.txt` and CI to avoid compatibility issues between recent `setuptools` version and `numpy.distutils`, due to `setuptools` now using it's own fork of `distutils` rather than the one from Python.

This is one step towards fixing CI:
- It also pin-point `PySide6` to version 6.2.1 in CI to avoid failures.
- It avoids `ci/info_platform.py` to import multiple Qt libs by running Qt checks in subprocesses.

### DIFF
--- a/ci/info_platform.py
+++ b/ci/info_platform.py
@@ -65,6 +65,7 @@ else:
 
 
 for binding_name in ("PyQt5", "PySide2", "PySide6"):
+    # Check Qt version in subprocess to avoid issues with importing multiple Qt bindins
     cmd = [
         sys.executable,
         "-c",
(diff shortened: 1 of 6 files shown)


t20100:
Back to previous state: everything passes except PySide2 on Windows with a seg fault when the test process ends.
Ready for review

ci/info_platform.py
```diff
@@ -62,28 +63,16 @@
...
-    import PySide6.QtCore
-    have_qt_binding = True
-    print("Qt (from PySide6): %s" % PySide6.QtCore.qVersion())
-except ImportError:
-    pass
-
-if not have_qt_binding:
-    print("No Qt binding")
+for binding_name in ("PyQt5", "PySide2", "PySide6"):
+    cmd = [
```
t20100, line 68:
```suggestion
    # Check Qt version in subprocess to avoid issues with importing multiple Qt bindins
    cmd = [
```

Issue #3574, opened 2021-12-08
Remove use of `distutils` from `setup.py`


t20100:
`distutils` is depracated and will be removed in Python 3.12 [PEP 632](https://www.python.org/dev/peps/pep-0632/), we should get rid of it.

t20100:
After PR #3585, remaining `distutils` usage will be: 

- `from distutils.command.clean import clean`: Inherited to remove `.c`, `.cpp` files generated by Cython 
- `from distutils.command.build import build`: Inherited to pass options for `force_cython` and `use_openmp`. Those options are used by `build_ext` and not `build`, so they could be moved to `build_ext` instead (which comes from `setuptools`) and/or provided as environment variable (I'm no fan of env. var. but passing options through pip's `--global-option` is cumbersome).

PR #3585, opened 2022-01-06
Build: `setup.py` cleanup

t20100:
This PR removes some now useless code (IMO) from `setup.py`.

A lot of changes in `setuptools` are to be expected (including #3574), so silx's `setup.py` will have to be updated.

This PR is preparing this by doing some clean-up first:
- removes python2 compatibility code
- removes lazy imports and dry-run: nowadays with wheels/conda and  `pyproject.toml` the main ways of installing packages do not need this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,27 +45,10 @@ logger = logging.getLogger("silx.setup")
 
 from distutils.command.clean import clean as Clean
 from distutils.command.build import build as _build
-try:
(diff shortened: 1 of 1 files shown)


payno:
Looks good to me. I was also wondering if we shoudn't get rid of the "sphinx buildDocCommand" redefinition ?

PR #3590, opened 2022-01-13
silx.gui.plot.PlotWidget: Fixed matplotlib backend replot failure under specific conditions

### LINKED ISSUE #3589: BackendMatplotlibQt has been deleted
HI.

Here is my first tests withs Qt 5.12 / Flint / silx 1.0.0

I have notices this problem sometimes. Reproducible in a very specific context (in CI and on my local computer).
I dont think I will be able to simplify the test to reproduce it.

I think i can't to much on my side.

Here it's sounds like a common asynchronous issue.
Could it be possible to protect `__deferredReplot` on your side? Or do you have another idea.

```
_________________________________________________________________________ TestScatterPlot.test_display_nothing __________________________________________________________________________

self = <tests.qt.flint.widgets.test_scatter_plot.TestScatterPlot testMethod=test_display_nothing>

    def tearDown(self):
        self.qapp.processEvents()
    
        if len(self.__class__._exceptions) > 0:
            messages = "\n".join(self.__class__._exceptions)
>           raise AssertionError("Exception occured in Qt thread:\n" + messages)
E           AssertionError: Exception occured in Qt thread:
E           Traceback (most recent call last):
E             File "/nobackup/lvalls1/valls/Software/miniconda3/envs/blissenv2/lib/python3.7/site-packages/silx/gui/plot/backends/BackendMatplotlib.py", line 1368, in __deferredReplot
E               self.replot()
E             File "/nobackup/lvalls1/valls/Software/miniconda3/envs/blissenv2/lib/python3.7/site-packages/silx/gui/plot/backends/BackendMatplotlib.py", line 1533, in replot
E               self.draw()
E             File "/nobackup/lvalls1/valls/Software/miniconda3/envs/blissenv2/lib/python3.7/site-packages/silx/gui/plot/backends/BackendMatplotlib.py", line 1486, in draw
E               FigureCanvasQTAgg.draw(self)
E             File "/nobackup/lvalls1/valls/Software/miniconda3/envs/blissenv2/lib/python3.7/site-packages/matplotlib/backends/backend_agg.py", line 439, in draw
E               super().draw()
E             File "/nobackup/lvalls1/valls/Software/miniconda3/envs/blissenv2/lib/python3.7/site-packages/matplotlib/backends/backend_qt.py", line 424, in draw
E               self.update()
E           RuntimeError: wrapped C/C++ object of type BackendMatplotlibQt has been deleted

/nobackup/lvalls1/valls/Software/miniconda3/envs/blissenv2/lib/python3.7/site-packages/silx/gui/utils/testutils.py:185: AssertionError
```

t20100:
This PR adds a test in `replot` method to make sure the underlying matplotlib backend is valid before reploting.
This can occur because of the async execution of `replot` through `__deferredReplot`.

@vallsv let me know whether it fixes your issue or not?

closes #3589

### DIFF
--- a/src/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/src/silx/gui/plot/backends/BackendMatplotlib.py
@@ -64,6 +64,7 @@ from . import BackendBase
 from .. import items
 from .._utils import FLOAT32_MINPOS
 from .._utils.dtime_ticklayout import calcTicks, bestFormatString, timestamp
+from ...qt import inspect as qt_inspect
 
 _PATCH_LINESTYLE = {
     "-": 'solid',
@@ -1516,6 +1517,10 @@ class BackendMatplotlibQt(FigureCanvasQTAgg, BackendMatplotlib):
         self._drawOverlays()
 
     def replot(self):
+        if not qt_inspect.isValid(self):
+            _logger.info("replot requested but widget no longer exists")
+            return
+
         with self._plot._paintContext():
             BackendMatplotlib._replot(self)
 


vallsv:
The ref became invalidated inside the draw method of the `backend_agg`. Only the `backend_qt` can check that the ref is invalidated

See the sequence
```
agg1 True
agg2 True
agg3 True
agg4 False
qt False
```


`Matplotlib backend_agg`
```
    def draw(self):
        from silx.gui.qt import inspect as qt_inspect; print("agg1", qt_inspect.isValid(self))
        # docstring inherited
        self.renderer = self.get_renderer(cleared=True)
        # Acquire a lock on the shared font cache.
        from silx.gui.qt import inspect as qt_inspect; print("agg2", qt_inspect.isValid(self))
        with RendererAgg.lock, \
             (self.toolbar._wait_cursor_for_draw_cm() if self.toolbar
              else nullcontext()):
            from silx.gui.qt import inspect as qt_inspect; print("agg3", qt_inspect.isValid(self))
            self.figure.draw(self.renderer)
            # A GUI class may be need to update a window using this draw, so
            # don't forget to call the superclass.
            from silx.gui.qt import inspect as qt_inspect; print("agg4", qt_inspect.isValid(self))
            super().draw()
```

Maybe that's too difficult to fix for nothing.

t20100:
Not sure what we can do if widget gets released while inside matplotlib code...

payno:
I agree that trying to go further might come with something over-complicated and no real improvement. This is more qt binding issue here that we can' do much about it.

PR #3593, opened 2022-01-31
`silx.gui.PlotWidget` OpenGL backend: Make text label background less transparent

t20100:
Change alpha from 0.5 to 0.75


vallsv (approved):

vallsv:
Thanks!

PR #3596, opened 2022-02-07
silx.gui.plot.PlotWidget: Fixed dashed curve rendering with the OpenGL backend

### LINKED ISSUE #3595: Some line style are not properly displayed in OpenGL
No idea if it's a regression.

I don't expect a perfect compatibility, but here it's really unreadable.

### OpenGL
image

### Matplotlib
image

t20100:
This PR reworks the OpenGL curve rendering to improve dashed curve display.
Beforehand dashes were correctly rendered only for aspect ratio Y/X = 1.

This PR checks if the aspect ratio of the plot changed since previous rendering and if the change is above 5%, it re-compute the distances along the curve used to render the dashes.
The 5% tolerance is to avoid continuous re-computation while resizing the plot.

- [x] It still need a bit of testing, especially checking if there is any issue with high/low DPI screen support.

closes #3595

### DIFF
--- a/src/silx/gui/plot/backends/BackendOpenGL.py
+++ b/src/silx/gui/plot/backends/BackendOpenGL.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2014-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2014-2022 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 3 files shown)


t20100:
ready for review

vallsv:
Thx. Nice job.

`-.` was implemented in OpenGL?

vallsv (approved):

t20100:
> -. was implemented in OpenGL?

Yes

PR #3597, opened 2022-02-08
silx.gui.plot.PlotWidget: Fixed time axis with values outside of supported range ]0, 10000[ years

### LINKED ISSUE #3570: [plot] ValueError: year 0 is out of range
Hi all,

We encountered an error for a `PlotWidget` with a datetime x-axis. The issue occurs when you zoom out too much. Sometimes this can happen when the mouse wheel is rotated too quickly.

A `ValueError` occurs when a year outside the range [1, 9999] is passed to this `datetime.datetime()` constructor: https://github.com/silx-kit/silx/blob/master/src/silx/gui/plot/_utils/dtime_ticklayout.py#L163.

This happens when on line 363 `niceSpacing` exceeds `dVal`. For example: if the lower limit of the x-axis is 1500 years, `niceSpacing` can be 1000 but it cannot become 2000 years. If the lower limit is 400 years, `niceSpacing` can be 200 but it cannot become 500 years. Such a condition occurs more rapidly if the plot widget is small (small `nTicks`).

Perhaps a similar solution could be used as for issue #3192.

I have included the traceback. I am using silx version 0.9.0 on Debian 10. For the time being, I have solved our issue by setting limits and range constraints for the x-axis.

[traceback.txt](https://github.com/silx-kit/silx/files/7661184/traceback.txt)

t20100:
This PR fixes 2 issues causing exceptions related to time axis:
- When the initial range of the plot is out of the range supported by Python `datetime`. In this case, there is no tick and values displayed: no ideal, but definitely better than an exception.
- When the data range leads to "nice" ticks outside of the `datetime` range. In this case, those ticks are ignored, but others are still displayed.

closes  #3570

### DIFF
--- a/src/silx/gui/plot/_utils/dtime_ticklayout.py
+++ b/src/silx/gui/plot/_utils/dtime_ticklayout.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2014-2018 European Synchrotron Radiation Facility
+# Copyright (c) 2014-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


payno:
Looks good to me

Issue #3598, opened 2022-02-08
Histogramnd fails on large arrays


pierrepaleo:
`Histogramnd` crashes on arrays with more than `2**31 - 1` elements:

```python
import numpy as np
from silx.math import Histogramnd

def test_histogram(n_samples, n_bins):
    data = np.random.randint(0, high=1000, size=(n_samples,), dtype=np.int32)
    dmin, dmax = (0, 1000)
    histogrammer = Histogramnd(
        data, n_bins=n_bins, histo_range=(dmin, dmax), last_bin_closed=True
    )
    return histogrammer.histo, histogrammer.edges[0]

test_histogram(2**31 - 1, 100) # succeeds
test_histogram(2**31, 100) # fails
```

This is probably due to the data type for indices (`i_n_elem`) in the histogramnd functions:

```C
int TEMPLATE(histogramnd, HISTO_SAMPLE_T, HISTO_WEIGHT_T, HISTO_CUMUL_T)
                        (HISTO_SAMPLE_T *i_sample,
                         HISTO_WEIGHT_T *i_weights,
                         int i_n_dim,
                         int i_n_elem,  // <== should be size_t
```

Using `int` for `i_n_dim` should be fine.

t20100:
closed by #3599

PR #3599, opened 2022-02-08
Histogramnd on large arrays

pierrepaleo:
This PR is an attempt at fixing #3598.

## To do:
  - [ ] `src` and `include`: use `size_t` instead of `int` for sample indices (`i_n_elem`)
  - [ ] Update `chistogramnd.pyx` and `chistogramnd_c.pxd`
  - [ ] `histogram.py`: Update `Histogrammd` and `HistogramndLut`
  - [ ] Add unit test

## Questions

I might need some help here. I'm not even sure that the "todolist" will be enough.

  - Will `#include <stddef.h>` work outside of Linux ?
  - Are the aforementioned updates enough to comprehensively fix the issue ?

Changelog: Fix `Histogramnd` computation on arrays with more than 2**31-1 samples


src/silx/math/histogramnd/src/histogramnd_template.c
```diff
@@ -36,7 +36,7 @@ int TEMPLATE(histogramnd, HISTO_SAMPLE_T, HISTO_WEIGHT_T, HISTO_CUMUL_T)
                         (HISTO_SAMPLE_T *i_sample,
                          HISTO_WEIGHT_T *i_weights,
                          int i_n_dim,
-                         int i_n_elem,
+                         size_t i_n_elem,
```
t20100, line 39: For consistency you'd better also change the type of `elem_idx` below:
```c
long elem_idx = 0;
```

t20100:
> Will #include <stddef.h> work outside of Linux ?

CI is happy with it so, it sounds good

> Are the aforementioned updates enough to comprehensively fix the issue ?

It looks fine to me. Unless I missed something, I don't think you need to change anything in `histogram.py`

src/silx/math/histogramnd/src/histogramnd_template.c
pierrepaleo: Thanks, I updated it.  
In `chistogramnd_lut.pyx` however, `elem_idx` should be left as `long` as it can be negative:

```C
elem_idx = 0 - i_n_dims
```

t20100 (approved):
LGTM

pierrepaleo:
A test was added for this special case. 

Of course, this test requires a large amount of memory (2.15 GB for chars, 8.6 GB for float/int, etc). 
The test is limited to
  - One dimensional case
  - `itemsize <= 4`

And by default the test is deactivated. I'd like to use a mechanism like `SILX_TEST_LOW_MEM`, but it's not clear how to use it with `unittest`. For now I'm using the plain `os.environ.get`, but I guess a better way would be to get `_TestOptions`, something like  
```python
from silx.test.utils import _TestOptions

test_options = _TestOptions()
test_options.configure()
if test_options.TEST_LOW_MEM:
    # ...
```
By the way, the behavior for `SILX_TEST_LOW_MEM` is unclear:
```python
        elif os.environ.get('SILX_TEST_LOW_MEM', 'True') == 'False':
            self.TEST_LOW_MEM = True
            self.TEST_LOW_MEM_REASON = "Skipped by SILX_TEST_LOW_MEM env var"
```
`self.TEST_LOW_MEM` is set to `True` when the env variable `SILX_TEST_LOW_MEM` is `"False"` which sounds weird.

t20100:
> `elif os.environ.get('SILX_TEST_LOW_MEM', 'True') == 'False':`

This sounds like a bug, I bet this is never used... `== "True"` would make more sense indeed.

You can use the [use_large_memory](https://github.com/silx-kit/silx/blob/master/src/silx/conftest.py#L87) fixture instead.

src/silx/math/test/test_histogramnd_vs_np.py
```diff
@@ -609,6 +610,41 @@ def test_reuse_cumul_float(self):
                                     result_np_w_1[0].sum(dtype=np.float64)),
                         msg=self.state_msg)
 
+    def test_histo_big_array(self):
```
t20100, line 613:
```suggestion
    @pytest.mark.usefixtures("use_large_memory")
    def test_histo_big_array(self):
```

src/silx/math/test/test_histogramnd_vs_np.py
```diff
@@ -26,6 +26,7 @@
 Results are compared to numpy's histogramdd.
 """
 
+import os
 import unittest
 import operator
```
t20100, line 31:
```suggestion
import os
import unittest
import operator
import pytest
```

pierrepaleo:
Thanks @t20100!

As a last thing, I would skip this test by default.  
For now the `low_mem` setting is disabled by default, so the test makes CI fail: 

```python
def pytest_addoption(parser):
    # ...
    parser.addoption("--low-mem", dest="low_mem", default=False,
                     action="store_true",
                     help="Disable test with large memory consumption (>100Mbyte")
```

should I use another fixture, or set the default value of `low_mem` to `True` ?
The latter will impact the following:
```
src/silx/opencl/sift/test/test_transform.py:    @pytest.mark.usefixtures("use_large_memory")
src/silx/opencl/sift/test/test_reductions.py:    @pytest.mark.usefixtures("use_large_memory")
src/silx/math/test/test_histogramnd_vs_np.py:    @pytest.mark.usefixtures("use_large_memory")
```

t20100:
> should I use another fixture, or set the default value of low_mem to True ?

The risk is then that those tests are never run...

Maybe this is not something to test?
Or limit the test case to (u)int8 and 1d: since we use `pytest`, the test can be a separate function rather than a method of the base class `TestCase` for histogramnd.

t20100:
On Github actions, Linux and Windows machine have 7GB RAM (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#cloud-hosts-for-github-hosted-runners)
On appveyor, it's at least 6GB (https://www.appveyor.com/docs/build-environment/#build-vm-configurations)

> Perhaps we should go for a dedicated fixture then ?

There is already a low mem option, IMO it's probably simpler to reuse it for this case as well and disable large memory tests on CI and enable them on weekly/release tests.
An alternative is to make a fixture to probe available memory (e.g., with `psutil` (https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory)) and skip the test if there is not enough free memory.
This can be done in steps (low mem here and memory probing in another PR).

PR #3600, opened 2022-02-09
silx.gui.plot.PlotWidget: Added support of math syntax to text with OpenGL

### LINKED ISSUE #845: [plot gl] Add support of math formula in axes labels and title
matplotlib supports a subset of TeX to support math formula.

It seems possible to use matplotlib text render (class MathTextParser in https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/mathtext.py#L3210 and https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/texmanager.py for TeX-based rendering) to do the text rasterisation in a numpy array (the same way Qt is used to do the text rasterisation). This array can then be provided as a texture to OpenGL to display the text. This would enable the support of matplotlib text formatting.

t20100:
This PR adds support for math syntax in plot texts (labels, title, etc...) with the OpenGL backend.
It does so by leveraging `matplotlib` parser (if present) to raster the text instead of using Qt.

It also fixes an issue in Qt raster which was producing larger image than necessary.

closes #845

### DIFF
--- a/src/silx/gui/_glutils/font.py
+++ b/src/silx/gui/_glutils/font.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -33,6 +33,7 @@ import logging
 import numpy
 
 from ..utils.image import convertQImageToArray
+from ..utils.matplotlib import rasterMathText
 from .. import qt
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/silx/gui/_glutils/font.py
```diff
@@ -32,8 +32,13 @@
...
 import numpy
 
-from ..utils.image import convertQImageToArray
 from .. import qt
+from ..utils.image import convertQImageToArray
+
+try:
+    from ..utils.matplotlib import rasterMathText
+except ImportError:
+    rasterMathText = None
```
t20100, line 41: Keep matplotlib option, since the OpenGL backend can run without it.

src/silx/gui/utils/matplotlib.py
```diff
@@ -38,27 +38,113 @@
...
+    )
+
+    fig = figure.Figure(figsize=(width / dpi, height / dpi))
+    fig.text(0, depth / height, stripped_text, fontproperties=prop)
+    with io.BytesIO() as buffer:
+        fig.savefig(buffer, dpi=dpi, format="raw")
+        buffer.seek(0)
+        image = numpy.frombuffer(buffer.read(), dtype=numpy.uint8).reshape(
+            int(height), int(width), 4
+        )
```
t20100, line 113: This is not nice, but it is the way it is done in `matplotlib` (except for the BytesIO buffer):
https://github.com/matplotlib/matplotlib/blob/d624571a19aec7c7d4a24123643288fc27db17e7/lib/matplotlib/mathtext.py#L264-L296

payno:
works smoothly on my side. Nice :)

payno:
works smoothly on my side. Nice :)

PR #3602, opened 2022-02-10
build: Added SILX_INSTALL_REQUIRES_STRIP build configuration environment variable

### LINKED ISSUE #3601: hdf5plugin is optional but not in the setup.py
Hello, hdf5plugin is an optional dependency and this is right.

```
for info on Debian/Ubuntu/etc.. we are working to install the hdf5 plugin at the system level.

https://packages.debian.org/unstable/hdf5-plugin-lzf
https://tracker.debian.org/news/1301128/accepted-bitshuffle-035-4-source-into-unstable/

This way it now possible to read out of the box the dectris images without special python modules.
```

Now the setup.py does not considere hdf5plugin as an optional dependency.
Due to this the generated entry points try to find hdf5plugin and fail to start silx.

I can maintain a patch on top of silx, but for consistency it seems to me that it should be clear that hdf5plugin is optional, and
it should be installed with something like pip install[hdf5plugin], or something else, I am not a user of pip.

so the setup.py must be consistent with the status of hdf5plugin.

Cheers

Fred

t20100:
This PR adds the `SILX_INSTALL_REQUIRES_STRIP` env. var. to strip a list of packages from the `install_requires`.

close #3601

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 # coding: utf8
 # /*##########################################################################
 #
-# Copyright (c) 2015-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2015-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


t20100:
@picca this should fit your needs now.

picca:
thanks a lot.

Can we have a release now ? ;)

t20100:
We can easily put a beta tag on github (and eventually push it to pypi if needed), but for a proper release it is different.

picca:
> We can easily put a beta tag on github (and eventually push it to pypi if needed), but for a proper release it is different.

Do you know when a proper release will be done ?

I can maintain a patch until then.

cheers

picca:
thanks, I already uploaded a silx with a patch :))

cheers

----- Thomas VINCENT ***@***.***> a écrit :

PR #3609, opened 2022-03-11
Rework colormap dialog

vallsv:
This PR improve some how the look end feel of the colormap.

This PR can be split if you prefer.

Screenshot from 2022-03-11 09-27-30

### Gamma value
I would like to be able to play with the gamma correction value without thinking at the number it-self.

The gamma value is theoretically symmetric, from exp to log with a linear transfer function at 1.

The idea is to linearise the value between min/max range in order to localize the color of the lut which is at 128.

This value can be displayed in the histogram.

### Refresh event

The colormap dialog and histogram is refreshed when the data change.

Unfortunately we don't have specific event, and most of the time the histogram don't have to be recomputed.

The PR tries to skip refresh when the colormapped data from the item is still the same (i.e. the same object)

This change have to be checked carefully, but i think it is fine while this array is created internally and should not be edited manually.

Changelog: 
- Changed the colormap dialog layout
- Changed min/max autoscale check now integrated inside the edit text
- Added gamma value edition using a marker in the histogram
- Changed the refresh event to reduce computation
- Added example of image plot refresh using gevent
- Added local histogram when using setting fixed range


src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1696,12 +1863,7 @@ def _handleScaleToVisibleAreaClicked(self):
         item = self._getItem()
         if item is None:
             return  # no-op
-
-        bounds = item.getVisibleBounds()
-        if bounds is None:
-            return  # no-op
-
```
t20100, line 1703: I guess that was here to handle the case the whole item is not visible (see other comment).

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1358,10 +1507,8 @@ def setColormapRangeFromDataBounds(self, bounds):
 
         :param Union[List[float],None] bounds:
             (xmin, xmax, ymin, ymax) Rectangular region in data space
+            If None, the full view is used
```
t20100, line 1510: It's not the full view, it's the item's visible area.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -634,18 +670,52 @@ def _updateMarkerPosition(self):
...
+
+        if colormap is None:
+            gamma = None
+        else:
+            if colormap.getNormalization() == Colormap.GAMMA:
+                gamma = colormap.getGammaNormalizationParameter()
+            else:
+                gamma = None
+
+        if gamma is not None:
```
t20100, line 698:
```suggestion
        if colormap is not None and colormap.getNormalization() == Colormap.GAMMA:
            gamma = colormap.getGammaNormalizationParameter()
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -634,18 +670,52 @@ def _updateMarkerPosition(self):
...
+            if not self._dragging[2]:
+                posRange = posMax - posMin
+                if posRange > 0:
+                    gammaPos = numpy.log(gamma) + 0.5
+                    gammaPos = posMin + (posMax - posMin) * gammaPos
+                else:
+                    gammaPos = posMin
```
t20100, line 705:
```suggestion
                gammaPos = posMin + max(0, posMax - posMin) * (numpy.log(gamma) + 0.5)
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -908,11 +1032,15 @@ def __init__(self, parent=None, title="Colormap Dialog"):
...
-        rangeLayout.addWidget(self._autoButtons, 2, 0, 1, -1, qt.Qt.AlignCenter)
+        rangeLayout.addWidget(labelMin, 0, 1)
+        rangeLayout.addWidget(labelMax, 0, 3)
+        rangeLayout.addWidget(self._minValue, 1, 1)
+        rangeLayout.addWidget(self._maxValue, 1, 3)
+        rangeLayout.setColumnStretch(0, 1)
+        rangeLayout.setColumnStretch(1, 2)
+        rangeLayout.setColumnStretch(2, 1)
+        rangeLayout.setColumnStretch(3, 2)
+        rangeLayout.setColumnStretch(4, 1)
```
t20100, line 1043:
```suggestion
        rangeLayout.setColumnStretch(4, 1)
```

Why not keeping the min/max with the other widgets on the left side when enlarging the colormap dialog?
Else, to me it would be best to center everything.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -127,6 +127,20 @@ def __init__(self, parent=None, value=0.0):
         self.setLayout(qt.QHBoxLayout())
         self.layout().setContentsMargins(0, 0, 0, 0)
         self._numVal = FloatEdit(parent=self, value=value)
+
+        self._iconAuto = icons.getQIcon('scale-auto')
+        self._iconFixed = icons.getQIcon('scale-fixed')
```
t20100, line 132: Maybe have the `scale-fixed` with something like: ~~AUTO~~

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -266,25 +293,25 @@ def __init__(self, parent=None):
...
 
-        self._minAuto = qt.QCheckBox(self)
-        self._minAuto.setText("")
+        self._minAuto = qt.QPushButton(self)
+        self._minAuto.setText("Min")
         self._minAuto.setToolTip("Enable/disable the autoscale for min")
+        self._minAuto.setCheckable(True)
         self._minAuto.toggled[bool].connect(self.__minToggled)
         self._minAuto.setFocusPolicy(qt.Qt.TabFocus)
+        self._minAuto.setVisible(False)
```
t20100, line 302: Unless I missed something, `self._minAuto` can be completely removed, it is never shown.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -266,25 +293,25 @@ def __init__(self, parent=None):
...
 
-        self._maxAuto = qt.QCheckBox(self)
-        self._maxAuto.setText("")
+        self._maxAuto = qt.QPushButton(self)
+        self._maxAuto.setText("Max")
         self._maxAuto.setToolTip("Enable/disable the autoscale for max")
+        self._maxAuto.setCheckable(True)
         self._maxAuto.toggled[bool].connect(self.__maxToggled)
         self._maxAuto.setFocusPolicy(qt.Qt.TabFocus)
+        self._maxAuto.setVisible(False)
```
t20100, line 310: Same here, `_maxAuto` looks to be no longer used.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1195,10 +1326,21 @@ def setItem(self, item):
...
+                array = item.getColormappedData(copy=False)
+            else:
+                array = None
+            colormapped = self._colormapped
+            if colormapped is not None:
+                oldArray = colormapped()
+            else:
+                oldArray = None
+            if oldArray is array:
+                return
```
t20100, line 1342: It's usually not a good idea to rely on `is` to ensure the array content didn't changed... but in this case, looking at `getColormappedData` and `_setColormappedData` it should work since a new `numpy.ndarray` object is created each time, so it should work but it can be fragile.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1195,10 +1326,21 @@ def setItem(self, item):
         the data range or the histogram of the data using :meth:`setDataRange`
         and :meth:`setHistogram`
         """
-        # While event from items are not supported, we can't ignore dup items
-        # old = self._getItem()
-        # if old is item:
-        #     return
+        old = self._getItem()
+        if old is item:
+            # While event from items are not supported, we can't ignore dup items
```
t20100, line 1331: I don't recall what this refers to, I guess it's because there is no even specific event for updates of `getColormappedData` data.

t20100:
It looks good: the gamma control is convenient, and the wider histogram area is nice (I miss the abscissa but that's fine).

src/silx/gui/dialog/ColormapDialog.py
vallsv: I have removed the orange color for better contrast, and used a strocked auto for the fixed.
It's much better.

src/silx/gui/dialog/ColormapDialog.py
vallsv: Yes, but i am afraid of breaking external stuffs using private stuffs.

src/silx/gui/dialog/ColormapDialog.py
vallsv: Yes

src/silx/gui/dialog/ColormapDialog.py
vallsv: To me it have to be aligned as much as possible to the min and max location of the plot.

It's the result of many tries.

If you want to change it, i let you play with.

src/silx/gui/dialog/ColormapDialog.py
vallsv: I will respel it, but it's the exact same meaning.

The "view" is the visible area, else you talk about "data"

vallsv:
I have removed the the local histogram. + plus expand the histogram vertically to fit the dialog size.

I let you check the result.
We can take a look together if you like.

payno:
very nice improvements :+1:

src/silx/gui/dialog/ColormapDialog.py
t20100: That's a private attribute in a private class instantiated as a private attribute.... Using it externally is looking for trouble.

I made a PR on your branch for it.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1671,11 +1781,25 @@ def _histogramRangeMoving(self, vmin, vmax):
...
                 colormap.setVMax(vmax)
             self._maxValue.setValue(vmax)
+        if gammaPos is not None:
+            vmin, vmax = self._histoWidget.getFiniteRange()
+            vrange = vmax - vmin
+            if vrange > 0:
+                g = (gammaPos - vmin) / (vmax - vmin)
+            else:
+                g = 0.5
+            gamma = numpy.exp(g - 0.5)
```
t20100: Same as above, see PR https://github.com/vallsv/silx/pull/34 for an improved implementation.

src/silx/gui/widgets/FormGridLayout.py
```diff
@@ -0,0 +1,74 @@
...
+__authors__ = ["V. Valls"]
+__license__ = "MIT"
+__date__ = "29/09/2022"
+
+
+import typing
+from .. import qt
+
+
+class FormGridLayout(qt.QGridLayout):
```
t20100, line 36: Do you need this elsewhere? Else I'd keep it private so we can change it more easily.

t20100 (approved):

vallsv:
> To me it is fine to keep it like this for now, this brings a lot of improvements, we can see how to fix it later.

The LUT in the histogram could be reworked with a solid scatter rendering. It sound not be so difficult. But i agree with you, to me it's already fine.

Issue #3611, opened 2022-03-16
specfile mca module not loading last data


physwkim:
I've noticed that SPEC's array_dump function puts a trailing space at the end of the mca array.
As a brief example, if there is a space before line-feed, the last data is not read.

```
#S 1 mythen_save Point:0 Scan:1 Datafile:/temp.dat
#T 5 (Seconds)
#@MCA %25C
#@CTIME 5
#I 0.000000
@A 253 253 253 279 284 286 
```
There are 6 data, but only 5 are actually loaded.
```
from silx.io.specfile import SpecFile
sf = SpecFile('test.dat')
sf[0].mca[0]

Out: array([253., 253., 253., 279., 284.])
```
If there is no trailing space at the end of the data file, it load normally.

llohse:
I ran into the same problem. 

+1 for fixing this

vasole:
closed by #3612

PR #3612, opened 2022-03-16
Fix missing data if there is a trailing space in the mca array

physwkim:
Fixed the problem that the last data was not loaded if there was a non-numeric trailing character when loading mca data.

Please refer to https://github.com/silx-kit/silx/issues/3611#issue-1170627169.

### DIFF
--- a/src/silx/io/specfile/src/sfmca.c
+++ b/src/silx/io/specfile/src/sfmca.c
@@ -298,6 +298,11 @@ SfGetMca( SpecFile *sf, long index, long number, double **retdata, int *error )
        val = PyMcaAtof(strval);
        data[vals] = val;
        vals++;
+     } else if (i>0) {
+         strval[i] = '\0';
+         val = PyMcaAtof(strval);
+         data[vals] = val;
+         vals++;
      }
 #ifndef _GNU_SOURCE
 #ifdef PYMCA_POSIX


src/silx/io/specfile/src/sfmca.c
```diff
@@ -298,6 +298,11 @@ SfGetMca( SpecFile *sf, long index, long number, double **retdata, int *error )
        val = PyMcaAtof(strval);
        data[vals] = val;
        vals++;
+     } else if (i>0) {
+         strval[i] = '\0';
+         val = PyMcaAtof(strval);
+         data[vals] = val;
```
llohse, line 304: This could potentially lead to a buffer overflow if the mca has 2049 entries. (Same as the block at line 295)

physwkim: The data pointer seems to be resized when it becomes larger than initsize(2048).

https://github.com/silx-kit/silx/blob/8f146767a22bb100f8ce6adf73377fea91b899ff/src/silx/io/specfile/src/sfmca.c#L269-L282

In my case, there was no buffer overflow even at 2560 entries.

llohse: Exactly my point. The code that resizes the buffer is not reached when the mca has 2049 entries.

vasole:
I find:

- the path works
- realloc is called always even when the number of channels of the MCA is smaller than 2048 so there is an excess buffer of 2048 channels every time.

vasole:
I still do not get the buffer overflow triggered.

On the first value read, the buffer is increased to 2 * initsize, so we are allocating space 4096 values despite having reading one. When we reach 2049, we are allocating up to 6144 despite having still enough room. So, when we reach 4097 we still have room up to 6144. Despite that, we reallocate again to 8192 and so on.  At no point there can be a buffer overflow since we are "wasting" a portion of memory of size initsize - 1. The 1 not wasted is for the value that would cause the overflow if we would not have that spare buffer.

llohse:
You are, of course, correct. Sorry for the confusion.

vasole:
@llohse No problem! I really appreciate your code review.

Issue #3577, opened 2021-12-17
Allow wildcards on the command line together with a dataset path


sdebionne:
The information that I need is not on the default plot but on the dataset itself. A little enhancement that would be handy for me:

```
$ silx view /tmp/test_hdf5_gray8s_*.h5::/entry_0000/instrument/Simulator/data
ERROR:silx.app.view.main:Filename '/tmp/test_hdf5_gray8s_*.h5' must be a file path
```

I try several variants with quotes but without luck.

t20100:
The `*` is indeed not supported in this case, but it is supposed to work if you provide the full filename.

sdebionne:
> it is supposed to work if you provide the full filename

Correct, it works. What I suggest here is a little extension to allow opening a given dataset for a set of (pattern matching) files. Just for convenience. It's probably doable with a little bit of bash scripting too...

payno:
It would also be nice to be able to combine data file pattern and data path pattern like
```
silx view file_pattern*.hdf::entry*/data
```

Furthermore in the case where we have a valid nexus dataset we could request a specific slice.
For example if we have a set of entries containing a 3D `/data` dataset that can be interpreted as an `image` (`interpretation` == `image` in h5py dataset attribute) then we could ask to display the last frame of the stack.
We could provide this as another parameter like
```
silx view file_pattern*.hdf5::entry*/data::slice=-1
```
Or maybe update the current way to have something like:

```
silx view file_pattern*.hdf5::/entry*/data&slice=-1
```
Or maybe go to the silx DataUrl way like:
```
silx view file_pattern*.hdf5?path=/entry*/data&slice=-1
```

vasole:
@payno I would recommend you to contact @cpascual He was working sometime ago on URIs related to that.

cpascual:
Hi. I don't know about the internal implementations of the URIs (or URI-like) in silx, so I can only talk in abstract terms based on my experience with [taurus](https://www.taurus-scada.org), and more specifically, with the [h5file plugin for taurus](https://gitlab.com/taurus-org/h5file-scheme).

In taurus we try to be strict in using URIs that conform to the [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) (although we allow some exceptions in some specific cases). This may be interesting for standarisation, but it imposes restrictions that may not be worth to you... so just decide what better suits your case.

> It would also be nice to be able to combine data file pattern and data path pattern like

[...]

PR #3615, opened 2022-04-05
Draft: silx view: add a first way to handle linux style shell character as w…

payno:
related to https://github.com/silx-kit/silx/issues/3577

TODO
=====
- [x] prototype to handle simple use cases
- [ ] check file management into more details. This use case was not really expected (having several item related to the same file). I guess this could mess up will the file management (and bring serious issues with HDF5 files). Might let the file open or stop working when we remove an item. I guess we would need some kind of a counter on top of the existing system. Not sure how complicated it can become.
- [ ] see ways to handle  `data_slice` 
For now handles simple uses cases as:
- improve management of `Viewer.appendFile` to handle DataUrl directly. Avoid some translation from `DataUrl` to `str`.
- improve wild card on the file path. For now uses glob. Might be better to use fnmatch for all. But seems less direct and might require creation of another function calling `os.listdir` ?
- handle slice. Today item are created but I guess when we provide slices we would like to have it automatically "selected" and displayed ? Behavior to be discussed.
  
```
clear; silx view 5.06_crayon_W150_60_Al2_W0.25_xc1000_.nx::entry/data/image*
clear; silx view 5.06_crayon_W150_60_Al2_W0.25_xc1000_.nx::entry/*
clear; silx view 5.06_crayon_W150_60_Al2_W0.25_xc*_.nx
clear; silx view 5.06_crayon_W150_60_Al2_W0.25_xc*_.nx::entry/* # !!! fails on closing
```

### DIFF
--- a/src/silx/app/view/main.py
+++ b/src/silx/app/view/main.py
@@ -32,6 +32,8 @@ import logging
 import os
 import signal
 import sys
+import glob
+from silx.io.url import DataUrl
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


t20100:
PR #3663 improves support for "*" in filename and adds support for "*" in data path.
Adding support for slicing seems more complicated since the model manages groups and datasets... so not sure what would be the best approach to support it.

t20100:
PR #3677 adds support for slicing

payno:
All has been done in other PR thank to @t20100 . We can close this PR

PR #3622, opened 2022-05-16
silx.io.specfile: Fixed buffer overflow for too long motor or label

### LINKED ISSUE #3621: specfile: is a label limited to 40 characters in windows?
Hello, 
I found some crashes in Windows (not in linux) using spec files that I created.

With this simple spec file: 
[test.txt](https://github.com/silx-kit/silx/files/8668975/test.txt)

the code: 
```
from silx.io.specfile import SpecFile

sf = SpecFile("test.txt")

print(len(sf))
for index in range(len(sf)):
    s1 = sf[index]
    name1 = s1.scan_header_dict["S"]
    print(index, name1)
```

gives me the correct results: 
```
C:\Users\srio\Miniconda3\python.exe C:/Users/srio/OASYS1.2/dabax/dabax/test.py
1
0 1 H

Process finished with exit code 0
```

But if I add the 40th character in the label (replace 012345678901234567890123456789012345678 by 0123456789012345678901234567890123456789) it crashes:

```
C:\Users\srio\Miniconda3\python.exe C:/Users/srio/OASYS1.2/dabax/dabax/test.py
1

Process finished with exit code -1073740791 (0xC0000409)
``` 

There is no problem in linux. Any idea? 

It is not a big problem (I can reduce the labels) but I wanted to report it. 

Many thanks in advance!

t20100:
This PR makes the temporary buffer larger (256 chars) and cuts the string in case of overflow.
There is certainly a better way to fix this by copying directly the right part of `buf` into the newly allocated `onelabel` buffer and avoiding the use of a temporary buffer, but it involves more code changes.

closes #3621

### DIFF
--- a/src/silx/io/specfile/src/sflabel.c
+++ b/src/silx/io/specfile/src/sflabel.c
@@ -1,5 +1,5 @@
 # /*##########################################################################
-# Copyright (C) 1995-2019 European Synchrotron Radiation Facility
+# Copyright (C) 1995-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


vasole:
I wonder if SPEC itself allows to have a label of more than 40 characters.

t20100:
Provided this issue is there since "ever", I would say no, but from a very quick look at the documentation I could not find something about it.

vasole (approved):

PR #3623, opened 2022-06-15
Utility to import/export FFTW plans

pierrepaleo:
This PR adds a small utility for easily importing/exporting FFTW plans, called [wisdom](https://www.fftw.org/fftw3_doc/Words-of-Wisdom_002dSaving-Plans.html).

The wisdom data structure is saved in a numpy `npz` file, along with some "metadata" (eg. machine name). The rationale of this metadata is to avoid using plans created on different platforms - this can hamper performance.

The unit test is probably too simple. In theory, we should check that instantiating `fftw.FFTW` object is faster after importing wisdom (for the corresponding configuration) - but that can be tricky. Suggestions welcome :-)

Changelog: Add export_wisdom() and import_wisdom() in silx.math.fft.fftw


t20100 (approved):
LGTM, can you fix the conflict?

pierrepaleo:
Sorry didn't see it. I'll fix it

PR #3624, opened 2022-06-21
Allow to change axes for nxdata stack of images

t20100:
This needs more testing to make sure it works under all condition this widget is used

### DIFF
--- a/src/silx/gui/data/NXdataWidgets.py
+++ b/src/silx/gui/data/NXdataWidgets.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -381,7 +381,6 @@ class ArrayImagePlot(qt.QWidget):
 
         # not closable
         self._selector = NumpyAxesSelector(self)
-        self._selector.setNamedAxesSelectorVisibility(False)
         self._selector.selectionChanged.connect(self._updateImage)
 
         self._auxSigSlider = HorizontalSliderWithBrowser(parent=self)


t20100:
As it is, it is not working because then `@axes` is not supported.

t20100:
Moving this to next release. It's probably worth merging "Image" and "Stack" views at the same time.

loichuder:
Encountered #4432 on the way. I'll wait until the fix #4433 is merged before continuing this.

src/silx/gui/data/DataViews.py
```diff
@@ -1763,27 +1764,15 @@ def setData(self, data):
 
         self._updateColormap(nxd)
 
-        # last two axes are Y & X
```
loichuder, line 1751: The association between axes and `Y` and `X` image axes is now done dynamically in the `ArrayImagePlot`.

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -595,3 +595,11 @@ def setNamedAxesSelectorVisibility(self, visible):
...
         for axis in self.__axis:
             axis.setNamedAxisSelectorVisibility(visible)
+
+    def getIndicesOfNamedAxes(self) -> dict[str, int]:
+        result: dict[str, int] = {}
+        for i, axis in enumerate(self.__axis):
+            name = axis.axisName()
+            if name:
+                result[name] = i
+        return result
```
loichuder, line 605: To be able to retrieve the index of an axis from its name

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -404,10 +401,9 @@ def __init__(self, parent=None):
         maskToolWidget = self._plot.getMaskToolsDockWidget().widget()
         maskToolWidget.setItemMaskUpdated(True)
 
-        # not closable
         self._axesSelector = NumpyAxesSelector(self)
-        self._axesSelector.setNamedAxesSelectorVisibility(False)
         self._axesSelector.selectionChanged.connect(self._updateImage)
+        self._axesSelector.selectedAxisChanged.connect(self._updateImageAxes)
```
loichuder, line 407: Note that there are two updating methods:
- `self._updateImage` is called when the slider value changes (the axes stay the same)
- `self._updateImageAxes` is the new method called when the displayed axis changes. It takes care of resetting the plot properly before slicing the image accordingly.

In normal operation, `self._updateImage` will always be preceded by a call to `self._updateImageAxes`.

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -453,48 +449,34 @@ def getPlot(self):
...
         self.__signals = signals
         self.__signals_names = signals_names
-        self.__axis_names = axes_names
-        self.__x_axis = x_axis
-        self.__x_axis_name = xlabel
-        self.__y_axis = y_axis
-        self.__y_axis_name = ylabel
+        self.__axes = axes
+        self.__axes_names = axes_names
+        self.__axes_scales = axes_scales
```
loichuder, line 480: Using more consistent naming and storing the scales now.

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -627,23 +626,39 @@ def _updateImage(self):
...
-        if self.__title:
-            title = self.__title
-            if len(self.__signals_names) > 1:
-                # Append dataset name only when there is many datasets
-                title += "\n" + self.__signals_names[signal_index]
-        else:
-            title = self.__signals_names[signal_index]
-        self._plot.setGraphTitle(title)
-        self._plot.getXAxis().setLabel(self.__x_axis_name)
-        self._plot.getYAxis().setLabel(self.__y_axis_name)
```
loichuder, line 639: Abstracted away in new method `_graphTitle`

loichuder:
@t20100 Now ready for review

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -367,14 +370,12 @@ class ArrayImagePlot(qt.QWidget):
...
 
-    The signal array can have an arbitrary number of dimensions, the only
-    limitation being that the last two dimensions must have the same length as
-    the axes arrays.
-
     Sliders are provided to select indices on the first (n - 2) dimensions of
     the signal array, and the plot is updated to show the image corresponding
     to the selection.
 
+    The dimensions can be changed with the signal array has more than 2 dimensions.
```
t20100:
```suggestion
    The dimensions can be changed when the signal array has more than 2 dimensions.
```

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -453,48 +449,34 @@ def getPlot(self):
...
-        xscale=None,
-        yscale=None,
-        keep_ratio: bool = True,
+        signals: list[h5py.Dataset],
+        axes: list[h5py.Dataset] | None = None,
+        signals_names: list[str] | None = None,
+        axes_names: list[str] | None = None,
+        axes_scales: list[Literal["linear", "log"] | None] | None = None,
+        title: str | None = None,
+        isRgba: bool = False,
```
t20100: This class is not reference in the doc and from a [quick search](https://github.com/search?q=ArrayImagePlot&type=code) the only place where it is used is:
https://github.com/silx-kit/pymca/blob/a07233366d76ad8e490a48e5b800493ee9db77be/src/PyMca5/PyMcaGui/io/hdf5/Hdf5NodeView.py#L139 and it should not be impacted by this change, so OK for me.

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -627,23 +626,39 @@ def _updateImage(self):
...
+    def _updateImage(self):
+        """Updates the image itself. Called when the user slices through the image without changing the axes."""
+        axes_selection = self._axesSelector.selection()
+        signal_index = self._signalSelector.getSignalIndex()
+        images = [img[axes_selection] for img in self.__signals]
+        image = images[signal_index]
+
+        self._plot.getActiveImage().setData(image)
+
+    def _graphTitle(self):
```
t20100:
```suggestion
    def _graphTitle(self) -> str:
```

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -453,48 +449,34 @@ def getPlot(self):
...
-        axes_names=None,
-        xlabel=None,
-        ylabel=None,
-        title=None,
-        isRgba=False,
-        xscale=None,
-        yscale=None,
-        keep_ratio: bool = True,
+        signals: list[h5py.Dataset],
+        axes: list[h5py.Dataset] | None = None,
```
t20100: I think that correct typing here would be `h5py.Dataset | silx.io.commonh5.Dataset` instead of `h5py.Dataset`.

https://github.com/silx-kit/silx/blob/731158f028952d6b5d0ed7376053f783daec7381/src/silx/io/commonh5.py#L164

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -539,8 +520,21 @@ def _updateImage(self):
         images = [img[axes_selection] for img in self.__signals]
         image = images[signal_index]
 
-        x_axis = self.__x_axis
-        y_axis = self.__y_axis
+        axis_indices = self._axesSelector.getIndicesOfNamedAxes()
+        x_axis_index = axis_indices["X"]
+        y_axis_index = axis_indices["Y"]
```
t20100: `X` and `Y` are always available?

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -627,23 +626,39 @@ def _updateImage(self):
...
         self._axesSelector.clear()
         self._axesSelector.blockSignals(old)
         self._plot.clear()
 
+    def _updateImage(self):
+        """Updates the image itself. Called when the user slices through the image without changing the axes."""
+        axes_selection = self._axesSelector.selection()
+        signal_index = self._signalSelector.getSignalIndex()
+        images = [img[axes_selection] for img in self.__signals]
+        image = images[signal_index]
```
t20100: Can't you just do?
```suggestion
        signal_index = self._signalSelector.getSignalIndex()
        image = self.__signals[signal_index][axes_selection]
```
or
```suggestion
        signal = self.__signals[self._signalSelector.getSignalIndex()]
        image = signal[axes_selection]
```

src/silx/gui/data/NXdataWidgets.py
loichuder: Yeah, I didn't really think about it but do you wish to go through a proper deprecation step?

src/silx/gui/data/NXdataWidgets.py
loichuder: If `setImageData` was called beforehand, yes.

And it should always be the case since the connection to the callback `_updateImageAxes` is done in `setImageData`. So if `setImageData` was not called, the `_updateImageAxes` slot is not connected to anything.

I will add additional checks for good measure though.

src/silx/gui/data/NXdataWidgets.py
loichuder: Seems to work. Thanks 👍

src/silx/gui/data/NXdataWidgets.py
t20100: No, I think it is OK to change the API here.
As I said: This is not in the doc, doesn't seems to be used directly... and it's a major release.

loichuder:
Actually, I have to rework this a bit. It can raise errors for RGBA images

PR #3631, opened 2022-07-20
silx.gui.PlotWindow: Fixed management of DockWidgets when showing/hiding the PlotWindow

### LINKED ISSUE #3629: Some PlotWindow dock widgets gets hidden when PlotWidget in a tab widgets
First raised in https://gitlab.esrf.fr/bliss/bliss/-/issues/3444

To reproduce:
```python
from silx import sx
w = sx.qt.QTabWidget()                                                  
w.addTab(sx.plot(), "p1")                                              
w.addTab(sx.plot(), "p2")                                              
w.show()
```
Then open the "Curves stats" from the `options` bottom menu, switch to tab "p2" and switch back to "p1", the curves stats dock widget is gone.

t20100:
This PR reworks the handling of the tools embedded in dock widgets in the `PlotWindow` to fix it when `PlotWindow` is displayed inside a tab and the visibility of the tab change.

- All opened tabbed dock widgets remain opened (it was not the case for the "Curves Stats" and the "Console" tabs)
- Dock widgets are only raised when it is shown due to a user action. Before the raised dock widget was changed when switch back and forth from a tab containing a `PlotWindow`.

attn @woutdenolf @vallsv 

closes #3629

### DIFF
--- a/src/silx/gui/console.py
+++ b/src/silx/gui/console.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2004-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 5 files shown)


woutdenolf (approved):
I can confirm it solves the issue. Thanks @t20100 LGTM

PR #3632, opened 2022-07-22
Update for Python 3.11

vasole:
Reported (for PyMca) in https://github.com/vasole/pymca/issues/889

For the implemented solution see:

https://stackoverflow.com/questions/4414234/getting-pythons-unittest-results-in-a-teardown-method


vallsv:
Nice to know.
Thanks.

PR #3639, opened 2022-08-23
silx.sx: Fixes compatibility with matplotlib>=3.6.0 and PySide6; Fixed continuous integration

t20100:
This PR fixes compatibility with the upcoming matplotlib version 3.6.0 and the display of `silx.sx` widgets which requires to enable matplotib (here through pylab) with matplotlib >=3.5.2 which uses PySide6 by default.

Tested on Linux with matplotlib v3.0 to v3.6.

It also updates tests and CI config to fix failing tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,43 +23,43 @@ jobs:
         include:
           - name-suffix: "PySide2 sdist"
             os: ubuntu-20.04
-            python-version: 3.6
+            python-version: '3.6'
             BUILD_COMMAND: sdist
             QT_BINDING: PySide2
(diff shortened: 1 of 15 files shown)


t20100:
Can someone review this and merge it?
It fixes all the CI.

vallsv:
Nice job.

There was maybe a way to use https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html instead of trying to use `rmtree`.
But it's probably still difficult to mix `unittest` and `pytest`

t20100:
Yes I agree. I fixed it with minimal changes to the code base, but more usage of `pytest` would be good.

PR #3644, opened 2022-08-26
silx.gui.plot: Removed error logs for scatter triangle visualisation with aligned points

### LINKED ISSUE #3608: Delaunay tesselation failed
Got a lot logs like the following i Flint, for some reason.

Any idea if it is useful for something? Or if this could be fixed in silx side?

I guess it's related to scatter rendering.

Else, i think ill filter out the message inside Flint.

```
2022-02-25 00:45:15,378 eh3 19785 silx.gui.plot._utils.delaunay ERROR : Delaunay tesselation failed: QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)

While executing:  | qhull d Qt Qz Qbb Qc Q12
Options selected for Qhull 2019.1.r 2019/06/21:
  run-id 468942845  delaunay  Qtriangulate  Qz-infinity-point  Qbbound-last
  Qcoplanar-keep  Q12-allow-wide  _pre-merge  _zero-centrum  Qinterior-keep
  Pgood  _max-width 4.1  Error-roundoff 3.5e-15  _one-merge 2.4e-14
  Visible-distance 6.9e-15  U-max-coplanar 6.9e-15  Width-outside 1.4e-14
  _wide-facet 4.2e-14  _maxoutside 2.8e-14

The input to qhull appears to be less than 3 dimensional, or a
computation has overflowed.

Qhull could not construct a clearly convex simplex from points:
- p164(v4): -0.46     0   2.5
- p100(v3): 1.9e-05     0     0
- p163(v2):   1.6     0   0.9
- p0(v1):  -2.5     0   2.3

The center point is coplanar with a facet, or a vertex is coplanar
with a neighboring facet.  The maximum round off error for
computing distances is 3.5e-15.  The center point, facets and distances
to the center point are as follows:

center point  -0.3469        0    1.419

facet p100 p163 p0 distance=    0
facet p164 p163 p0 distance=    0
facet p164 p100 p0 distance=    0
facet p164 p100 p163 distance=    0

These points either have a maximum or minimum x-coordinate, or
they maximize the determinant for k coordinates.  Trial points
are first selected from points that maximize a coordinate.

The min and max coordinates for each dimension are:
  0:      -2.5     1.575  difference= 4.075
  1:         0         0  difference=    0
  2:         0       2.5  difference=  2.5

If the input should be full dimensional, you have several options that
may determine an initial simplex:
  - use 'QJ'  to joggle the input and make it full dimensional
  - use 'QbB' to scale the points to the unit cube
  - use 'QR0' to randomly rotate the input for different maximum points
  - use 'Qs'  to search all points for the initial simplex
  - use 'En'  to specify a maximum roundoff error less than 3.5e-15.
  - trace execution with 'T3' to see the determinant for each point.

If the input is lower dimensional:
  - use 'QJ' to joggle the input and make it full dimensional
  - use 'Qbk:0Bk:0' to delete coordinate k from the input.  You should
    pick the coordinate with the least range.  The hull will have the
    correct topology.
  - determine the flat containing the points, rotate the points
    into a coordinate plane, and delete the other coordinates.
  - add one or more points to make the input full dimensional.
```

t20100:
This PR removes the error message displayed when `delaunay` fails when displaying a scatter plot as a mesh.

It changes `delaunay` error log from error to debug level (returns None in case of error anyway)

closes #3608

### DIFF
--- a/src/silx/gui/plot/_utils/delaunay.py
+++ b/src/silx/gui/plot/_utils/delaunay.py
@@ -55,8 +55,7 @@ def delaunay(x, y):
     try:
         delaunay = _Delaunay(points)
     except (RuntimeError, ValueError):
-        _logger.error("Delaunay tesselation failed: %s",
-                      sys.exc_info()[1])
+        _logger.debug("Delaunay tesselation failed: %s", sys.exc_info()[1])
         delaunay = None
 
     return delaunay


vallsv:
Thanls

PR #3649, opened 2022-08-29
Build: Removed usage of `numpy.distutils`

t20100:
This PR replaces the use of `numpy.distutils` by a `ext_modules` list in the top level `setup.py`.
It also imports the `build` command from `setuptools` if available (since v62.4.0).

The only remaining dependency to `distutils` is: `from distutils.command.clean import clean`.

related to #3574


t20100:
This looks to work.

Full build for release passes, see https://gitlab.esrf.fr/silx/bob/silx/-/pipelines/85428

t20100:
I didn't unpin `setuptools` in `pyproject.toml` since it is under heavy updates:
https://github.com/silx-kit/silx/blob/bf182fd14eb79b0458220f79a5e3b6d27ed7475d/pyproject.toml#L4

But by not using `numpy.distutils` anymore, it spares some issues like recently with `setuptools` v65 and it is a big step towards stopping using `distutils`.

BTW, this does not prevent us from moving to another tool like meson, but it buys us some more time to do so.

vallsv:
@t20100 an appveyor is failing, related to mask UI. No idea if it's important or not.
I am not able to relaunch it.

vallsv:
Sounds like it fails often and sometime. So let say it's fine.
Probably not related to the build itself.

t20100:
Yes, CI currently has random failures on appveyor, this is not related to this PR

PR #3650, opened 2022-08-30
ElideLabel: Fix API inherited from QLabel

### LINKED ISSUE #3641: inconsistent behaviour of toolTip for silx.gui.widgets.ElidedLabel and generic QWidget
Hi,

the silx.gui.widgets.ElidedLabel redefines the toolTip as a qt.Property which is inconsistent with the generic toolTip method (using Qt5).
Is there a reason for this behaviour? This raises exceptions when implementing an eventFilter for toolTips because the property is not callable.
As the ElidedLabel is a subclassed QLabel, I think the toolTip should be consistent in both child and parent.

The ElidedLabel is used as base class for the _PositionInfoLabel used in the PositionInfo widget in the Plot2D widget and I noticed the issue when implementing an eventFilter for an application with a Plot2D widget.

```
label = QtWidgets.QLabel()
elided_label = silx.gui.widgets.ElidedLabel.ElidedLabel()

type(label.toolTip)
Out: builtin_function_or_method

type(elided_label.toolTip)
Out: str
```

vallsv:
This PR fixes the ElideLabel API inherited from QLabel.

For consistency and interoperability purpose.

Closes #3641

Changelog: 
- ElideLabel: Fix API inherited from QLabel

### DIFF
--- a/src/silx/gui/widgets/ElidedLabel.py
+++ b/src/silx/gui/widgets/ElidedLabel.py
@@ -31,13 +31,13 @@ from silx.gui import qt
 
 
 class ElidedLabel(qt.QLabel):
-    """QLabel with an edile property.
+    """QLabel with an elide property.
 
-    By default if the text is too big, it is elided on the right.
+    By default if the text is too long, it is elided on the right.
 
     This mode can be changed with :func:`setElideMode`.
 
-    In case the text is elided, the full content is displayed as part of the
+    In this case the text is elided, the full content is displayed as part of the
     tool tip. This behavior can be disabled with :func:`setTextAsToolTip`.
     """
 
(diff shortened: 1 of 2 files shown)


vallsv:
I assume it is working with QtDesigner.
We don't have the setup for anyway, and we dont use it this way anyway.

src/silx/gui/widgets/ElidedLabel.py
```diff
@@ -83,25 +83,33 @@ def __updateToolTip(self):
         else:
             qt.QLabel.setToolTip(self, self.__toolTip)
 
-    # Properties
+    # Inherited properties
+
+    def text(self):
+        """Returns the text defined by the user.
+
+        It can defier from the one really displayed, depending on the
```
t20100, line 91:
```suggestion
        It can be different from the one really displayed, depending on the
```

src/silx/gui/widgets/ElidedLabel.py
```diff
@@ -83,25 +83,33 @@ def __updateToolTip(self):
...
+        `elideMode` defined for this widget.
+        """
+        return self.__text
 
     def setText(self, text):
         self.__text = text
         self.__updateText()
 
-    def getText(self):
```
t20100, line 92: It would be better to keep `getText` for compatibility with a deprecation warning.

src/silx/gui/widgets/ElidedLabel.py
```diff
@@ -83,25 +83,33 @@ def __updateToolTip(self):
...
+        It can defier from the one really displayed, if `textAsToolTip` was
+        set to true.
+        """
+        return self.__toolTip
 
     def setToolTip(self, toolTip):
         self.__toolTip = toolTip
         self.__updateToolTip()
 
-    def getToolTip(self):
```
t20100, line 101: Same for `getToolTip` as for `getText`.

t20100 (commented):
To me, it's not worth breaking the API beyond replacing the properties by methods.
+ tests are failing due to: https://github.com/silx-kit/silx/blob/master/src/silx/gui/plot/actions/histogram.py#L67

t20100 (approved):
LGTM

PR #3655, opened 2022-09-02
silx.gui: Added support of PyQt6

t20100:
This PR adds support for `PyQt6` >= 6.3 in that the tests should pass.

Tests fails with PyQt6 < 6.3.0 due to errors related to `enum` (so it would require more patches/specific handling).
Since the version of (Py)Qt packaged in debian/ubuntu is >= 6.3 (see https://packages.debian.org/bookworm/python3-pyqt6, https://packages.ubuntu.com/kinetic/python3-pyqt6), it sounds reasonable to me to not support lower versions.

related to #3432


t20100:
Ready for review.

There is still probably issues in corner cases with PyQt6, but at least all tests pass.

vasole:
To reproduce:

silx view 
zoom on an image
right-click and zoom back

image

t20100:
Thanks for the testing!

That should be fixed, `PyQt6` is more picky than the others regarding `QPoint` vs `QPointF`...

PR #3659, opened 2022-09-05
silx.gui.plot.stats: Fixed warnings when all data is outside the selected stats region

### LINKED ISSUE #3607: Warning on statshandler
No idea if it was already fixed, but here is a warning from python 3.7, and silx 1.0.0 (maybe beta)

```
2022-01-27 19:03:04,091 eh3 3065 py.warnings WARNING : 
/.../bliss_dev/lib/python3.7/site-packages/silx/gui/plot/stats/statshandler.py:78:
FutureWarning: Format strings passed to MaskedConstant are ignored,
but in future may error or produce different behavior
  return self.formatter.format(val)
```

t20100:
This PR fixes deprecation warnings in `StatFormatter` that occur when all the data of an item is outside the Stat computation region.
It also updates the display to show: `--` instead of `None` or `masked` (for numpy `MaskedConstant`).

The `StatCOM` also check for completely masked array to return `NaNs` in this case.

closes #3607

### DIFF
--- a/src/silx/gui/plot/stats/stats.py
+++ b/src/silx/gui/plot/stats/stats.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2022 European Synchrotron Radiation Facility
 #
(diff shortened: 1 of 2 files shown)


payno:
thanks for the bug fix

PR #3660, opened 2022-09-05
silx.gui.data: Do not keep aspect ratio in `NXdata` image views when axes `@units` are different

### LINKED ISSUE #3626: How about a posibility to save the aspect ration option in SILX_style?
Working with 2d datasets in NeXus hdf5 with different units on the x- and y-axis I was wondering if there is a way to tell silx view to open a NXdata image view with the option "Do not keep data aspect ration" activated? By default it seems to keep the aspect ratio.

Would the the `SILX_style` be an appropriate place?

https://github.com/silx-kit/silx/blob/360f890a617676a92f0bed6a28b718d09e70ec03/src/silx/io/nxdata/parse.py#L64

t20100:
This PR does not keep aspect ratio by default in the case the 2 axes of an image have different `@units` (either both are defined and different or only one is defined).
In all other cases, the aspect ratio is kept by default.
This is done for the image and complex image widgets, and not for the stack or 3D views.

closes #3626

### DIFF
--- a/src/silx/gui/data/DataViews.py
+++ b/src/silx/gui/data/DataViews.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 # /*##########################################################################
 #
-# Copyright (c) 2016-2020 European Synchrotron Radiation Facility
+# Copyright (c) 2016-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


t20100:
attn @linupi, would that work for your use case?

linupi:
Yes @t20100 , this would do the job in my use case. Thank you for considering my issue!

PR #3663, opened 2022-09-08
silx view: Improved wildcard support in filename and data path

### LINKED ISSUE #3577: Allow wildcards on the command line together with a dataset path

t20100:
This PR adds "*" support for `silx view` `files` argument for filename even when a data path is provided and for data path. Some examples of supported syntax:
`myfiles_*.h5::/data`, `myfiles_*.h5::/entry_*/data`, `myfile.h5::/entry_*/data/*`

It also adds a `silx.io.utils.match` function to find data path matching a pattern in a file.

It is similar to PR #3615 and borrows some of the approach.

closes  #3577

### DIFF
--- a/src/silx/app/view/main.py
+++ b/src/silx/app/view/main.py
@@ -28,10 +28,12 @@ __license__ = "MIT"
 __date__ = "17/01/2019"
 
 import argparse
+import glob
 import logging
 import os
 import signal
 import sys
+from typing import Generator, Iterable
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/silx/app/view/main.py
```diff
@@ -71,6 +73,39 @@ def createParser():
     return parser
 
 
+def filesArgToUrls(filenames: Iterable[str]) -> Generator[object, None, None]:
+    """Expand filenames and HDF5 data path in files input argument"""
+    import silx.io
+    from silx.io.utils import match
+    from silx.io.url import DataUrl
+    import silx.utils.files
```
payno, line 81: might be moved to top of the file

src/silx/app/view/main.py
```diff
@@ -71,6 +73,39 @@ def createParser():
...
+
+    for filename in filenames:
+        url = DataUrl(filename)
+
+        for file_path in sorted(silx.utils.files.expand_filenames([url.file_path()])):
+            if glob.has_magic(url.data_path()):
```
payno, line 87: will fail if `data_path()` is None must add something like 
```
if url.data_path() and glob.has_magic(url.data_path())
```

payno (commented):
seems to work fine on the few test I did on my sid

src/silx/app/view/main.py
t20100: Some might, but anything importing `h5py` should be imported after `HDF5_USE_FILE_LOCKING` is set from the command line arguments.
But it's definitely worth a comment!

src/silx/app/view/main.py
t20100: Indeed

t20100: Thanks for the testing! This is fixed.

PR #3664, opened 2022-09-12
silx.io.dictdump.dicttoh5: Added "info" logs when an entity is not copied to the output hdf5 file

### LINKED ISSUE #3462: [silx.io.dictdump.dicttoh5] unexpected behavior using update_mode="modify"
The function [`silx.io.dictdump.dicttoh5`](https://github.com/silx-kit/silx/blob/fe38fd3c712a311876ce8804e144621c9dad7c4f/src/silx/io/dictdump.py#L171) has an expected behavior when `update_mode="modify"` in the following scenario:

1. The file `mystuff.h5` already exists and has the following internal tree:

```
/A (Group)
/A/data (Group)
/A/data/a (Dataset)
/A/data/data (Dataset)
```

2. I want to flatten the structure, so I update it such that the second nested group's (`A/data`) content will be saved in the first one (`A`). Resulting in:

```
/A (Group)
/A/a (Dataset)
/A/data (Dataset)
```

But I'd like not to use `update_mode="replace` because other things might be updated as well during the same call.

Because of [this condition (L343)](https://github.com/silx-kit/silx/blob/fe38fd3c712a311876ce8804e144621c9dad7c4f/src/silx/io/dictdump.py#L343) , the update will not happen and no warning is yeilded.

t20100:
This PR adds some `INFO` level logs when `dicttoh5` is skipping some elements from the input dict because of the selected `update_mode`.

closes #3462

### DIFF
--- a/src/silx/io/dictdump.py
+++ b/src/silx/io/dictdump.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 # /*##########################################################################
-# Copyright (C) 2016-2020 European Synchrotron Radiation Facility
+# Copyright (C) 2016-2022 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


woutdenolf (approved):
LGTM

PR #3665, opened 2022-09-12
silx view: Fixed refresh for content opened as `file.h5::/path`

### LINKED ISSUE #3616: silxview: Refresh on a file open with subdataset is not working properly
Silx view allow to open a sub dataset/group from a hdf5 file.

In this case only this subtree is displayed.

But when you click on refresh, the parent file is appended as a new source.

Probably the refresh is not properly working too.

## To reproduce

- Open silx with an HDF5 with a sub path

image

- Select the source
- Click on the refresh button

image

t20100:
This PR fixes the refresh of selected files for files which where opened with a subpath (e.g., `file.h5::/entry/data`)

closes #3616

### DIFF
--- a/src/silx/app/view/Viewer.py
+++ b/src/silx/app/view/Viewer.py
@@ -266,7 +266,6 @@ class Viewer(qt.QMainWindow):
             index = indexes.pop(0)
             if index.column() != 0:
                 continue
-            h5 = model.data(index, role=silx.gui.hdf5.Hdf5TreeModel.H5PY_OBJECT_ROLE)
             rootIndex = index
             # Reach the root of the tree
             while rootIndex.parent().isValid():
@@ -274,7 +273,8 @@ class Viewer(qt.QMainWindow):
             rootRow = rootIndex.row()
             relativePath = self.__getRelativePath(model, rootIndex, index)
             selectedItems.append((rootRow, relativePath))
-            h5files.add(h5.file)
+            h5 = model.data(rootIndex, role=silx.gui.hdf5.Hdf5TreeModel.H5PY_OBJECT_ROLE)
+            h5files.add(h5)
 
         if len(h5files) == 0:
             qt.QApplication.restoreOverrideCursor()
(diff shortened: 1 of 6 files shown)


src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -557,7 +558,7 @@ def synchronizeIndex(self, index):
     def h5pyObjectRow(self, h5pyObject):
         for row in range(self.__root.childCount()):
             item = self.__root.child(row)
-            if item.obj == h5pyObject:
+            if numpy.all(item.obj == h5pyObject):
```
t20100, line 561: Needed as now it can be dataset so it compares arrays

t20100:
It was need to add `silx.io.commonh5.Node.__hash__` for the refresh to work with such input and not only with `h5py`.

src/silx/app/view/Viewer.py
```diff
@@ -305,7 +305,7 @@ def __synchronizeH5pyObject(self, h5):
         # FIXME: The architecture have to be reworked to support this feature.
         # model.synchronizeH5pyObject(h5)
 
-        filename = h5.filename
+        filename = f"{h5.file.filename}::{h5.name}"
```
t20100, line 308: This need some reworks: It would be better to keep the initial url used to reload (it would also work for #3677)

t20100:
I also store the filename our url provided as input in the model in order to retrieve it when refreshing a file.
This turned out to be useful in PR #3677.

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -557,7 +564,7 @@ def synchronizeIndex(self, index):
     def h5pyObjectRow(self, h5pyObject):
         for row in range(self.__root.childCount()):
             item = self.__root.child(row)
-            if item.obj == h5pyObject:
+            if hash(item.obj) == hash(h5pyObject):
```
vallsv, line 567: Same as the other PR. I don't think it is needed.
Or you are about to change the previous behaviour, which is maybe not bad, i can't tell.

t20100: It was removed, instead `__eq__` was rewritten to NOT compare the data but the filename/dataset name instead.

t20100:
Replaced commonh5 `__eq__` with a method performing the check to avoid changing the behavior of commh5 objects.

PR #3668, opened 2022-09-12
Build: Added optional use of `sphinx_autodoc_typehints` to generate the documentation

### LINKED ISSUE #2996: What about using python typing
So now we are in Python 3.

What about changing the documentation guideline to use typing?

It's not about reworking everything, nor enforcing typing. It's just about changing the way we write documentation for the new code we design.

In Flint i mostly type everything. That's really useful. And as we already do it in the docstring, what about doing it with the python parser. 

So this code:
```
    def getAlternativeImageData(self, copy=True):
        """Get the optional RGBA image that is displayed instead of the data

        :param bool copy: True (Default) to get a copy,
            False to use internal representation (do not modify!)
        :rtype: Union[None,numpy.ndarray]
        """
        pass
```
Can be written:
```
    def getAlternativeImageData(self, copy:bool=True) -> typing.Optional[numpy.ndarray]:
        """Get the optional RGBA image that is displayed instead of the data

        :param copy: True (Default) to get a copy,
            False to use internal representation (do not modify!)
        """
        pass
```

I dont have yet tried the result in sphinx.

What do you think?

t20100:
This PR adds an optional use of the [sphinx_autodoc_typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) extension to leverage Python type hints in the documentation.
It's optional because it is not available as a package in older Debian (e.g. 10) that we still support.

The only limitation is that if the type hint is provided for an argument, its type should not be defined in the corresponding docstring's as `:param <type> <name>:` (or I guess as `:type <name>: <type>`) or it's duplicated in the doc. To me this is an affordable limitation as the idea is not duplicate the type in the signature and in the docstring.
BTW, it is not the case for `:rtype:`: The one in the doctring takes over the type hint.

closes #2996


vallsv:
The library will also probably be improved on it's own.

PR #3671, opened 2022-09-14
Upgrade colormap mix max selection

### LINKED ISSUE #3669: colormap dialog: min-max ROI selection
Having to press selection each time is not very conveniant when you want to switch from one region to another one (in the case of tomography for example).

Having a 'min-max' colormap mode would be more conveniant (a button to keep this mode cf the zoom mode like this is done in imageJ)

payno:
Changelog: 

  update ColormapDialog 'selection' QPushButton to make it checkable
This  allow picking a ROI several time and updating min / max within the same sequence.

close #3669


t20100 (commented):
Nice! It's way better to use IMO.

I'm not sure the wait icon and the `...` are still needed here since it is now an interaction mode by itself (as the zoom or pan).

Also, in `silx view` when switching to another dataset with the selection tool enabled, it stays enabled. It's good when the new view is for an image, but for other views, I found that weird to have the wait icon keeping turning while you can't do something with it, because the interaction is zoom in this view.

t20100 (approved):
LGTM

PR #3673, opened 2022-09-19
Misc: Removed Python2 compatibilty code

t20100:
This PR removes Python2 compatibility that is no longer needed:
- `from __future__ import ...`
-  `# coding: utf-8`-like header comments: No longer needed, UTF-8 is the default encoding


kif:
LGTM

PR #3679, opened 2022-09-22
Retry for generator functions

woutdenolf:
We already had retry decorators for functions and context managers

```python
@retry()
def method(...):
    ....

@retry_contextmanager()
def method(...):
    yield ...
```

This PR adds a decorator for generator functions

```python
@retry()
def method(...):
    yield ...
    yield ...
```

### DIFF
--- a/src/silx/utils/retry.py
+++ b/src/silx/utils/retry.py
@@ -157,12 +157,12 @@ def retry_contextmanager(
                 retry_on_error=_retry_on_error,
             ):
                 with _handle_exception(options):
-                    gen = method(*args, **kw)
-                    result = next(gen)
+                    ctx = method(*args, **kw)
+                    result = next(ctx)
                     options["retry_on_error"] = None
                     yield result
                     try:
-                        next(gen)
+                        next(ctx)
                     except StopIteration:
                         return
                     else:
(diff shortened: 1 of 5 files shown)


src/silx/utils/retry.py
```diff
@@ -127,6 +129,59 @@ def wrapper(*args, **kw):
     return decorator
 
 
+def retry_iterator(
```
t20100, line 132: From what I tested, it seems possible to add support for iterator to `retry` instead of having a second one:

```python
def printer(prefix=''): 
     def d(f): 
         if inspect.isgeneratorfunction(f): 
             @wraps(f) 
             def w(*args, **kwargs): 
                 print(f"{prefix} generatorfunction {args}, {kwargs}") 
                 for r in f(*args, **kwargs): 
                     print('yield', r) 
                     yield r 
             return w 
         @wraps(f) 
         def w(*args, **kwargs): 
             print(f"{prefix} function {args}, {kwargs}") 
             return f(*args, **kwargs) 
         return w 
     return d
```

```python
In [22]: @printer("test") 
    ...: def f(): 
    ...:     return 1 
    ...:                                                                        

In [23]: f()                                                                    
test function (), {}
Out[23]: 1

In [25]: @printer("test") 
    ...: def g(): 
    ...:     yield 1 
    ...:                                                         

In [27]: for i in g(): 
    ...:     print(i) 
    ...:                                                                        
test generatorfunction (), {}
yield 1
1
```

woutdenolf: Thanks for the suggestion.

t20100 (approved):
LGTM

PR #3680, opened 2022-09-23
silx.gui.plot.tools.roi: Added `BandROI` for rotated rectangle regions of interest

### LINKED ISSUE #3638: a "BandROI" proposed
Hello,

Herewith I 'd like to ask for a new ROI class that is a line ROI _with width_. It would look similar to ArcROI, just with three handles, not four, and with a straight shape. I suggest to optionally have "open" or "closed" state for both ends, when the roi continues infinitely ("open") or up to the end point ("closed"). I also suggest to add "polar1" and "polar2" modes of handle dragging when the roi is rotated about the corresponding end point.

Thank you in advance!

t20100:
This PR provides a ROI object for bounded band or rectangle ROI that can be rotated.

The first implementation is as follows:
- square anchors move freely and allow to rotate and extend/reduce the ROI.
- diamond anchors move perpendicular to the ROI and allow only to enlarge/reduce it.
- The cross anchor moves freely and allow to move the ROI

bandroi

I did that as a "band" ROI, but I'm thinking of changing the anchors interaction to make like a rectangle + anchor(s) to rotate it.
I'll also look how to make an unbound "band" ROI.

attn @kklmn

closes #3638


t20100:
@kklmn does that fit your need?

vallsv:
Ideally it would be good to add a basic test.
We probably have some for other ROis.
Like testing basic geometry state.

PR #3683, opened 2022-09-26
silx.io.dictdump: Added support of `pint` in `dicttoh5` and `dicttonx`

t20100:
f24df26ad86f3a470c04fdaa1da5e884260223f4: This PR avoids warnings when passing `pint.Quantity` to `dicttoh5` by using the `.magnitude`.

641c0d0fb5b2008820bf1c9f334a849bf67f6c4f: This PR also proposes to store `pint` units as `@units` HDF5 attributes when calling `dicttonx`.
There is an issue here since [pint units string formatting](https://pint.readthedocs.io/en/stable/formatting.html#unit-format-specifications) is not strictly compatible with NeXus (see https://manual.nexusformat.org/nxdl-types.html?highlight=units#unit-categories-allowed-in-nxdl-specifications, https://manual.nexusformat.org/datarules.html#nexus-data-units): For example power is either `**` or `^`. BTW does anyone have a link to the specification of Nexus units format?

What do you think? Should I remove support in `dicttonx`?

attn @sdebionne

### DIFF
--- a/src/silx/io/dictdump.py
+++ b/src/silx/io/dictdump.py
@@ -460,6 +460,7 @@ def nexus_to_h5_dict(
                     value = h5py.SoftLink(first)
             elif is_link(value):
                 key = key[1:]
+
         if isinstance(value, Mapping):
             # HDF5 group
             key_has_nx_class = add_nx_class and _has_nx_class(treedict, key)
@@ -468,9 +469,15 @@ def nexus_to_h5_dict(
                 parents=parents+(key,),
                 add_nx_class=add_nx_class,
                 has_nx_class=key_has_nx_class)
+
+        elif pint is not None and isinstance(value, pint.quantity.Quantity):
+            copy[key] = value.magnitude
+            copy[(key, "units")] = f"{value.units:~}"
+
         else:
             # HDF5 dataset or link
             copy[key] = value
+
     if add_nx_class and not has_nx_class:
         _ensure_nx_class(copy, parents)
     return copy
--- a/src/silx/io/dictdump.py
+++ b/src/silx/io/dictdump.py
@@ -30,8 +30,11 @@ import json
 import logging
 import numpy
 import os.path
-import sys
 import h5py
+try:
+    import pint
+except ImportError:
+    pint = None
 
 from .configdict import ConfigDict
 from .utils import is_group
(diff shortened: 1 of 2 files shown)


woutdenolf:
https://manual.nexusformat.org/datarules.html#nexus-data-units

> Values for the NeXus units must be specified in a format compatible with [Unidata UDunits](http://www.unidata.ucar.edu/software/udunits) ... At present, the contents of NeXus units attributes are not validated in data files.

> Should I remove support in dicttonx?

I would say keep it and don't worry too much about the string formatting for now.

In fact I would go even further and optionally support pint in `nxtodict`.

vallsv:
BTW pint also provides a `Measurement` object with the error. It maybe also fit another nx field?

woutdenolf:
As for uncertainties or errors: https://manual.nexusformat.org/datarules.html#uncertainties-or-errors

So if the name of the dataset is `field` then the uncertainties are stored in the datasets `field_errors`.

I haven't seen anyone use `Measurement` but I have seen https://pypi.org/project/uncertainties/.

woutdenolf:
As for the unit formatting, we could register an NeXus formatter with pint

```python
from pint import UnitRegistry
from pint import formatting

@formatting.register_unit_format("N")
def format_nexus(unit, registry, **options):
    return formatting.formatter(
        unit.items(),
        as_ratio=True,
        single_denominator=False,
        product_fmt=" * ",
        division_fmt=" / ",
        power_fmt="{} ^ {}",
        parentheses_fmt=r"({})",
        **options,
    )

ureg = UnitRegistry()

print(f"{ureg.cm**2:~N}")
```

t20100:
Thanks for the feedbacks

> Values for the NeXus units must be specified in a format compatible with [Unidata UDunits](http://www.unidata.ucar.edu/software/udunits) ...

I was looking for a syntax specification, I found the udunits2 one: https://docs.unidata.ucar.edu/udunits/udunits-2.2.28/udunits2lib.html#Syntax

I'll check how far the default formatter is from this syntax (e.g., it looks like both `**` and `^` are valid for power) and go for a custom formatter.
I'll leave everything else for other PRs if need be.

sdebionne:
Thanks for looking into this. The last time I had to produce datasets with a `units` attribute (netCDF with the Climate and Forecast convention) , I found out that the grammar is complex enough that you really need to use the UDunits library... Just FIY, a [Python binding exists](https://github.com/ckhroulev/py_udunits2).

t20100:
Looking at the [udunits2 syntax](https://docs.unidata.ucar.edu/udunits/udunits-2.2.28/udunits2lib.html#Syntax), it seems that `pint` formatting `~C`  (see [~](https://pint.readthedocs.io/en/stable/formatting.html#modifiers) and [C](https://pint.readthedocs.io/en/stable/formatting.html#unit-format-types)) is compatible with the udunits2 syntax (I couldn't check the log operator, if someone know how to make a unit with log?).

I propose to use this `~C` to stick to a syntax support by `pint` rather than using a custom formatter.

woutdenolf (approved):
LGTM

PR #3686, opened 2022-09-29
ArrayTableWidget editing without clearing previous data

### LINKED ISSUE #3682: ArrayTableWidget editing without clearing previous data
Hi everybody,

If the data in a cell of a `silx.gui.data.ArrayTableWidget` is about to be edited (for example by double clicking it), the current data in this cell is not shown in the editor:
grafik
grafik

I'd say that it would be better if the editor would be filled with the current data:
grafik

### Possible implementation
To enable this behaviour, the `silx.gui.data.ArrayTableWidget.data` method can return the data of the cell for the Qt.EditRole.

Here is my version, where I just return the data as str (since it is anyways converted back to the array dtype in `setData` ):
tifuchs@22158acc45b60b281b42ae162df29694002f50b7

Can I create a PR for this?

Cheers,
Timo

tifuchs:
Return contents of array with `Qt.EditRole` in `silx.gui.data.ArrayTableWidget.data` method to retain contents of the cell, if edited.

grafik

Closes #3682.


vallsv:
I think it's fine.

I was expecting sure default behaviour with None result.

PR #3688, opened 2022-10-04
Allow to check AA_ShareOpenGLContexts in isOpenGLAvailable

vallsv:
Allow to check `AA_ShareOpenGLContexts` in `isOpenGLAvailable`

Default is false.

Changelog: 
- Allow to check AA_ShareOpenGLContexts in isOpenGLAvailable


t20100 (approved):
LGTM

PR #3693, opened 2022-10-11
Fixed Debian/Ubuntu packaging

### LINKED ISSUE #3687: `silx view` depends on `hdf5plugin`
`silx view` crashes at start-up when installed from a debian package:
```
pkg_resources.DistributionNotFound: The 'hdf5plugin' distribution was not found and is required by silx
```

t20100:
This should close #3687


t20100:
Tested on debian10, it fixes the issue #3687.

kif:
I confirm it work ... tested on scisoft12 without hdf5plugin installed.

PR #3695, opened 2022-10-14
silx.gui.plot.PlotWidget: Fixed OpenGL rendering

### LINKED ISSUE #3681: [plot gl] ROIs not displayed with solid lines
Might be related to #3595.

grafik

grafik

### Way to reproduce:

1. Before running `examples/plotInteractiveImageROI.py`, change line https://github.com/silx-kit/silx/blob/1ae49bae6ef218a73b15e555563713bee06ed341/examples/plotInteractiveImageROI.py#L81 to `roi.setLineStyle('-')`.
2. Toggle on OpenGL rendering. Or create Plot2D directly with `backend='opengl'`

Sometimes the rois still appear with the opengl backend on, but only distorted.

### Reproduced:

on Win10, silx 1.0.0 and latest version 1ae49bae6ef218a73b15e555563713bee06ed341. On two machines.

t20100:
This PR changes the way anchors are rendered which was causing issues with some GPUs.

It is now passing vertex attributes directly from memory (the old way) as the other rendering do rather than creating/deleting a vertex buffer object.

closes #3681

### DIFF
--- a/ci/requirements-pinned.txt
+++ b/ci/requirements-pinned.txt
@@ -9,3 +9,6 @@ pybind11  # Required to build pyopencl
 # Anyway, we don't test OpenCL on appveyor
 pyopencl == 2020.3.1; sys_platform == 'win32' and python_version < '3.8'
 pyopencl == 2022.1.6; sys_platform == 'win32' and python_version >= '3.8'
+
+# Workaround https://github.com/matplotlib/matplotlib/issues/24155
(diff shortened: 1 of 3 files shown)


vasole:
There is an error in AppVeyor that might be relevant.

t20100:
Not sure, it is one of the recurrent one.
It would be good to fix all those CI random issues... but it requires time to devote to it.

PR #3702, opened 2022-10-24
silx.gui.plot: Updated `BandROI` item

t20100:
This PR addresses post-merge comments on PR #3680 (https://github.com/silx-kit/silx/pull/3680#issuecomment-1282635661 and https://github.com/silx-kit/silx/issues/3638#issuecomment-1284199029), it:
- Adds some basic tests
- Changes the rotation center from the center of the ROI to the other edge
- Inherits from `InteractionModeMixIn` in order to integrate the bounds mode change with context menu, click on selection ROI,.... This also allows to extend later for half-bounded ROI.

I also added a `setAvailableInteractionModes` so from the code, one can force the user to do either bounded or unbounded ROIs. This can be put in the `InteractionModeMixIn` so the `ArcROI` can also benefit from it.. but being close to the release I choose not to change the existing ROI, this can be done later if needed.

attn @kklmn

### DIFF
--- a/src/silx/gui/plot/items/_band_roi.py
+++ b/src/silx/gui/plot/items/_band_roi.py
@@ -28,7 +28,7 @@ import logging
 from typing import Iterable, List, NamedTuple, Optional, Sequence, Tuple
 import numpy
 
-from ... import utils
+from ... import qt, utils
 from .. import items
 from ...colors import rgba
 from silx.image.shapes import Polygon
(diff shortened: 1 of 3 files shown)


kklmn:
Hi,
I've tested the latest BandROI. Looks good. Pulling one end behaves as expected. Clicking on the handles to toggle the bounded state is lovely.
A suggestion: dragging with Shift pressed should orient it to the nearest 'round' angle: 0, 45 or 90.
Thank you for adding new features!

t20100:
Added support of the Shift key modifier, ready for review.

kklmn:
Tested. Works beautifully!

kklmn:
Hi,
I've just found one problem with the latest update of rois: `plot.clearImages()` also clears roi lines (only lines, not roi handles), and I think it shouldn't. At least it didn't in v1.0.0.

PR #3707, opened 2022-10-26
silx.gui: Fixes

t20100:
This PR fixes some issues found while testing `flint`:

- Remove call to `ElidedLabel` deprecated API (related to PR #3650)
- Disable OpenGL rendering if testing OpenGL version/extension fails to avoid exceptions at each render
- Fix `Shape` and `Line` items overlay (introduced in PR #3680)

Thanks @vallsv for the testing.


vallsv:
:+1:

PR #3725, opened 2022-11-22
Fixed support of `pint` >= 0.20

### LINKED ISSUE #3724: nexus_to_h5_dict function fails with Pint 0.20.1
Nexus writer uses `nexus_to_h5_dict` from `silx.io.dictdump`, but it fails with `AttributeError`:

```
   [...]
  File "/home/matias/miniconda3/envs/bliss_demo_env/lib/python3.9/site-packages/nexus_writer_service/io/nexus.py", line 712, in nxInit
    dicttonx(nxdict, h5group, update_mode="modify")
  File "/home/matias/miniconda3/envs/bliss_demo_env/lib/python3.9/site-packages/silx/io/dictdump.py", line 728, in dicttonx
    nxtreedict = nexus_to_h5_dict(
  File "/home/matias/miniconda3/envs/bliss_demo_env/lib/python3.9/site-packages/silx/io/dictdump.py", line 473, in nexus_to_h5_dict
    elif pint is not None and isinstance(value, pint.quantity.Quantity):
AttributeError: module 'pint' has no attribute 'quantity'
```

t20100:
This PR fixes `silx.io.dictdump` support of `pint` for versions >= 0.20.

At the same time, older versions of `pint` where `Quantity` is not available through `pint.Quantity` but through `pint.quantity.Quantity` are no longer supported (it's possible to add it if needed).

`pint` is also added to `requirements.txt` and so it is installed when running the CI... This would have spotted the removal of `pint.quantity` (see #3724).

closes #3724

### DIFF
--- a/src/silx/io/dictdump.py
+++ b/src/silx/io/dictdump.py
@@ -32,9 +32,12 @@ import numpy
 import os.path
 import h5py
 try:
-    import pint
+    from pint import Quantity as PintQuantity
 except ImportError:
-    pint = None
+    try:
+        from pint.quantity import Quantity as PintQuantity
+    except ImportError:
+        PintQuantity = None
 
 from .configdict import ConfigDict
 from .utils import is_group
@@ -67,7 +70,7 @@ def _prepare_hdf5_write_value(array_like):
         ``numpy.array()`` (`str`, `list`, `numpy.ndarray`…)
     :return: ``numpy.ndarray`` ready to be written as an HDF5 dataset
     """
-    if pint is not None and isinstance(array_like, pint.quantity.Quantity):
+    if PintQuantity is not None and isinstance(array_like, PintQuantity):
         return numpy.array(array_like.magnitude)
     array = numpy.asarray(array_like)
     if numpy.issubdtype(array.dtype, numpy.bytes_):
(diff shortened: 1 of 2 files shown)


vallsv:
I suggest to support both here. BLISS 1.10 will not disappear in a month.
If for some reason i want to the new silx in BLISS 1.10, i expect to be able to use it.

Or at least please check that an installation of BLISS 1.10 supports the new way to import the quantity.

vallsv:
I did it for you.

```
lbm05loka:~ % . blissenv
setting BLISS environment
Using BLISS development
(bliss_dev) lbm05loka:~ % python -c "from pint import Quantity"
(bliss_dev) lbm05loka:~ % python -c "import pint; print(pint.__version__)"
0.17
```

So i think this change is save.

src/silx/io/dictdump.py
```diff
@@ -32,9 +32,9 @@
 import os.path
 import h5py
 try:
-    import pint
+    from pint import Quantity as PintQuantity
 except ImportError:
-    pint = None
+    PintQuantity = None
```
woutdenolf, line 37: I propose

```python
try:
    from pint import Quantity as PintQuantity
except ImportError:
    try:
        from pint.quantity import Quantity as PintQuantity
    except ImportError:
        PintQuantity = None
```

Also it is no longer needed to `import pint`.

woutdenolf (changes requested):

src/silx/io/dictdump.py
t20100, line 37:
```suggestion
    try:
        from pint.quantity import Quantity as PintQuantity
    except ImportError:
        PintQuantity = None
```

t20100: Done

woutdenolf (approved):
LGTM

t20100:
Rebased to fix CI

PR #3727, opened 2022-11-29
CI: Use `intel-opencl-icd` package

t20100:
Use the `intel-opencl-icd` from ubuntu rather than downloading an older version.
This fixes the CI that was broken with ubuntu22.04


pierrepaleo:
Thanks! It's a good thing that we could replace this old `icd` file with a packaged one.

PR #3729, opened 2022-11-30
Updated changelog, bump to version 1.1.1

t20100:
This PR prepares a v1.1.1 version.


vallsv:
Do you think it really makes sense to log in the change log that you change the change log?

PR #3732, opened 2022-12-12
Update rules

kif:
Could someone check if i is the proper way ?

### DIFF
--- a/package/debian10/rules
+++ b/package/debian10/rules
@@ -70,7 +70,7 @@ override_dh_python3:
 # https://lists.debian.org/debian-python/2017/08/msg00095.html
 override_dh_auto_test:
 	mkdir -p $(POCL_CACHE_DIR) # create POCL cachedir in order to avoid an FTBFS in sbuild
-	dh_auto_test -- -s custom --test-args="env PYTHONPATH={build_dir} GPU=False WITH_QT_TEST=False SILX_OPENCL=False SILX_TEST_LAW_MEM=True xvfb-run -a --server-args=\"-screen 0 1024x768x24\" {interpreter} run_tests.py -vv --installed"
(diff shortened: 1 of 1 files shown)


package/debian10/rules
```diff
@@ -70,7 +70,7 @@ override_dh_python3:
 # https://lists.debian.org/debian-python/2017/08/msg00095.html
 override_dh_auto_test:
 	mkdir -p $(POCL_CACHE_DIR) # create POCL cachedir in order to avoid an FTBFS in sbuild
-	dh_auto_test -- -s custom --test-args="env PYTHONPATH={build_dir} GPU=False WITH_QT_TEST=False SILX_OPENCL=False SILX_TEST_LAW_MEM=True xvfb-run -a --server-args=\"-screen 0 1024x768x24\" {interpreter} run_tests.py -vv --installed"
+	dh_auto_test -- -s custom --test-args="env PYTHONPATH={build_dir} GPU=False WITH_QT_TEST=False SILX_OPENCL=False SILX_TEST_LOW_MEM=True xvfb-run -a --server-args=\"-screen 0 1024x768x24\" {interpreter} run_tests.py -vv --installed"
```
t20100, line 73: There is an issue in the way we interpret the env. var., to disable it, it's:
```suggestion
	dh_auto_test -- -s custom --test-args="env PYTHONPATH={build_dir} GPU=False WITH_QT_TEST=False SILX_OPENCL=False SILX_TEST_LOW_MEM=False xvfb-run -a --server-args=\"-screen 0 1024x768x24\" {interpreter} run_tests.py -vv --installed"
```
As used in https://salsa.debian.org/science-team/silx/-/blob/master/debian/rules

PR #3733, opened 2022-12-12
Dedicated function to parse gl version

vallsv:
I found that the GL version was not properly parsed.

It was returning `51, 48` for a string version `3.0`.
Because the string was bytes and indexes was returning ordinal of char `3` and `0`

This parsing issue was probably broken in python 3 since the beginning. Or it is a change in the behavior of PyOpenGL.

This PR create a dedicated function to return this version.

Changelog: 
- Fix GL version parsing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,7 @@ jobs:
         run: |
           pip install pytest
           pip install pytest-xvfb
+          pip install pytest-mock
 
       - name: Install silx package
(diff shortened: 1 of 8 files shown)


t20100 (approved):
LGTM

t20100:
`pytest-mock` is missing in github actions config (same needs to be done as for appveyor)

t20100:
It would be good to put that into 1.1.2 as well.

PR #3737, opened 2022-12-15
silx.gui: Fixed support of PySide6.4 enums

### LINKED ISSUE #3730: [GUI] PySide6 6.4 issues with file dialog enums
Screenshot_2022-12-01_16-28-29

t20100:
This PR solves issue #3730 for `QFileDialog`.

I also checked similar cases for all `QDialog` and inherited classes.

closes #3730 even if there is probably other enum issues but the one reported should be fixed.


t20100:
BTW, I'm using `qt.QFileDialog.DontUseNativeDialog` rather than `qt.QFileDialog.Option.DontUseNativeDialog` to keep supporting PyQt5 < 5.11

vasole (approved):
Enough for the time being.

PR #3739, opened 2022-12-16
Prepare v1.1.2

t20100:
Backports:
- #3737
- #3736
- #3735
- #3733 + adding `pytest-mock` to Debian packaging `Build-depends`
- #3732

Update changelog and bump to v1.1.2

PR #3740, opened 2022-12-16
Added `pytest-mock` to Debian build dependencies

t20100:
`pytest-mock` was missing in Debian `Build-Depends` to run the tests during packaging.

Related to #3733


t20100:
Also update copyright date as in #3739

PR #3742, opened 2022-12-21
Fix histogram visibility

vallsv:
This PR fixes the `setSideHistogramDisplayed(False)` when the widget or one of the ancestor is not visible.

`isVisible()` is one of the property which is not symmetric to `setVisible()`.

As result `isSideHistogramDisplayed` was false is some cases for false reason.
Which was inhibiting `setSideHistogramDisplayed`.

Changelog: 
- Fix ImageView histogram visibility


vallsv:
The following code from ImageStack`looks to have the same problem
```
    def urlSelectionTableIsVisible(self):
        return self._urlsTable.isVisible()
```

Issue #3747, opened 2023-01-17
h5todict string issue


woutdenolf:
```
Traceback (most recent call last):
  File "inspect_h5master.py", line 6, in <module>
    adict = nxtodict(master, errors="ignore", asarray=False)
  File "/users/denolf/virtualenvs/ewoks/ubuntu_20_04/py38/lib/python3.8/site-packages/silx/io/dictdump.py", line 746, in nxtodict
    nxtreedict = h5todict(h5file, include_attributes=include_attributes, **kw)
  File "/users/denolf/virtualenvs/ewoks/ubuntu_20_04/py38/lib/python3.8/site-packages/silx/io/dictdump.py", line 621, in h5todict
    for aname, avalue in attrs.items():
  File "/users/denolf/virtualenvs/ewoks/ubuntu_20_04/py38/lib/python3.8/site-packages/silx/io/utils.py", line 1170, in items
    yield k, self[k]
  File "/users/denolf/virtualenvs/ewoks/ubuntu_20_04/py38/lib/python3.8/site-packages/silx/io/utils.py", line 1159, in __getitem__
    return h5py_encode_value(value, encoding="utf-8")
  File "/users/denolf/virtualenvs/ewoks/ubuntu_20_04/py38/lib/python3.8/site-packages/silx/io/utils.py", line 1053, in h5py_encode_value
    return value.encode(encoding, errors=errors)
AttributeError: 'numpy.bytes_' object has no attribute 'encode'
```

t20100:
Do you have a file to reproduce?
Which version of h5py do you have the problem with?

woutdenolf:
I'm not sure I can share the file. It was not produced by `h5py`.

t20100:
Looking at this, the problem is with "fixed-length" ASCII strings attributes which does not get decoded as "variable-length" ASCII strings.
This raises the same exception:

```python
f = h5py.File("test.h5", "w")
f.attrs["ascii_fl"] = numpy.array("text".encode("ascii"), dtype=h5py.string_dtype('ascii', 5))
# f.attrs["ascii_vl"] = numpy.array("text".encode("ascii"), dtype=h5py.string_dtype('ascii', None))
f.close()
```

See https://docs.h5py.org/en/stable/strings.html

t20100:
The behavior of ASCII fixed/variable length string is not consistent with datasets:
ASCII string attributes:
- Fixed-length -> numpy.bytes_
- Variable-length -> str
ASCII string datasets:
[...]

PR #3748, opened 2023-01-17
3747 h5todict string issue

woutdenolf:
Changelog: fix h5todict fixed-length string issue

### DIFF
--- a/src/silx/io/utils.py
+++ b/src/silx/io/utils.py
@@ -1025,6 +1025,25 @@ def vol_to_h5_external_dataset(vol_file, output_url, info_file=None,
                                           overwrite=overwrite)
 
 
+def hdf5_to_python_type(value, decode_ascii, encoding):
+    """Convert HDF5 type to proper python type.
+
+    :param value:
+    :param bool decode_ascii:
+    :param encoding str:
+    """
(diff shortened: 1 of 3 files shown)


woutdenolf:
I need to add a test for this: https://github.com/silx-kit/silx/issues/3747#issuecomment-1396683564

src/silx/io/utils.py
```diff
@@ -1050,13 +1050,47 @@ def h5py_encode_value(value, encoding="utf-8", errors="surrogateescape"):
...
+    """
+    try:
+        string = string.item()
+    except AttributeError:
+        pass  # not a numpy type
+    try:
+        return string.encode(encoding, errors=errors)
+    except AttributeError:
+        pass  # not bytes
+    return string
```
t20100, line 1091: What about this:
```suggestion
    if isinstance(string, bytes):
        return string
    return string.encode(encoding, errors=errors)
```

It should work since:
```python
In [70]: isinstance(numpy.bytes_(b"aaa"), bytes)
Out[70]: True
```

woutdenolf:
I made sure it works for `h5py >=2.8`

t20100 (approved):
LGTM

PR #3770, opened 2023-03-02
Fixed support of cython 3.0.0b1

t20100:
This PR makes silx build with cython3.0.0b1 regarding the support of power `**`.
This PR makes direct use of the `pow` function from `libc.math` since this looks to work across different versions of cython (0.29.x and 3.0).

Some background information for the record:

See https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#power-operator

Also cython 3.0.0b1 introduced a `cpow` [compiler directive](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives) to use the C `pow` function for `**`... but this directive is not available in previous versions of Cython.

<details>
<summary>Details of different behaviors from cython.cpow doc</summary>

Type of a | Type of b | cpow==True | cpow==False
-- | -- | -- | --
C integer | Negative integer compile-time constant | Return type is C double | Return type is C double (special case)
C integer | C integer (known to be >= 0 at compile time) | Return type is integer | Return type is integer
C integer | C integer (may be negative) | Return type is integer | Return type is C double (note that Python would dynamically pick int or float here, while Cython doesn’t)
C floating point | C integer | Return type is floating point | Return type is floating point
C floating point (or C integer) | C floating point | Return type is floating point, result is NaN if the result would be complex | Either a C  real or complex number at cost of some speed

</details>

### DIFF
--- a/src/silx/math/_colormap.pyx
+++ b/src/silx/math/_colormap.pyx
@@ -1,6 +1,6 @@
 # /*##########################################################################
 #
-# Copyright (c) 2018-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2018-2023 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -34,6 +34,7 @@ cimport cython
 from cython.parallel import prange
 cimport numpy as cnumpy
 from libc.math cimport frexp, sinh, sqrt
+from libc.math cimport pow as c_pow
 from .math_compatibility cimport asinh, isnan, isfinite, lrint, INFINITY, NAN
 
 import logging
@@ -227,7 +228,7 @@ cdef class LogarithmicNormalization(Normalization):
         return result
 
     cdef double revert_double(self, double value, double vmin, double vmax) nogil:
-        return 10**value
+        return c_pow(10, value)
 
 
 cdef class ArcsinhNormalization(Normalization):
@@ -247,7 +248,7 @@ cdef class SqrtNormalization(Normalization):
         return sqrt(value)
 
     cdef double revert_double(self, double value, double vmin, double vmax) nogil:
-        return value**2
+        return value*value
 
 
 cdef class PowerNormalization(Normalization):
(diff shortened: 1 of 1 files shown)


src/silx/math/_colormap.pyx
```diff
@@ -247,7 +248,7 @@ cdef class SqrtNormalization(Normalization):
         return sqrt(value)
 
     cdef double revert_double(self, double value, double vmin, double vmax) nogil:
-        return value**2
+        return value*value
```
t20100, line 251: Not needed but I'd expect `*` to be faster than `pow`

src/silx/math/_colormap.pyx
```diff
@@ -268,6 +269,7 @@ cdef class PowerNormalization(Normalization):
         # Needed for multiple inheritance to work
         pass
 
+    @cython.cdivision(True)
```
t20100, line 272: Simplifies generated C code

src/silx/math/_colormap.pyx
```diff
@@ -276,15 +278,16 @@ cdef class PowerNormalization(Normalization):
...
     cdef double revert_double(self, double value, double vmin, double vmax) nogil:
         if value <= 0.:
             return vmin
         elif value >= 1.:
             return vmax
         else:
-            return vmin + (vmax - vmin) * value**(1.0/self.gamma)
+            return vmin + (vmax - vmin) * c_pow(value, (1.0/self.gamma))
```
t20100, line 290: That's the only one that was actually failing

src/silx/math/_colormap.pyx
```diff
@@ -34,6 +34,7 @@ cimport cython
 from cython.parallel import prange
 cimport numpy as cnumpy
 from libc.math cimport frexp, sinh, sqrt
+from libc.math cimport pow as c_pow
```
t20100, line 37: Using the `pow` function directly avoids issues and extra code that handles complex.

src/silx/math/_colormap.pyx
kif: makes sense

kif:
LGTM

PR #3772, opened 2023-03-08
Fixed `build-deb.sh`: Avoid to run Python from `src/silx`

### LINKED ISSUE #3771: build-deb.sh fails to run
Raised in https://github.com/silx-kit/silx/issues/3443#issuecomment-1452052407 by @tobiasvos:

I get the following error when trying to build a Debian package (and even with just `./build-deb.sh --help`):

Fatal Python error: init_sys_streams: can't initialize sys standard streams
Python runtime state: core initialized
```
  File "silx/src/silx/io/__init__.py", line 36, in <module>
  File "silx/src/silx/io/utils.py", line 34, in <module>
  File "/usr/lib/python3.9/logging/__init__.py", line 26, in <module>
  File "/usr/lib/python3.9/traceback.py", line 5, in <module>
  File "/usr/lib/python3.9/linecache.py", line 11, in <module>
  File "/usr/lib/python3.9/tokenize.py", line 27, in <module>
ImportError: cannot import name 'open' from 'builtins' (unknown location)
```
The issue is that the silx/src/io subpackage shadows the standard library's io module. This issue can be circumvented by renaming src/silx/io to e.g. src/silx/_io, but then the build process still fails because silx.io.xxx is expected in many places.

Any clue how to get around this?

t20100:
This PR provides a small fix to retrieve the version information without starting Python from the `src/silx` folder to avoid issues with e.g. `import io` during Python start-up.

closes #3771

### DIFF
--- a/build-deb.sh
+++ b/build-deb.sh
@@ -63,12 +63,12 @@ project_directory="`( cd \"$project_directory\" && pwd )`" # absolutized
 dist_directory=${project_directory}/dist/${target_system}
 build_directory=${project_directory}/build/${target_system}
 
-# Get version info
-cd ${project_directory}/src/${project}
-version=$(python3 -c"import _version; print(_version.version)")
-strictversion=$(python3 -c"import _version; print(_version.strictversion)")
-debianversion=$(python3 -c"import _version; print(_version.debianversion)")
 cd ${project_directory}
+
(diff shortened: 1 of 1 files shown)


kif:
LGTM

PR #3788, opened 2023-04-17
[silx compare] Create a dedicated app to compare images

### LINKED ISSUE #3759: Promote "example/compareImages" into "silx compare"
This is just convenient.

There is no equivalent as far as i can say.

There is few improvement i would like to add.

- Better handling of the autoscale
- Autoscale applyed on the full frame, and not on half of the frame
- In autoscale, the autoscale have to be applied independently to the 2 images, actually it is kind of weird
- Support URLs with `*`
- Order the URLs taking care of the numerical order in strings:  `2` is before `10`
- Dont use resetzoom when images are selected

vallsv:
This PR introduce a `silx compare` UI to make is easy to use.

```
silx compare --help
usage: silx compare [-h] [--debug] [--use-opengl-plot] [files [files ...]]

GUI to compare images

positional arguments:
  files              Image data to compare (HDF5 file with path, EDF files,
                     JPEG/PNG image files). Data from HDF5 files can be
                     accessed using dataset path and slicing as an URL:
                     silx:../my_file.h5?path=/entry/data&slice=10 EDF file
                     frames also can can be accessed using URL:
                     fabio:../my_file.edf?slice=10 Using URL in command like
                     usually have to be quoted: "URL".

optional arguments:
  -h, --help         show this help message and exit
  --debug            Set logging system in debug mode
  --use-opengl-plot  Use OpenGL for plots (instead of matplotlib)
```

```
silx compare ../bliss.git/blissdemo/src/blissdemo/demo_data/ID11_MA4752.h5::**/data ../bliss.git/blissdemo/src/blissdemo/demo_data/ID13_kevlar.h5::**/data
```

image

```
silx compare /tmp/RECONSTRUCTIONS.h5::ad1b7ad99b94749cd4a92a75eef06f18a69c0201/*
```

Screenshot from 2023-04-17 10-27-44

Changelog:
- `silx compare`:
    -  Create a `silx compare` UI from an example

closes #3759

### DIFF
--- a/src/silx/gui/plot/CompareImages.py
+++ b/src/silx/gui/plot/CompareImages.py
@@ -947,17 +947,17 @@ class CompareImages(qt.QMainWindow):
         alignmentMode = self.getAlignmentMode()
         self.__transformation = None
 
-        if self.__raw1 is None or self.__raw2 is None:
+        if raw1 is None or raw2 is None:
             # No need to realign the 2 images
             # But create a dummy image when there is None for simplification
             if raw1 is None:
-                raw1 = numpy.empty((1, 1))
-                raw1[0, 0] = numpy.nan
+                data1 = numpy.empty((0, 0))
+            else:
+                data1 = raw1
             if raw2 is None:
-                raw2 = numpy.empty((1, 1))
-                raw2[0, 0] = numpy.nan
-            data1 = raw1
-            data2 = raw2
+                data2 = numpy.empty((0, 0))
+            else:
+                data2 = raw2
             self.__matching_keypoints = None
         else:
             if alignmentMode == AlignmentMode.ORIGIN:
@@ -1004,14 +1004,22 @@ class CompareImages(qt.QMainWindow):
 
         mode = self.getVisualizationMode()
         if mode == VisualizationMode.COMPOSITE_RED_BLUE_GRAY_NEG:
-            data1 = data1.astype(numpy.float64) - data2.astype(numpy.float64)
+            data1 = self.__composeImage(data1, data2, mode)
             data2 = numpy.empty((0, 0))
         elif mode == VisualizationMode.COMPOSITE_RED_BLUE_GRAY:
             data1 = self.__composeImage(data1, data2, mode)
             data2 = numpy.empty((0, 0))
         elif mode == VisualizationMode.COMPOSITE_A_MINUS_B:
-            data1 = self.__composeImage(data1, data2, mode)
-            data2 = numpy.empty((0, 0))
+            data1 = self.__asIntensityImage(data1)
+            data2 = self.__asIntensityImage(data2)
+            if raw1 is None:
+                data1 = data2
+                data2 = numpy.empty((0, 0))
+            elif raw2 is None:
+                data2 = numpy.empty((0, 0))
+            else:
+                data1 = data1.astype(numpy.float32) - data2.astype(numpy.float32)
+                data2 = numpy.empty((0, 0))
         elif mode == VisualizationMode.ONLY_A:
             data2 = numpy.empty((0, 0))
         elif mode == VisualizationMode.ONLY_B:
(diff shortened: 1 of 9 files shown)


vallsv:
- Some of this commits could be merged independently (launcher, app utils, colors, io utils)
- The breaking change on UrlSelection is needed because it overwrite a Qt method (we optionally could create a UrlSelection2 to avoid that)

src/silx/gui/plot/CompareImages.py
```diff
@@ -927,8 +931,12 @@ def __updateData(self, updateColormap):
...
         """
         raw1, raw2 = self.__raw1, self.__raw2
-        if raw1 is None or raw2 is None:
-            return
+        if raw1 is None:
+            raw1 = numpy.empty((1, 1))
+            raw1[0, 0] = numpy.nan
+        if raw2 is None:
+            raw2 = numpy.empty((1, 1))
+            raw2[0, 0] = numpy.nan
```
payno, line 939: This modification seems to make some stuff failing on my side. Because `setVisualizationMode` is called and this triggers `self.__updateColormap` which get issues because of data1 and data2 values.

But I don't really see why raw1 and raw2 become 2D numpy arrays with a nan ?

payno (commented):
Nice, what is the scope of the application ? Because I have some use cases where it would be nice to browse a file (hdf5) to collect dataset, potentially 3D and allow user to compare between frames.

Regarding the app name maybe I would highlight more the fact that it allow comparison of images. Otherwise maybe some will complain that plot comparison for example is not possible. So maybe compare-imgs

src/silx/gui/plot/CompareImages.py
vallsv: The aim is to allow to compare a set image1 with an unset image2 (in case image2 is not yet set or a problem occured during the loading of image2)

So the user can see that one or the other image is missing.

Previously the previous state of the image2 was still displayed.

To make it easy, i have just replaced the None with a value which behave like an normal image (with NaN to make sure it is not taken by the user as a real image), so i know that any further processing will work.

But we can take a look at it together if you like.

vallsv:
> Regarding the app name maybe I would highlight more the fact that it allow comparison of images. Otherwise maybe some will complain that plot comparison for example is not possible. So maybe compare-imgs

I prefer to make it simple. Because we still can implement the comparison of 1D arrays or 3D arrays if we like, without the need of creating something new. But will this tool is usable somehow i am fine.

vallsv:
Here is a clean up of the branch.
This can be reviewed.
@payno tell me if you have some time, so what we can check together your regression.

t20100 (approved):

src/silx/gui/plot/CompareImages.py
```diff
@@ -1012,15 +1020,27 @@ def __updateColormap(self):
...
+                    if vmin is None:
+                        return vmax
+                    if vmax is None:
+                        return vmin
+                    return func(vmin, vmax)
+                vmin = vreduce(range1.minimum, range2.minimum, min)
+                vmax = vreduce(range1.maximum, range2.maximum, max)
+                if vmin is None or vmax is None:
+                    return 0, 1
+                return vmin, vmax
```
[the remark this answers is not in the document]
vallsv: :+1: Yes this could just be a function.

vallsv:
Added some extra tests and fix.
This looks to be fine now. But @payno will take a look again next week.

vallsv:
@t20100 the last remaining thing is if the image composition display an RGB image, the colormap will display the colormap of the scatter i use to display the control points. I would prefer it to be empty. Any idea how to disable this behaviour?

That's due to the action widget opening the colormap dialog: `_updateColormap`.
What about some extra setter to allow to filter some stuffs on this control?
One idea would be to allow to filter unselectable items.

Anyway i prefer to fix it later. Better to merge this PR first if it is possible.

payno:
LGTM.

t20100:
> One idea would be to allow to filter unselectable items.
@vallsv there is already a way to do this by not making the items "active" which is the default with `PlotWidget.add*` methods

PR #3794, opened 2023-05-04
color: Exception if vrange is not finite

vallsv:
Changelog: 
- `Colormap.setVRange` raises an exception if the range is not finite


t20100 (approved):
LGTM

PR #3795, opened 2023-05-04
ioutils: Supports recursive match "**"

vallsv:
Changelog: 
- silx view:
    - Supports data URL containing `**` to match multiple sub groups


t20100 (approved):
LGTM

Issue #3754, opened 2023-01-24
Remove deprecated code


t20100:
There's quite a bit of methods/functions/modules that are deprecated in `silx` (I counted more than a hundred...) since quite some time (the oldest deprecation I found was from v0.5.0 - May 2017)...

That would be good to do some clean-up, there's been plenty of time to mitigate this in code using silx.

I'd keep this for a major release, i.e., v2.0

t20100:
Next release should be 2.0.
So it's time for this kind of clean-up.
I propose to remove all depreacted features from the v0.x versions.

t20100:
That should be it for all deprecated features from v0.x

PR #3799, opened 2023-05-05
 Removed features deprecated from v0.8 to v0.12

t20100:
Second round of deprecated features removal marked as such since at least January 2020.

related to #3754


t20100:
I think it was to be inline with other `Item`s having a `setName`.

I removed this change from here and put it in a separate PR (#3810) for discussion.

vallsv (approved):

PR #3800, opened 2023-05-05
silx.gui.data.DataView: Removed patch for pymca <v5.3.0 support

t20100:
This should no longer be needed.


vasole:
I have not seen the warning displayed. Therefore I conclude that piece of code is not needed.

PR #3803, opened 2023-05-05
silx.gui.plot: Deprecate PlotWidget's Curve and Image sequence-like access

t20100:
This PR deprecates the access of Curve and Images PlotWidget Items as sequence-like that was there initially.

The same information is accessible through those Item's API, so let's deprecate the compatibilty mechanism.

attn @vasole do you rely on this compatibility in pymca?

### DIFF
--- a/src/silx/gui/plot/items/curve.py
+++ b/src/silx/gui/plot/items/curve.py
@@ -1,6 +1,6 @@
 # /*##########################################################################
 #
-# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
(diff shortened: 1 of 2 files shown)


t20100:
That's maybe been merged a bit too fast, I should have put it as draft, sorry.

There is some issue in the CI that does not seems to be the usual random ones and it would have been good to make sure this is not used in pymca.

vasole:
PyMca only access the via the plot getAllCurves and similar higher level methods but not directly via `__getitem__`

PR #3805, opened 2023-05-05
silx.gui.plot.StackView: Removed `setColormap` `autoscale` argument

t20100:
This PR removes the `autoscale` argument of `StackView.setColormap` that is marked as deprecated since  v0.14.

related to #3754

### DIFF
--- a/src/silx/gui/plot/StackView.py
+++ b/src/silx/gui/plot/StackView.py
@@ -56,7 +56,7 @@ Example::
 
 
     sv = StackViewMainWindow()
-    sv.setColormap("viridis", vmin=numpy.min(mystack), vmax=numpy.max(mystack))
+    sv.setColormap("viridis", vmin=-4, vmax=4)
     sv.setStack(mystack)
     sv.setLabels(["1st dim (0-99)", "2nd dim (0-199)",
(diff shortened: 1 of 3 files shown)


src/silx/gui/plot/StackView.py
```diff
@@ -56,7 +56,7 @@
 
 
     sv = StackViewMainWindow()
-    sv.setColormap("jet", autoscale=True)
+    sv.setColormap("viridis", vmin=numpy.min(mystack), vmax=numpy.max(mystack))
```
vallsv, line 59: Maybe safer to use `minmax` (for nan/inf pbm)

vallsv: jet hihihi

t20100: Yes, but that's for sample code, then I would hard-code the value.

t20100:
Added an update of `actions/checkout` to `v3` because v2 relies on node12 which is soon to be removed from github.

src/silx/gui/plot/StackView.py
vallsv: Aaaaoutch, that's a big bunch of code for the documentation...

PR #3806, opened 2023-05-05
silx.io.dictdump: Removed `dicttoh5` `overwrite_data` argument

t20100:
This PR removes `dicttoh5`'s `overwrite_data` argument that is deprecated since v0.15.0.

Do you think it's OK, or it's best to keep maximum compatibility here?

related to #3754


woutdenolf (approved):
LGTM

PR #3808, opened 2023-05-05
Removed copy of scipy's Delaunay from third_party

### LINKED ISSUE #3675: Remove Qhull from third_party and use scipy
What about removing the copy of scipy's QHull Delaunay wrapper in `silx.third_party` and use `scipy` directly?
`scipy` can be kept optional by disabling features based on its availability.

That would simplify `setup.py` and specific stripping for Linux packaging (see #3674).

t20100:
This PR removes the `Delaunay` implementation from `third_party`, so there is no more a fallback in case `scipy` is not installed (yet it does not raise an exception).

It concerns:
- `PlotWidget` for the "Solid" visualization mode of `Scatter`
- `plot3d` for the "Solid" and "Lines" visualization modes of `Scatter2D`

In this scipy is not available, the feature relying on Delaunay are disabled: i.e., not available as an option in the associated widgets and displaying nothing (as if tesselation has failed) in case it is set from the API.

closes #3675


t20100:
> What has one to do to pass a "triangulator" to silx to recover the functionality without scipy?

We can add a function like `set_delaunay_function` to allow overriding the default behavior, or add a `silx.config.DELAUNAY_FUNCTION` (default `None` or "scipy") that one can set to a custom function/class.
The down side of the later is that it introduces using something that is not a basic type in the `config`.

I can propose something and we can discuss further.

t20100:
From what I tested so far, it seems we could easily switch to `matplotlib.tri` which would make it simpler for everyone.
I'll rework this PR to do so.

t20100:
This PR replaces the use of a copy of `scipy`'s `Delaunay` with usage of `matplotlib.tri.Triangulation`.
This removes a lot of code from third_party and simplifies the existing code which already had partial usage of matplotlib.

Ready for review.

vallsv (approved):

vallsv:
Sounds good, as it is already part of our dependencies.

PR #3809, opened 2023-05-09
This test fails on edna site with a CPU device

kif:
using PortableCL... disable it for now.

Changelog: skip test on broken platform


kif:
This test breaks on `edna-site`, fixed with this patch.

PR #3810, opened 2023-05-10
Remove deprecated methods RegionOfInterest.get|setLabel

t20100:
This deprecation removal was initially in PR #3799 but is extracted as a separate PR for discussion.

Related to PR #3799 and issue #3754


vallsv:
What i mean for the ROIs it is used to displayed a text like a label.

- I guess you could have 2 ROIs with the same `name`
- You can have space in the `name`

Maybe we just have to distinguish the 2 concepts

t20100:
This was deprecated in PR #2684 when unifying the API with those of `CurvesROIWidget` and to have something similar to plot `Item`s where the `name` is said to be the "legend".

BTW, the name does not need to be unique and can be any `str`.

I agree that there can be 2 slightly different purposes:
- A `name` that can be used to represent the ROI item in a list widget for example
- A `label` that is displayed next to the ROI in the plot

With current implementation there a single string for both, and since `name` is used elsewhere I would stick to it now.

If we need it, we can later add a `get|setLabel` again but with a different behaviour: if `label` is not `None`, it is used instead of `name` for the display in the plot.

vallsv:
I feel like it's a good idea.

The displayed stuff could be overwritten by setLabel in the future if we need.

But i think i already need such thing. Right now i have to hold some key/roi mapping on Flint.
This could be replaced by such `setName/setLabel`.

Let's merge that PR for now.

vallsv (approved):

vasole:
Well, there are two different concepts:

- One is to uniquely identify the ROI.
- The other one is to set what is displayed in the plot.

For instance, you can have multiple ROIs in a 1D plot, but all of them can display MIN and MAX when activated.

If silx is just a library for flint, you can forget my comment above.

I find all these plot deprecation going in the opposite direction to what silx intended to be.

PR #3811, opened 2023-05-10
Refactor app utils

vallsv:
This PR creates a `silx.app.utils` to share it with others silx app.

In my point of view everything inside `silx.app.*` is private stuff, so people should not use it.
So there is no backward compatibility here.

Plus, there is small change on `silx.io`, which is mostly typing stuff.

Changelog: 
-


t20100:
I agree that `silx.app` is private.. could even be renamed `silx._app` IMO

vallsv:
It's kind of big change, but it would make it really explicit.

t20100 (approved):

vallsv:
Here is the PR with cleaned history.

PR #3812, opened 2023-05-10
Debian12

kif:
Changelog: provide automatic packaging for debian12


t20100:
This also needs updating [build-deb.sh](https://github.com/silx-kit/silx/blob/main/build-deb.sh) to be effective, right?

kif:
Not necessary ... but neverthless a good comment.

PR #3814, opened 2023-05-11
Remove unused imports

t20100:
This PR cleans-up most of the unused imports spotted by `pylint`.
I left the ones with side effects (matplotlib, Qt bindings) and didn't touch `silx.opencl`.

### DIFF
--- a/src/silx/app/utils/test/test_parseutils.py
+++ b/src/silx/app/utils/test/test_parseutils.py
@@ -21,8 +21,6 @@
 #
 # ############################################################################*/
 
-import glob
-import logging
 import pytest
 import h5py
 from ..parseutils import filenames_to_dataurls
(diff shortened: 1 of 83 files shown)


vallsv:
:+1: 

Any idea if it is easy to drop this `import *`?

t20100:
> +1
> 
> Any idea if it is easy to drop this `import *`?

Which one?

vallsv:
The ones part of your changes. I think it was inside sift or processing.

t20100:
OK, if it's in the diff of this PR, then it's the `src/silx/image/*.py` which are wrapper to their counterparts in `src/silx/opencl/`.

IMO those modules could be deprecated and use directly `silx.opencl.*`.

PR #3815, opened 2023-05-11
Remove python2 support

t20100:
This PR removes most of the remaining code for supporting Python 2.

I have not removed code with a comment stating it's for supporting Python which can be hard to test under all conditions.


vallsv:
`src/silx/utils/test/test_enum.py` is about to test that Python enum works like Python enum?

In this case this could be dropped. No?

vallsv (approved):

PR #3818, opened 2023-05-11
silx.opencl.convolution: Removed Python 2 compatible code

t20100:
Replacing Python2 compatible code with commented Python3-only code.


t20100:
CI fails on unrelated tests

PR #3821, opened 2023-05-12
Fix url with slices

vallsv:
This was just not implemented.

Changelog:


src/silx/io/url.py
```diff
@@ -37,6 +37,37 @@
 _logger = logging.getLogger(__name__)
 
 
+def slice_to_string(s):
```
t20100, line 40: Best to keep it private:
```suggestion
def _slice_to_string(s):
```

src/silx/io/url.py
```diff
@@ -37,6 +37,37 @@
...
+    else:
+        raise TypeError("Unexpected slicing type. Found %s" % type(s))
+
+
+def slice_sequence_to_string(data_slice):
+    """Convert a Python slice sequence or a slice into a string"""
+    if isinstance(data_slice, Iterable):
+        return ",".join([slice_to_string(s) for s in data_slice])
+    else:
+        return slice_to_string(data_slice)
```
t20100, line 68:
```suggestion
        return ",".join([_slice_to_string(s) for s in data_slice])
    else:
        return _slice_to_string(data_slice)
```

src/silx/io/url.py
```diff
@@ -37,6 +37,37 @@
...
+        else:
+            result += f":{s.step}"
+        return result
+    elif isinstance(s, int):
+        return str(s)
+    else:
+        raise TypeError("Unexpected slicing type. Found %s" % type(s))
+
+
+def slice_sequence_to_string(data_slice):
```
t20100, line 63: Same, base private:
```suggestion
def _slice_sequence_to_string(data_slice):
```

src/silx/io/url.py
```diff
@@ -302,27 +333,14 @@ def path(self):
...
         else:
             queries = []
             if self.__data_path is not None:
                 queries.append("path=" + self.__data_path)
             if self.__data_slice is not None:
-                if isinstance(self.__data_slice, Iterable):
-                    data_slice = ",".join([slice_to_string(s) for s in self.__data_slice])
-                else:
-                    data_slice = slice_to_string(self.__data_slice)
+                data_slice = slice_sequence_to_string(self.__data_slice)
```
t20100, line 343:
```suggestion
                data_slice = _slice_sequence_to_string(self.__data_slice)
```

src/silx/io/url.py
t20100, line 63: Same, base private:
```suggestion
def _slice_sequence_to_string(data_slice):
```

src/silx/io/test/test_url.py
```diff
@@ -27,190 +27,236 @@
...
+        ((1,slice(1,2)), "silx:///foo.h5?slice=1,1:2"),
+    ]
+)
+def test_path_creation(data):
+    """make sure the construction of path succeed and that we can
+    recreate a DataUrl from a path"""
+    data_slice, expected_path = data
+    url = DataUrl(scheme="silx", file_path="/foo.h5", data_slice=data_slice)
+    path = url.path()
+    DataUrl(path=path)
```
t20100, line 261: This is not used, is it a typo?

src/silx/io/url.py
```diff
@@ -37,6 +37,37 @@
...
+        if s.start is None:
+            result += ":"
+        else:
+            result += f"{s.start}:"
+        if s.stop is not None:
+            result += f"{s.stop}"
+        if s.step is None:
+            result += ""
+        else:
+            result += f":{s.step}"
```
t20100, line 55:
```suggestion
        if s.step is not None:
            result += f":{s.step}"
```

src/silx/io/url.py
```diff
@@ -37,6 +37,37 @@
...
+def slice_to_string(s):
+    """Convert a Python slice into a string"""
+    if s == Ellipsis:
+        return "..."
+    elif isinstance(s, slice):
+        result = ""
+        if s.start is None:
+            result += ":"
+        else:
+            result += f"{s.start}:"
```
t20100, line 49: Could be:
```suggestion
        if s.start is None:
            result = ":"
        else:
            result = f"{s.start}:"
```

t20100 (approved):
Indeed support of slice was missing.

Let's keep helper functions (`slice_to_string` and `slice_sequence_to_string`) private to make it clear it's not part of the API.

src/silx/io/url.py
vallsv: I have to reuse it at another place. That's why it's public

src/silx/io/url.py
vallsv: I don't like, it breaks the flow logic

src/silx/io/test/test_url.py
vallsv: It is used, it checked that the url can be reparsed.

It's part of the previous implementation

src/silx/io/url.py
t20100: Up to you

src/silx/io/url.py
t20100: All right then.

t20100:
One corner case: How does it work with `()` as slicing?

vallsv:
(basically there is the same bug in UrlSelectionTable, cause it was a copy paste, i prefer to use the same base code for now)

vallsv:
@t20100  what `()` should do? i have no idea.

vallsv:
Actually

- `()` -> `slice=`
- `(1, )` -> `slice=1`

I suggest to rework that to have

- `()` -> `slice=,`
- `(1, )` -> `slice=1,`

but i am really not sure, it's maybe better to change that in another PR

t20100:
> what () should do? 

From https://numpy.org/doc/stable/user/basics.indexing.html :

> An empty (tuple) index is a full scalar index into a zero-dimensional array. x[()] returns a scalar if x is zero-dimensional and a view otherwise. On the other hand, x[...] always returns a view.

Basically it returns everything but it also works with scalars:

```
In [6]: s = numpy.array(1)
In [7]: s[()]
Out[7]: 1
In [8]: s[...]
Out[8]: array(1)
In [9]: s[:]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
```

```
In [10]: a = numpy.array([1])
In [11]: a[:]
Out[11]: array([1])
In [12]: a[...]
Out[12]: array([1])
In [13]: a[()]
Out[13]: array([1])
```

t20100:
Acutally it should be OK as it is.

vallsv:
Ok, nice, i prefer not to touch this semantic.

It's just that there is an inconsistency because `1` is the same as `(1,)`, which is a bit weird. But with this 2 entries the  result should also be the same so it's fine.

PR #3822, opened 2023-05-12
Urlselectiontable with 3 columns

vallsv:
Split the URL column in 3 columns for filename/datapath/slice.

This is just for readability.

And it make it easy to hide one of the columns from the API.

I have removed the `URL_COLUMN`, but we also can restore it by default if you prefer.
So it could be easy to display it and to hide the others on your side.

image

image

Changelog: 
- UrlSelectionTable: Split the URL column in 3 columns

### DIFF
--- a/src/silx/gui/widgets/UrlSelectionTable.py
+++ b/src/silx/gui/widgets/UrlSelectionTable.py
@@ -35,7 +35,7 @@ import logging
 from silx.gui import qt
 from silx.gui import utils as qtutils
 from silx.gui.widgets.TableWidget import TableWidget
-from silx.io.url import DataUrl
+from silx.io.url import DataUrl, slice_sequence_to_string
 from silx.utils.deprecation import deprecated, deprecated_warning
 from silx.gui import constants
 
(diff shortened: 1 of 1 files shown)


src/silx/gui/widgets/UrlSelectionTable.py
```diff
@@ -71,28 +71,31 @@ def isChecked(self) -> bool:
...
-                raise TypeError("Unexpected slicing type. Found %s" % type(data_slice))
-
-        text = os.path.basename(url.file_path())
-        if url.data_path() is not None:
-            text += f" {url.data_path()}"
-        if url.data_slice() is not None:
-            text += f" [{slice_to_string(url.data_slice())}]"
+        toolTip = self._url.path()
+
+        print(text)
```
payno, line 95: looks like a typo :)

payno (changes requested):
Fine with expect the typo.
Fine as well to remove `URL_COLUMN` from the API (thanks to moving to a major.)

payno (approved):

PR #3823, opened 2023-05-12
CompareImage: Feature a consistent autoscale

vallsv:
Feature a consistent autoscale for the CompareImage.

Now
- the autoscale statistics are computer globally for the 2 images
- it is also computed on the raw image (instead of in the sliced left/right image)
- This makes the autoscale usable, from the ColormapDialog, including 3stddev
- subsequently when you are about to slice, the statistics are not recomputed -> that's faster

Changelog: 
- CompareImage: Colormap autoscale is now global to the 2 raw images to compare
- CompareImage: `updateColormap=` was dropped, this can be set with `getColormap()`

### DIFF
--- a/src/silx/gui/plot/CompareImages.py
+++ b/src/silx/gui/plot/CompareImages.py
@@ -39,6 +39,7 @@ from silx.gui import qt
 from silx.gui import plot
 from silx.gui.colors import Colormap
 from silx.gui.plot import tools
+from silx.utils.deprecation import deprecated_warning
 from silx.utils.weakref import WeakMethodProxy
 from silx.gui.plot.items import Scatter
 
@@ -445,7 +446,7 @@ class CompareImages(qt.QMainWindow):
     def clear(self):
         self.setData(None, None)
 
-    def setData(self, image1, image2):
+    def setData(self, image1, image2, updateColormap="deprecated"):
         """Set images to compare.
 
         Images can contains floating-point or integer values, or RGB and RGBA
@@ -457,13 +458,16 @@ class CompareImages(qt.QMainWindow):
         :param numpy.ndarray image1: The first image
         :param numpy.ndarray image2: The second image
         """
+        if updateColormap != "deprecated":
+            deprecated_warning("Argument", "setData's updateColormap argument", since_version="2.0.0")
+
         self.__raw1 = image1
         self.__raw2 = image2
         self.__updateData()
         if self.isAutoResetZoom():
             self.__plot.resetZoom()
 
-    def setImage1(self, image1):
+    def setImage1(self, image1, updateColormap="deprecated"):
         """Set image1 to be compared.
 
         Images can contains floating-point or integer values, or RGB and RGBA
@@ -474,12 +478,15 @@ class CompareImages(qt.QMainWindow):
 
         :param numpy.ndarray image1: The first image
         """
+        if updateColormap != "deprecated":
+            deprecated_warning("Argument", "setImage1's updateColormap argument", since_version="2.0.0")
+
         self.__raw1 = image1
         self.__updateData()
         if self.isAutoResetZoom():
             self.__plot.resetZoom()
 
-    def setImage2(self, image2):
+    def setImage2(self, image2, updateColormap="deprecated"):
         """Set image2 to be compared.
 
         Images can contains floating-point or integer values, or RGB and RGBA
@@ -490,6 +497,9 @@ class CompareImages(qt.QMainWindow):
 
         :param numpy.ndarray image2: The second image
         """
+        if updateColormap != "deprecated":
+            deprecated_warning("Argument", "setImage2's updateColormap argument", since_version="2.0.0")
+
         self.__raw2 = image2
         self.__updateData()
         if self.isAutoResetZoom():
--- a/src/silx/gui/plot/CompareImages.py
+++ b/src/silx/gui/plot/CompareImages.py
@@ -32,6 +32,7 @@ __date__ = "23/07/2018"
 import logging
 import numpy
 import math
+from typing import Optional
 
 import silx.image.bilinear
 from silx.gui import qt
@@ -39,7 +40,6 @@ from silx.gui import plot
 from silx.gui.colors import Colormap
 from silx.gui.plot import tools
 from silx.utils.weakref import WeakMethodProxy
-from silx.math.combo import min_max
 from silx.gui.plot.items import Scatter
 
 from .tools.compare.core import sift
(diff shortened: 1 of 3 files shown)


src/silx/gui/plot/CompareImages.py
```diff
@@ -1132,9 +1176,22 @@ def __updateData(self, updateColormap):
...
+            colormap1 = None
+        self.__plot.addImage(data1, z=0, legend="image1", resetzoom=False, colormap=colormap1)
         self.__image1 = self.__plot.getImage("image1")
+
         if data2 is not None:
+            mode2 = self.__getImageMode(data2)
+            if mode2 == "intensity":
+                colormap2 = colormap
+            else:
+                colormap2 = None
```
t20100, line 1194: `colormap2` is never used.

t20100 (commented):
There's a miss-spelling of `pinned` and probably a missing change regarding `colormap2`.

Otherwise, it's OK for me. We can live with `ColormapWithPinnedData` as long as it is not exposed publicly.

vallsv:
I have committed an alternative implementation to drop the custom colormap class.

- The virtual item is used to hold the concatenation of the 2 images
- When the colormap is changed it update a sealed version of the colormap for the displayed images

That maybe a bit less efficient than the previous implementation, but this could be improved, and for now who cares.

The good thing is that the displayed histogram is properly computed from the concatenation (previously the histogram was computed from the crop of one of the images)

t20100:
Can you rebase to include commit https://github.com/silx-kit/silx/commit/ccdc13bbd4b8aa5097958ad8f3e0fd6e20b93cf6 from `main` to fix CI on Windows?

src/silx/gui/plot/CompareImages.py
```diff
@@ -32,14 +32,14 @@
...
 import numpy
 import math
+from typing import Optional
 
 import silx.image.bilinear
 from silx.gui import qt
 from silx.gui import plot
 from silx.gui.colors import Colormap
 from silx.gui.plot import tools
 from silx.utils.weakref import WeakMethodProxy
```
t20100, line 42:
```suggestion
from silx.utils.deprecation import deprecated_warning
from silx.utils.weakref import WeakMethodProxy
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -430,53 +445,53 @@ def __separatorMoved(self, pos):
     def clear(self):
         self.setData(None, None)
 
-    def setData(self, image1, image2, updateColormap=True):
+    def setData(self, image1, image2):
```
t20100, line 448: I'd keep compatibility here:
```suggestion
    def setData(self, image1, image2, updateColormap="deprecated"):
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -430,53 +445,53 @@ def __separatorMoved(self, pos):
...
         values, but should have comparable intensities.
 
         RGB and RGBA images are provided as an array as `[width,height,channels]`
-        of usigned integer 8-bits or floating-points between 0.0 to 1.0.
+        of unsigned integer 8-bits or floating-points between 0.0 to 1.0.
 
         :param numpy.ndarray image1: The first image
         :param numpy.ndarray image2: The second image
         """
         self.__raw1 = image1
```
t20100, line 460:
```suggestion
        if updateColormap != "deprecated":
            deprecated_warning("Argument", "setData's updateColormap argument", since_version="2.0.0")

        self.__raw1 = image1
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -430,53 +445,53 @@ def __separatorMoved(self, pos):
...
         self.__raw2 = image2
-        self.__updateData(updateColormap=updateColormap)
+        self.__updateData()
         if self.isAutoResetZoom():
             self.__plot.resetZoom()
 
-    def setImage1(self, image1, updateColormap=True):
+    def setImage1(self, image1):
```
t20100, line 466:
```suggestion
    def setImage1(self, image1, updateColormap="deprecated"):
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -430,53 +445,53 @@ def __separatorMoved(self, pos):
...
         Images can contains floating-point or integer values, or RGB and RGBA
         values, but should have comparable intensities.
 
         RGB and RGBA images are provided as an array as `[width,height,channels]`
-        of usigned integer 8-bits or floating-points between 0.0 to 1.0.
+        of unsigned integer 8-bits or floating-points between 0.0 to 1.0.
 
         :param numpy.ndarray image1: The first image
         """
         self.__raw1 = image1
```
t20100, line 477:
```suggestion
        if updateColormap != "deprecated":
            deprecated_warning("Argument", "setImage1's updateColormap argument", since_version="2.0.0")

        self.__raw1 = image1
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -430,53 +445,53 @@ def __separatorMoved(self, pos):
...
-        self.__updateData(updateColormap=updateColormap)
+        self.__updateData()
         if self.isAutoResetZoom():
             self.__plot.resetZoom()
 
-    def setImage2(self, image2, updateColormap=True):
+    def setImage2(self, image2):
```
t20100, line 482:
```suggestion
    def setImage2(self, image2, updateColormap="deprecated"):
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -430,53 +445,53 @@ def __separatorMoved(self, pos):
...
         Images can contains floating-point or integer values, or RGB and RGBA
         values, but should have comparable intensities.
 
         RGB and RGBA images are provided as an array as `[width,height,channels]`
-        of usigned integer 8-bits or floating-points between 0.0 to 1.0.
+        of unsigned integer 8-bits or floating-points between 0.0 to 1.0.
 
         :param numpy.ndarray image2: The second image
         """
         self.__raw2 = image2
```
t20100, line 493:
```suggestion
        if updateColormap != "deprecated":
            deprecated_warning("Argument", "setImage2's updateColormap argument", since_version="2.0.0")

        self.__raw2 = image2
```

t20100 (approved):
Great! Colormap range handling is as expected, thanks!

I'd keep backward compatibility for the `updateColormap` arguments, it's doesn't hurt.

src/silx/gui/plot/CompareImages.py
vallsv: I prefer not. It's pointless, and @payno is agree to drop that.

vallsv: Ok let's try it anyway, i have some time before the rain

vallsv:
While i was here, i have used the auto scale to improve the contrast on the composite image comparison.

The result is probably not extactly the same for the B&W composition. @payno would you have some times to check if it's fine for you?

vallsv:
@payno thinking about that a bit more, i think it is better to revert my change on the compare image for the B&W composition, it would be better to talk about together before merging.

t20100:
I'd be +1 to merge this PR as it was and work on additional feature/fix in another PR.

vallsv:
@t20100 i am fine with it, let's merge it as it is

t20100 (approved):
LGTM

PR #3827, opened 2023-05-15
Polish image compare app

vallsv:
Here is few fixes to make the UI a bit more usable.

This also includes a refactoring the to toolbuttons used to selection the modes (in order to expose them).

Main changes are:
- save/restore settings
- expose toolbutton as public
    - `getActionFromMode`, for general use case
    - `setVisibleModes`, for convenience
- fix broken `--use-opengl-plot`
- plus polish

### DIFF
--- a/src/silx/app/compare/CompareImagesWindow.py
+++ b/src/silx/app/compare/CompareImagesWindow.py
@@ -47,15 +47,15 @@ class CompareImagesWindow(qt.QMainWindow):
         silxIcon = icons.getQIcon("silx")
         self.setWindowIcon(silxIcon)
 
-        self.__settings = settings
-        if settings:
-            self.restoreSettings(settings)
-
         self._plot = CompareImages(parent=self, backend=backend)
         self._plot.setAutoResetZoom(False)
 
         self._selectionTable = UrlSelectionTable(parent=self)
         self._selectionTable.setAcceptDrops(True)
+
+        self.__settings = settings
+        if settings:
+            self.restoreSettings(settings)
 
         spliter = qt.QSplitter(self)
         spliter.addWidget(self._selectionTable)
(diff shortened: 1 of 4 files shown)


t20100 (approved):
LGTM

src/silx/app/compare/CompareImagesWindow.py
```diff
@@ -196,3 +221,7 @@ def restoreSettings(self, settings):
             self.resize(size)
         if isFullScreen:
             self.showFullScreen()
+        self._plot.setVisualizationMode(visualizationMode)
+        self._plot.setAlignmentMode(alignmentMode)
+        print(displayKeypoints)
```
payno, line 226:
```suggestion
```

src/silx/gui/plot/CompareImages.py
```diff
@@ -79,151 +81,297 @@ class AlignmentMode(enum.Enum):
...
+    :class:`CompareImages` widget.
+
+    :param Union[qt.QWidget,None] parent: Parent of this widget.
+    """
+    def __init__(self, parent=None):
+        qt.QToolBar.__init__(self, parent)
+
+        self.__compareWidget = None
+
+        self.__visualizationToolButton = VisualizationModeToolButton(self)
```
payno, line 366: I would need to get some control over the different buttons / actions (__visualizationToolButton, __alignmentToolButton and __displayKeypoints) in order for example to hide some when needed.
Either by moving them to protected or even better by having getters

payno (changes requested):
Nice improvement.

src/silx/gui/plot/CompareImages.py
vallsv: I think for you it is better to create your own toolbar. So you are free to do what you want.
But you can reuse the sub components.

vallsv: If you realy want to reuse everything, i will propose to move that code into `silx.gui.plot.tools`m because the module start to be huge.

We can talk about it into another PR.

PR #3828, opened 2023-05-16
`silx.gui.plot`: Added `margins` argument to `PlotWidget.setLimits`

### LINKED ISSUE #3753: PlotWidget: Add 'useDataMargins' to setLimits
Allow to use `setLimits` automatically with the default defined data margins.

Flint is actually using some private function and module to implement the same without redefining sstuffs.

This change would simplify few things, in a versatile way.

t20100:
This PR adds a `margins` argument to `PlotWidget.setLimits` which allows to choose between:
- `margins=True`: to apply the "default" margins from `PlotWigdet.getDataMargins()`
- `margins=False`: to apply no margins
- `margins=4-tuple of float`: to apply provided margins.

To achieve this, it moves some logic from `_forceResetZoom` to `setLimits`.

There is a few more minor things (see the different commits):
- Added typing to a few methods and some tests related to `get|setDataMargins`.
- To write those, I added a `qWidgetFactory` fixture to handle the life-cycle of QWidget in tests.
- Fix a warning in the tests 

closes #3753

### DIFF
--- a/src/silx/gui/plot/PlotWidget.py
+++ b/src/silx/gui/plot/PlotWidget.py
@@ -36,10 +36,10 @@ _logger = logging.getLogger(__name__)
 
 from collections import OrderedDict, namedtuple
 from contextlib import contextmanager
+from typing import Optional, Tuple
 import datetime as dt
 import itertools
 import numbers
-import typing
 import warnings
 
 import numpy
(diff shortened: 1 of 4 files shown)


src/silx/gui/plot/PlotWidget.py
```diff
@@ -2491,36 +2491,78 @@ def setGraphYLimits(self, ymin, ymax, axis='left'):
...
-    def setLimits(self, xmin, xmax, ymin, ymax, y2min=None, y2max=None):
+    def setLimits(
+        self,
+        xmin: float,
+        xmax: float,
+        ymin: float,
+        ymax: float,
+        y2min: Optional[float]=None,
+        y2max: Optional[float]=None,
+        margins: Union[bool, Tuple[float, float, float, float]]=False,
```
t20100, line 2502: New argument.
I hesitated with using `None` / `"default"` rather than `False` / `True`.

vallsv:
Looks good, thanks a lot.

PR #3830, opened 2023-05-17
Replace `OrderedDict` by `dict`

t20100:
There is no longer need to use `OrderedDict`, so replace it with `dict`.


payno:
LGTM. Can it be merged ?

vallsv:
There is no reason not to merge it i think.
If there is remaining stuffs, this can be fixed later.

PR #3833, opened 2023-05-22
`silx.gui.dialog`: Fixed `ColormapDialog` state when updating `Item`

### LINKED ISSUE #3745: Wrong colormap range in the dialog
THere is maybe a problem with the update of the colormap range in the dialog.

Noticed in Flint.

I have changed the order of some function to make it work (in Flint), but the problem is maybe still there.

My guess is the problem happen when:

```
item = Imageitem()
item.setColormap(a_colormap)
assert item.getData() is None  # At this place the data is None
dialog.setItem(item)
dialog.setColormap(item.getColormap())
item.setData(a_2d_array)

plot.add_the_item(item)
```

As result the histogram is right, but the range is wrong.

image

t20100:
This PR fixes the update of the `ColormapDialog` when the associated item is updated.

closes #3745

### DIFF
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -1263,8 +1263,7 @@ class ColormapDialog(qt.QDialog):
             if oldArray is array:
                 return
 
-        self._data = None
-        self._itemHolder = None
+        self.__resetItem()
         try:
             if item is None:
                 self._item = None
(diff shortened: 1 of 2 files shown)


vallsv (approved):

vallsv:
Do you think this expect an extra test, to avoid regression?

vallsv:
I think it is better to use a single factory which can provide different helper function.
It makes the code simpler, and easier to use in different context.

Anyway, it's fine like it is.

t20100:
There was some issues while using the factory fixture for existing colormap dialog tests (widget not being deleted in some cases) anyway, so I removed the modification for now and only use `qWidgetFactory` in the new test.

vallsv:
I think it's better to use a weakref instead of a proxy.

- The test will be released before the end of the fixture, so i think it is not a problem to pass the real object
- The test have to remove anyway the object from other stuffs, in case there is Qt signal registred or what

It will stay trick anyway.

t20100:
> I think it's better to use a weakref instead of a proxy

If it's not a proxy, then I would pass the widget instance directly... and the test is responsible for not keeping reference on it.

vallsv:
> > I think it's better to use a weakref instead of a proxy
> 
> If it's not a proxy, then I would pass the widget instance directly... and the test is responsible for not keeping reference on it.

I mean `qWidgetFactory` returns a proxy.

PR #3835, opened 2023-05-23
`silx.gui.plot`: Added `Plotwidget.get}setDefaultColors` and updated default colors behavior

### LINKED ISSUE #3757: Try not to copy config.DEFAULT_PLOT_CURVE_COLORS
The Plot1D offer 3 level of use customization for the color palette.

- `silx.config.DEFAULT_PLOT_CURVE_COLORS`
- `silx.gui.plot.PlotWidget._COLORLIST`
-`silx.gui.plot.PlotWidget.PlotWidget._colorList`

Each level is set at initialization/instantiation time, which make it difficult to setup.

I propose to let them set to `None` until somebody wants to override it. And to move the logic at the place it is needed (if that is defined, use that, else use that, else use that).

The same implementation could also be done transparently with properties.

This way, you could set `silx.config.DEFAULT_PLOT_CURVE_COLORS` when ever you want.

t20100:
This PR replaces `PlotWidget.colorList` class attribute with `PlotWidget.get|setDefaultColors` methods in order to simplify the ways to configure curve/histogram default colors.

With this PR, there is 2 ways:
- Globally with `silx.config.DEFAULT_PLOT_CURVE_COLORS`
- In a `PlotWidget` instance with `setDefaultColors`.

By default `PlotWidget` uses the current value of `silx.config.DEFAULT_PLOT_CURVE_COLORS`.

closes #3757

### DIFF
--- a/src/silx/gui/plot/PlotWidget.py
+++ b/src/silx/gui/plot/PlotWidget.py
@@ -68,10 +68,6 @@ from ._utils.panzoom import ViewConstraints
 from ...gui.plot._utils.dtime_ticklayout import timestamp
 
 
-
-_COLORDICT = colors.COLORDICT
-_COLORLIST = silx.config.DEFAULT_PLOT_CURVE_COLORS
-
 """
 Object returned when requesting the data range.
 """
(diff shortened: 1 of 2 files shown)


vallsv:
:+1: Nice, sounds simpler to understand and use.

vallsv (approved):

PR #3836, opened 2023-05-23
silx.gui.colors, silx.gui.plot: Added indexed color names in `rgba` and `PlotWidget`

### LINKED ISSUE #3755: Support an index as color in `PlotWidget.addCurve`
Since there is a list of default colors in `PlotWidget`, the idea is to leverage it by allowing to select colors from it in applications.

First raised here: https://gitlab.esrf.fr/bliss/bliss/-/issues/3665

Alternative: Makes sure there is an API for getting/setting the list of default colors and use it like:
```python
plot = PlotWidget()
plot.addCurve(x, y, color=plot.getDefaultColors(1))
```

Note: This is provided through PR #3835

t20100:
~~! Merge PR #3835 first!~~

This PR should close #3755.
It adds support for `color{index}` name in both `colors.rgba` function and `PlotWidget.addCurve|Histogram`.

To do so, `rgba` takes an extra optional `colors` argument

It all sounds a bit complex to me (e.g., to use `PlotWidget.getDefaultColors()` at some place and use `silx.config.DEFAULT_PLOT_CURVE_COLORS` everywhere else). So need to check how this is going to be used and maybe consider adding a `silx.config.DEFAULT_COLORS`.

Maybe that's better as a `PlotWidget.addCurve|Histogram` only feature at first...

### DIFF
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -24,13 +24,21 @@
 """This module provides API to manage colors.
 """
 
+from __future__ import annotations
+
 __authors__ = ["T. Vincent", "H.Payno"]
 __license__ = "MIT"
 __date__ = "29/01/2019"
 
+
 import numpy
 import logging
+import numbers
+import re
+from collections.abc import Sequence
+from typing import Optional, Union
 
+import silx
 from silx.gui import qt
 from silx.gui.utils import blockSignals
 from silx.math import colormap as _colormap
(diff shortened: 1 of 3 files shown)


src/silx/gui/colors.py
```diff
@@ -24,13 +24,21 @@
 """This module provides API to manage colors.
 """
 
+from __future__ import annotations
```
t20100, line 27: This enable using `tuple`, `dict` or `collections.abc.Sequence` as types with Python 3.7.
See [PEP585](https://peps.python.org/pep-0585/#implementation)

vallsv (approved):

vallsv:
You could allocate the matcher in the module scope, to make it faster, i think

src/silx/gui/plot/PlotWidget.py
```diff
@@ -1168,7 +1168,10 @@ def addCurve(self, x, y, legend=None, info=None,
             # Override previous/default values with provided ones
             curve.setInfo(info)
             if color is not None:
-                curve.setColor(color)
+                curve.setColor(
+                    colors.rgba(color, colors=self.getDefaultColors())
+                    if isinstance(color, str) else color
+                )
```
t20100, line 1174: Only convert `str` since `color` can also be an array of colors, one for each point..

t20100:
Fixed issue with passing an array of colors to `addCurve`/`addHistogram`

PR #3841, opened 2023-05-25
`silx.third_party`: Deprecated `EdfFile` and `TiffIO`

### LINKED ISSUE #248: [Plot] Use fabio to load/save image
If available, use fabio from plot widgets (maskTool, saveAction) to load/save image data.

### LINKED ISSUE #3820: Make `silx.app` and `silx.third_party` private
Having `silx._app` and `silx._third_party` instead would be explicit about their usage.

t20100:
This PR replaces usage on `EdfFile` by `fabio.edfimage.EdfImage` and silx's copy of `TiffIO` by `fabio.TiffIO` inside the `silx` project.
As a result, it also deprecates `silx.third_party`, `EdfFile` and `TiffIO` that are no longer used by the project.
Also silx's `TiffIO.py` is now `from fabio.TiffIO import *` since the copy of `TiffIO` in `fabio` is more recent than the one in silx.

This PR also fixes minor issues found while testing:
- Fixes issues in the `imageview.py` example
- Add support of `tif` extension in `PlotWidget.saveGraph` (only `tiff` with 2`f` was supported).

closes #3820
closes #248


vallsv:
Sounds good to me.

Issue #3743, opened 2023-01-03
Optional individual zooming on left and right Y axes


kklmn:
Hi!
In a 1D plot, when both left and right Y axes are in use, zooming acts on both of them.
Please consider adding a 3-state option "zoom both Y's", "zoom left Y" and "zoom right Y".
Thank you!

t20100:
Hi,

Having single-axis zoom interaction is indeed a missing generic feature.
Unfortunately, no time for me to look at it now.

Best

t20100:
Hi,

> Please consider adding a 3-state option "zoom both Y's", "zoom left Y" and "zoom right Y".

Could you give more information on how you intend to use it?
E.g., is it only for the zoom-on-wheel or also for the zoom selection? To be selected by the user?

kklmn:
Hi,
I typically zoom by selection but I think it should do equally by mouse wheel.
Yes, the zoom state (left, right, both) should be selectable by the user. Maybe a zoom state button would be good but first of all, I imagine a menu section.

t20100:
BTW, related to this, it's already possible to constrain the limits and range of an axis of the plot from the API with 
[Axis.setLimitsConstraints](http://www.silx.org/doc/silx/latest/modules/gui/plot/items.html#silx.gui.plot.items.Axis.setLimitsConstraints) and [Axis.setRangeConstraints](http://www.silx.org/doc/silx/latest/modules/gui/plot/items.html#silx.gui.plot.items.Axis.setRangeConstraints) methods (To retrieve the axes [PlotWidget.getXAxis](http://www.silx.org/doc/silx/latest/modules/gui/plot/plotwidget.html#silx.gui.plot.PlotWidget.PlotWidget.getXAxis) and [PlotWidget.getYAxis](http://www.silx.org/doc/silx/latest/modules/gui/plot/plotwidget.html#silx.gui.plot.PlotWidget.PlotWidget.getYAxis)).

PR #3842, opened 2023-05-25
`silx.gui.plot`: Added per-axis toggle of `PlotWidget` zoom

t20100:
This PR aims at allowing to configure zoom on the different axes independently.

With this PR, only zoom on wheel is managed.
To provide more control over the interaction, it adds a `PlotWidget.interaction` method to retrieve the interaction handler.
For now it only allows to toggle zoom on wheel and to select which axes can be zoomed.
This paves the way for adding more control over the interaction.

Related to #3743

### DIFF
--- a/src/silx/gui/plot/PlotInteraction.py
+++ b/src/silx/gui/plot/PlotInteraction.py
@@ -117,18 +117,13 @@ class _PlotInteraction(object):
 
 # Zoom/Pan ####################################################################
 
-class _ZoomOnWheel(ClickOrDrag, _PlotInteraction):
-    """:class:`ClickOrDrag` state machine with zooming on mouse wheel.
+class _PlotInteractionWithClickEvents(ClickOrDrag, _PlotInteraction):
+    """:class:`ClickOrDrag` state machine emitting click and double click events.
 
     Base class for :class:`Pan` and :class:`Zoom`
     """
 
     _DOUBLE_CLICK_TIMEOUT = 0.4
-
-    class Idle(ClickOrDrag.Idle):
-        def onWheel(self, x, y, angle):
-            scaleF = 1.1 if angle > 0 else 1. / 1.1
-            applyZoomToPlot(self.machine.plot, scaleF, (x, y))
 
     def click(self, x, y, btn):
         """Handle clicks by sending events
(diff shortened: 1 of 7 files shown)


src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1636,14 +1620,15 @@ def getDescription(self):
         return params
 
 
-class PlotInteraction(object):
-    """Proxy to currently use state machine for interaction.
-
-    This allows to switch interactive mode.
+class PlotInteraction(qt.QObject):
```
t20100, line 1623: This is added to the API as it can be retrieved through `PlotWidget.interaction()`, so some methods are made private.

src/silx/gui/plot/PlotWidget.py
```diff
@@ -25,6 +25,8 @@
 The :class:`PlotWidget` implements the plot API initially provided in PyMca.
 """
 
+from __future__ import annotations
```
t20100, line 28: Allows to use `tuple` as type with python3.7

src/silx/gui/plot/PlotWidget.py
```diff
@@ -3472,6 +3477,15 @@ def onMouseLeaveWidget(self):
...
     # Interaction modes #
 
+    def interaction(self) -> PlotInteraction:
+        """Returns the interaction handler for this PlotWidget"""
+        return self._eventHandler
+
+    def __interactionChanged(self):
+        """Handle PlotInteraction updates"""
+        if self.__isInteractionSignalForwarded:
+            self.sigInteractiveModeChanged.emit(None)
```
t20100, line 3487: In case this results from a call to `setInteractiveMode`, do not forward the event

src/silx/gui/plot/_utils/panzoom.py
```diff
@@ -116,16 +120,33 @@ def scale1DRange(min_, max_, center, scale, isLog):
...
+    def isDisabled(self) -> bool:
+        """True only if all axes are disabled"""
+        return not (self.xaxis or self.yaxis or self.y2axis)
+
+
+def applyZoomToPlot(
+    plot,
+    scale: float,
+    center: tuple[float, float]=None,
+    enabled: ZoomOnAxes=ZoomOnAxes(),
```
t20100, line 138: New argument to select axes for which zoom is enabled.

PR #3845, opened 2023-06-08
Draft: Added profile to compare image

vallsv:
Screenshot from 2023-06-08 18-46-01

Changelog: 
- Added profile to compare image


src/silx/gui/plot/tools/compare/profile.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+COLOR_A = "#0080FF"  # azure
+COLOR_B = "#702963"  # byzantine
+
+
+class ProfileImageLineROI(rois.ProfileImageLineROI):
+    """ROI for a compare image profile between 2 points.
+
+    The X profile of this ROI is the projecting into one of the x/y axes,
```
t20100, line 46:
```suggestion
    The X profile of this ROI is the projection into one of the x/y axes,
```

src/silx/gui/plot/tools/compare/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ###########################################################################*/
+"""This module provides tools related to the compare image plot.
+"""
+
+__authors__ = ["V. Valls"]
+__license__ = "MIT"
+__date__ = "09/06/2023"
```
t20100, line 29: Maybe `tools.compare` should be private (i.e., `tools._compare`): it sounds to be implementation details of the `CompareImages` widget.

t20100 (commented):
There's an issue with typing in `silx.gui.plot.tools.profile.core`.
I'm also wondering if `silx.gui.plot.tools.compare` shouldn't be private.

PR #3847, opened 2023-06-09
silx.gui.plot.items.roi: Added `RegionOfInterest`'s `getText` and `setText` methods

### LINKED ISSUE #3817: Add setLabel to the ROIs
It would be convenient to add an extra `setLabel` to the ROIs.

This way we programatically use the name to identify a ROI, and a label to display something.

Set a value to the label would only change the label displayed by a ROI.

- If `label=None` (default) the ROI would display it's `name` (for compatibility)
- If `label="a string"`, this label will be used instead of the name
- If `label=""`, that could be a way to display nothing

This would simplify some binding in FLint like https://gitlab.esrf.fr/bliss/bliss/-/blob/master/bliss/flint/widgets/image_plot.py#L287

t20100:
Added 2 methods to change the displayed text of ROIs without changing its name.
The default behavior (`setText(None)`) is to display the ROI's name.

This PR also contains a fix for PySide6 for RegionOfInterest.

closes #3817

### DIFF
--- a/src/silx/gui/plot/test/testItem.py
+++ b/src/silx/gui/plot/test/testItem.py
@@ -1,6 +1,6 @@
 # /*##########################################################################
 #
-# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
+# Copyright (c) 2017-2023 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -31,6 +31,7 @@ __date__ = "01/09/2017"
 import numpy
 
 from silx.gui.utils.testutils import SignalListener
+from silx.gui.plot.items.roi import RegionOfInterest
 from silx.gui.plot.items import ItemChangedType
 from silx.gui.plot import items
 from .utils import PlotWidgetTestCase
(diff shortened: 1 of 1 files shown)


src/silx/gui/plot/items/_roi_base.py
```diff
@@ -66,7 +67,9 @@ class _RegionOfInterestBase(qt.QObject):
     """
 
     def __init__(self, parent=None):
-        qt.QObject.__init__(self, parent=parent)
+        qt.QObject.__init__(self)
+        if parent is not None:
+            self.setParent(parent)
```
t20100, line 72: This was needed with PySide6 6.5.0 on macos with python3.10

src/silx/gui/plot/items/_roi_base.py
```diff
@@ -437,6 +441,26 @@ def setVisible(self, visible):
...
 
+    def getText(self) -> str:
+        """Returns the currently displayed text for this ROI"""
+        return self.getName() if self.__text is None else self.__text
+
+    def setText(self, text: Optional[str] = None) -> None:
+        """Set the displayed text for this ROI.
+
+        If None (the default), the ROI name is used.
+        """
```
t20100, line 452: I choose `set|getText` rather than `set|getLabel` since plot markers already have some `set|getText` methods.

src/silx/gui/plot/items/_roi_base.py
```diff
@@ -437,6 +441,26 @@ def setVisible(self, visible):
...
+        if self.__text != text:
+            self.__text = text
+            self._updated(items.ItemChangedType.TEXT)
+
+    def _updateText(self, text: str) -> None:
+        """Update the text displayed by this ROI
+
+        Override in subclass to custom text display
+        """
+        pass
```
t20100, line 462: This has moved from `HandleBasedROI`

src/silx/gui/plot/items/_roi_base.py
```diff
@@ -437,6 +441,26 @@ def setVisible(self, visible):
             self._visible = visible
             self._updated(items.ItemChangedType.VISIBLE)
 
+    def getText(self) -> str:
+        """Returns the currently displayed text for this ROI"""
+        return self.getName() if self.__text is None else self.__text
```
vallsv, line 446: I would prefer to decouple the `text` property from the `name`.

You could see the use of the name as a display mode (which we don't have to handle explicitly).

What do you think?

But i dont see any problem with the actual implementation.

vallsv (approved):

vallsv:
Very nice. I think this will simplify more things than i was expecting.

src/silx/gui/plot/items/_roi_base.py
t20100: What do you mean? Having `getText` eventually returning `self.__text` and an extra method, e.g., `getDisplayedText` to retrieve what should be displayed?

t20100: After discussion, this is fine as it is

PR #3853, opened 2023-06-14
Update the curve color cycle

### LINKED ISSUE #3850: Update the default curve color palette
The actual palette used to identify the curves have some lacks.

- It is mostly compound by saturated colors
- The yellow is not really readable on top of the white

A good palette is stricky. It have to match constraints which are not easy to solve.

- Colors have to be distinguishable but are the same time consistent
- It have to be distinguishable in a b&w print
- It have to take care of the common color vision deficiency

Luckily there is some research on that. So i propose to use the one i have selected for Flint:  The normal, 12 colors from http://tsitsul.in/blog/coloropt/

That's a convenient one because it provides 12 colors (which is already a lot), and silx use about 19 colors.

- This could became the default to the library (it is still possible to change it programmatically anyway)
- This could became the default to silx view only

What do you think?

## Actually

image

```
    DEFAULT_PLOT_CURVE_COLORS = ['#000000',  # black
                                 '#0000ff',  # blue
                                 '#ff0000',  # red
                                 '#00ff00',  # green
                                 '#ff66ff',  # pink
                                 '#ffff00',  # yellow
                                 '#a52a2a',  # brown
                                 '#00ffff',  # cyan
                                 '#ff00ff',  # magenta
                                 '#ff9900',  # orange
                                 '#6600ff',  # violet
                                 '#a0a0a4',  # grey
                                 '#000080',  # darkBlue
                                 '#800000',  # darkRed
                                 '#008000',  # darkGreen
                                 '#008080',  # darkCyan
                                 '#800080',  # darkMagenta
                                 '#808000',  # darkYellow
                                 '#660000']  # darkBrown
```

## Proposed

image

Here is with a reorder (from Flint source code)
```
NORMAL_PALETTE_12 = [
    (87, 81, 212),
    (235, 171, 33),
    (176, 69, 0),
    (0, 197, 248),
    (207, 97, 230),
    (0, 166, 107),
    (184, 0, 87),
    (0, 138, 248),
    (0, 110, 0),
    (0, 186, 171),
    (255, 145, 133),
    (133, 133, 0),
]
```

vallsv:
Closes #3850

This PR updates the color cycle with the one from matplotlib 2.0.

# Previously

image

# Now

image

Changelog: 
- Update the color cycle with the one from matplotlib 2.0

### DIFF
--- a/src/silx/_config.py
+++ b/src/silx/_config.py
@@ -125,7 +125,7 @@ class Config(object):
     It will have an influence on PlotWidget items
 
     Supported symbols:
-    
+
         - 'o', 'Circle'
         - 'd', 'Diamond'
         - 's', 'Square'
--- a/src/silx/_config.py
+++ b/src/silx/_config.py
@@ -83,26 +83,22 @@ class Config(object):
     .. versionadded:: 0.8
     """
 
-    DEFAULT_PLOT_CURVE_COLORS = ['#000000',  # black
-                                 '#0000ff',  # blue
-                                 '#ff0000',  # red
-                                 '#00ff00',  # green
-                                 '#ff66ff',  # pink
-                                 '#ffff00',  # yellow
-                                 '#a52a2a',  # brown
-                                 '#00ffff',  # cyan
-                                 '#ff00ff',  # magenta
-                                 '#ff9900',  # orange
-                                 '#6600ff',  # violet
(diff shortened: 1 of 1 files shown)


t20100 (approved):
LGTM, thanks!

PR #3854, opened 2023-06-14
Use linewidth2 as default selected style

### LINKED ISSUE #3849: Update linewidth for the default style of a selected curve
The actual style for the selected curve is setup with a selection color which is black.

I think that's not a very good idea, because it's a color like any others, and usually a desaturation means that the stuff is not selected or disabled.

As we are updating to 2.0 i propose to change the default to use the line width to show the selection (that's what i do in Flint), and let the original color of the curve.

- This could became the default to the library (it is still possible to change it programatically anyway)
- This could became the default to silx view only

What do you think?

## Actually

image

## Proposed

image

vallsv:
Closes #3849

This PR uses linewidth2 as default selected style

# Previously

image

# Now

image

Changelog:
- Added `DEFAULT_PLOT_ACTIVE_CURVE_COLOR` to `silx.config`
- Added `DEFAULT_PLOT_ACTIVE_CURVE_LINEWIDTH` to `silx.config`
- Changed default active curve style as linewidth=2

### DIFF
--- a/src/silx/_config.py
+++ b/src/silx/_config.py
@@ -145,3 +145,19 @@ class Config(object):
 
     .. versionadded:: 0.10
     """
+
+    DEFAULT_PLOT_ACTIVE_CURVE_COLOR = "#000000"
+    """Default color for the active curve.
+
+    It will have an influence on PlotWidget curve items
+
+    .. versionadded:: 2.0
+    """
+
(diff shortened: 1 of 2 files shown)


t20100 (approved):
LGTM

PR #3856, opened 2023-06-14
CompareImages: Fix the A-B visualization mode

vallsv:
The previous implementation was trying to constraint the data on the same data range, anyway the visualization mode.

This approach is fine for most of the modes, except A-B.

- Now A-B is handled in a special way, in order to use the basic operation `A-B`
- The colormap is now updated (when we switch between the visualization mode) in order to stay in the same data range
- Plus an RGB improvement to use any colormap normaization (before it was only a linear scale)

No changelog is needed as it's just a kind of partial revert of the previous change

Changelog:


t20100 (approved):
LGTM
Good to have `set|getVisualizationMode` on the item

PR #3858, opened 2023-06-14
Fix colormap dialog data bound

### LINKED ISSUE #3857: CompareImages: Error when using colormap fixed scale on ROI
The composite item does not store the data the way it is expected by the `ColormapDialog`.

As result it fails when we use the fixed scale tool from the `ColormapDialog`

We could fix that by
- Creating a dedicated API for that use case in `Item`
- Disabling the tool for the `CompareImages`
- Skipping unknown items from the `ColormapDialog`

```
  File "/nobackup/lvalls1/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py", line 1836, in __roiFinalized
    self.setColormapRangeFromDataBounds((ox, ox+width, oy, oy+height))
  File "/nobackup/lvalls1/valls/workspace/silx.git/build/lib.linux-x86_64-3.7/silx/gui/dialog/ColormapDialog.py", line 1458, in setColormapRangeFromDataBounds
    subset = data[ystart:ystop, xstart:xstop]
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

vallsv:
Closes #3857 

Make the colormap dialog safe with derivated image item.

Changelog: 
- Robustness on `ColormapDialog` tool with derivated `ImageBase`

### DIFF
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -1432,7 +1432,7 @@ class ColormapDialog(qt.QDialog):
             (xmin, xmax, ymin, ymax) Rectangular region in data space
         """
         if bounds is None:
-            return None  # no-op
+            return  # no-op
 
         colormap = self.getColormap()
         if colormap is None:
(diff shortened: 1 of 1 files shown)


vallsv:
This fix is a workaround, if we are lazy to make it better.

t20100 (approved):
LGTM

PR #3859, opened 2023-06-14
Get data without schema

### LINKED ISSUE #3852: Move compareimagewindow readData into get_data
This code from `CompareImagesWindow` could be moved into `get_data`.

- This could be done by default
- This could be only enabled by attribute

It looks like there is no special problem to move that code.

```
        if url.scheme() is None:
            for scheme in ('silx', 'fabio'):
                specificUrl = DataUrl(
                    file_path=url.file_path(),
                    data_slice=url.data_slice(),
                    data_path=url.data_path(),
                    scheme=scheme
                )
                try:
                    data = silx.io.utils.get_data(specificUrl)
                except Exception:
                    _logger.debug("Error while trying to loading %s as %s",
                                  urlPath, scheme, exc_info=True)
                else:
                    break
            else:
                raise ValueError(f"Data from '{urlPath}' is not readable as silx nor fabio")
```

vallsv:
This PR move back a feature from `CompareImagesWindow` to the core `silx.io.utils.get_data`.

Closes #3852

Changelog: 
- Added support of bare file from `silx.io.utils.get_data` with `check_schemas=True`

### DIFF
--- a/src/silx/io/utils.py
+++ b/src/silx/io/utils.py
@@ -32,7 +32,7 @@ import os.path
 import sys
 import time
 import logging
-from typing import Generator
+from typing import Generator, Union
 import urllib.parse
 
 import numpy
@@ -854,7 +854,7 @@ def match(group, path_pattern: str) -> Generator[str, None, None]:
                 yield f"{matching_path}/{matching_subpath}"
 
 
-def get_data(url):
+def get_data(url: Union[str, silx.io.url.DataUrl]):
     """Returns a numpy data from an URL.
 
     Examples:
@@ -879,7 +879,7 @@ def get_data(url):
 
     .. seealso:: :class:`silx.io.url.DataUrl`
 
-    :param Union[str,silx.io.url.DataUrl]: A data URL
+    :param url: A data URL
     :rtype: Union[numpy.ndarray, numpy.generic]
     :raises ImportError: If the mandatory library to read the file is not
         available.
@@ -942,6 +942,23 @@ def get_data(url):
         # There is no explicit close
         fabio_file = None
 
+    elif url.scheme() is None:
+        for scheme in ('silx', 'fabio'):
+            specificUrl = silx.io.url.DataUrl(
+                file_path=url.file_path(),
+                data_slice=url.data_slice(),
+                data_path=url.data_path(),
+                scheme=scheme
+            )
+            try:
+                data = get_data(specificUrl)
+            except Exception:
+                logger.debug("Error while trying to loading %s as %s",
+                             url, scheme, exc_info=True)
+            else:
+                break
+        else:
+            raise ValueError(f"Data from '{url}' is not readable as silx nor fabio")
     else:
         raise ValueError("Scheme '%s' not supported" % url.scheme())
 
--- a/src/silx/io/utils.py
+++ b/src/silx/io/utils.py
@@ -32,13 +32,13 @@ import os.path
 import sys
 import time
 import logging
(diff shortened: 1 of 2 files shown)


t20100:
CI fails:
```
silx/io/utils.py:857: in <module>
    def get_data(url: Union[str,silx.io.url.DataUrl], check_schemas: bool=False):
E   AttributeError: module 'silx' has no attribute 'io'
```
:shrug:

src/silx/io/utils.py
```diff
@@ -854,7 +854,7 @@ def match(group, path_pattern: str) -> Generator[str, None, None]:
                 yield f"{matching_path}/{matching_subpath}"
 
 
-def get_data(url):
+def get_data(url: Union[str,silx.io.url.DataUrl], check_schemas: bool=False):
```
t20100, line 857: Why adding a `check_schemas=False` argument rather than always trying with `silx` and `fabio` schemas?
To me the improvement you did makes sense all the time.

+ `silx.io.url.DataUrl` typing looks to cause trouble to `pytest` loading mechanism...

src/silx/io/utils.py
```diff
@@ -879,7 +879,9 @@ def get_data(url):
 
     .. seealso:: :class:`silx.io.url.DataUrl`
 
-    :param Union[str,silx.io.url.DataUrl]: A data URL
+    :param url: A data URL
+    :param check_schemas: If true, the url with be checked with known schemas
+                          even is no schema was specified
```
t20100, line 884: I feel, this makes it more complicated than it should.

t20100:
This will makes `get_data` simpler to use!

src/silx/io/utils.py
vallsv: It's up to you.

Just in case it introduce problem.

t20100: The new feature replaces `raise ValueError...` so it shouldn't be a problem.

vallsv:
Here is an update

vallsv:
> CI fails:
> 
> ```
> silx/io/utils.py:857: in <module>
>     def get_data(url: Union[str,silx.io.url.DataUrl], check_schemas: bool=False):
> E   AttributeError: module 'silx' has no attribute 'io'
> ```
> 
> depistes `import silx.io.url`...
> 
> shrug

I think it's because `utils` was imported by the `silx.io.__init__`, so it was maybe creating weird behaviour depending the way you import it.

src/silx/io/utils.py
```diff
@@ -32,13 +32,13 @@
...
 import logging
-from typing import Generator
+from typing import Generator, Union, Optional
 import urllib.parse
 
 import numpy
 
 from silx.utils.proxy import Proxy
-import silx.io.url
+from silx.io.url import DataUrl
```
t20100, line 41: Provided `silx.io` import `utils`, maybe relative import is more robust:
```suggestion
from .url import DataUrl
```

src/silx/io/utils.py
```diff
@@ -854,7 +854,7 @@ def match(group, path_pattern: str) -> Generator[str, None, None]:
                 yield f"{matching_path}/{matching_subpath}"
 
 
-def get_data(url):
+def get_data(url: Union[str, DataUrl], check_schemas: bool=False):
```
t20100, line 857:
```suggestion
def get_data(url: Union[str, DataUrl]):
```

t20100 (approved):
LGTM

vallsv:
The PR was updated.

PR #3861, opened 2023-06-15
Use "C" prefix to identify a color from the default palette

vallsv:
This change is done to match the matplotlib syntax "CN".

See https://matplotlib.org/stable/tutorials/colors/colors.html#cn-color-selection

Changelog: 
- Changed indexed color pattern to match "CN" color selection from matplotlib


t20100 (approved):
LGTM

PR #3865, opened 2023-06-16
silx.gui.plot.PlotWidget: Use matplotlib tight layout as experimental feature

### LINKED ISSUE #3404: Tight layout
I would like to know is we could use the tight layout provided by matplotlib now.

matplotlib use to provide a lot of empty space.
This is one way to reduce that.

vallsv:
Since the start of the project, the layout is based on percentage.

As result
- at low resolution part of the titles are cropped or outside of the widget
- at high resolution a lot of space is lost

Here is a proposal to use optionally the tight layout provided by matplotlib.

Closes #3404

I have put it as optional, and disabled by default because it looks to slow down a bit when we rescale the window.

# Previously / or without the tight layout enable (default)

image

# With the tight layout enable

image

Changelog: 
- Added `silx.config` option to enable tight layout for matplotlib backend
- Added `silx view` menu to enable/disable tight layout for matplotlib


vallsv:
Have to be checked
- [ ] Is it still fine when we hide the axis
- [ ] Special plots like colormap dialog?
- [ ] ImageView with side histogram

vallsv:
Screenshot from 2023-06-16 10-17-23

- [ ] The right histogram is not perfectly aligned because the plot histogram dont have title
- [ ] The layout change when the axes label goes outside, as result it creates glitches at interaction

vallsv:
image

- [ ] The tick at 1.0 on the y-axis and the histogram data is blinking when the mouse enter and leave the plot
- [ ] There is a weird margin on the y-axis

vallsv:
image

- [ ] `setAxesDisplayed(False)` is not properly handled (ticks, title, and labels are still displayed) only the axis lines are hide

vallsv:
I think it would be good to drop or to deprecate `setAxesMargins`.
If we agree with that i think we can provide something compatible for now.

vallsv:
Talking with @t20100 

- Maybe rework the displayed axis as a mode, `HIDDEN`/`PERCENTAGE`/`CONSTRAINED`

vallsv:
That's probably one of the most important feature for users.

I would like to have some progress here if possible. What to do with that?

No matter if it stay as an experimental stuff, but it's true that if it is not backposrted to opengl, it's maybe not as useful as i think.

t20100:
It sounds fine to me to have this as an experimental/matplotlib-only feature first, we can then see for something similar for OpenGL support (that obviously more work).

t20100:
There's a systematic error in appveyor CI:

<details><summary>Logs</summary>

```
_______________ TestPlotWidget.testChangeLimitsWithAspectRatio ________________
self = <silx.gui.plot.test.testPlotWidget.TestPlotWidget testMethod=testChangeLimitsWithAspectRatio>
    def testChangeLimitsWithAspectRatio(self):
        self.plot.setKeepDataAspectRatio()
        self.qapp.processEvents()
        xlim = self.plot.getXAxis().getLimits()
        ylim = self.plot.getYAxis().getLimits()
        defaultRatio = abs(xlim[1] - xlim[0]) / abs(ylim[1] - ylim[0])
    
        self.plot.getXAxis().setLimits(1., 10.)
        self._checkLimits(expectedXLim=(1., 10.), expectedRatio=defaultRatio)
        self.qapp.processEvents()
>       self._checkLimits(expectedXLim=(1., 10.), expectedRatio=defaultRatio)
venv_test\lib\site-packages\silx\gui\plot\test\testPlotWidget.py:121: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <silx.gui.plot.test.testPlotWidget.TestPlotWidget testMethod=testChangeLimitsWithAspectRatio>
expectedXLim = (1.0, 10.0), expectedYLim = None
expectedRatio = 1.3816463583494878
    def _checkLimits(self,
                     expectedXLim=None,
                     expectedYLim=None,
                     expectedRatio=None):
        """Assert that limits are as expected"""
        xlim = self.plot.getXAxis().getLimits()
        ylim = self.plot.getYAxis().getLimits()
        ratio = abs(xlim[1] - xlim[0]) / abs(ylim[1] - ylim[0])
    
        if expectedXLim is not None:
            self.assertEqual(expectedXLim, xlim)
    
        if expectedYLim is not None:
            self.assertEqual(expectedYLim, ylim)
    
        if expectedRatio is not None:
>           self.assertTrue(
                numpy.allclose(expectedRatio, ratio, atol=0.01))
E           AssertionError: False is not true
venv_test\lib\site-packages\silx\gui\plot\test\testPlotWidget.py:108: AssertionError
```

</details>

t20100:
If we consider this feature as completely experimental and we won't keep the `config.MPL_TIGHT_LAYOUT` as it is, it LGTM (We could even rename it `config.EXPERIMENTAL_MPL_TIGHT_LAYOUT` or something like this).

If we want to target a more stable API, that would need some rework of the `PlotWidget` API as you propose, e.g., `setAxesLayout(NONE|PERCENTAGE|TIGHT|...)` (and the config `MPL_TIGHT_LAYOUT` can be renamed `DEFAULT_PLOT_AXES_LAYOUT`) and deprecate `setAxesDisplayed`. I'd keep `setAxesMargins` for the default mode.

t20100:
See PR https://github.com/vallsv/silx/pull/36 with a few updates.

t20100:
Renamed to remove Draft status

t20100 (approved):
LGTM

PR #3870, opened 2023-06-16
Fix some pyside6 issues

payno:
Changelog:

### DIFF
--- a/src/silx/gui/dialog/AbstractDataFileDialog.py
+++ b/src/silx/gui/dialog/AbstractDataFileDialog.py
@@ -443,13 +443,8 @@ class _Browser(qt.QStackedWidget):
         stream.writeInt32(self.__serialVersion)
         stream.writeQVariant(self.__detailView.header().saveState())
         viewMode = self.viewMode()
-        if qt.BINDING == 'PyQt6':
+        if qt.BINDING in ('PyQt6', 'PySide6'):
             viewMode = viewMode.value
-        if qt.BINDING == 'PySide6':  # No auto conversion to int
-            from PySide6 import __version__ as pyside6_version
-            from pkg_resources import parse_version as _parse_version
-            if parse_version(__version__) >= parse_version('6.4'):
-                viewMode = viewMode.value
         stream.writeInt32(viewMode)
 
         return data
@@ -1728,13 +1723,8 @@ class AbstractDataFileDialog(qt.QDialog):
         stream.writeQString(u"%s" % self.directory())
         stream.writeQVariant(self.__browser.saveState())
         viewMode = self.viewMode()
-        if qt.BINDING == 'PyQt6':
+        if qt.BINDING in ('PyQt6', 'PySide6'):
             viewMode = viewMode.value
-        if qt.BINDING == 'PySide6':  # No auto conversion to int
-            from PySide6 import __version__ as pyside6_version
-            from pkg_resources import parse_version as _parse_version
-            if parse_version(__version__) >= parse_version('6.4'):
-                viewMode = viewMode.value
         stream.writeInt32(viewMode)
         colormap = self.colormap()
         if colormap is not None:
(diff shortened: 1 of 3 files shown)


src/silx/gui/dialog/AbstractDataFileDialog.py
```diff
@@ -443,7 +443,7 @@ def saveState(self):
         stream.writeInt32(self.__serialVersion)
         stream.writeQVariant(self.__detailView.header().saveState())
         viewMode = self.viewMode()
-        if qt.BINDING == 'PyQt6':  # No auto conversion to int
+        if qt.BINDING in ('PyQt6', 'PySide6'):  # No auto conversion to int
```
payno, line 446: @t20100 this probably comes with this Enum redefinition. So this mean this code will only work for recent PySide6 version. But couldn't find the version

payno: error is:
```
        viewMode = self.viewMode()
        if qt.BINDING == 'PyQt6':  # No auto conversion to int
            viewMode = viewMode.value
>       stream.writeInt32(viewMode)
E       TypeError: 'PySide6.QtCore.QDataStream.writeInt32' called with wrong argument types:
E         PySide6.QtCore.QDataStream.writeInt32(ViewMode)
E       Supported signatures:
E         PySide6.QtCore.QDataStream.writeInt32(int)

```

payno: looks like matplotlib get this kind of handling:
```
    elif QT_API == QT_API_PYSIDE6:
        from PySide6 import QtCore, QtGui, QtWidgets, __version__
        import shiboken6
        def _isdeleted(obj): return not shiboken6.isValid(obj)
        if parse_version(__version__) >= parse_version('6.4'):
            _to_int = operator.attrgetter('value')
        else:
            _to_int = int
```

t20100:
There seems to be an issue with the `--qt-binding` argument which is not really taken into account...

t20100:
IMO provided we are making a major version, we should keep the compatibility simple and drop the support of PySide6 < v6.4 because this is the kind of code that will be never tested again.

Where PySide6<v6.4 compatibility is a matter?
- It is not packaged in Debian
- It is available in conda-forge from v6.4 onward only
- So that's "only" an issue for older installation with `pip`

src/silx/gui/dialog/AbstractDataFileDialog.py
```diff
@@ -443,8 +443,13 @@ def saveState(self):
...
         stream.writeQVariant(self.__detailView.header().saveState())
         viewMode = self.viewMode()
-        if qt.BINDING == 'PyQt6':  # No auto conversion to int
+        if qt.BINDING == 'PyQt6':
             viewMode = viewMode.value
+        if qt.BINDING == 'PySide6':  # No auto conversion to int
+            from PySide6 import __version__ as pyside6_version
+            from pkg_resources import parse_version as _parse_version
+            if parse_version(__version__) >= parse_version('6.4'):
+                viewMode = viewMode.value
```
vallsv, line 452: Why dont use just use `viewMode.value` always? this should be retrocompatible no?

Or the following probably also work, no?
```int(viewMode)```

I mean there is probably a compatible way to get the int in this way.

Comment (t20100) [reply src/silx/gui/dialog/AbstractDataFileDialog.py:452]: `int(viewMode)` doesn't work..

Comment (vallsv) [reply src/silx/gui/dialog/AbstractDataFileDialog.py:452]: What about
```
if hasattr(viewMode, "value"):
    viewMode = viewMode.value
```

payno: yes, should do the work. But as was saying @t20100 this is a major release and to my knowledge the usage of pyside6 is very limited for now. So it would be better / simpler to force `pyside6 > 6.4` as requirements and fix what needs to be fix only for `pyside6 > 6.4` only. This would simplify the management.

t20100: The issue with `hasattr` is that you don't know for which binding/version it is needed and so it's gonna stay forever.

I would be in favor of dropping support of previous version of `pyside6`, I doubt it is used a lot yet...

vallsv: Contrary to kson, since python 1 you can introcide some comments in the code with `#` ;-)

payno:
Let's go for requirements PySide >= 6.4

vallsv:
Is this PR can be merged?

payno:
finally it will be done in https://github.com/silx-kit/silx/pull/3872

PR #3872, opened 2023-06-19
silx.gui: For PySide6 required version is >=6.4

t20100:
Let's see how this changes the results of the CI making sure only the selected Qt binding is used


t20100:
Added fixes taken from PR #3870

t20100:
Test PySide6 on Windows with a version of Python for which matplotlib provides wheel for latest releases (to get pyside6 related fixes...)

vasole:
I would not uninstall any Qt binding in CI.

If having multiple Qt bindings installed is a problem, it will also be a problem for the end users.

t20100:
Having multiple backends is not an issue for users, `silx.gui.qt`'s probing of already loaded binding works fine.

I can remove the uninstall Qt5, I initially added it to be sure I was debugging the right binding.
But it sounds safer to me to keep it, if for whatever reason tests are not picking the right binding, it's not gonna be noticed easily...

t20100:
This fixes CI with PySide6 and update the min required version of PySide6 to >=6.4.

vasole:
> Having multiple backends is not an issue for users, `silx.gui.qt`'s probing of already loaded binding works fine.

My point is that if 'we' are not able to respect the chosen binding or have difficulties when PyQt and PySide are present, our users will face the same difficulties.

t20100:
Our users does not run the tests from the source repository ;)

vasole:
> Our users do not run the tests from the source repository ;)

Be careful with Debian packagers ;)

t20100:
Indeed, my point is I prefer to be "ceinture et bretelle" about which binding is tested.
And anyway, by having multiple bindings installed, we actually don't test that we pick the right one, we just take the risk that it happens without noticing it...

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -61,11 +61,9 @@ class InterpolationMixIn(ItemMixInBase):
 
     def __init__(self, mode=NEAREST_INTERPOLATION, primitive=None):
         self.__primitive = primitive
+        self.__interpolationMode = mode
         self._syncPrimitiveInterpolation()
 
-        self.__interpolationMode = None
-        self.setInterpolation(mode)
-
```
payno, line 68: I am fine with all but this one seems strange to me. There is no need for calling `_syncPrimitiveInterpolation` or `_updated` in this case ?

t20100: This is to fix an inheritance issue.
`_syncPrimitiveInterpolation` is already called and is needed to sync the `__primitive`.
And since it's in the `__init__`, there is no need to call `_updated` (= trigger a signal) because it's initializing it, and if I recall well, that was the issue: emitting a signal from a not fully initialized object.

t20100:
Rebase to fix conflict

t20100:
I ended-up skipping the test that seg fault with PySide6, not ideal but I couldn't find a way to workaround this.

PR #3887, opened 2023-06-26
ArcROI: Hide the handler instead of hidding the symbol

### LINKED ISSUE #3875: Some ArcROI handlers use a none symbol instead of hidding the handler
Noticed with the ArcRoi.

In translation mode, if you hover the start of end handler (which are not visible) the mouse cursor change to a hand pointer.

The ROI also have to be the active one.

vallsv:
This PR makre sure the handle are now hoverable by the mouse.

Closes #3875 

Changelog:


t20100 (approved):
LGTM

PR #3889, opened 2023-06-26
Added extra functionality to FloatEdit

### LINKED ISSUE #3888: Extended FloatEdit
A common problem is to provide a convinient `FloatEdit` widget

The actual one could be improved.

- Support optional value else `None`
- min max including or not the boundaries
- background color if not valid
- Use the value set by `setValue` and not a rounded version
- Easy way to move the `.`

vallsv:
Closes #3888

This PR mostly provides extra validators that i have designed during the years.

It allows configurable behaviours like:
- Empty input as `None`
- Fast edition of the dot `.` from the input
- Included/excluded float min/max boundaries
- Basic support for quantities
- Plus a bit more accurate value will is it not edited by the user

This PR also add some tests to the `FloatEdit`.

Changelog: 
- Added support to custom validators to `FloatEdit` widget
- Added extra validators to share common behaviours

### DIFF
--- a/src/silx/gui/utils/validators.py
+++ b/src/silx/gui/utils/validators.py
@@ -1,0 +1,432 @@
+# /*##########################################################################
+#
+# Copyright (c) 2016-2022 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ###########################################################################*/
+"""Validator class to write Qt widget."""
+
+__authors__ = ["V. Valls"]
+__license__ = "MIT"
+__date__ = "21/06/2020"
+
+
+from typing import Tuple, Optional
+from silx.gui import qt
+
+
+class CustomValidator:
+    """Extra API for a QValidator in order to put more logic inside."""
+
+    def toValue(self, text: str) -> Tuple[object, bool]:
+        """Convert the input string into an interpreted value.
+
+        :param text: Input string
+        :returns: A tuple containing the resulting object and True if the
+            string is valid
+        """
+        raise NotImplementedError()
+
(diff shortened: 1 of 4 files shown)


vallsv:
Tested manually with `pyside6-6.5.1.1`, shoulds like it's ok

src/silx/gui/widgets/FloatEdit.py
```diff
@@ -29,40 +29,67 @@
...
         if value is not None:
             self.setValue(value)
 
-    def value(self):
+    def keyPressEvent(self, event: qt.QEvent):
+        result = super(FloatEdit, self).keyPressEvent(event)
+        if event.isAccepted():
+            self.__wasModified = True
+        return result
+
```
t20100, line 61: This looks to be useless (`__wasModified` is never used)
```suggestion
```

src/silx/gui/utils/validators.py
```diff
@@ -0,0 +1,432 @@
...
+        """
+        Allow the field to be empty. Default is false.
+
+        An empty field is represented as a `None` value.
+
+        :param allow: New state.
+        """
+        self.__allowEmpty = allow
+
+    def setIncludedBound(self, minBoundIncluded: bool, maxBoundIncluded: bool):
```
t20100, line 365: It would be good to have a getter

src/silx/gui/utils/validators.py
```diff
@@ -0,0 +1,432 @@
...
+    """
+    Validate double values and provides features to allow or disable other
+    things.
+    """
+    def __init__(self, parent=None):
+        super(AdvancedDoublePintValidator, self).__init__(parent=parent)
+        self.__allowEmpty = False
+        self.__boundIncluded = True, True
+
+    def setAllowEmpty(self, allow: bool):
```
t20100, line 355: It would be good to have a getter

src/silx/gui/utils/validators.py
```diff
@@ -0,0 +1,432 @@
...
+        """
+        Allow the field to be empty. Default is false.
+
+        An empty field is represented as a `None` value.
+
+        :param allow: New state.
+        """
+        self.__allowEmpty = allow
+
+    def setIncludedBound(self, minBoundIncluded: bool, maxBoundIncluded: bool):
```
t20100, line 164: It would be good to have a getter

src/silx/gui/utils/validators.py
```diff
@@ -0,0 +1,432 @@
...
+    """
+    Validate double values and provides features to allow or disable other
+    things.
+    """
+    def __init__(self, parent=None):
+        super(AdvancedDoubleValidator, self).__init__(parent=parent)
+        self.__allowEmpty: bool = False
+        self.__boundIncluded: Tuple[bool, bool] = (True, True)
+
+    def setAllowEmpty(self, allow: bool):
```
t20100, line 154: It would be good to have a getter

src/silx/gui/utils/validators.py
```diff
@@ -0,0 +1,432 @@
...
+
+            # If the typed character is a dot, move the dot instead of ignoring it
+            if inputText[pos - 1] == locale.decimalPoint():
+                beforeDot = inputText[0:pos].count(locale.decimalPoint())
+                pos -= beforeDot
+                inputText = inputText.replace(locale.decimalPoint(), "")
+                inputText = inputText[0:pos] + locale.decimalPoint() + inputText[pos:]
+                pos = pos + 1
+
+            print(pos, inputText, inputText[pos - 1])
```
t20100, line 98:
```suggestion
```

src/silx/gui/utils/validators.py
```diff
@@ -0,0 +1,432 @@
...
+__authors__ = ["V. Valls"]
+__license__ = "MIT"
+__date__ = "21/06/2020"
+
+
+from typing import Tuple, Optional
+from silx.gui import qt
+
+
+class CustomValidator:
```
t20100, line 35: Maybe use a name with more meaning, e.g. `ConverterValidatorMixIn`

t20100 (commented):
There seems to be a bit of duplicated code between `AdvancedDoubleValidator` and `AdvancedDoublePintValidator`. If possible that'd be best to share it.
Also why providing 2 classes (e.g., `DoubleValidator` and  `AdvancedDoubleValidator`) rather than only the advanced one which by default provides what the default one does?

PR #3890, opened 2023-06-27
PlotWidget: Added sigBackendChanged

### LINKED ISSUE #3883: Create a sigBackendChanged to the PlotWidget
When the backend is updated, the widget handle is not the same.

In some case i have to register this widget to capture some qt events.

It would be useful to provide a signal in this case.

Related to #3876

vallsv:
Added `sigBackendChanged` to `PlotWidget`

Related to #3876 

Closes #3883

Changelog: 
- Added `sigBackendChanged` to `PlotWidget`

### DIFF
--- a/src/silx/gui/plot/PlotWidget.py
+++ b/src/silx/gui/plot/PlotWidget.py
@@ -360,6 +360,9 @@ class PlotWidget(qt.QMainWindow):
     It provides the menu which will be displayed.
     """
 
+    sigBackendChanged = qt.Signal()
+    """Signal emitted when the backend have changed."""
+
     def __init__(self, parent=None, backend=None):
         self._autoreplot = False
         self._dirty = False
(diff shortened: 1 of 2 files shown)


vallsv:
I was thinking of added `current, previous` params to the signal.

Usually if you don't do that, the user of the API have to store the backend reference on its side. So i think it could be done without much problem.

t20100 (approved):
LGTM

The `OpenGLAction` can be improved with this

vallsv:
That's a good idea i have updated the stuff.

t20100:
> I was thinking of added current, previous params to the signal.

If it's needed, why not.
It's not Qt-like but some other signals of `PlotWidget` already provides current, previous params.

But this means you'll give back a `previous` backend that is no longer used by a `PlotWidget`, not sure what it's useful for, and risky if used.

t20100:
> aboutToConnect/aboutToDisconnect, but i feel like it's overkill.

I agree.

`sigBackendChanged` can also only provide the current backend, that more Qt-like and does cause trouble with giving access to a disconnected backend.

It's fine as it for me.

PR #3893, opened 2023-06-27
Make sure the cursor is updated after an interaction

### LINKED ISSUE #3892: After an interaction the mouse cursor is not restored
- Create a ROI
- Move an handle
- Release the mouse
- The cursor hover the handle (and could drag it) but does not have the right cursor pointer

vallsv:
Closes #3892 

This PR make sure the cursor have the proper shape after an interaction.

- If you create a ROI, at the end of the creation your mouse probably hover a marker, the mouse cursor have to be updated
- If you drag a marker, at the end of the drag your mouse is still hovering the same marker

I can't say it's a very good patch, but i guess there is no way to make it better if we don't use a kind of stack based hierarchical state machine, or maybe messages between interactions. Because we lose some states between interactions.

Changelog: 
- Fixes mouse cursor hover a marker after an interaction

### DIFF
--- a/src/silx/gui/plot/PlotInteraction.py
+++ b/src/silx/gui/plot/PlotInteraction.py
@@ -1084,6 +1084,8 @@ class SelectFreeLine(ClickOrDrag, _PlotInteraction):
 
 # ItemInteraction #############################################################
 
+PREVIOUS_HOVERED_MARKER = None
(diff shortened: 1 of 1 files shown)


t20100:
closing as supersed by #3904

PR #3894, opened 2023-06-27
Arc roi getPositionRole

### LINKED ISSUE #3881: Extra API for ROIs
In Flint i use

```
        x0, y0 = roi._handleStart.getPosition()
        x1, y1 = roi._handleEnd.getPosition()
```

I was thinking of creating a dedicated API for that.

This could be something like the following which can easely be adapted by different geometry
```
getPositionRole(END)
getPositionRole(START)
```

Or an API to expose the handles by roles directly. But i dont like much this idea.

vallsv:
Closes #3881

Added `getPositionRole` to the `ArcROI` item

This is just an helper which can reduce some code in the app side.

Finally that was no need to make it global to any ROIs, so it's just part of the `ArcROI` API. This could be changed in the future.

Plus items ROI tests converted to pytest

Changelog: 
- Added `getPositionRole` to the `ArcROI` item


src/silx/gui/plot/items/_arc_roi.py
```diff
@@ -714,6 +727,20 @@ def getGeometry(self):
             raise ValueError("This ROI can't be represented as a section of circle")
         return geometry.center, self.getInnerRadius(), self.getOuterRadius(), geometry.startAngle, geometry.endAngle
 
+    def getPositionRole(self, role: Role) -> Tuple[float, float]:
```
t20100, line 730: `getPositionRole` make thinks it returns a role, what about:
```suggestion
    def getPosition(self, role: Role) -> Tuple[float, float]:
```
or eventually `getPositionFromRole`.

src/silx/gui/plot/tools/test/testRoiItems.py
```diff
@@ -0,0 +1,304 @@
...
+def testArc_position_role(qapp):
+    """Test validity of getPositionRole"""
+    item = roi_items.ArcROI()
+    center = numpy.array([10, 20])
+    innerRadius, outerRadius, startAngle, endAngle = 1, 100, numpy.pi * 0.5, numpy.pi
+    item.setGeometry(center, innerRadius, outerRadius, startAngle, endAngle)
+    assert item.getPositionRole(roi_items.ArcROI.Role.START) == pytest.approx((10.0, 70.5))
+    assert item.getPositionRole(roi_items.ArcROI.Role.STOP) == pytest.approx((-40.5, 20.0))
+    assert item.getPositionRole(roi_items.ArcROI.Role.MIDDLE) == pytest.approx((-25.71, 55.71), abs=0.1)
+    assert item.getPositionRole(roi_items.ArcROI.Role.CENTER) == pytest.approx((10.0, 20), abs=0.1)
```
t20100, line 287:
```suggestion
def testArc_position(qapp):
    """Test validity of getPosition"""
    item = roi_items.ArcROI()
    center = numpy.array([10, 20])
    innerRadius, outerRadius, startAngle, endAngle = 1, 100, numpy.pi * 0.5, numpy.pi
    item.setGeometry(center, innerRadius, outerRadius, startAngle, endAngle)
    assert item.getPosition(roi_items.ArcROI.Role.START) == pytest.approx((10.0, 70.5))
    assert item.getPosition(roi_items.ArcROI.Role.STOP) == pytest.approx((-40.5, 20.0))
    assert item.getPosition(roi_items.ArcROI.Role.MIDDLE) == pytest.approx((-25.71, 55.71), abs=0.1)
    assert item.getPosition(roi_items.ArcROI.Role.CENTER) == pytest.approx((10.0, 20), abs=0.1)
```

t20100 (commented):
Good to keep in `ArcROI` for now, that can be extended to other ROIs when needed.
I'd propose to rename `getPositionRole` to `getPosition`.

vallsv:
Dont you think it's a too generic name?

Or in this case `role` could be optional and return by default something like the center.

t20100:
`getPosition` without arguments is provided by markers (`MarkerBase`), and the foloowing ROIs: `PointROI`, `CrossROI`, `HorizontalLineROI`, `VerticalLineROI`.

To me it sounds coherent to use the same name here, and yes, the center could be the default role.

vallsv:
I have updated the role to have a default value.

t20100 (approved):
LGTM

PR #3898, opened 2023-06-30
Added group item

vallsv:
This PR feature a `Group` plot item together with an example.

It is an intermediate implementation which doesn't stores the item as a tree representation, but instead still use the flat representation. In the other word, a group have no special meaning for a plot.

The link between a group and another item is just a relationship between 2 plot items.

This anyway allow to group items together.
- A group added/removed from a plot will update the plot with its childrens
- The visibility of the group will impact the visibility of the children

There is improvement that could be optionally tested:
- Expose the relationshit in the 2 directions (create a `Item.getParentItem()`)
- Handle the bouding box as a tree
- Handle the visibility as a tree

Changelog: 
- Added a Group plot item

### DIFF
--- a/examples/plotGroups.py
+++ b/examples/plotGroups.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python
+# /*##########################################################################
+#
+# Copyright (c) 2016-2019 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


t20100:
Adding a `Group` item while keeping the flat structure is not a good move IMO.

If the flat structure is maintained, the `Group` should not be an item of the plot, but more an object that allows to handle a couple of items (like changing visibility/adding-removing from plot) at once like matplotlib has `Container` to group `Artist`s. That would be the intermediate step for me.

If `Group` is a plot item, then we need a proper tree structure of items.

vallsv:
A group which is not a plot item is pointless to me.
Because anybody can already create such structure, but will still have to manage the life cycle of the object on it's own.

I would prefer not to create another extra layer of indirection in Flint. Because that's stuff which will never be shared back to silx.

loichuder:
Superseded by #3908

PR #3900, opened 2023-07-06
[Draft] Opencl lz4

kif:
Changelog: LZ4 compressor on GPU

* [ ] OpenCL code targeted to run on GPU
     * [X] Find position of interest
     * [X] Create segment descriptor (position in input, length of litterals and matches)
     * [x] Concatenate several descriptor (between passes, in the case there are only litterals)
     * [x] Write compressed stream, including tokens, offsets, overflows in litteral and matches
     * [x] Manage the beginning of stream (4 bytes with uncompressed buffer size)
     * [x] Manage end of stream (cannot end with matches, has to finish with litterals
     * [x] Concatenate output stream to make it contiguous (1 or 2 kernel options ...)
* [x] Python code to call the OpenCL code
* [ ] Non regression test
* [ ] Profiling and performance assessment


t20100:
I've added a commit to fix the conflict introduced by #3991.
A rebase would be best in order to sparse the merge of `main` in this branch.

payno:
@kif did the rebase was done ? I don't see it. Else could you do it please ? Thanks

Issue #3901, opened 2023-07-07
Split pseudo-voigt: allow splitting the lorentzian fraction


woutdenolf:
Formula: https://repository.uantwerpen.be/docman/irua/c9f8c6/6141.pdf#subsection.A.1.4

Requested by ID26.

PR #3902, opened 2023-07-08
add split pseudo-voigt with split lorentzian fraction

woutdenolf:
Allows the Lorentzian fraction to be different on the left and on the right of the centroid:

image

Closes https://github.com/silx-kit/silx/issues/3901

Changelog: add split pseudo-voigt with split lorentzian fraction

### DIFF
--- a/src/silx/math/fit/fittheories.py
+++ b/src/silx/math/fit/fittheories.py
@@ -694,6 +694,97 @@ class FitTheories(object):
                 newcons[5 * i + 4, 2] = 1.0
         return newpar, newcons
 
+    def estimate_splitpvoigt2(self, x, y):
+        """Estimation of *Height, Position, FWHM1, FWHM2, eta1, eta2* of peaks, for
+        asymmetric pseudo-Voigt curves.
+
+        This functions uses :meth:`estimate_height_position_fwhm`, then
+        adds an identical FWHM2 parameter and a constant estimation of
+        *eta1* and *eta2* (0.5) to the fit parameters for each peak, and the corresponding
+        constraints.
+
+        Constraint for the eta parameter can be set to QUOTED (0.--1.)
+        by setting :attr:`config`['QuotedEtaFlag'] to ``True``.
+        If this is not the case, the constraint code is set to FREE.
+
+        :param x: Array of abscissa values
+        :param y: Array of ordinate values (``y = f(x)``)
+        :return: Tuple of estimated fit parameters and fit constraints.
+            Parameters to be estimated for each peak are:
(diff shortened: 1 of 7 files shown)


src/silx/math/fit/functions.pyx
```diff
@@ -472,6 +473,62 @@ def sum_splitpvoigt(x, *params):
...
+        *(height1, centroid1, fwhm11, fwhm21, eta11, eta21,...)*
+    :return: Array of sum of split pseudo-Voigt functions at each ``x``
+        coordinate
+    """
+    cdef:
+        double[::1] x_c
+        double[::1] params_c
+        double[::1] y_c
+
+    if not len(params):
```
t20100, line 506: According to the exception, this would be better suited:
```suggestion
    if len(params) < 6:
```

src/silx/math/fit/fittheories.py
```diff
@@ -1263,6 +1354,13 @@ def configure(self, **kw):
                               'HighFWHM', 'Eta'),
                   estimate=fitfuns.estimate_splitpvoigt,
                   configure=fitfuns.configure)),
+    ('Split Pseudo-Voigt 2',
+        FitTheory(description='Asymmetric pseudo-Voigt functions',
```
t20100, line 1358: Maybe have a different `description` that the one for `functions.sum_splitpvoigt`? :shrug:

t20100 (approved):

src/silx/math/fit/functions.pyx
woutdenolf: Applied that change to all functions, also modifying the error message accordingly

```
-    if not len(params):
-        raise IndexError("No parameters specified. " +
-                         "At least 6 parameters are required.")
+    if len(params) < 6:
+        raise IndexError("At least 6 parameters are required.")
```

src/silx/math/fit/fittheories.py
woutdenolf: - `Asymmetric pseudo-Voigt functions with split Lorentzian factor`
- `Asymmetric pseudo-Voigt functions with two Lorentzian factors`
- ???

vasole:
In fact one should test to have a multiple of 6 parameters.

`if len(params) and not params % 6:`

src/silx/math/fit/fittheories.py
woutdenolf: I went for 'Asymmetric pseudo-Voigt functions with split Lorentzian fraction'

woutdenolf:
> In fact one should test to have a multiple of 6 parameters.
> 
> `if len(params) and not params % 6:`

Done

vasole (approved):

PR #3904, opened 2023-07-10
silx.gui.plot: Fixed PlotWidget mouse cursor update

### LINKED ISSUE #3892: After an interaction the mouse cursor is not restored

t20100:
This PR fixes the update of the mouse cursor when finishing an interaction with a ROI.
This is an alternative to PR #3893.

Compared to PR #3893, it does not need to use a global variable, removes a blinking cursor (set to default in `__terminateDrag` and changed in `Idle.enterState`) and updates the cursor according to the item that interaction will occur on (= the top-most marker) rather than the last one used for interaction.

closes #3892

### DIFF
--- a/src/silx/gui/plot/PlotInteraction.py
+++ b/src/silx/gui/plot/PlotInteraction.py
@@ -1098,8 +1098,10 @@ class ItemsInteraction(ClickOrDrag, _PlotInteraction):
             self._hoverMarker = None
 
         def enterState(self):
-            plot = self.machine.plot
-            position = plot.getWidgetHandle().mapFromGlobal(qt.QCursor.pos())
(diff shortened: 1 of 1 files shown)


vallsv (approved):

vallsv:
I would prefer to still use the same marker, because it will be really surprising. But still my PR is anyway not enough

So go for it.
Feel free to close by PR, it was mostly a proof of concept.

t20100:
> I would prefer to still use the same marker

Yes, I see, but that would break the pick the top-most marker policy..

PR #3906, opened 2023-07-12
silx.gui.plot.PlotWidget: Fixed display of `Shape` with dashes and a background color

### LINKED ISSUE #3879: plot: Shape can be drawn outside of the plot
```
                ringItem = silxItems.Shape("polylines")
                ringItem.setPoints(numpy.array((x, y)).T)
                ringItem.setName(key)
                ringItem.setLineStyle(":")
                ringItem.setColor("#FFFFFFA0")
                ringItem.setLineBgColor("#000000A0")
                plot.addItem(ringItem)
```

image

t20100:
This PR fixes the display of `Shape` items with the matplotlib backend when they are display with dash and a background color.
It does so by changing the implementation of the feature to avoid using a proxy item in the backend

closes #3879

### DIFF
--- a/src/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/src/silx/gui/plot/backends/BackendMatplotlib.py
@@ -393,42 +393,47 @@ class _MarkerContainer(_PickableContainer):
         return self.line.contains(mouseevent)
 
 
-class _DoubleColoredLinePatch(matplotlib.patches.Patch):
-    """Matplotlib patch to display any patch using double color."""
+class SecondEdgeColorPatchMixIn:
+    """Mix-in class to add a second color for patches with dashed lines"""
 
-    def __init__(self, patch):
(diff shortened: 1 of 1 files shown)


src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -393,42 +393,47 @@ def contains(self, mouseevent):
...
+        self.set_edgecolor(self.get_second_edgecolor())
+        self.set_hatch(None)
+        super().draw(renderer)
 
-    def set_transform(self, transform):
-        self.__patch.set_transform(transform)
+        self.set_linestyle(linestyle)
+        self.set_edgecolor(edgecolor)
+        self.set_hatch(hatch)
+        super().draw(renderer)
```
t20100, line 428: I kept the same rendering as before, there's probably a better way, but this has been working fine so far.

PR #3909, opened 2023-07-19
Build: Removed constraint on `setuptools` version

t20100:
This PR removes the constraint of using `setuptools<60` that was introduced in PR #3583 to avoid issues with `numpy.distutils`.
`silx` no longer relies on `numpy.distutils`, so let's lift this constraint which should no longer be needed.

This PR also modernises CI config to use `python -m build` rather than `python setup.py`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,46 +24,46 @@ jobs:
           - name-suffix: "PyQt5 sdist"
             os: ubuntu-20.04
             python-version: '3.7'
-            BUILD_COMMAND: sdist
+            BUILD_OPTION: --sdist
             QT_BINDING: PyQt5
             RUN_TESTS_OPTIONS: --qt-binding=PyQt5 --no-opencl --low-mem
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -110,7 +109,7 @@ jobs:
           if [ ${{ runner.os }} == 'macOS' ]; then
               export MACOSX_DEPLOYMENT_TARGET=10.9;
           fi
-          python setup.py ${{ matrix.BUILD_COMMAND }}
```
t20100, line 113: Using `--no-isolation` to spare creating a virtual env and to use pre-release versions of `numpy` and `cython`

ci/appveyor.yml
```diff
@@ -75,7 +75,7 @@ build_script:
     - "pip list --format=columns"
 
     # Build
-    - "python setup.py bdist_wheel"
+    - "pythom -m build --no-isolation --wheel"
```
t20100, line 78: same here

PR #3910, opened 2023-07-19
Dependency: Replaced `setuptools`'s `pkg_resources` with `packaging` as runtime dependency

### LINKED ISSUE #3895: handle `pkg_resources` deprecation
looks like `pkg_resources` is deprecated. For parsing version `packaging` seems to the common solution

t20100:
This PR replaces usage of `pkg_resources` (deprecated) with `packaging` (new dependency), `importlib.resources` and `importlib.metadata`.

As a result, `setuptools` is no longer needed at runtime and so it is removed from `setup.py`'s `install_requires`.

closes #3895


requirements.txt
```diff
@@ -5,10 +5,14 @@
...
+# Required dependencies
+# From pyproject.toml
+wheel
 setuptools
+numpy >= 1.12
 Cython >= 0.21.1
+# From setup.py install_requires
+packaging
 h5py
 fabio >= 0.9
```
t20100, line 17: Reorder to document where it comes from

src/silx/gui/hdf5/test/test_hdf5.py
```diff
@@ -32,7 +32,7 @@
 import os
 import tempfile
 import numpy
-from pkg_resources import parse_version
+from packaging.version import Version
```
t20100, line 35: See https://packaging.pypa.io/en/stable/version.html#packaging.version.Version

src/silx/gui/utils/matplotlib.py
```diff
@@ -36,7 +36,6 @@
 
 
 import io
-from pkg_resources import parse_version
```
t20100, line 39: Unused import

src/silx/resources/__init__.py
```diff
@@ -52,28 +52,29 @@
...
-# pkg_resources is useful when this package is stored in a zip
-# When pkg_resources is not available, the resources dir defaults to the
-# directory containing this module.
-try:
-    import pkg_resources
-except ImportError:
-    pkg_resources = None
+# Manage resource files life-cycle
+_FILE_MANAGER = contextlib.ExitStack()
+atexit.register(_FILE_MANAGER.close)
```
t20100, line 77: See https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename for explanation

src/silx/resources/__init__.py
```diff
@@ -96,45 +97,36 @@
...
-                if pkg_resources is None:
-                    # In this case we have to compute the package path
-                    # Else it will not be used
-                    module = importlib.import_module(package_name)
-                    package_path = os.path.abspath(os.path.dirname(module.__file__))
-        self.package_name = package_name
-        self.package_path = package_path
-        self.forced_path = forced_path
+    package_name : str
+    forced_path : Optional[str] = None
```
t20100, line 103: `package_path` is no longer used because there is no more a fallback as there was if `pkg_resources` was not available.

src/silx/resources/__init__.py
```diff
@@ -155,60 +147,54 @@ def list_dir(resource):
...
-        path = resource_filename(resource)
-        return os.listdir(path)
-    else:
-        # Preferred way to get resources as it supports zipfile package
-        package_name = resource_directory.package_name
-        return pkg_resources.resource_listdir(package_name, resource_name)
+
+    # Preferred way to get resources as it supports zipfile package
+    package_name = '.'.join([resource_directory.package_name] + resource_name.split('/'))
+    return [entry.name for entry in importlib.resources.files(package_name).iterdir()]
```
t20100, line 162: See https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-listdir

src/silx/resources/__init__.py
```diff
@@ -270,15 +256,10 @@ def _resource_filename(resource, default_directory=None):
...
-        resource_path = os.path.join(base_dir, *resource_name.split('/'))
-        return resource_path
-    else:
-        # Preferred way to get resources as it supports zipfile package
-        package_name = resource_directory.package_name
-        return pkg_resources.resource_filename(package_name, resource_name)
+
+    package_name = resource_directory.package_name
+    ref = importlib.resources.files(package_name) / resource_name
+    return _FILE_MANAGER.enter_context(importlib.resources.as_file(ref))
```
t20100, line 262: See https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename

src/silx/gui/icons.py
```diff
@@ -399,7 +399,7 @@ def getQFile(name):
     for format_ in _supported_formats:
         format_ = str(format_)
         filename = silx.resources._resource_filename('%s.%s' % (name, format_),
-                                                     default_directory=os.path.join('gui', 'icons'))
```
t20100, line 402: It should always use `/` and not `\\`

src/silx/resources/__init__.py
```diff
@@ -233,18 +215,27 @@ def resource_filename(resource):
...
     :return: Absolute resource path in the file system
-    :rtype: str
     """
     return _resource_filename(resource, default_directory=None)
 
 
-def _resource_filename(resource, default_directory=None):
+# Manage resource files life-cycle
+_file_manager = contextlib.ExitStack()
+atexit.register(_file_manager.close)
```
t20100, line 230: See https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename for explanation

t20100:
Impossible to do without `pkg_resources` before python3.9 unless using `importlib_resources` backport package.... which is not packaged for ubuntu20.04...

So `pkg_resources` will still be used for python<3.9.

src/silx/resources/__init__.py
```diff
@@ -217,14 +203,14 @@ def _get_package_and_resource(resource, default_directory=None):
     else:
         prefix = "silx"
         if default_directory is not None:
-            resource = os.path.join(default_directory, resource)
+            resource = f"{default_directory}/{resource}"
```
payno, line 206: this one I don't get

src/silx/resources/__init__.py
```diff
@@ -258,9 +261,17 @@ def _resource_filename(
...
+    if cached_path is not None:
+        return cached_path
+
     ref = importlib.resources.files(package_name) / resource_name
-    path = _FILE_MANAGER.enter_context(importlib.resources.as_file(ref))
-    return str(path.absolute())
+    path = _file_manager.enter_context(importlib.resources.as_file(ref))
+    path_string = str(path.absolute())
+    _file_cache[cache_key] = path_string
+    return path_string
```
payno: question: is using `functools.cache` wouldn't simplify things ? something like:

```
def _resource_filename(
    resource: str, default_directory: Optional[str] = None
) -> str:
    resource_directory, resource_name = _get_package_and_resource(resource,
                                                                  default_directory=default_directory)

    if resource_directory.forced_path is not None:
        # if set, use this directory
        base_dir = resource_directory.forced_path
        resource_path = os.path.join(base_dir, *resource_name.split('/'))
        return resource_path

    package_name = resource_directory.package_name
    return  _get_cached_path(package_name, resource_name)

@cache
def _get_cached_path(package_name, resource_name):
       if sys.version_info < (3, 9):
        return pkg_resources.resource_filename(package_name, resource_name)

    file_context = importlib.resources.as_file(
        importlib.resources.files(package_name) / resource_name)
    path = _file_manager.enter_context(file_context)
    path_string = str(path.absolute())
    return path_string
```

payno (commented):
seems to work on my side.
If I understand why the os.path is replaced by `/` then I guess we can merge it.

src/silx/resources/__init__.py
t20100: Having a 2nd look at this, I think we can remove the `default_directory` stuff, it's only accessible from private API and shouldn't be needed.

src/silx/resources/__init__.py
t20100: Yes! I'll use it

src/silx/resources/__init__.py
t20100: Forget it, this is need to simplify loading icons without passing `gui/icons/` as a prefix

src/silx/resources/__init__.py
t20100: Done

t20100:
Resource relative path to package are expected with `/`

It was written in the doc:
https://github.com/silx-kit/silx/blob/4e64c9acdb2c30e0b3cd01e030f2c3a9636a5b44/src/silx/resources/__init__.py#L236-L238

And is parsed with `split('/')` to be converted to a proper path:
https://github.com/silx-kit/silx/blob/4e64c9acdb2c30e0b3cd01e030f2c3a9636a5b44/src/silx/resources/__init__.py#L271

PR #3912, opened 2023-07-21
WaiterOverlay updates

t20100:
A few updates of PR #3876:
- Change way to handle parent/setParent
- Defer resize handling in eventHandler
- For PlotWidget parent: Support change of backend and toolbars/dockwidgets
- Changed name to `WaitingOverlay` to be inline with `WaitingPushButton`
- Added a `text` getter and a test.

There is an issue when `text` is set: the size is not enlarged. It probably comes from `WaitingPushButton`.


payno (approved):

vallsv:
:+1:

PR #3924, opened 2023-08-22
Do not probe GPU interfaced with PoCL

kif:
Changelog: prevent crash at init when loading silx with PortableCL + Cuda devices. Work-around


pierrepaleo:
Good to know. Thanks

PR #3926, opened 2023-08-30
Fix: Datasets with many rows fail to load due to incorrect variable type

### LINKED ISSUE #3925: Datasets with many rows fail to load due to incorrect variable type
## Issue

silx fails to load the details of datasets with more than 10 million rows with error:

```invalid result from RecordTableModel.rowCount(), 'float' object cannot be interpreted as an integer```

## Root cause

`MAX_NUMBER_OF_ROWS` is defined with a float literal, thus if the dataset reaches `MAX_NUMBER_OF_ROWS`, then `rowCount()` returns a `float` instead of `int`.
https://github.com/silx-kit/silx/blob/5e33cb69afd2a8b1cfe3183282acdd8b34c1a74f/src/silx/gui/data/RecordTableView.py#L151

### PR DESCRIPTION
**Pull request by carez**
## Issue 

Fixes #3925

`MAX_NUMBER_OF_ROWS` is defined with a float literal, thus if a dataset has more than `MAX_NUMBER_OF_ROWS`, `rowCount()` returns a `float` instead of `int` and the dataset does not load.

```
>>> type(10e6)
<class 'float'>
```

## Fix

Set limits as integers, not floats.

### DIFF
--- a/examples/plot3dUpdateScatterFromThread.py
+++ b/examples/plot3dUpdateScatterFromThread.py
@@ -49,7 +49,7 @@ from silx.gui.plot3d.SceneWindow import SceneWindow
 from silx.gui.plot3d import items
 
 
-MAX_NUMBER_OF_POINTS = int(10**6)
+MAX_NUMBER_OF_POINTS = 10**6
 
 
 class UpdateScatterThread(threading.Thread):


vallsv:
Thanks a lot for that fix.
I wasn't aware of that.
I maybe depend on the Python version.

vallsv:
Be careful i have sckatched your changes

t20100:
Thanks for the fix!

That can also depends on pyqt versions, and it's not something we're getting into so often.

PR #3937, opened 2023-09-14
[dicttoh5] better error message for non-serializable data

woutdenolf:
Changelog: better error message for dicttoh5 with non-serializable data

### DIFF
--- a/src/silx/io/dictdump.py
+++ b/src/silx/io/dictdump.py
@@ -361,14 +361,14 @@ def dicttoh5(
                         h5f.create_dataset(h5name, data=data)
                     else:
                         h5f.create_dataset(h5name, data=data, **create_dataset_args)
-                except Exception:
+                except Exception as e:
                     if isinstance(data, numpy.ndarray):
                         dtype = f"numpy.ndarray-{data.dtype}"
                     else:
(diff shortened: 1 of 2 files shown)


sdebionne:
This will pretty be useful, thanks!

src/silx/io/dictdump.py
```diff
@@ -334,20 +347,28 @@ def _iter_treedict(attributes=False):
...
-                else:
-                    h5f.create_dataset(h5name,
-                                       data=data,
-                                       **create_dataset_args)
+                try:
+                    if data.shape == () or create_dataset_args is None:
+                        h5f.create_dataset(h5name, data=data)
+                    else:
+                        h5f.create_dataset(h5name, data=data, **create_dataset_args)
+                except Exception as e:
```
t20100, line 364: No way to have a stricter selection of exceptions?

t20100 (approved):
LGTM

src/silx/io/dictdump.py
woutdenolf: I reduced it to `TypeError` only. Ok?

t20100: Yes :+1:

Issue #3942, opened 2023-09-28
silx compare - typo


payno:
Looks like there is a small typo on a set definition. Get the following error:

```
TypeError: set expected at most 1 argument, got 2
ERROR:silx.app.compare.CompareImagesWindow:Error while loading URL bambou_hercules_0001_rec_2058.800_00000.hdf5?/entry0000/reconstruction/results/data
Traceback (most recent call last):
  File "/home/lesaint/envs/tomotools/lib/python3.9/site-packages/silx/app/compare/CompareImagesWindow.py", line 116, in _updateImageB
    data = self.readData(urlPath)
  File "/home/lesaint/envs/tomotools/lib/python3.9/site-packages/silx/app/compare/CompareImagesWindow.py", line 141, in readData
    if data.ndim == 3 and data.shape[0] in set(3, 4):
TypeError: set expected at most 1 argument, got 2
```

lesaintjerome:
OK. I probably need to be more precise on the issue. The error happened when with `silx compare "bambou_hercules_0001_rec_2040.800_00000.hdf5?path=/entry0000/reconstruction/results/data" "bambou_hercules_0001_rec_2058.800_00000.hdf5?path=/entry0000/reconstruction/results/data"`, i.e. with no slice number. But when the command `silx compare "bambou_hercules_0001_rec_2040.800_00000.hdf5?path=/entry0000/reconstruction/results/data&slice=0" "bambou_hercules_0001_rec_2058.800_00000.hdf5?path=/entry0000/reconstruction/results/data&slice=0"`is used (i.e. with slice number), then it works fine. That's why I thought stacks were possible....

lesaintjerome:
It would be nice to have the possibility to compare stacks (with a slider to run through the slices, simultaneously in both stacks). Might be an enhancement request?

vallsv:
BTW i don't think stacks are supported

vallsv:
Closed by #3943

PR #3943, opened 2023-09-28
Issue #3942 - silx compare typo

lesaintjerome:
Fix python `set(3,4)` command with `set([3,4])`.

### DIFF
--- a/src/silx/app/compare/CompareImagesWindow.py
+++ b/src/silx/app/compare/CompareImagesWindow.py
@@ -138,7 +138,7 @@ class CompareImagesWindow(qt.QMainWindow):
 
         if data.ndim == 2:
             return data
-        if data.ndim == 3 and data.shape[0] in set(3, 4):
+        if data.ndim == 3 and data.shape[0] in set([3, 4]):
             return data
 
         raise ValueError(f"URL '{urlPath}' does not link to an numpy image")


src/silx/app/compare/CompareImagesWindow.py
```diff
@@ -138,7 +138,7 @@ def readData(self, urlPath: str):
 
         if data.ndim == 2:
             return data
-        if data.ndim == 3 and data.shape[0] in set(3, 4):
+        if data.ndim == 3 and data.shape[0] in set([3, 4]):
```
t20100, line 141: Also, I guess this is to support RGB(A) images, and so checking the number of channels which is usually the last dimension:
```suggestion
        if data.ndim == 3 and data.shape[2] in set([3, 4]):
```

attn @vallsv

lesaintjerome: My use case was rather about comparing 2 (gray level) 3D volumes, possibly each of them with only one slice. Slice being the first index.

t20100: OK, I don't think this is supported: It is meant to support comparing 2 images..
Otherwise the limit to 3 or 4 slices makes no sense, while it does for RGB(A).

What could be added for stack with only one slice could be:
```python
if data.ndim == 3 and data.shape[0] == 1:
    return data[0]
```

t20100: What's loaded here is then passed to:

https://github.com/silx-kit/silx/blob/e912d5e520262a73095252a6e825b42bc0b6d20d/src/silx/gui/plot/CompareImages.py#L473-L482

lesaintjerome: OK. I probably need to be more precise on the issue. The error happened when with 
```
silx compare "bambou_hercules_0001_rec_2040.800_00000.hdf5?path=/entry0000/reconstruction/results/data" "bambou_hercules_0001_rec_2058.800_00000.hdf5?path=/entry0000/reconstruction/results/data"
```
, i.e. with no slice number. But when the command 
```
silx compare "bambou_hercules_0001_rec_2040.800_00000.hdf5?path=/entry0000/reconstruction/results/data&slice=0" "bambou_hercules_0001_rec_2058.800_00000.hdf5?path=/entry0000/reconstruction/results/data&slice=0"
```
is used (i.e. with slice number), then it works fine. That's why I thought stacks were possible....

vallsv: Yes, to support slicing with a slider for example, you better to create another dedicated issue.

I dont think we will have tome to do it now.

Optionally you could try to see if `slice=*` works, but i am pretty sure it is not supported.

vallsv: @t20100 i agree there is something weird with the shape index which is supposed to be ordered `y,x,rgb[a]`. Ill check on my side.

vallsv: BTW, this `shape[0]` is maybe the layout provided by fabio or another lib when we load a RGB image. I have to check.

vallsv:
I merge this PR for convenience.
I will rework the other proposals with the PR #3944

PR #3953, opened 2023-10-23
silx view: Fixed setting focus at startup when opening a dataset

### LINKED ISSUE #3916: silx view: focus is not done when specifying the internal HDF5 path
When doing `silx view somefile.h5`, the focus is done in the left pane by default (the one showing the file structure), on the first entry.
Therefore, when the viewer window pops up, the file structure can be navigated directly with the keyboard only.

Now when doing `silx view somefile.h5::entry0000/data/data`, the focus is not done - the left pane has to be clicked at least once.

Probably the default focus is missing when specifying a HDF5 data path.

t20100:
This PR aims at opening the vis at startup when providing a data path along with the file to open: closes  #3916

To achieve so, this PR adds an extra argument to `Hdf5TreeModel.sigH5pyObjectLoaded` in order to pass the full filename used to open the item.

### DIFF
--- a/src/silx/app/view/Viewer.py
+++ b/src/silx/app/view/Viewer.py
@@ -453,9 +453,9 @@ class Viewer(qt.QMainWindow):
         layout.addWidget(customNxdataWidget)
         return widget
 
-    def __h5FileLoaded(self, loadedH5):
+    def __h5FileLoaded(self, loadedH5, filename):
         self.__context.pushRecentFile(loadedH5.file.filename)
-        if loadedH5.file.filename == self.__displayIt:
+        if filename == self.__displayIt:
             self.__displayIt = None
             self.displayData(loadedH5)
 
--- a/src/silx/gui/hdf5/Hdf5TreeModel.py
+++ b/src/silx/gui/hdf5/Hdf5TreeModel.py
@@ -250,7 +250,6 @@ class Hdf5TreeModel(qt.QAbstractItemModel):
         """Static method to close explicit references to internal objects."""
         _logger.debug("Clear Hdf5TreeModel")
         for obj in fileList:
-            _logger.debug("Close file %s", obj.filename)
             obj.close()
         fileList[:] = []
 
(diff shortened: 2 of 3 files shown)


src/silx/app/view/Viewer.py
```diff
@@ -453,9 +453,9 @@ def __createCustomNxdataWindow(self, customNxdataWidget):
         layout.addWidget(customNxdataWidget)
         return widget
 
-    def __h5FileLoaded(self, loadedH5):
+    def __h5FileLoaded(self, loadedH5, filename):
```
t20100, line 456: Retrieved from `Hdf5TreeModel.sigH5pyObjectLoaded`

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -250,7 +250,6 @@ def _closeFileList(fileList):
         """Static method to close explicit references to internal objects."""
         _logger.debug("Clear Hdf5TreeModel")
         for obj in fileList:
-            _logger.debug("Close file %s", obj.filename)
```
t20100, line 253: This was causing issues with the tests since `h5py.Dataset` do not have a `filename` attribute.

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -703,7 +709,7 @@ def insertFile(self, filename, row=-1):
             h5file = silx_io.open(filename)
             if self.__ownFiles:
                 self.__openedFiles.append(h5file)
-            self.sigH5pyObjectLoaded.emit(h5file)
+            self.sigH5pyObjectLoaded.emit(h5file, filename)
             self.insertH5pyObject(h5file, row=row, filename=filename)
```
t20100, line 713: Not sure here why the signal is emitted **before** inserting the item into the model? :shrug:

pierrepaleo:
I can't really review the code, but I gave a few tries and it seems to be working. 

Thanks!

PR #3956, opened 2023-10-26
silx.gui.plot.items: Added Marker item font configuration; silx.config: Added `DEFAULT_PLOT_MARKER_TEXT_FONT_SIZE` config

### LINKED ISSUE #3931: plot: editable default text size
With big screen it could be nice to grow up text size of markers

We could provide that as a `silx.config.DEFAULT_MARKER_TEXT_SIZE` or something like that.

Such config you have to setup before creating a plot.

t20100:
This PR adds:
- Control over marker text's `QFont` through `MarkerBase.get|setFont` methods.
- General configuration of marker text's `QFont` through `silx.config.DEFAULT_PLOT_MARKER_TEXT_FONT`

A marker QFont can be `None` (the default) to use the default. This is useful for letting `matplotlib` pick its own default.

closes #3931

### DIFF
--- a/src/silx/gui/plot/backends/glutils/GLPlotFrame.py
+++ b/src/silx/gui/plot/backends/glutils/GLPlotFrame.py
@@ -1,6 +1,6 @@
 # /*##########################################################################
 #
-# Copyright (c) 2014-2022 European Synchrotron Radiation Facility
+# Copyright (c) 2014-2023 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -44,6 +44,7 @@ from collections import namedtuple
 
 import numpy
 
+from .... import qt
 from ...._glutils import gl, Program
 from ..._utils import checkAxisLimits, FLOAT32_MINPOS
 from .GLSupport import mat4Ortho
@@ -215,6 +216,8 @@ class PlotAxis(object):
         labels = []
         tickLabelsSize = [0., 0.]
 
+        font = qt.QApplication.instance().font()
+
         xTickLength, yTickLength = self._tickLength
         xTickLength *= self.devicePixelRatio
         yTickLength *= self.devicePixelRatio
@@ -225,6 +228,7 @@ class PlotAxis(object):
                 tickScale = 1.
 
                 label = Text2D(text=text,
+                               font=font,
                                color=self._foregroundColor,
                                x=xPixel - xTickLength,
                                y=yPixel - yTickLength,
@@ -258,6 +262,7 @@ class PlotAxis(object):
         # yOffset -= 3 * yTickLength
 
         axisTitle = Text2D(text=self.title,
+                           font=font,
                            color=self._foregroundColor,
                            x=xAxisCenter + xOffset,
                            y=yAxisCenter + yOffset,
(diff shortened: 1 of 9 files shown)


src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -215,6 +216,8 @@ def getVerticesAndLabels(self):
         labels = []
         tickLabelsSize = [0., 0.]
 
+        font = qt.QApplication.instance().font()
```
t20100, line 219: Using QApplication default for plot labels/title. This can be made configurable if needed

src/silx/gui/plot/backends/glutils/GLText.py
```diff
@@ -161,9 +165,13 @@ def __init__(self, text, x=0, y=0,
 
         self._rotate = numpy.radians(rotate)
 
-    def _getTexture(self, text, devicePixelRatio):
+    def _textureKey(self) -> tuple[str, str, float]:
+        """Returns the current texture key"""
+        return self.text, self.font.key(), self.devicePixelRatio
```
t20100, line 170: font is added to the texture cache key

src/silx/gui/utils/matplotlib.py
```diff
@@ -41,6 +41,36 @@
 
 from .. import qt
 
+# This must be performed before any import from matplotlib
+if qt.BINDING in ("PySide6", "PyQt6", "PyQt5"):
+    matplotlib.use("Qt5Agg", force=False)
+    from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg  # noqa
+
+else:
+    raise ImportError("Unsupported Qt binding: %s" % qt.BINDING)
```
t20100, line 50: Moved this first to avoid lazy loading in function

src/silx/gui/utils/matplotlib.py
```diff
@@ -41,6 +41,36 @@
...
+
+def qFontToFontProperties(font: qt.QFont):
+    """Convert a QFont to a matplotlib FontProperties"""
+    weightFactor = 10 if qt.BINDING == "PyQt5" else 1
+    return FontProperties(
+        family=[font.family(), font.defaultFamily()],
+        style=_FONT_STYLES[font.style()],
+        weight=weightFactor * font.weight(),
+        size=font.pointSizeF(),
+    )
```
t20100, line 72: Extracted from `rasterMathText` and improved conversion... but font family look-up might vary between Qt and matplotlib

vallsv:
I am not a huge fan of the fact `DEFAULT_PLOT_MARKER_TEXT_FONT` will not be anymore a bar basic python types, but it's up to you.

src/silx/gui/utils/matplotlib.py
vallsv: I don't understand why you put such comment in the PR.

If it's useful to understand the code, better to put such comment in the code itself, no?

vallsv (approved):

src/silx/gui/utils/matplotlib.py
t20100: It's to explain the modification: It's no longer needed.
A comment in the code saying I first import modules is pointless.

t20100:
I replaced `DEFAULT_PLOT_MARKER_TEXT_FONT` with `DEFAULT_PLOT_MARKER_TEXT_FONT_SIZE` to only use basic types in the config.

PR #3958, opened 2023-10-26
silx.gui.utils.image: Added support of `QImage.Format_Grayscale8` to `convertQImageToArray`

t20100:
Using grayscale8 QImage (available in Qt>=5.5) avoids some array copies/conversion when dealing with text in the OpenGL backend.

This moves minimum required version from Qt5.3 to Qt5.5... I documented minimum version to 5.6 ton Linux to be inline with Windows and macOS. BTW, we don't test with an old version of pyQt... so the effective min version requirement is probably higher.

Small improvement/clean-up done while working on PR #3956

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -10,11 +10,11 @@ This table summarizes the support matrix of silx:
 +------------+--------------+--------------------------------+
 | System     | Python vers. | Qt and its bindings            |
 +------------+--------------+--------------------------------+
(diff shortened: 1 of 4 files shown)


vallsv:
Dependency `PyQt5.6` looks really optimistic no?

I am pretty sure there is dependency with at least `5.9`, which is imo not a problem.

t20100:
> Dependency with PyQt5.6 looks optimistic no?

+1, never checked I'd say.
I agree 5.9 is more realistic, I'll update it.

PR #3959, opened 2023-10-30
PlotWindow: add an option to get tape measure

payno:
Changelog: 

add tape measure action

# screenshot

Screenshot from 2023-11-16 12-56-40


vallsv:
Sounds good to me.

And to answer to your message about displaying or not start and end points, i think it is not needed, and it's counter intuitive to display them. If you want to mesure something, just display the measure. If you need coords, you already have the tooltip or the status bar.

For other stuffs like `coef` or `ndigits`, this can be handled by overwrite of method.
You just have to provide the default behaviour.

```
class Ruler:
    ...

    # One can easely override that to change the digits, the scale, the unit
    def getLabelFromDeltaPixels(self, dx: float, dy: float):
        return f"{sqrt(x*x + y*y):0.2f} px"
```

If you provide such component, that's something i will integrate directly in flint.

vallsv:
About the code
- I would not add it by default in the plots, that's a tool you can integrate yourself
- I would split what is about the toolbutton and what is about the ruler behaviour itself (but i am fine with it for now, it's easier to override with a single class)
- I would create an even minimalistic icon, else it will not be readable (i.e. https://www.creativefabrica.com/fr/product/ruler-icon-2/ with even less detail) tell me if you want i can do the stuff.

payno:
thanks, don't worry I can do it all.

vallsv:
I haven't cheak the interaction, i have not much time but if think:

- `format_distance` -> `formatDistance`
- `format_distance` must not be a class method, else it's pointless i can't override it
- you should cut the ruler icon by 3, there is no need to have realistic image -> only 2 or 3 tick is enough -> as result you can have a larger ruler for readability

I also would like to tune the line style, but i could try to do it later

I also think it's not the best idea to use a ROI here, because you need a dedicated roi manager for a single roi, but if it does the job, it's fine.

vallsv:
I think `format_distance` should have `dx` and `dy` as argument, you can't expect the coefficient is the same for x and y. But well, this can be considered as another use case.

payno:
Having 'format_distance' not being a 'staticmethod' is not that straightforward sadly.

using `RulerROI` with the roi manager is 'breaking' the link between the `RulerToolButton` and the `RulerROI`.
So we can change the roi manager to allow providing other parameters or avoid using manager (as you suggested on a previous command). But in this case we need to redo all the management of interaction.

An alternative to avoid having heavy modifications is the commit '917ff73fb'. This is not very elegant but it will do the job and should allow us to extend the class as we wish. Else I guess we can also do something with metaclasses but again it will be against readability.

vallsv:
Thanks for the fix.
Looks good to me.

vallsv (approved):

PR #3960, opened 2023-11-03
[GUI] Histogram Action. QHoverEvent does not have attribute pos

vasole:
Changelog: RangeSlider compatibility with PyQt6

### DIFF
--- a/src/silx/gui/plot/actions/histogram.py
+++ b/src/silx/gui/plot/actions/histogram.py
@@ -277,7 +277,7 @@ class HistogramWidget(qt.QWidget):
         controlsLayout.addWidget(self.__rangeSlider)
         self.__weightCheckBox = qt.QCheckBox(self)
         self.__weightCheckBox.setText("Use weights")
-        self.__weightCheckBox.stateChanged[int].connect(self.__weightChanged)
+        self.__weightCheckBox.clicked.connect(self.__weightChanged)
         controlsLayout.addWidget(self.__weightCheckBox)
         controlsLayout.addStretch(1)
 
--- a/src/silx/gui/plot/actions/histogram.py
+++ b/src/silx/gui/plot/actions/histogram.py
@@ -401,11 +401,12 @@ class HistogramWidget(qt.QWidget):
         self.__rangeSlider.setRange(*range_)
         self.__rangeSlider.setPositions(*previousPositions)
 
+        data = array.ravel().astype(numpy.float32)
         histogram = Histogramnd(
-            array.ravel().astype(numpy.float32),
+            data,
             n_bins=max(2, self.__nbinsLineEdit.getValue()),
             histo_range=self.__rangeSlider.getValues(),
-            weights=array.ravel().astype(numpy.float32),
+            weights=data,
         )
         if len(histogram.edges) != 1:
             _logger.error("Error while computing the histogram")
--- a/src/silx/gui/plot/actions/histogram.py
+++ b/src/silx/gui/plot/actions/histogram.py
@@ -31,7 +31,7 @@ The following QAction are available:
 """
 
 __authors__ = ["V.A. Sole", "T. Vincent", "P. Knobel"]
-__date__ = "01/12/2020"
+__date__ = "07/11/2023"
 __license__ = "MIT"
 
 from typing import Optional, Tuple
@@ -275,6 +275,10 @@ class HistogramWidget(qt.QWidget):
             self.__updateHistogramFromControls)
         self.__rangeSlider.sigValueChanged.connect(self.__rangeChanged)
         controlsLayout.addWidget(self.__rangeSlider)
+        self.__weightCheckBox = qt.QCheckBox(self)
+        self.__weightCheckBox.setText("Use weights")
+        self.__weightCheckBox.stateChanged[int].connect(self.__weightChanged)
+        controlsLayout.addWidget(self.__weightCheckBox)
         controlsLayout.addStretch(1)
 
         # Stats display
@@ -347,6 +351,9 @@ class HistogramWidget(qt.QWidget):
         self.__rangeSlider.setToolTip(tooltip)
         self.__rangeLabel.setToolTip(tooltip)
 
+    def __weightChanged(self, value):
+        self._updateFromItem()
+
     def _updateFromItem(self):
         """Update histogram and stats from the item"""
         item = self.getItem()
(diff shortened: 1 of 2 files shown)


vasole:
Problem triggered under PyQt6.

vasole:
closes #3961

src/silx/gui/plot/actions/histogram.py
```diff
@@ -398,13 +405,19 @@ def _updateFromItem(self):
             array.ravel().astype(numpy.float32),
             n_bins=max(2, self.__nbinsLineEdit.getValue()),
             histo_range=self.__rangeSlider.getValues(),
+            weights=array.ravel().astype(numpy.float32),
```
payno, line 408: maybe we could create an intermediary array like
```
data = array.ravel().astype(numpy.float32)
        histogram = Histogramnd(
            data,
            n_bins=max(2, self.__nbinsLineEdit.getValue()),
            histo_range=self.__rangeSlider.getValues(),
            weights=data,
        )
``` 
as we are using it twice this would avoid to do the operation several time.

src/silx/gui/plot/actions/histogram.py
```diff
@@ -275,6 +275,10 @@ def __init__(self, *args, **kwargs):
             self.__updateHistogramFromControls)
         self.__rangeSlider.sigValueChanged.connect(self.__rangeChanged)
         controlsLayout.addWidget(self.__rangeSlider)
+        self.__weightCheckBox = qt.QCheckBox(self)
+        self.__weightCheckBox.setText("Use weights")
+        self.__weightCheckBox.stateChanged[int].connect(self.__weightChanged)
```
payno, line 280: small detail but I think the 'silx policy' would be more to use the 'clicked' signal no @t20100 ?

payno:
looks like pyqt6 skipped the deprecation part (https://doc.qt.io/qt-6/qhoverevent-obsolete.html#pos)

src/silx/gui/widgets/RangeSlider.py
```diff
@@ -500,6 +505,14 @@ def setGroovePixmapFromProfile(self, profile, colormap=None):
         self.setGroovePixmap(qpixmap)
 
     # Handle interaction
+    def _mouseEventPosition(self, event):
+        if hasattr(event, "pos"):
+            # qt-5 returns QPoint
+            position = event.pos()
+        else:
+            # qt-6 returns QPointF
+            position = event.position()
```
payno, line 514: I think the following would be more coherent with existing code'. and from what I understood we should also use `position` for Pyside6
```suggestion
        if qt.BINDING in ('PyQt6', 'PySide6'):
            position = event.position()
            # qt-6 returns QPointF
        else:
            # qt-5 returns QPoint
            position = event.pos()
```

payno (commented):
Except some small details looks good to me :)

src/silx/gui/widgets/RangeSlider.py
vasole: I took the least invasive approach but I can change it.

However, testing for the binding instead of the presence of the method is not future proven.

If you want to use position for PyQt6 and PySide6, then one has to test for `position` and not for `pos`.

src/silx/gui/plot/actions/histogram.py
vasole: OK

src/silx/gui/widgets/RangeSlider.py
vasole: If one wants to stick to using qt.BINDING, the test should be for using PyQt5 so that one is defaulting to the newer method.

payno: I agree but I think for the time being Qt5 is still the default / mostly used. I don't think we switched to Qt6 as the default Qt version yet.

payno: > I took the least invasive approach but I can change it.
> 
> However, testing for the binding instead of the presence of the method is not future proven.
> 
> If you want to use position for PyQt6 and PySide6, then one has to test for `position` and not for `pos`.

My point of view regarding the check of the biding is that you might be safer for future checking the binding.
Because if somehow this object get `pos` and `position` attribute then we would validate the condition anyway (I agree this is a highly unlikely ...)

vasole: In the past there have been versions of PyQt/PySide with the capability to handle two different versions of Qt behind.

Personally if I have to use a method, I prefer to test for the presence of the method and not for the version.

@t20100 any opinion on this?

src/silx/gui/plot/actions/histogram.py
vasole: For a checkbox that is going to be accessed through the user interface, the two signals are indeed the same.

However, stateChanged is emitted when the checkbox changes either programmatically or via the interface. clicked is not emitted when the checkbox is modified via the interface.

So, in this case, I do not care.

vasole: https://doc.qt.io/qt-6/qabstractbutton.html#clicked

src/silx/gui/widgets/RangeSlider.py
```diff
@@ -122,7 +122,12 @@ def __init__(self, parent=None):
     def event(self, event):
         t = event.type()
         if t == qt.QEvent.HoverEnter or t == qt.QEvent.HoverLeave or t == qt.QEvent.HoverMove:
-            return self.__updateHoverControl(event.pos())
+            if hasattr(event, "position"):
+               # qt-6
+               return self.__updateHoverControl(event.position().toPoint())
+            else:
+               # qt-5
+               return self.__updateHoverControl(event.pos())
```
payno, line 130:
```suggestion
            return self.__updateHoverControl(
                self._mouseEventPosition(event)
            )
```

vasole: No, it is a different type of event and it has to be handled differently. One is a mouseEvent and the other a HoverEvent.

payno: indeed you are right, I was thinking that the binding should also be used here

payno, line 130:
```suggestion
            if qt.BINDING in ("PyQt5",):
               # qt-5
               return self.__updateHoverControl(event.pos())
            else:
               # qt-6
               return self.__updateHoverControl(event.position().toPoint())
```

PR #3964, opened 2023-11-16
ColormapDialog: upgrade `_DataInPlotMode` to `DisplayMode`

### LINKED ISSUE #3963: Using 'histogram' mode of the ColorMapDialog from PlotWindow
I would like to use the 'histogram' mode of the ColormapDialog by default in a Plot.

From what I see for now I would have to:

- either redefine the ColormapAction
- either force the creation of the ColorMapDialog and force the histogram to be used (using the protected API of _DataInPlotMode and _ColormapHistogram._setDataInPlotMode ).
 
Is it correct ? Does someone know other ways ?

payno:
The goal of this PR is to allow user select the display mode of the ColormapDialog for the min / max value (`data histogram` or `data range`)

# TODO

- [x] move _DataInPlotMode to public API
- [x] add getter for `_dataInPlotMode`
- [x] rename `setColorDialog` to `setColormapDialog` ? This would make more sense to me. Especially as we are moving to 2.0
    - [x] add deprecation warning
- [x] add a getColor(map)Dialog to ColormapAction

# EXTRA
Changelog: 

move _DataInPlotMode to public API.
close #3963

### DIFF
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -66,7 +66,7 @@ import enum
 import logging
 
 import numpy
-from typing import Union
+from future import __annotations__
 
 from .. import qt
 from .. import utils
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -567,6 +567,7 @@ class _ColormapHistogram(qt.QWidget):
         self._plotToolbar.addAction(action)
         group.addAction(action)
         self._dataHistogramAction = action
+        group.setExclusive(True)
         group.triggered.connect(self._displayDataInPlotModeChanged)
 
         plotBoxLayout = qt.QHBoxLayout()
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -755,7 +755,7 @@ class _ColormapHistogram(qt.QWidget):
             x = min(x, vmax)
         return x, y
 
-    def setDataInPlotMode(self, mode: Union[str, DataInPlotMode]):
+    def setDataInPlotMode(self, mode: str | DataInPlotMode):
         mode = DataInPlotMode.from_value(mode)
         if mode is DataInPlotMode.HISTOGRAM:
             action = self._dataHistogramAction
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -547,7 +547,7 @@ class _ColormapHistogram(qt.QWidget):
 
         group = qt.QActionGroup(self._plotToolbar)
         group.setExclusive(True)
-
+        # data range mode
         action = qt.QAction("Data range", self)
         action.setToolTip("Display the data range within the colormap range. A fast data processing have to be done.")
         action.setIcon(icons.getQIcon('colormap-range'))
@@ -556,6 +556,8 @@ class _ColormapHistogram(qt.QWidget):
         action.setChecked(action.data() == self._dataInPlotMode)
         self._plotToolbar.addAction(action)
         group.addAction(action)
+        self._dataRangeAction = action
+        # histogram mode
         action = qt.QAction("Histogram", self)
         action.setToolTip("Display the data histogram within the colormap range. A slow data processing have to be done. ")
         action.setIcon(icons.getQIcon('colormap-histogram'))
@@ -564,6 +566,7 @@ class _ColormapHistogram(qt.QWidget):
         action.setChecked(action.data() == self._dataInPlotMode)
         self._plotToolbar.addAction(action)
         group.addAction(action)
+        self._dataHistogramAction = action
         group.triggered.connect(self._displayDataInPlotModeChanged)
 
         plotBoxLayout = qt.QHBoxLayout()
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -85,6 +85,7 @@ from silx.gui.widgets.FormGridLayout import FormGridLayout
 from silx.math.histogram import Histogramnd
 from silx.gui.plot.items.roi import RectangleROI
 from silx.gui.plot.tools.roi import RegionOfInterestManager
+from silx.utils.enum import Enum as _Enum
 
 _logger = logging.getLogger(__name__)
 
@@ -301,7 +302,7 @@ class _AutoScaleButton(qt.QPushButton):
             self.setChecked(autoRange[0] if autoRange[0] == autoRange[1] else False)
 
 @enum.unique
-class DataInPlotMode(enum.Enum):
+class DataInPlotMode(_Enum):
     """Enum for each mode of display of the data in the plot."""
     RANGE = 'range'
     HISTOGRAM = 'histogram'
--- a/src/silx/gui/dialog/ColormapDialog.py
+++ b/src/silx/gui/dialog/ColormapDialog.py
@@ -66,6 +66,7 @@ import enum
 import logging
 
 import numpy
+from typing import Union
 
 from .. import qt
 from .. import utils
@@ -300,7 +301,7 @@ class _AutoScaleButton(qt.QPushButton):
             self.setChecked(autoRange[0] if autoRange[0] == autoRange[1] else False)
 
 @enum.unique
-class _DataInPlotMode(enum.Enum):
+class DataInPlotMode(enum.Enum):
     """Enum for each mode of display of the data in the plot."""
     RANGE = 'range'
     HISTOGRAM = 'histogram'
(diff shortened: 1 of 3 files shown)


vasole:
| rename setColorDialog to setColormapDialog ? This would make more sense to me

That is change to a public API without any deprecation warning. I would appreciate more conservative changes.

For instance, that can break PyMca:

image

vasole:
@payno 

Are you sure it is not actually a `Color` dialog (to choose a color and not a colormap)?

payno:
Well the expected object is a `ColormapDialog` and the associated action is named `ColormapAction` so at least it would be coherent. But then depends on the lexical language people are using probably

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -749,12 +755,26 @@ def _plotGammaMarkerConstraint(self, x, y):
             x = min(x, vmax)
         return x, y
 
+    def setDataInPlotMode(self, mode: Union[str, DataInPlotMode]):
```
t20100, line 758: By adding a `from future import __annotations__` you can already spare the `from typing import Union` and write:
```suggestion
    def setDataInPlotMode(self, mode: str | DataInPlotMode):
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1054,6 +1074,9 @@ def __init__(self, parent=None, title="Colormap Dialog"):
 
         self._applyColormap()
 
+    def getHistoWidget(self):
```
t20100, line 1077:
```suggestion
    def getHistogramWidget(self):
```

?

src/silx/gui/plot/actions/control.py
```diff
@@ -330,14 +331,26 @@ def __init__(self, plot, parent=None):
         self.plot.sigActiveImageChanged.connect(self._updateColormap)
         self.plot.sigActiveScatterChanged.connect(self._updateColormap)
 
-    def setColorDialog(self, colorDialog):
+    def setColormapDialog(self, colormapDialog):
```
t20100, line 334:
```suggestion
    def setColormapDialog(self, dialog):
```

?

src/silx/gui/plot/actions/control.py
```diff
@@ -330,14 +331,26 @@ def __init__(self, plot, parent=None):
...
         """Set a specific color dialog instead of using the default dialog."""
```
t20100, line 335:
```suggestion
        """Set a specific colormap dialog instead of using the default one."""
```

src/silx/gui/plot/actions/control.py
```diff
@@ -330,14 +331,26 @@ def __init__(self, plot, parent=None):
...
-        assert(colorDialog is not None)
-        assert(self._dialog is None)
-        self._dialog = colorDialog
-        self._dialog.visibleChanged.connect(self._dialogVisibleChanged)
+        assert(colormapDialog is not None)
```
t20100, line 336:
```suggestion
        assert colormapDialog is not None
```

src/silx/gui/plot/actions/control.py
```diff
@@ -330,14 +331,26 @@ def __init__(self, plot, parent=None):
...
         self.setChecked(self._dialog.isVisible())
 
+    @deprecated(replacement="setColormapDialog", since_version="2.0")
+    def setColorDialog(self, colorDialog):
+        self.setColormapDialog(colormapDialog=colorDialog)
+
+    def getColormapDialog(self):
+        if self._dialog is None:
+            self._dialog = self._createDialog(self.plot)
+            self._dialog.visibleChanged.connect(self._dialogVisibleChanged)
```
t20100, line 351: It sounds an unexpected place to connect the signal, it would be best in `_createDialog`, but I guess there is a reason for not doing it there.

t20100 (commented):
It is worth checking where `setColorDialog` is used inside `silx` and update to `setColormapDialog` to avoid deprecation warnings.
It would be nice to see where `_createDialog` is used and if it can't be replaced by usage `setColormapDialog`... but I guess not (or it would not be there).

t20100:
Regarding public API changes, we should make explicit what we actually do:

No public API (= stuff without a `_` prefix) change without deprecation warning.*

*: Unless impossible to deprecate, but it should be the exception.

t20100:
BTW, +1 for the method renaming to use a consistent name

payno:
> It is worth checking where `setColorDialog` is used inside `silx` and update to `setColormapDialog` to avoid deprecation warnings. It would be nice to see where `_createDialog` is used and if it can't be replaced by usage `setColormapDialog`... but I guess not (or it would not be there).

For the first one clearly. For `_createDialog` I don't see it used internally (except for the `ColormapAction` ). I guess it is used outside by other projects (flint ?)

src/silx/gui/plot/actions/control.py
payno: I agree. But it cannot be done on the `_createDialog` which is static. And we want to build the dialog during runtime. So for now sounds like the safer solution but I agree this is weird.

t20100: Yes, let's keep it as it is

t20100:
I propose to keep a possible change of `_createDialog` for another PR.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -749,12 +755,26 @@ def _plotGammaMarkerConstraint(self, x, y):
...
         return x, y
 
+    def setDataInPlotMode(self, mode: str | DataInPlotMode):
+        mode = DataInPlotMode.from_value(mode)
+        if mode is DataInPlotMode.HISTOGRAM:
+            action = self._dataHistogramAction
+        elif mode is DataInPlotMode.RANGE:
+            action = self._dataRangeAction
+        else:
+            raise NotImplementedError
```
t20100, line 765: Missing instanciation + it's not a missing implementation:
```suggestion
            raise ValueError("Mode not supported")
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -749,12 +755,26 @@ def _plotGammaMarkerConstraint(self, x, y):
             x = min(x, vmax)
         return x, y
 
+    def setDataInPlotMode(self, mode: str | DataInPlotMode):
```
t20100, line 758: Maybe there is a better naming for: `setDataInPlotMode` and `DataInPlotMode`?
`PlotMode`? `DisplayMode`? `RangeDisplayMode`?

t20100 (approved):

src/silx/gui/dialog/ColormapDialog.py
payno: +1 for `DisplayMode`

payno: But then I would also rename `DataInPlotMode` by `DisplayMode`

payno: @vallsv do you mind if we change `_DataInPlotMode` to `DisplayMode` ? As those are Enum we won't be able to deprecate them

t20100: `_DataInPlotMode` is private, so no need for deprecation for me

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -66,6 +66,7 @@
 import logging
 
 import numpy
+from future import __annotations__
```
t20100, line 69: This needs to be the very first of executable line of code right after the module docstring.

And it's: `from __future__ import annotations` ;)

payno: nice one ^^

src/silx/gui/plot/actions/control.py
```diff
@@ -330,14 +331,26 @@ def __init__(self, plot, parent=None):
...
-        assert(self._dialog is None)
-        self._dialog = colorDialog
-        self._dialog.visibleChanged.connect(self._dialogVisibleChanged)
+    def setColormapDialog(self, dialog):
+        """Set a specific colormap dialog instead of using the default one."""
+        assert colormapDialog is not None
+        if self._dialog is not None:
+            self._dialog.visibleChanged.disconnect(self._dialogVisibleChanged)
+
+        self._dialog = colormapDialog
```
t20100, line 340:
```suggestion
        assert dialog is not None
        if self._dialog is not None:
            self._dialog.visibleChanged.disconnect(self._dialogVisibleChanged)

        self._dialog = dialog
```

src/silx/gui/plot/actions/control.py
```diff
@@ -330,14 +331,26 @@ def __init__(self, plot, parent=None):
...
+        self._dialog.visibleChanged.connect(self._dialogVisibleChanged, qt.Qt.UniqueConnection)
         self.setChecked(self._dialog.isVisible())
 
+    @deprecated(replacement="setColormapDialog", since_version="2.0")
+    def setColorDialog(self, colorDialog):
+        self.setColormapDialog(colormapDialog=colorDialog)
```
t20100, line 346:
```suggestion
        self.setColormapDialog(colorDialog)
```

t20100 (approved):

PR #3966, opened 2023-11-21
silx.gui: Fixed PyQt6 support by not using deprecated `qt.QPalette.Normal`

### LINKED ISSUE #3962: [GUI] Colormap under PyQt6: QPalette has no attribute Normal
At the very least, the problem affects ColormapDialog.py

t20100:
Replace `qt.QPalette.Normal` by `qt.QPalette.Active` to support pyqt6

closes #3962


t20100:
I guess it's because `QPalette.Normal` is deprecated (see https://doc.qt.io/qt-6/qpalette.html#ColorGroup-enum) but I don't really see the point.

vasole (approved):

Issue #3946, opened 2023-10-03
plot: Slow down with big polyline Shapes


vallsv:
Hi,

I am displaying a lot of polylines `Shape` to mark some locations in an image.

It sounds like it slow down a lot the rendering.

Playing with the flag `setOverlay` does'nt change much.

Previously i was using `Curve` for the rendering, but this does not feature double colors.

Here is a piece of code

```
                    if True:
                        ringItem = silxItems.Shape("polylines")
                        ringItem.setPoints(numpy.array((x, y)).T)
                        ringItem.setName(key)
                        ringItem.setLineStyle("--")
                        ringItem.setColor("#FF0000A0")
                        ringItem.setLineBgColor("#FFFFFFA0")
                    else:
                        # The implementation with shape have some lacks
                        # Maybe we will have to fall down with curves
                        ringItem = FlintRingCurve()
                        ringItem.setRing(ring)
                        ringItem.setData(x, y)
                        ringItem.setName(key)
                        ringItem.setColor("#FFFFFFA0")
                        ringItem.setLineStyle("--")
```

The use case is really to put some location on top of the image. It is not supposed to be updated often (up until now), so it is really a problem of rendering.

![Screenshot from 2023-10-02 18-43-01](https://github.com/silx-kit/silx/assets/7579321/c2986701-48d8-4245-b0c7-42b254105f49)

PR #3973, opened 2023-11-27
silx.gui.plot.PlotWidget: Added support of dashed line gap color

t20100:
This PR adds 2 methods: `get|setLineGapColor` to `Curve` and `Histogram` items to configure the gap color of dashed lines.
This is implemented for the OpenGL backend and for the matplotlib backend provided that matplotlib is at least v3.6.0.

related to #3946

### DIFF
--- a/src/silx/gui/plot/items/core.py
+++ b/src/silx/gui/plot/items/core.py
@@ -91,6 +91,9 @@ class ItemChangedType(enum.Enum):
     LINE_BG_COLOR = 'lineBgColorChanged'
     """Item's line background color changed flag."""
 
+    LINE_GAP_COLOR = 'lineGapColorChanged'
+    """Item's dashed line gap color changed flag."""
+
     YAXIS = 'yAxisChanged'
     """Item's Y axis binding changed flag."""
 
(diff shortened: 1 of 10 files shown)


src/silx/gui/plot/items/core.py
```diff
@@ -935,6 +938,39 @@ def setColor(self, color, copy=True):
...
+class LineGapColorMixIn(ItemMixInBase):
+    """Mix-in class for dashed line gap color"""
+
+    _DEFAULT_LINE_GAP_COLOR = None
+    """Default dashed line gap color of the item"""
+
+    def __init__(self):
+        self.__lineGapColor = self._DEFAULT_LINE_GAP_COLOR
+
+    def getLineGapColor(self):
```
t20100, line 950: I hesitated to reuse shapes method naming: `get|setLineBgColor` but it's implemented slightly differently (it's not a background color here) and I choose to stick to matplotlib naming.

payno:
I think add of 'gapcolor' parameter to `silx.gui.plot.PlotWidget.addCurve` function is missing  (except if you want to limit this function signature) else fine for me

payno:
I think add of 'gapcolor' parameter to `silx.gui.plot.PlotWidget.addCurve` function is missing  (except if you want to limit this function signature)

t20100:
I didn't wanted to add more arguments to `addCurve` (there's already quite a few). I expect this gap color will not be widely used and so I preferred not to expose it there.

vallsv (approved):

PR #3974, opened 2023-11-27
silx.gui.colors: Added typing

t20100:
This PR adds typing to `silx.gui.colors` and defines input and output types for `colors.rgba`.
Since this function is used everywhere for colors, it should help to type colors in the plot API.

### DIFF
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -249,6 +249,33 @@ def _getColormap(name: str) -> numpy.ndarray:
     return _colormap.get_colormap_lut(name)
 
 
+class _Colormappable:
+    """Class for objects that can be colormapped by a :class:`Colormap`
+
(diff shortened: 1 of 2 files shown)


src/silx/gui/colors.py
```diff
@@ -238,21 +249,46 @@ def _getColormap(name):
     return _colormap.get_colormap_lut(name)
 
 
+class _Colormappable:
```
t20100, line 252: Base class for objects that interacts with `Colormap` i.e, `ColormapMixIn` plot items: This avoids lazy imports and ease typing.

payno:
Looks good. Can we merge it ?

t20100:
Yes, if you approuve it ;)

payno:
I am all for typing :)

PR #3979, opened 2023-11-28
Lazy-loading of multiprocessing module

vasole:
Prevent issues with the multiprocessing module with frozen applications


woutdenolf (approved):
LGTM

PR #3981, opened 2023-11-28
silx.gui: Added support for `QT_API` environment variable

t20100:
This PR adds support for selecting the Qt binding with `QT_API` env. var. (case-insensitive) the same way as [matplotlib](https://matplotlib.org/stable/api/backend_qt_api.html#qt-bindings) and close to [qtpy](https://github.com/spyder-ide/qtpy#requirements) - which is case sensitive.

### DIFF
--- a/src/silx/gui/qt/__init__.py
+++ b/src/silx/gui/qt/__init__.py
@@ -27,8 +27,10 @@
 - `PySide6 <https://pypi.org/project/PySide6/>`_
 - `PyQt6 <https://pypi.org/project/PyQt6/>`_
 
-If a Qt binding is already loaded, it will use it, otherwise the different
-Qt bindings are tried in this order: PyQt5, PySide6, PyQt6.
+If a Qt binding is already loaded, it will be used.
+If the `QT_API` environment variable is set to one of the supported Qt bindings
+(case insensitive), this binding is loaded if available, otherwise the
+different Qt bindings are tried in this order: PyQt5, PySide6, PyQt6.
 
 The name of the loaded Qt binding is stored in the BINDING variable.
 
--- a/src/silx/gui/qt/_qt.py
+++ b/src/silx/gui/qt/_qt.py
@@ -28,7 +28,9 @@ __license__ = "MIT"
 __date__ = "12/01/2022"
 
 
+import importlib
 import logging
+import os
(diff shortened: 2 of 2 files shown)


vasole (commented):
If silx choses the binding, QtPy and matplotlib should be informed to prevent warnings.

t20100:
There is no need to set/change the environment variable, both qtpy and matplotlib first check for already imported bindings, so I would prefer not to change the env. var. If it is fine for you, I would revert your change.

- [qtpy](https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L12):
```
If one of the APIs has already been imported, then it will be used.

Otherwise, the shim will automatically select the first available API (PyQt5, PySide2,
PyQt6 and PySide6); in that case, you can force the use of one
specific bindings (e.g. if your application is using one specific bindings and
you need to use library that use QtPy) by setting up the ``QT_API`` environment
variable.
```

- [matplotlib](https://matplotlib.org/stable/api/backend_qt_api.html#qt-bindings):
```
At runtime, these backends select the actual binding used as follows:

    If a binding's QtCore subpackage is already imported, that binding is selected (the order for the check is PyQt6, PySide6, PyQt5, PySide2).

    If the [QT_API](https://matplotlib.org/stable/users/installing/environment_variables_faq.html#envvar-QT_API) environment variable is set to one of "PyQt6", "PySide6", "PyQt5", "PySide2" (case-insensitive), that binding is selected. (See also the documentation on [Environment variables](https://matplotlib.org/stable/users/installing/environment_variables_faq.html#environment-variables).)

    Otherwise, the first available backend in the order PyQt6, PySide6, PyQt5, PySide2 is selected.
```

vasole:
> There is no need to set/change the environment variable, both qtpy and matplotlib first check for already imported bindings, so I > would prefer not to change the env. var. If it is fine for you, I would revert your change.

There is no **need**, but I was so annoyed of seeing the warning _"selected binding pyqt5 but pyside6 already imported"_ that I did not realize the warning seems to have been removed in recent versions.

You can revert my commit if you wish.

mretegan:
Thank you @t20100! I tested it and it works as expected.

mretegan (approved):

PR #3984, opened 2023-11-29
Do not call multiprocessing module in frozen binaries

vasole:
Calls to multiprocessing fail when using PyInstaller frozen applications


woutdenolf:
Ah ok, multiprocessing will not work at all, so yes your solution is better.

PR #3986, opened 2023-12-05
silx.gui.plot.PlotWidget: Refactored item management

### LINKED ISSUE #3896: Use addItem + custom item are not listed with getItems
For Flint i deal with a lot of custom item classes.

Sounds like if an item is created without a legend, added to the plot, the item displayed but not listed by `getItems`

As result
- Every items are displayed
- But the `getDataRange` is wrong

In fact only the last or the first one is listed because i guess the plot drops the dup empty keys `""`.

To be checked.

```
plot = PlotWidget()

item = items.Curve()
item.setData([0,10], [0,10])
plot.addItem(item)

item = items.Curve()
item.setData([0,-10], [0,-10])
plot.addItem(item)

plot.resetZoom()

print(list(plot.getItems()))
print(plot.getDataRange())
```

t20100:
This PR is changing the way items are stored inside a `PlotWidget`. This aims at making it simpler and solves issues like #3896. On the down side, it might change the behaviour of item management when not using the `PlotWidget.add*` methods...

Still needs testing

closes #3896

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
 
           - name-suffix: "PyQt6 wheel"
             os: ubuntu-latest
-            python-version: '3.7'
+            python-version: '3.11'
             BUILD_OPTION: --wheel
             QT_BINDING: PyQt6
             RUN_TESTS_OPTIONS: --qt-binding=PyQt6 --no-opengl --low-mem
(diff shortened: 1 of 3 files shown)


vallsv (approved):

vallsv:
As you said, maybe an extra test could be added? If it makes sens.

t20100:
I added a test from the sample code in #3896

PR #3988, opened 2023-12-08
silx.gui.plot.PlotWidget: Refactored management of active items

t20100:
~~Merge PR #3987 first!~~

This PR follows PR #3986 and enable passing items rather than only their legend to `PlotWidget.setActiveCurve|Image|Scatter`.

### DIFF
--- a/doc/source/modules/gui/plot/plotsignal.rst
+++ b/doc/source/modules/gui/plot/plotsignal.rst
@@ -91,6 +91,7 @@ Both share the following keys:
 
 - 'event': 'curveClicked' or 'imageClicked'
 - 'button': the mouse button that was pressed in 'left', 'middle', 'right'
+- 'item': The plot item object that was clicked
 - 'label': The legend associated with the clicked image or curve
 - 'type': The type of item in 'curve', 'image'
 - 'x' and 'y': The clicked position in data coordinates
--- a/src/silx/gui/plot/LegendSelector.py
+++ b/src/silx/gui/plot/LegendSelector.py
@@ -1,6 +1,6 @@
 # /*##########################################################################
 #
-# Copyright (c) 2004-2022 European Synchrotron Radiation Facility
+# Copyright (c) 2004-2023 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -39,6 +39,7 @@ import numpy
 from .. import qt, colors
 from ..widgets.LegendIconWidget import LegendIconWidget
 from . import items
+from ...utils.deprecation import deprecated
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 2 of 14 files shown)


src/silx/gui/plot/PlotWidget.py
```diff
@@ -3034,9 +3019,9 @@ def graphCallback(self, ddict=None):
             ddict = {}
         _logger.debug("Received dict keys = %s", str(ddict.keys()))
         _logger.debug(str(ddict))
-        if ddict['event'] in ["legendClicked", "curveClicked"]:
+        if ddict['event'] == "curveClicked":
```
t20100, line 3022: `legendClicked` is not used

src/silx/gui/plot/LegendSelector.py
```diff
@@ -148,6 +149,7 @@ class LegendModel(qt.QAbstractListModel):
     showLineRole = qt.Qt.UserRole + 3
     iconSymbolRole = qt.Qt.UserRole + 4
     showSymbolRole = qt.Qt.UserRole + 5
+    itemRole = qt.Qt.UserRole + 6
```
t20100, line 152: Add a role to pass the curve item in the model and to operate directly on the item.
The goal is to make the legend widget works with curves that might have the same legend.

There is plenty of room to refactor the legend widget since the item provides all the other required information, but it would require to break the API and rewrite a lot of this file, so not for now.

src/silx/gui/plot/LegendSelector.py
```diff
@@ -936,13 +957,17 @@ def renameCurve(self, oldLegend, newLegend):
     def _legendSignalHandler(self, ddict):
         """Handles events from the LegendListView signal"""
         _logger.debug("Legend signal ddict = %s", str(ddict))
+        # If item is not provided, retrieve it from its legend
+        curve = ddict.get("item", None)
+        if curve is None:
+            curve = self.plot.getCurve(ddict["legend"])
```
t20100, line 963: This maintains backward compatibility

src/silx/gui/plot/LegendSelector.py
```diff
@@ -951,43 +976,27 @@ def _legendSignalHandler(self, ddict):
             ret = dialog.exec()
             if ret:
                 newLegend = dialog.getText()
-                self.renameCurve(oldLegend, newLegend)
+                wasActive = self.plot.getActiveCurve() is curve
+                self.plot.removeItem(curve)
+                curve.setName(newLegend)
+                self.plot.addItem(curve)
```
t20100, line 982: It is still not possible to change the name while the item is in the plot.
This restriction can probably be lifted now, but let's keep it a bit longer.

doc/source/modules/gui/plot/plotsignal.rst
```diff
@@ -91,6 +91,7 @@ Both share the following keys:
 
 - 'event': 'curveClicked' or 'imageClicked'
 - 'button': the mouse button that was pressed in 'left', 'middle', 'right'
+- 'item': The plot item object that was clicked
```
t20100, line 94: Pass item in plot curev|image clicked event so event handling does not have to rely on legends.

t20100:
It still need some tests.

t20100:
Added some tests, ready for review.

vallsv:
Let's do it

PR #3989, opened 2023-12-08
silx.gui: Fixed `QMouseEvent.globalPos()` not available with PyQt6

t20100:
[QMouseEvent.globalPos](https://doc.qt.io/qt-6/qmouseevent-obsolete.html#globalPos) is deprecated in Qt6 and removed from PyQt6.


vasole (approved):
LGTM. It is strange that globalPos is kept in QContextMenuEvent

t20100:
Yes... and QMouseEvent.pos() is deprecated in Qt6 but still available in pyqt6 🤷

PR #3996, opened 2023-12-13
`silx.gui.plot.PlotWidget`: Changed add* methods return value to return the item instead of its legend

### LINKED ISSUE #3990: Change `PlotWidget.add*` methods to return the corresponding `Item` instance
As it stands, `PlotWidget` add methods (namely `addCurve`, `addImage`, `addHistogram`, `addScatter`, `addShape`, `addMarker`, `addXMarker`, `addYMarker`) returns the legend of the add/updated item as a `str`.

This issue proposes to change those `add*` methods to return the added/updated item instance instead.

Pros:
- Since the `legend` is passed as input argument, returning it is only useful when using the default legend. So I would expect its usage to be limited. As a matter of fact, in the `silx` project, when the returned legend is used, it is only used to retrieve the corresponding `Item` object...
- The `legend` can be retrieved from the item through: `item.getName()`.
- The `legend` is not enough to retrieve the item, its `kind` (i.e., `curve`, `scatter`...) is also needed.

Cons:
- This would break the current API.

How would this affect your project?

t20100:
Merge PR #3988 first!

This PR introduces a breaking API change: `PlotWidget.add*` methods (namely `addCurve`, `addImage`, `addHistogram`, `addScatter`, `addShape`, `addMarker`, `addXMarker`, `addYMarker`) now returns the created/updated PlotWidget item instead of its legend.

To mitigate this API break, `PlotWidget.remove*` methods and `PlotWidget.get*` methods were update to accept an item instance as their `legend` attribute. Together with `setActive*` methods accepting item, this should make the API change seamless for code that was using the legend as an identifier of the item in the `PlotWidget` without using it otherwise.

`PlotWidget.get*` methods log a warning when passed an item instance since this call is no longer needed.

Changes to `examples/` and tests were minimal and actually not needed to make it run.

closes #3990

### DIFF
--- a/examples/plotWidget.py
+++ b/examples/plotWidget.py
@@ -160,8 +160,7 @@ class MyPlotWindow(qt.QMainWindow):
         x = numpy.random.rand(nbPoints)
         y = numpy.random.rand(nbPoints)
         value = numpy.random.rand(nbPoints)
-        legend = plot.addScatter(x=x, y=y, value=value)
-        scatter = plot.getScatter(legend)
+        scatter = plot.addScatter(x=x, y=y, value=value)
         scatter.setVisualization("solid")
         plot.resetZoom()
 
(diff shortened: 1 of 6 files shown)


t20100:
Added some tests, ready for review.

vallsv (approved):

PR #4005, opened 2023-12-14
Improve ruler

vallsv:
This PR improve the ruler from the plot with a default fixed dashed style

image

No need changelog, it's a new feature

- Move PlotToolButton as an independent module
- Move RulerToolButton as an independent module
- Move inner class as a private class
- Update the style

### DIFF
--- a/src/silx/gui/plot/PlotToolButtons.py
+++ b/src/silx/gui/plot/PlotToolButtons.py
@@ -41,8 +41,6 @@ __date__ = "27/06/2017"
 
 import functools
 import logging
-import weakref
-import numpy
 
 from .. import icons
 from .. import qt
@@ -50,8 +48,6 @@ from ... import config
 from .tools.PlotToolButton import PlotToolButton
 
 from .items import SymbolMixIn, Scatter
-from silx.gui.plot.tools.roi import RegionOfInterestManager
-from silx.gui.plot.items.roi import LineROI
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


payno:
short answer: fine with me :+1: . Try to get something as much as possible 'readable' by default for colormap like 'grayscale' and that's it.

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -624,13 +626,33 @@ class RulerToolButton(PlotToolButton):
...
-        def registerFormatFunction(self, fct):
-            """fct is expected to be a function taking (startPoint, endPoint) as parameter"""
+            self._formatFunction: typing.Optional[
+                typing.Callable[
+                    [numpy.ndarray, numpy.ndarray], str
+                ]
+            ] = None
+            self.setColor("#001122")  # Only there to trig updateStyle
+
+        def registerFormatFunction(
```
t20100, line 636: Would better be named `set` since it is a setter:
```suggestion
        def setFormatFunction(
```
but fine since this can be considered as private API

t20100:
Regarding dashed line support, as a general rule, I prefer to stick to matplotlib naming/default style whenever possible and to have the OpenGL backend behave as close as possible to it.

So I'd prefer to add support for (offset, (dashes)) as in matplotlib `set_linestyle` than defining our own names.
I'll check how complex this is (it shouldn't be) and sync OpenGL dashes with default from matplotlib.
Anyway, this is not for this PR.

t20100:
BTW, did you try with dashed dot: `-.`?

vallsv:
> BTW, did you try with dashed dot: `-.`?

I just try, that's not very nice

vallsv:
> Regarding dashed line support, as a general rule, I prefer to stick to matplotlib naming/default style whenever possible and to have the OpenGL backend behave as close as possible to it.
> 
> So I'd prefer to add support for (offset, (dashes)) as in matplotlib `set_linestyle` than defining our own names. I'll check how complex this is (it shouldn't be) and sync OpenGL dashes with default from matplotlib. Anyway, this is not for this PR.

Only supporting the subset `(offset, (dash1, dash2)))` would be already very nice, and i will try to use it for sure.

t20100:
ok, then I'll add this

vallsv:
Everything was rebased using the new features from the main branch.

Still some commit could be cherry picked alone to the main branch, but it's mostly typing.

This can be reviewed.

payno:
Fine with the source code, works well on my side. I don't have any strong opinion regarding splitting out the `PlotToolButton`.

PR #4006, opened 2023-12-15
FloatEdit: Added widgetResizable

vallsv:
Added a `widgetResizable`/`setWidgetResizable` to `FloatEdit` in order to always display the full value

- With this flag the widget width will grow/shrink during the user typing
- `sizeHint` is reimplemented as result the default layout also fit the content
- The implementation is based on `minimumSize` to inforce the parent layout to resize them self (windows and dialog will also grow and shrink)
- The property name `widgetResizable` was picked from `QScrollBar`

Changelog: 
- Added `widgetResizable` to `FloatEdit` (default is false)

### DIFF
--- a/src/silx/gui/widgets/test/test_floatedit.py
+++ b/src/silx/gui/widgets/test/test_floatedit.py
@@ -1,0 +1,88 @@
+# /*##########################################################################
+#
+# Copyright (c) 2018 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ###########################################################################*/
+"""Tests for FloatEdit"""
+
+__license__ = "MIT"
+
+import pytest
+import weakref
+from silx.gui import qt
+from silx.gui.widgets.FloatEdit import FloatEdit
+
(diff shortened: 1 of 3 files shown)


t20100:
Ci fails with:
```
================================== FAILURES ===================================
______________________________ test_widgetresize ______________________________
holder = <PyQt6.QtWidgets.QWidget object at 0x00000228B7ADF250>
floatEdit = <silx.gui.widgets.FloatEdit.FloatEdit object at 0x00000228B7ADF400>
    def test_widgetresize(qapp_utils, holder, floatEdit):
        holder.layout().addWidget(floatEdit)
        holder.resize(100, 100)
        holder.show()
        qapp_utils.qWaitForWindowExposed(holder)
        floatEdit.setWidgetResizable(True)
        floatEdit.setValue(123)
        a = floatEdit.width()
        floatEdit.setValue(123456789123456789.123456789123456789)
        b = floatEdit.width()
>       assert b > a
E       assert 98 > 98
venv_test\Lib\site-packages\silx\gui\widgets\test\test_floatedit.py:88: AssertionError
```

src/silx/gui/widgets/test/test_floatedit.py
```diff
@@ -0,0 +1,88 @@
...
+
+@pytest.fixture
+def floatEdit(qapp, qapp_utils):
+    widget = FloatEdit()
+    widget.setAttribute(qt.Qt.WA_DeleteOnClose)
+    yield widget
+    widget.close()
+    ref = weakref.ref(widget)
+    widget = None
+    qapp_utils.qWaitForDestroy(ref)
```
t20100, line 42: There is a [qWidgetFactory](https://github.com/silx-kit/silx/blob/12dc3152e41eab819873899d9869c29479dab730/src/silx/gui/conftest.py#L14) fixture that can be used for this in tests or in the fixture: `floatEdit = qWidgetFactory(FloatEdit)`, e.g.,:

```suggestion
@pytest.fixture
def floatEdit(qWidgetFactory):
    yield qWidgetFactory(FloatEdit)
```

Note: The last part is missing from `qWidgetFactory`, it can be added:
```
    ref = weakref.ref(widget)
    widget = None
    qapp_utils.qWaitForDestroy(ref)
```

src/silx/gui/widgets/test/test_floatedit.py
```diff
@@ -0,0 +1,88 @@
...
+    holder.layout().addWidget(floatEdit)
+    holder.resize(100, 100)
+    holder.show()
+    qapp_utils.qWaitForWindowExposed(holder)
+    floatEdit.setWidgetResizable(True)
+    floatEdit.setValue(123)
+    a = floatEdit.width()
+    floatEdit.setValue(123456789123456789.123456789123456789)
+    b = floatEdit.width()
+    assert b > a
```
t20100, line 88: This fails with Qt6, maybe a `Qapplication.processEvent` call is needed?

t20100 (approved):

src/silx/gui/widgets/test/test_floatedit.py
vallsv: This have to be checked manually i think.

vallsv: The test factory exposes a proxy, that's not really convenient i think. I will try to use it anyway.

vallsv: Also if you what to check the API, you dont have to "show" it, so you loose some time at this place

vallsv:
I have used such `proxy.__repr__.__self__` to deref the proxies.

I think you should remove the proxy from your factory.

t20100:
> I think you should remove the proxy from your factory.

If it causes some issues, yes

vallsv:
Here is with the fixes

PR #4007, opened 2023-12-15
silx.gui.plot.PlotWidget: Updated plot axes tick labels behavior

### LINKED ISSUE #3997: PlotWidget tick labels are all the same for very small data range
To reproduce:
```python
from silx import sx
sx.plot((0, 1), (0, 1e-9))
```

Related to https://gitlab.esrf.fr/bliss/bliss/-/issues/4059

t20100:
This PR changes the configuration of matplotlib tick label `ScalarFormatter` to enable offset, scientific notation and math text to format tick labels.

The behavior of the formatter has changed over time and here are some examples of the behavior as of today: no offset/scientific notation | scientific notation only around 0 | ofset + scientific notation

Screen Shot 2023-12-15 at 11 34 01

closes #3997

### DIFF
--- a/src/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/src/silx/gui/plot/backends/BackendMatplotlib.py
@@ -42,7 +42,11 @@ _logger = logging.getLogger(__name__)
 from ... import qt
 
 # First of all init matplotlib and set its backend
-from ...utils.matplotlib import FigureCanvasQTAgg, qFontToFontProperties
+from ...utils.matplotlib import (
+    DefaultTickFormatter,
+    FigureCanvasQTAgg,
+    qFontToFontProperties,
+)
 import matplotlib
 from matplotlib.container import Container
 from matplotlib.figure import Figure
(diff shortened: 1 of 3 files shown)


src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -541,21 +541,9 @@ def __init__(self, plot, parent=None):
...
-                self.ax2.get_xaxis().get_major_formatter(),
-            ]
-            for axis in axes:
-                axis.set_useOffset(False)
-                axis.set_scientific(False)
-        except:
-            _logger.warning(
-                "Cannot disabled axes offsets in %s " % matplotlib.__version__
-            )
+        # Configure axes tick label formatter
```
t20100, line 544: There is no need for `try/execpt` at least from matplotlib >= 1.2.0 which is the minimum version advertised in `setup.py`.

vallsv:
If mpl switch automatically for us, it's fine for me.

It's anyway better than having the same labels in the axis.

t20100:
Added: using the same `Formatter` class for OpenGL backend, so the plot tick labels and order of magnitude+offset text behaves the same for both matplotlib and OpenGL backends.

vallsv:
NIce :+1:

t20100:
Updated to fix an issue when switching linear/log scale, the "formatter" was not correctly updated (this issue was there before this PR).

t20100:
Ready for review

PR #4010, opened 2023-12-15
Draft: Improve line rendering

vallsv:
This PR is an intent to improve line rendering by

- Trying to have an extra `(offset, (dash1, dash2))` sypported type instead of hardcoded
- Trying to fix line width between backend
- Trying to have consistent linestyle between backend
- Trying to have consistent linestyle between curve/shape

An example was added to simplify the checks

image

I have tried to to do as much as i can, but i can't finalise that
- [ ] (1) In opengl the period looks hardcoded, maybe there is a missing thing in the gl program
- [ ] (2) In opengl `Curve` linestyle have twice the period of `Shape` linestyle
- [ ] The hardcoded offset +0.33 is probably not at the right location
- [ ] Was not tested with screen with different pixel densities

Changelog:

### DIFF
--- a/examples/compareBackend.py
+++ b/examples/compareBackend.py
@@ -1,0 +1,261 @@
+# /*##########################################################################
+#
+# Copyright (c) 2017-2021 European Synchrotron Radiation Facility
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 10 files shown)


vallsv:
This PR have to be cherry picked in small commits

t20100:
closing this PR since its proposition has been reused and reworked in other PRs.

PR #4015, opened 2023-12-18
silx.gui.plot.PlotWidget: Improved line dash rendering for OpenGL backend

t20100:
This PR reworks the rendering of dashed lines in the OpenGL backend to make it more consistent with the matplotlib backend.
It aims at making dashes behaves the same as the matplotlib backend and the same between `Shape` and `Curve`.
There is a limitation due to OpenGL is that the available line width can be limited by the OpenGL implementation (this PR does not aim at mitigating that).
It also paves the way in the OpenGL backend to provide the line style as `(offset, dash pattern)` by centralizing the conversion of `linestyle` to dash pattern early.

Screen Shot 2023-12-18 at 16 05 17

related to #4010

### DIFF
--- a/src/silx/gui/plot/backends/BackendOpenGL.py
+++ b/src/silx/gui/plot/backends/BackendOpenGL.py
@@ -58,7 +58,7 @@ _logger = logging.getLogger(__name__)
 
 class _ShapeItem(dict):
     def __init__(
-        self, x, y, shape, color, fill, overlay, linestyle, linewidth, gapcolor
+        self, x, y, shape, color, fill, overlay, linewidth, dashpattern, gapcolor
     ):
         super(_ShapeItem, self).__init__()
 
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -847,6 +847,22 @@ def _castArrayTo(v):
         else:
             raise ValueError("Unsupported data type")
 
+    _DASH_PATTERNS = {  # Convert from linestyle to dash pattern
+        "": None,
+        " ": None,
+        "-": (),
```
t20100, line 853: Empty tuple = solid line (no dashes). Not really nice but simple.

vallsv (approved):

t20100:
Rebase to fix conflict

PR #4023, opened 2023-12-20
Antialiasing on opengl markers

vallsv:
Reworks few opengl markers to use distance function.
This allow to implement easily antialiasing and fix some floating point glitches.

### Before

image

### After

image

Changelog: 
- Fixes glitches on OpenGL marker rendering

### DIFF
--- a/src/silx/gui/plot/backends/glutils/GLPlotCurve.py
+++ b/src/silx/gui/plot/backends/glutils/GLPlotCurve.py
@@ -650,14 +650,14 @@ class Points2D(object):
         PLUS: """
         float alphaSymbol(vec2 coord, float size) {
             vec2 d = abs(size * (coord - vec2(0.5, 0.5)));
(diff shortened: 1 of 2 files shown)


t20100:
Looks good, thanks! [`smoothstep`](https://docs.gl/sl4/smoothstep) is not available in (old) GLSL 1.20 (OpenGL2.1).
You might end-up with OpenGL2.1 when running through ssh and I'm not sure what version of GLSL you get then.

smoothstep doc provides a corresponding alternative, let's use it.
I can take care of this.

vallsv:
BTW, `smoothstep` was already used by one of the symbols

vallsv:
> smoothstep doc provides a corresponding alternative, let's use it. I can take care of this.

Yes please @t20100 if you know how to do it

t20100 (approved):

PR #4025, opened 2023-12-20
Try to use the default font from mpl

### LINKED ISSUE #4018: opengl: No minus sign in the axes
```
WARNING:matplotlib.mathtext:Font 'default' does not have a glyph for '\u2212' [U+2212], substituting with a dummy symbol.
```

No idea if it's a specificity of my env or a specificity of the master

One of the solution is to use the font packaged with matplotlib (see #4010)

image

vallsv:
Closes #4018 

The DejaVu font is packaged together with matplotlib. Better to take advantage of that font than expecting the user environment to provides font with every expected characters.

The opengl backend use a qfont and then try to get back a matplotlib font for the axes. This was not properly working.
So i have more or less forced a `sans-serif` font. That anyway the default for us, for readability.

Changelog: 
- Use matplotlib font for OpenGL backend


vallsv:
Notice that the axes labels use the `rasterMathText` which is already a mpl backend, which is maybe way there is still a problem.

vallsv:
Also the code import `rasterMathText` as optional, but this API is now mandatory

vallsv:
The real fix the prepend of the `sans-serif` in the QFont->mpl font

t20100 (approved):
LGTM

In the mid-term we should think of how to make both backends more compatible: It's probably worth upgrading this and handle all those kind of defaults once for both backends.

PR #4026, opened 2023-12-20
Draft: Size for markers

### LINKED ISSUE #4024: Marker items have a setSymbolSize but it is not propagated
If you set `marker.setSymbolSize(20)` it is the same as doing nothing, the marker will stay at size 10.0 (size 11 in the opengl backend)

```
marker.setSymbolSize(20)  # noop
```

vallsv:
Closes #4024

This API was exposed but not propagated the the backend.

Notice that the default marker size was not the same in mpl and opengl.
That's something which have to be checked.

Changelog: 
- Markers now supports `setSymbolSize`


src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -1135,6 +1138,7 @@ def addMarker(
             text,
             color,
             symbol,
+            int(symbolsize + 0),
```
t20100, line 1141: Cast to int is not needed, a `float` is used internally:

```suggestion
            symbolsize,
```

t20100 (approved):

vallsv:
Changed as draft, i dont expect it to be merged.

t20100:
It's not a big PR and it fixes an issue, so it can go in 2.0 for me

t20100:
CI fails, let's keep this for later indeed

t20100:
Closing as rebased and updated in PR  #4181

PR #4033, opened 2023-12-21
cython3 compatibility

### LINKED ISSUE #4019: Check cython compilation warnings
There's some `performance hint` that needs to be addressed, e.g. :

```
performance hint: src/silx/math/_colormap.pyx:348:46: Exception check will always require the GIL to be acquired. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
```

kif:
Remove compilation warnings
close #4019

### DIFF
--- a/src/silx/image/marchingsquares/_mergeimpl.pyx
+++ b/src/silx/image/marchingsquares/_mergeimpl.pyx
@@ -1,5 +1,11 @@
+#cython: embedsignature=True, language_level=3
+## This is for optimisation
+#cython: boundscheck=False, wraparound=False, cdivision=True, initializedcheck=False,
+## This is for developping:
+##cython: profile=True, warn.undeclared=True, warn.unused=True, warn.unused_result=False, warn.unused_arg=True
+
 # /*##########################################################################
-# Copyright (C) 2018-2020 European Synchrotron Radiation Facility
+# Copyright (C) 2018-2023 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


src/silx/image/marchingsquares/_mergeimpl.pyx
```diff
@@ -1,5 +1,11 @@
+#cython: embedsignature=True, language_level=3
+## This is for optimisation
+#cython: boundscheck=False, wraparound=False, cdivision=True, initializedcheck=False,
```
t20100, line 3: Why is this is needed globally?

src/silx/math/_colormap.pyx
```diff
@@ -1,3 +1,8 @@
+#cython: embedsignature=True, language_level=3
+## This is for optimisation
+##cython: boundscheck=False, wraparound=False, cdivision=True, initializedcheck=False,
```
t20100, line 3: That's disabled, right?

src/silx/math/_colormap.pyx
```diff
@@ -101,7 +106,7 @@ ctypedef fused image_types:
 
 # Normalization
 
-ctypedef double (*NormalizationFunction)(double) nogil
+# ctypedef double (*NormalizationFunction)(double) nogil
```
t20100, line 109: This is no longer a valid syntax?

t20100 (approved):
LGTM, thanks!

src/silx/math/_colormap.pyx
kif: It is still valid but it is unused ... so I commented it to ensure nothing breaks.
To be removed once we are sure it makes no regression anywhere.

PR #4011, opened 2023-12-18
opengl: Added padding for readability

vallsv:
Related to #4010 and #4005

### Before

image

### After

image

### Tested with

```
    @_register("text", "Text")
    def _setupText(self, plot):
        plot.clear()
        plot.getDefaultColormap().setName("viridis")

        # Add an image to the plot
        x = numpy.outer(numpy.linspace(-10, 10, 200), numpy.linspace(-10, 5, 150))
        image = numpy.sin(x) / x
        plot.addImage(image)

        label = items.Marker()
        label.setPosition(50, 50)
        label.setText("Foo bar\nmmmmmmmmmmmmmmmmmmmm")
        label.setBackgroundColor("#FFFFFF44")
        label.setPosition
        plot.addItem(label)

        label2 = items.XMarker()
        label2.setPosition(70, 70)
        label2.setText("X marker")
        label2.setColor("red")
        label2.setBackgroundColor("#00000044")
        plot.addItem(label2)

        label3 = items.YMarker()
        label3.setPosition(10, 70)
        label3.setText("Y marker")
        label3.setColor("yellow")
        label3.setBackgroundColor("#000000")
        plot.addItem(label3)
```

Changelog: 
- Added a fixed padding around the OpenGL text background color

PR #4034, opened 2023-12-21
silx.gui.plot.PlotWidget: Improved text background rendering for OpenGL backend

t20100:
Alternative implementation of PR #4011

This PR adds padding around text displayed by markers for the display of borders.
Instead of relying on 2 rendering as in PR #4011, it plays with the texture coordinates.

### DIFF
--- a/src/silx/gui/plot/backends/BackendOpenGL.py
+++ b/src/silx/gui/plot/backends/BackendOpenGL.py
@@ -912,7 +912,7 @@ class BackendOpenGL(BackendBase.BackendBase, glu.OpenGLWidget):
             pattern = ()
         if len(pattern) == 2:
             pattern = pattern * 2
-        return offset, pattern
+        return float(offset), tuple(float(v) for v in pattern)
 
     def addCurve(
         self,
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -907,7 +912,7 @@ def _lineStyleToDashOffsetPattern(
             pattern = ()
         if len(pattern) == 2:
             pattern = pattern * 2
-        return offset, pattern
+        return float(offset), tuple(float(v) for v in pattern)
```
t20100, line 915: Not related to this PR but missing from adding `int` as a accepted type for line style.

src/silx/gui/plot/backends/glutils/GLText.py
```diff
@@ -282,17 +292,31 @@ def render(self, matrix: numpy.ndarray):
...
+        xoffset = paddingOffset / width
+        yoffset = paddingOffset / height
+        texCoords = numpy.array(
+            (
+                (-xoffset, -yoffset),
+                (1.0 + xoffset, -yoffset),
+                (-xoffset, 1.0 + yoffset),
+                (1.0 + xoffset, 1.0 + yoffset),
+            ),
+            dtype=numpy.float32,
```
t20100, line 314: Extends texture coordinates, what's `<0` or `>1` is rendered with `bgColor`.

PR #4035, opened 2023-12-21
build: Updated build dependencies

t20100:
Small PR following PR #4033 to update build dependencies.
Using `oldest-supported-numpy` should fix building `manylinux2010` wheels which is now broken since the build tries to build the latest numpy and fails.


payno:
good to know that it exists :)

PR #4039, opened 2024-01-09
Documentation: Updated landing page

t20100:
This PR reworks the documentation landing page to use `sphinx-panels` tabs and grid.

I used the `sphinx-panels` package which is deprecated rather than `sphinx-design` (its successor) to be compatible with debian 12 which do not provide `sphinx-design`.

Screenshot from 2024-01-09 15-13-39


payno:
Looks very good :+1:

PR #4042, opened 2024-01-11
Documentation: Added a link to download the Windows app

t20100:
This PR reworks the documentation to add a link to download the Windows installer (or zip) in the installation section of the landing page. This was not advertised before.

Note: The link will be broken when building the doc for something else than a release (or if we misspell the .exe/.zip filename in github release)

image


doc/source/conf.py
```diff
@@ -110,6 +110,16 @@
...
 
+# Substitutions defined for all pages
+rst_prolog = f"""
+.. |silx_installer_btn| replace::
+   .. link-button:: https://github.com/silx-kit/silx/releases/download/v{version}/silx-{version}-windows-installer-x86_64.exe
+      :classes: btn-success
+      :text: Download Windows installer
+
+.. |silx_archive| replace:: :download:`silx ZIP archive <https://github.com/silx-kit/silx/releases/download/v{version}/silx-{version}-windows-application.zip>`
+"""
```
t20100, line 121: That's the way sphinx/rst offers to tune documentation content from the config file: it is not possible to do substitutions inside directives/roles (e.g. use `|version|` inside a link)...

payno (approved):

PR #4058, opened 2024-01-30
Documentation: Changed guideline to provide changelog in PR

t20100:
Checklist:

<hr>

This PR proposes to change the guidelines of how to provide information in PRs for the release notes.
The goal is:

- (1) to keep it simple and easy so we are not bored to do it for each PR (e.g., adding a specific file to each PR to document the changes or filling the changelog.rst file for each PR),
- (2) to ease the authoring of release notes. It's quite long to manually sort PRs by subpackages and by kind of changes, while with a formatting convention it can be automated.

I'd take any better approach/convention for naming the kind changes/etc! and if there is something already existing, even better!
What do you think?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,14 @@
 <!--
-You are encouraged to write a PR title that is meaningful by itself.
-It will be used to auto-generate the complete changelog.
+Thank you for your pull request!
 
-To highlight a change in the changelog, please add a line starting with `Changelog: ` in the PR description.
-It will be used to generate the changelog's summary.
-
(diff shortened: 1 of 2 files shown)


payno:
Good idea. But maybe we could even reduce it. I mean have either `topic` or `module` for all PR.
instead of having sometime a topic and sometime a module ?

t20100:
>  I mean have either topic or module for all PR

The idea was to use a topic for PRs that are not related to a silx subpackage like fixing CI, updating the documentation. This can be clarified, otherwise, how would you handle this?

PR #4059, opened 2024-02-02
CI: Added release workflow

t20100:
This PR proposes to add a github actions release workflow based on [cibuildwheel](https://cibuildwheel.readthedocs.io/).
This workflow generates the source tarball, the wheels and the documentation before uploading the tarball and wheels to pypi (this still needs some configuration on both pypi and github side). The documentation can be retrieved from the workflow artifacts.

The release workflow can be triggered either manually or when publishing a release on github (and will be set to only start for tags with a `v` as first letter).  When the workflow is started manually, the upload to pypi is skipped. When started on a release, a confirmation is requested before publishing to PyPi.

This PR also removes `bitshuffle` from `full` requirements: There is no wheel and it does not build on macos for Python3.11 and 3.12. There is an unmerged PR that looks to address this: https://github.com/kiyo-masui/bitshuffle/pull/142

For an example of run logs, see https://github.com/t20100/silx/actions/runs/7756788606

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,165 @@
+name: Build and deploy
+
+on:
+  workflow_dispatch:
+  release:
+    types:
+      - published
+
+jobs:
+  build_sdist:
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


t20100:
I completed the configuration of pypi and "github environment".
Next release will use the automated deployment.

Issue #4071, opened 2024-02-25
silx.gui.hdf5 and HDF5_USE_FILE_LOCKING


kklmn:
Hello,

The latest version of silx has broken my application ([ParSeq](https://github.com/kklmn/ParSeq)) where I join together QFileSystemModel with silx.gui.hdf5.Hdf5TreeModel. All external links are seen as faulty tree nodes. I see this new behaviour with silx 2.0.0 but not with 1.1.2 on the same Python 3.11.6. I cannot confirm the change on Python 3.12.2 because the older silx 1.1.2 is not installable there.

The reason is in line 184 of silx.gui.hdf5.Hdf5Item.py where `parent_obj.get(self.__key)` returns None for external links. One may say this is an h5py problem, but its version is the same (3.10.0) when I compare silx 2.0.0 with silx 1.1.2.

The workaround I have found is to do
```
os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"  # to work with external links
```
before the import of silx.gui.hdf5. If I also import hdf5plugin before importing silx, I have to do hdf5 file unlocking before importing hdf5plugin. And if I directly use h5py, I also have to do this beforehand.

### Question
What is the silx developers’ view on hdf5 file locking and on its interaction with silx.gui.hdf5, hdf5plugin and h5py? Is it "not silx's business"?

Thank you!

t20100:
Since this was working with silx 1.1.2, it sounds to be a regression in silx v2.

Did you change any other package when switching from silx v1 to v2?
To be complete, what is the version of libhdf5 you are using (`h5py.version.hdf5_version`)?

Are the HDF5 files you access being otherwise opened for writing or reading?

> What is the silx developers’ view on hdf5 file locking

We try to support HDF5's file locking disabled. We even provide [some helpers for accessing HDF5 files being written](http://www.silx.org/doc/silx/latest/modules/io/h5py_utils.html). We are relying on this for online data access (though my personal opinion is that it is not a good idea since we keep facing issues).

In silx view, we do the same "workaround" as you do: set HDF5_USE_FILE_LOCKING before anything imports h5py
https://github.com/silx-kit/silx/blob/0abd43eb9d15daa4cb764436856025cdf6eb1080/src/silx/app/view/main.py#L123

It's been working fine even with changes in libhdf5 related to file locking. The issue is that you should not write files without locking because they can get corrupted if another application is opening them.

kklmn:
Hi,

[...]

Issue #3938, opened 2023-09-14
silx.io.utils.open: disable HDF5 locking


woutdenolf:
>I am trying reading an HDF5 file while it is written by BLISS using the `silx.io.utils.open()` generic entry point and it fails with `OSError` because by default HDF5 wants to lock the file, which is already locked by BLISS.
>
>Would it be possible to add `locking=False` here ?
>
>https://github.com/silx-kit/silx/blob/6d1ed0d466aeac22d77f5b605d5e57ef7b277a62/src/silx/io/utils.py#L486
>
>This is very important because I use this entry point in [Larch](https://github.com/xraypy/xraylarch/blob/master/larch/io/specfile_reader.py).

woutdenolf:
This should use `silx.io.h5py.utils.File` instead of `h5py.File`.

maurov:
@woutdenolf thanks for looking into this. I do not understand why you opened a new issue. Anyway, I will close #3936 then. If this could be patched rapidly, it would be of great help.

woutdenolf:
Sorry for the duplicate, I didn't pay attention.

woutdenolf:
You can always patch it locally by replacing `h5py.File` with `silx.io.h5py_utils.File`. I'll make a MR asap.

Issue #3936, opened 2023-09-14
silx.io.utils.open() fails reading BLISS files during scans


maurov:
@woutdenolf,

I am trying reading an HDF5 file while it is written by BLISS using the `silx.io.utils.open()` generic entry point and it fails with `OSError` because by default HDF5 wants to lock the file, which is already locked by BLISS.

Would it be possible to add `locking=False` here ?

https://github.com/silx-kit/silx/blob/6d1ed0d466aeac22d77f5b605d5e57ef7b277a62/src/silx/io/utils.py#L486

This is very important because I use this entry point in [Larch](https://github.com/xraypy/xraylarch/blob/master/larch/io/specfile_reader.py).

maurov:
duplicate of #3938

PR #4074, opened 2024-03-06
silx.io: Changed `open` function behavior: lock HDF5 files when opening them

t20100:
Checklist:

<hr>

This PR reverts PR #3939 which disabled HDF5 file locking when opening a file with `silx.io.open`.
Disabling HDF5 file locking is causing trouble (see #4071) because there is an issue with external links when file locking is disabled (see https://forum.hdfgroup.org/t/external-link-and-file-locking-disabled-issue/12012).
So as of today, disabling file locking when opening a HDf5 file is not safe and IMO shouldn't be the default.

Looking at the [Larch](https://github.com/xraypy/xraylarch/blob/master/larch/io/specfile_reader.py) link in the issues PR #3939 solved(see issues #3938, #3936), it looks it is calling directly `h5py_utils.File` instead of `silx.io.open` for hdf5 file so it should not be affected.

attn @woutdenolf @maurov

### DIFF
--- a/src/silx/gui/hdf5/test/test_hdf5.py
+++ b/src/silx/gui/hdf5/test/test_hdf5.py
@@ -39,7 +39,6 @@ from silx.gui.utils.testutils import TestCaseQt
 from silx.gui import hdf5
 from silx.gui.utils.testutils import SignalListener
 from silx.io import commonh5
-from silx.io import h5py_utils
 from silx.io.url import DataUrl
 import weakref
 
@@ -55,7 +54,7 @@ def useH5File(request, tmpdir_factory):
     tmp = tmpdir_factory.mktemp("test_hdf5")
     request.cls.filename = os.path.join(tmp, "data.h5")
     # create h5 data
-    with h5py_utils.File(request.cls.filename, "w") as f:
+    with h5py.File(request.cls.filename, "w") as f:
         g = f.create_group("arrays")
         g.create_dataset("scalar", data=10)
     yield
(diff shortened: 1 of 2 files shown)


maurov:
@t20100 thanks for making me aware of this change. In Larch, the wanted default behavior is to have file locking disabled by default because we want to be able to read HDF5 files while their are open in BLISS for writing, without blocking or corrupting the file. My understanding is that `silx.io.h5py_utils.File` is the entry point for this behavior. Do you confirm it?

t20100:
> My understanding is that silx.io.h5py_utils.File is the entry point for this behavior. Do you confirm it?

Yes, I confirm. You can also use `h5py.File(... locking=False)` with recent versions of h5py and libhdf5 (with `h5py_utils.File` you don't have to care about the versions).

However beware that we recently found [an issue with external links when disabling file locking](https://forum.hdfgroup.org/t/external-link-and-file-locking-disabled-issue/12012): If you open a hdf5 file and access an external link from it, the "external" file is opened with file locking. This is causing issues which motivates reverting the change in `silx.io.open`.

maurov:
> However beware that we recently found [an issue with external links when disabling file locking](https://forum.hdfgroup.org/t/external-link-and-file-locking-disabled-issue/12012): If you open a hdf5 file and access an external link from it, the "external" file is opened with file locking. This is causing issues which motivates reverting the change in `silx.io.open`.

Thanks for letting me know about this bug. Please, could you tell me how to programmatically recognize that a group in a HDF5 is an external link?

t20100:
You can use [h5py.Group.get](https://docs.h5py.org/en/stable/high/group.html#h5py.Group.get) and [h5py.ExternalLink](https://docs.h5py.org/en/stable/high/group.html#h5py.ExternalLink):

```python
isinstance(h5group.get("/h5/path", getlink=True), h5py.ExternalLink)
# or
h5group.get("/h5/path", getclass=True, getlink=True) is h5py.ExternalLink
```

But beware (again!), this only tells you whether or not the last part of the path is an external link.
For instance, "/ext_link_group/data" will not be reported as an external link even if "/ext_link_group" is an external link to a group... You need to check each part of the path.

PR #4092, opened 2024-03-21
silx.io.h5py_utils.File: avoid emitting several time warning if incoherent hdf5 / h5py locking argument

payno:
avoid emitting several time warning if incoherent hdf5 / h5py locking argument.
Done with a decorator to try to improve readability

### DIFF
--- a/src/silx/io/h5py_utils.py
+++ b/src/silx/io/h5py_utils.py
@@ -64,6 +64,33 @@ HAS_LOCKING_ARGUMENT = HDF5_HAS_LOCKING_ARGUMENT & H5PY_HAS_LOCKING_ARGUMENT
 
 LATEST_LIBVER_IS_V108 = HDF5_HEX_VERSION < calc_hexversion(1, 10, 0)
 
+LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED = False
+
+
+def only_once(func):
+    """
+    simple decorator to make sure the warning is call at most one time.
+    """
+
+    def wrapper(*args, **kwargs):
+        global LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED
+        if not LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED:
+            LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED = True
+            return func(*args, **kwargs)
+
+    return wrapper
+
+
+@only_once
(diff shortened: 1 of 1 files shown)


payno:
I have to admit I am a bit uncertain about this PR.
To be considered:
    * If users have particular version on h5py / hdf5 en get incoherent locking parameter we will get a tone of warning at the moment. But the class will hardly failed. (as expressed in the warning). So it might be not that bad that logs are filled with the error.
    * but some users knowing very well what they are doing might that this cumbersome. But do we have such users ?
    * This will increase complexity of the class (I did a decorator to try to reduce it). Not sure if I succeed.

src/silx/io/h5py_utils.py
```diff
@@ -64,6 +64,33 @@
...
+    def wrapper(*args, **kwargs):
+        global LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED
+        if not LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED:
+            LOCKING_ARG_CHANGED_H5PY_VS_HDF5_WARMED = True
+            return func(*args, **kwargs)
+
+    return wrapper
+
+
+@only_once
```
t20100, line 84: You could (ab)use [functools.lru_cache](https://docs.python.org/3.8/library/functools.html#functools.lru_cache) here and save the global variable

src/silx/io/h5py_utils.py
```diff
@@ -363,12 +390,7 @@ def __init__(
         """
         # File locking behavior has changed in recent versions of libhdf5
         if HDF5_HAS_LOCKING_ARGUMENT != H5PY_HAS_LOCKING_ARGUMENT:
-            _logger.critical(
-                "The version of libhdf5 ({}) used by h5py ({}) is not supported: "
-                "Do not expect file locking to work.".format(
-                    h5py.version.hdf5_version, h5py.version.version
-                )
-            )
+            warm_incoherent_locking_argument()
```
t20100, line 393: I agree this makes it more complicated to have a function to log a specific message

t20100 (commented):
Provided it's a critical log message and the issues we have with HDF5 file locking, I'm not sure making this log quiet is a good idea.

What could be improved is that the message is logged without checking if locking is actually disabled, so moving the log below and checking the value of `locking` would save a few unwanted messages.

payno:
Agree, let close this one for the moment and we might rethink it later.

PR #4093, opened 2024-03-22
silx.opencl.cmmon: Changed `ocl` object for lasy initialization of OpenCL devices

### LINKED ISSUE #4088: silx.opencl: Avoid executing OpenCL kernels at import time
First raised in https://github.com/silx-kit/silx/issues/4077#issuecomment-1994509743 :

> then there is most likely a design issue with the silx.opencl.common.py module and more specifically about the OpenCL class.
> During import plenty of code will be executed. And this is what create a seg fault during import. Even if no-opencl is used when importing silx.opencl or other module linked.

Also, for advanced usage, 1it would be nice to be able to set the OpenCL platforms+devices to use to spare the probing of all platforms

kif:
- make the OpenCL class a borg (i.e. instantiating it again is for free)
- make atomic32/64 cached properties, evaluated at runtime rather than at import time
- make the ocl object creation at runtime, not at import-time

Overall the import time decreased from 2000ms to 200ms.

close #4088

PS: I hate the black formatting.

### DIFF
--- a/src/silx/opencl/__init__.py
+++ b/src/silx/opencl/__init__.py
@@ -39,13 +39,30 @@ __author__ = "Jerome Kieffer"
 __contact__ = "Jerome.Kieffer@ESRF.eu"
 __license__ = "MIT"
 __copyright__ = "European Synchrotron Radiation Facility, Grenoble, France"
-__date__ = "15/03/2017"
+__date__ = "22/03/2024"
 __status__ = "stable"
 
 import logging
 
-
 logger = logging.getLogger(__name__)
 
+from . import common
 
-from .common import *
+
+def __getattr__(name):
+    """lasy submodule import, replace `from .common import *`
+
+    :param name: str
+    :return: instanciated object.
+    """
+    if name in [
+        "ocl",
+        "pyopencl",
+        "mf",
+        "release_cl_buffers",
+        "allocate_cl_buffers",
+        "measure_workgroup_size",
+        "kernel_workgroup_size",
+    ]:
+        return getattr(common, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/src/silx/opencl/common.py
+++ b/src/silx/opencl/common.py
@@ -3,7 +3,7 @@
 #    Project: S I L X project
 #             https://github.com/silx-kit/silx
 #
-#    Copyright (C) 2012-2023 European Synchrotron Radiation Facility, Grenoble, France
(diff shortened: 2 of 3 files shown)


src/silx/opencl/common.py
```diff
@@ -28,28 +28,18 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 #
+from builtins import property
```
t20100, line 31: Why is this needed?

```suggestion
```

src/silx/opencl/__init__.py
```diff
@@ -39,13 +39,30 @@
...
+        "ocl",
+        "pyopencl",
+        "mf",
+        "release_cl_buffers",
+        "allocate_cl_buffers",
+        "measure_workgroup_size",
+        "kernel_workgroup_size",
+    ]:
+        return getattr(common, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
t20100, line 68: Unless I missed something, if the `OpenCL` does the checks lazily, there is no need for laziness here:

`from .common import *` should be fine.

src/silx/opencl/common.py
```diff
@@ -28,28 +28,18 @@
...
 __status__ = "stable"
-__all__ = [
-    "ocl",
-    "pyopencl",
-    "mf",
-    "release_cl_buffers",
-    "allocate_cl_buffers",
-    "measure_workgroup_size",
-    "kernel_workgroup_size",
-]
```
t20100, line 46: Same, if `OpenCL` is lazy, everything else can be kept as before.

t20100 (commented):
It seems to me the  `__getattr__` used in `__init__` is not needed.
That would make it simpler to avoid it.

src/silx/opencl/__init__.py
kif: I checked without it ... and it does not have the same behavior.
it is a replacement of `__all__= [ .... ] ` which is no more possible in common.

src/silx/opencl/common.py
kif: maybe, but declaring it here ... removes the lazyness !

t20100: OK, I see, it due to `if ocl.nb_devices == 0:` in `common.py`'s `def __getattr__(name):`

https://github.com/silx-kit/silx/pull/4093/files#diff-11a0e82f846b8e9f0ee348e05666f2f6c77b8927fa8dc0a42307112bc548f584R676

Issue #4098, opened 2024-04-03
overriding the match ratio in opencl/sift/match.py


ChannyClaus:
hi! it seems that the parameters for SIFT is hardcoded in https://github.com/silx-kit/silx/blob/eba3c0409ca36325ee34a56249ee83cea032333a/src/silx/opencl/sift/param.py - would it be possible to make them configurable via environment variables?

ChannyClaus:
https://github.com/silx-kit/silx/pull/4099#issuecomment-2044552887 🌵

PR #4099, opened 2024-04-03
make match ratio configurable in opencl/sift

ChannyClaus:
resolves https://github.com/silx-kit/silx/issues/4098

allows the users to configure match ratio via the environment variable `SILX_SIFT_MATCH_RATIO`. should be a pretty low-risk + perfectly backwards compatible change.

### DIFF
--- a/src/silx/opencl/sift/param.py
+++ b/src/silx/opencl/sift/param.py
@@ -31,6 +31,8 @@
 Contains the default parameters for the SIFT algorithm
 """
 
+import os
+
 __authors__ = ["Jérôme Kieffer", "Pierre Paleo"]
 __contact__ = "jerome.kieffer@esrf.eu"
 __license__ = "MIT"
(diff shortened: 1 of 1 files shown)


kif:
One should document this new environment variable ... 
beside this, the code LGTM

t20100:
An environment variable doesn't seems to be needed here.

Since `silx.opencl.sift.param.par` is a `dict`, it is possible to change the value programmatically:
```
import silx.opencl.sift.param

silx.opencl.sift.param.par["MatchRatio"] = 0.5
```

ChannyClaus:
> An environment variable doesn't seems to be needed here.
> 
> Since `silx.opencl.sift.param.par` is a `dict`, it is possible to change the value programmatically:
> 
> ```python
> import silx.opencl.sift.param
> 
> silx.opencl.sift.param.par["MatchRatio"] = 0.5
> ```

ahh darn great point. let me close this. thanks!

t20100:
> should we document this?

Yes, sounds like a good idea, Pull Request welcomed.

PR #4108, opened 2024-04-12
Dependencies: Fixed numpy v2 support on Windows; Continous integration: Added tests with numpy v2; Fixed test by using PySide<6.7

t20100:
Checklist:

<hr>

This PR aims at having CI pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
             os: macos-latest
             python-version: "3.9"
             BUILD_OPTION: --wheel
-            QT_BINDING: PySide6
+            QT_BINDING: "PySide6<6.7"
             RUN_TESTS_OPTIONS: --qt-binding=PySide6 --no-opencl --low-mem
 
           - name-suffix: "PyQt6 wheel"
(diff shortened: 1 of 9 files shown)


t20100:
This PR also fixes the cython code so that tests pass on Windows with numpy 2, since there is a breaking change regarding [Windows default integer](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#windows-default-integer)

t20100:
This PR also modifies: 
- The `strip` fit filter to use `int` instead of `long` (I expect int64 is not needed for indices here...): This avoids the difference between Windows and others regarding the definition of `long` 
- The median filter to use sized types to avoid. Again an issue with `long` on Windows.

t20100:
This is ready for review.

.github/workflows/ci.yml
```diff
@@ -120,15 +120,7 @@ jobs:
...
-          if [ "${{ matrix.QT_BINDING }}" == "PyQt5" ]; then
-            pip install --pre pyqt5;
-          fi
-          if [ "${{ matrix.QT_BINDING }}" == "PySide6" ]; then
-            pip install --pre pyside6;
-          fi
-          if [ "${{ matrix.QT_BINDING }}" == "PyQt6" ]; then
-            pip install --pre pyqt6;
-          fi
+          pip install --pre "${{ matrix.QT_BINDING }}"
```
payno, line 123: This is the kind of commit I like ;)

PR #4113, opened 2024-04-19
Tests: Improved tests by converting warnings to errors

t20100:
Checklist:

<hr>

This PR aim at running tests with stricter rules by raising errors for warnings (e.g. deprecation warnings).
This is done by configuring pytest, adding ignores on specific tests and silenting warnings in  `silx.utils.number`

Thanks @woutdenolf for the idea!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,3 +11,18 @@ build-backend = "setuptools.build_meta"
 required-version = 23
 target-version = ["py37", "py38", "py39", "py310", "py311", "py312"]
 safe = true
+
+[tool.pytest.ini_options]
+minversion = "6.0"
+python_files = [
+    "test/test*.py",
+    "test/Test*.py",
+]
+python_classes = "Test"
+python_functions = "test"
+filterwarnings = [
+    "error",
+    # note the use of single quote below to denote "raw" strings in TOML
+    'ignore:tostring\(\) is deprecated. Use tobytes\(\) instead\.:DeprecationWarning:OpenGL',
+    'ignore:Jupyter is migrating its paths to use standard platformdirs:DeprecationWarning',
+]
--- a/src/silx/gui/plot/test/testPlotWidget.py
+++ b/src/silx/gui/plot/test/testPlotWidget.py
@@ -1530,6 +1530,7 @@ class TestPlotAxes(TestCaseQt, ParametricTestCase):
         with self.assertRaises(Exception):
             item.setBounds((-1000, 1000, 2000, -2000))
 
+    @pytest.mark.filterwarnings("ignore:Attempting to set identical low and high ylims makes transformation singular; automatically expanding.:UserWarning")
     def testBoundingRectWithLog(self):
         item = BoundingRect()
         self.plot.addItem(item)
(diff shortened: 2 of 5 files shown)


src/silx/utils/number.py
```diff
@@ -105,13 +106,17 @@ def min_numerical_convertible_type(string, check_accuracy=True):
...
+        precision = _biggest_float(10) ** nb_precision_digits * 1.2
     previous_type = _biggest_float
     for numpy_type in _float_types:
         if numpy_type == _biggest_float:
             # value was already casted using the bigger type
             continue
-        reduced_value = numpy_type(value)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", "overflow encountered in cast", RuntimeWarning)
+            reduced_value = numpy_type(value)
```
t20100, line 119: Having tests for raise those warnings, I expects that overflows were taken into accounts when writing the code 🤷

src/silx/gui/plot/test/testPlotWidget.py
```diff
@@ -1530,6 +1530,7 @@ def testBoundingRectArguments(self):
         with self.assertRaises(Exception):
             item.setBounds((-1000, 1000, 2000, -2000))
 
+    @pytest.mark.filterwarnings("ignore:Attempting to set identical low and high ylims makes transformation singular; automatically expanding.:UserWarning")
```
t20100, line 1533: Ignore warnings only during tests: Those 3 warnings are about erroneous conditions so thet are not ignored in silx code.

pyproject.toml
```diff
@@ -11,3 +11,18 @@ build-backend = "setuptools.build_meta"
...
+    "test/test*.py",
+    "test/Test*.py",
+]
+python_classes = "Test"
+python_functions = "test"
+filterwarnings = [
+    "error",
+    # note the use of single quote below to denote "raw" strings in TOML
+    'ignore:tostring\(\) is deprecated. Use tobytes\(\) instead\.:DeprecationWarning:OpenGL',
+    'ignore:Jupyter is migrating its paths to use standard platformdirs:DeprecationWarning',
```
t20100, line 27: Listing here general warnings from our dependencies

woutdenolf: For the jupyter path warnings you do not specify the module but I guess there are several? I don't remember but in our CI I do this instead of ignoring the warnings:

```
test-3.11:
  extends: .test-3.11_glx
  variables:
    JUPYTER_PLATFORM_DIRS: 1
```

I'm fine with it either way.

src/silx/utils/number.py
woutdenolf: The docstring says "It has some known issues: precision loss." so I guess.

woutdenolf (approved):
That wasn't too bad. Nice!

pyproject.toml
t20100: AFAIR, it occurs once through an import, during the test collections I think.
But indeed, the env var does the job

t20100: I'll keep it like that for now: it's set at a single place instead of appveyor+github configs, and also works when running locally.

PR #4119, opened 2024-04-26
silx.io.url: Added HSDS URL support

felix-engelmann:
Silx internally supports h5pyd and therefore should be able to connect to a HSDS server with a http url.

However, when running e.g.

    silx view http://hsds-server.tld/home/live  --debug

it crashes with
```
DEBUG:silx.gui.hdf5.Hdf5TreeModel:File 'http:///home/live' can't be read.
  File ".../silx/build/lib.linux-x86_64-3.11/silx/gui/hdf5/Hdf5TreeModel.py", line 733, in insertFile
    h5file = silx_io.open(filename)
             ^^^^^^^^^^^^^^^^^^^^^^
  File ".../silx/build/lib.linux-x86_64-3.11/silx/io/utils.py", line 687, in open
    return h5pyd.File(path, "r", endpoint=endpoint)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.conda/envs/silx/lib/python3.11/site-packages/h5pyd/_hl/files.py", line 212, in __init__
    raise IOError(400, "relative paths are not valid")
OSError: [Errno 400] relative paths are not valid
ERROR:silx.app.view.main:400
DEBUG:silx.app.view.main:Backtrace
  File ".../silx/build/lib.linux-x86_64-3.11/silx/app/view/main.py", line 181, in mainQt
    window.appendFile(url.path())
  File ".../silx/build/lib.linux-x86_64-3.11/silx/app/view/Viewer.py", line 910, in appendFile
    self.__treeview.findHdf5TreeModel().appendFile(filename)
  File ".../silx/build/lib.linux-x86_64-3.11/silx/gui/hdf5/Hdf5TreeModel.py", line 749, in appendFile
    self.insertFile(filename, -1)
  File ".../silx/build/lib.linux-x86_64-3.11/silx/gui/hdf5/Hdf5TreeModel.py", line 733, in insertFile
    h5file = silx_io.open(filename)
             ^^^^^^^^^^^^^^^^^^^^^^
  File ".../silx/build/lib.linux-x86_64-3.11/silx/io/utils.py", line 687, in open
    return h5pyd.File(path, "r", endpoint=endpoint)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.conda/envs/silx/lib/python3.11/site-packages/h5pyd/_hl/files.py", line 212, in __init__
    raise IOError(400, "relative paths are not valid")
OSError: [Errno 400] relative paths are not valid
```

The main cause is, that `silx.io.url.DataUrl` is not able to parse a http url correctly.

This patch adds support for http urls which are then correctly displayed.

A minor fix is to provide h5pyd with an absolute url in `utils.py` instead of stripping the leading /

Unfortunately I did not find a public HSDS server to tests against and the leading / is not tripping any tests, so I assume it is still backwards compatible.

### DIFF
--- a/src/silx/io/utils.py
+++ b/src/silx/io/utils.py
@@ -28,7 +28,7 @@ __date__ = "03/12/2020"
 
 import enum
 import fnmatch
-import os.path
+import os
 import sys
 import time
 import logging
@@ -598,6 +598,32 @@ def _open_local_file(filename):
     raise IOError("File '%s' can't be read as HDF5" % filename)
 
 
+def _open_url_with_h5pyd(url: str):
+    """Open the given url with h5pyd, raises an exception on failures
+
+    :param url: URL of a "file" served by HSDS
+    :returns: h5pyd.File instance
+    """
+    if h5pyd is None:
+        raise IOError(f"URL '{url}' unsupported. Try to install h5pyd.")
+
+    # Retrieve configured HSDS endpoint if any
+    default_endpoint = os.environ.get("H5SERV_ENDPOINT", None)
+    if default_endpoint is None:
+        cfg = h5pyd.Config()
+        if "hs_endpoint" in cfg:
+            default_endpoint = cfg["hs_endpoint"].rstrip("/")
+    else:
+        default_endpoint = default_endpoint.rstrip("/") 
+
+    # Remove endpoint prefix from the URL:
+    # Needed for HSDS servers not exposed as the url netloc, e.g., example.com/hsds/
+    if default_endpoint and url.startswith(default_endpoint):
+        return h5pyd.File(url[len(default_endpoint):], "r")
+
+    return h5pyd.File(url, "r")
+
+
 class _MainNode(Proxy):
     """A main node is a sub node of the HDF5 tree which is responsible of the
     closure of the file.
@@ -673,18 +699,10 @@ def open(filename):  # pylint:disable=redefined-builtin
         if not url.is_valid():
             raise IOError("URL '%s' is not valid" % filename)
         h5_file = _open_local_file(url.file_path())
-    elif url.scheme() in ["fabio"]:
-        raise IOError("URL '%s' containing fabio scheme is not supported" % filename)
+    elif url.scheme() in ("http", "https"):
+        return _open_url_with_h5pyd(filename)
     else:
-        # That's maybe an URL supported by h5pyd
-        uri = urllib.parse.urlparse(filename)
-        if h5pyd is None:
-            raise IOError("URL '%s' unsupported. Try to install h5pyd." % filename)
-        path = uri.path
-        endpoint = "%s://%s" % (uri.scheme, uri.netloc)
-        if not uri.scheme.startswith("http") and path.startswith("/"):
-            path = path[1:]
-        return h5pyd.File(path, "r", endpoint=endpoint)
+        raise IOError(f"Unsupported URL scheme {url.scheme}: {filename}")
 
     if url.data_path() in [None, "/", ""]:  # The full file is requested
         if url.data_slice():


src/silx/io/url.py
```diff
@@ -111,6 +111,10 @@ class DataUrl(object):
     >>> DataUrl("./foo/bar/image.edf")
     >>> DataUrl("C:/data/")
 
+    >>> # HSDS urls are supported with a http schema
```
t20100, line 114:
```suggestion
    >>> # HSDS urls are supported with a http or https schema
```

src/silx/io/url.py
```diff
@@ -211,6 +215,9 @@ def __check_validity(self):
                 self.__data_path is None and self.__data_slice is None
             ) or self.__data_path is not None
             self.__is_valid = slice_implies_data
+        elif self.__scheme.startswith("http"):
```
t20100, line 218: It would be best to explicitly list valid schemes here:

```suggestion
        elif self.__scheme in ("http", "https"):
```

src/silx/io/url.py
```diff
@@ -277,7 +283,10 @@ def __parse_from_path(self, path: str):
             file_path = self.__path[0:pos] + url.path
         else:
             scheme = url.scheme if url.scheme != "" else None
-            file_path = url.path
+            if url.netloc != "":
+                file_path = f"{url.netloc}{url.path}"
+            else:
+                file_path = url.path
```
t20100, line 289: This can be simplified to:
```suggestion
            file_path = f"{url.netloc}{url.path}"
```

src/silx/io/url.py
```diff
@@ -367,7 +376,10 @@ def path(self) -> str:
                 else:
                     path = self.__scheme + ":///" + path
             else:
-                path = self.__scheme + ":" + path
+                if self.__scheme.startswith("http"):
+                    path = self.__scheme + "://" + path
+                else:
+                    path = self.__scheme + ":" + path
```
t20100, line 382: Unless I missed something, this should be fine:

```suggestion
                path = self.__scheme + "://" + path
```

src/silx/io/test/test_url.py
```diff
@@ -298,3 +298,23 @@ def test_scheme_none():
     assert url.file_path() == "my_file.hdf5"
     assert url.scheme() is None
     assert url.data_path() == "/path/to/data"
+
+
+def test_http_scheme():
+    url = DataUrl("http://hsds-server.tld/home/file")
+    print("url is", url)
```
t20100, line 305:
```suggestion
```

src/silx/io/test/test_url.py
```diff
@@ -298,3 +298,23 @@ def test_scheme_none():
...
+    expected = [True, False, "http", "hsds-server.tld/home/file", None, None]
+    assert_url(url, expected)
+    assert url.path() == "http://hsds-server.tld/home/file"
+
+def test_http_scheme_with_port():
+    url = DataUrl("http://hsds-server.tld:8080/home/file")
+    print("url is", url)
```
t20100, line 312:
```suggestion
```

src/silx/io/utils.py
```diff
@@ -682,7 +682,7 @@ def open(filename):  # pylint:disable=redefined-builtin
             raise IOError("URL '%s' unsupported. Try to install h5pyd." % filename)
         path = uri.path
         endpoint = "%s://%s" % (uri.scheme, uri.netloc)
-        if path.startswith("/"):
+        if not uri.scheme.startswith("http") and path.startswith("/"):
```
t20100, line 685: It would be better to check the `http(s)` scheme early, then this leading `/` removal would not be needed.
Also I encountered an issue while testing this PR locally because the HSDS server I have access to is served as `example.com/hsds/`.

I'll make a PR to your branch to fix both.

t20100 (commented):
Thanks for the Pull Request!

The support of `h5pyd` was made several years ago to assess it can work but we actually never used it.
So, thanks for looking into this!

I made a few comments and I will make a PR on your branch to issues in `silx.io.utils`.

t20100:
Here is the PR on your branch: https://github.com/felix-engelmann/silx/pull/1

Issue #4106, opened 2024-04-10
Add `Tiled` data sources to browse Bluesky runs


padraic-shafer:
There has been a recent burst of interest and activity in integrating `bluesky/tiled` data sources into `silx` (and by extension into pyMCA). I'm summarizing some of those discussions here to get feedback from `silx` developers.

## Background

Several light sources are looking into using pyMCA as a browser of data collected during [Bluesky runs](https://blueskyproject.io/event-model/main/user/explanations/data-model.html). From discussions with @linupi @vasole @t20100 it was suggested that modifying `silx` to accept a [Tiled](https://blueskyproject.io/tiled/) data catalog would be an elegant way to do this for pyMCA, silx-view, and any other apps depending on silx.

@t20100 has started a proof-of-concept [branch](https://github.com/silx-kit/silx/compare/main...t20100:silx:tiled) that shows a pathway for adapting a Tiled Container to a HDF5-like interface.

---
## Preliminary scope (to be refined)
_Discussion on 2024-03-26_ @whs92 @danielballan @abbygi @vshekar @padraic-shafer [...missing handles for more BESSY-II participants]

During a chat between several developers at NSLS-II and BESSY-II, we recognized a common interest in using pyMCA as a "bluesky-supported" visual explorer of Tiled datasets for beamline experimenters. We identified several preliminary goals for a development sprint.

1. Connect to a tiled server over HTTP -- Accept a URL; handle Auth
2. Browse contents, with ability to filter and sort
    * Should identify bluesky runs
    * Will likely need a per-endstation configuration of metadata "projections" (flattened subset of important metadata)
3. View baseline data for selected run(s)
4. Plot scan data using existing plot tools
    * Use hinted data by default
    * User can assign "any" channel to a plot axis
5. "Live plot" of data being captured
    * More than one bluesky run may be active at once (nested scans)
    * Initially target a poling loop ~1 second
    * Leave a path open to tiled-stream / websocket
    * Must be able to resume viewing a scan-in-progress if client restarts

---
## Refined goals
_[Discussion on 2024-04-09](https://hackmd.io/hoQ-XJCgT3y6TBtVK4ENUA?view)_ @whs92 @danielballan @abbygi @vshekar @padraic-shafer

1. Use an isolated "Open" dialog or similar entrypoint that can cope with paginated access to large Catalogs, generating a smaller dataset (a tiled client with filters applied) that can be passed down to the rest of the silx/PyMca stack. This can also hand down authentication state.
[...]

PR #4121, opened 2024-05-03
silx.io.open: Added basic support for tiled URLs

t20100:
Checklist:
<hr>

This PR adds minimal support for [tiled](https://blueskyproject.io/tiled/) in `silx.io.open` through a `silx.io.tiledh5` module.
As it stand, it only supports `Container` and `ArrayClient`.

There is a way to limit the number of entries retrieved for each group by setting `silx.io.tiledh5.TiledH5.MAX_CHILDREN_PER_GROUP` (default: all).

`silx.io.open` tries to open http(s) URLs with `h5pyd` and `tiled` when available and raises an `IOError` exception when both fail.
Other exceptions when accessing content (timeout, connection errors) are not catched.

There a shortcut in `silx.io.open` added in first poc when passing an URL with a `tiled-` prefix that I would prefer to remove now that the basic URL can be used. (Note: since first poc in this branch, I changed the prefix from `tiled:` to `tiled-` to avoid issues with URL parsing) attn @vasole

Related to #4106

### DIFF
--- a/src/silx/io/tiledh5.py
+++ b/src/silx/io/tiledh5.py
@@ -44,7 +44,7 @@ def _get_children(
     container: tiled.client.container.Container,
     max_children: int | None = None,
 ):
-    """Return first max_children entries of given container as commonh5 wrappers.
+    """Return first max_children entries of given container as a dict of commonh5 wrappers.
 
     :param parent: The commonh5 wrapper for which to retrieve children.
     :param container: The tiled container from which to retrieve the entries.
@@ -77,7 +77,7 @@ def _get_children(
 class TiledH5(commonh5.File):
     """tiled client wrapper"""
 
-    MAX_CHILDREN: int | None = None
+    MAX_CHILDREN_PER_GROUP: int | None = None
     """Maximum number of children to instantiate for each group.
 
     Set to None for allowing an unbound number of children per group.
@@ -102,7 +102,7 @@ class TiledH5(commonh5.File):
 
     @lru_cache
     def _get_items(self):
-        return _get_children(self, self.__container, self.MAX_CHILDREN)
+        return _get_children(self, self.__container, self.MAX_CHILDREN_PER_GROUP)
 
 
 class TiledGroup(commonh5.Group):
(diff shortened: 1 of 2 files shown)


vasole:
> I changed the prefix from `tiled:` to `tiled-` to avoid issues with URL parsing) attn @vasole

https://github.com/vasole/pymca/pull/1075 should deal with it

danielballan:
Hello! I am just writing to provide a heartbeat...we're very interested in this work. We have recently had our time consumed with some urgent operational tasks, and NSLS-II heads back into operations next week. We look forward to digging into this.

src/silx/io/tiledh5.py
```diff
@@ -0,0 +1,158 @@
...
+    _cache = None
+    """Shared tiled cache with lazy initialization"""
+
+    def __init__(
+        self,
+        name: str,
+        mode: str | None = None,
+        attrs: dict | None = None,
+    ):
+        assert mode in ("r", None)
```
payno, line 96: I am not very comfortable having a mode that could be 'None'.
I saw it comes from legacy and that 'File' class allows it. And then it falls back to 'r' mode. This is probably too late because a new major has been done but maybe remove `None` as a valid mode could be considered (for next major maybe)

src/silx/io/tiledh5.py
```diff
@@ -0,0 +1,158 @@
...
+        super().__init__(name, mode, attrs)
+        if self._cache is None:
+            TiledH5._cache = Cache()  # Use tiled cache default
```
payno, line 99: I would expect to have it defined at the class scope. Like
```python
class TiledH5(commonh5.File):
        _cache = Cache()
```

payno (approved):

src/silx/io/tiledh5.py
t20100: I fully agree but it should be done in a separate PR.

src/silx/io/tiledh5.py
t20100: It was done this way for lazy init

Issue #4130, opened 2024-06-12
Issue with DataViewer when removing views


malte-storm:
When I remove a registered view from the ``silx.gui.data.DataViewer.DataViewer`` using the ``removeView`` method, the display breaks down after switching between views, see the image below for the display of an 3D volume as image stack

A minimum code example is given below (with debugging prints). The reason is that the ``DataViewer.__getStackIndex`` method always calls the ``__stack.addWidget`` method of the QStackedWidget when the view is not listed in the ``__index`` (which is invalidated after the call to ``removeView``). The ``addWidget`` always appends the given widget to the end of the QStackedWidgets list which moves all other widgets "up" in the list without updating the index.

```
from qtpy import QtWidgets

from silx.gui.data.DataViewer import DataViewer
from silx.gui.data import DataViews
import numpy as np


def get_view_indices(viewer):
    return {
        v.__class__.__name__: (
            viewer._DataViewer__stack.indexOf(v.getWidget()),
            viewer._DataViewer__index.get(v, "None"),
        ) for v in viewer.availableViews()
    }


if __name__ == "__main__":
    data = np.random.random((100, 100, 10))
    app = QtWidgets.QApplication([])
    viewer = DataViewer()

    viewer.setData(data)
    # Force creation of views
    for v in viewer.currentAvailableViews():
        viewer.setDisplayedView(v)

    for v in viewer.availableViews():
        if v.__class__.__name__ == "_ImageView":
            viewer.removeView(v)

    viewer.show()

    viewer.setData(data)
    print(get_view_indices(viewer))
    viewer.setDisplayMode(DataViews.STACK_MODE)
    print(get_view_indices(viewer))
    viewer.setDisplayMode(DataViews.PLOT1D_MODE)
    print(get_view_indices(viewer))
    viewer.setDisplayMode(DataViews.STACK_MODE)
    print(get_view_indices(viewer))
    app.exec()
```

![grafik](https://github.com/silx-kit/silx/assets/10544058/a125c637-a6f7-453c-a19f-21c4dc55f3e4)

t20100:
closed by #4131

PR #4131, opened 2024-06-12
silx.gui.data.DataViewer.DataViewer: Fixed issue with accessing views after using the removeView method.

malte-storm:
Fixed issue #4130 which occured after using the ``silx.gui.data.DataViewer.DataViewer.removeView`` method which invalidated the ``index`` dictionary for accessing widgets.
Switching between views modified the referenced ``__stack`` without updating the ``__index``.

### DIFF
--- a/src/silx/gui/data/DataViewer.py
+++ b/src/silx/gui/data/DataViewer.py
@@ -195,7 +195,9 @@ class DataViewer(qt.QFrame):
         """
         if view not in self.__index:
             widget = view.getWidget()
-            index = self.__stack.addWidget(widget)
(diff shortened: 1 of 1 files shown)


t20100:
Thanks for the investigation and the fix!

Looking at it, it seems `self.__index` is not needed and the following would simplify the code:
```python
    def __getStackIndex(self, view):
        widget = view.getWidget()
        index = self.__stack.indexOf(widget)
        if index != -1:
            return index
        return self.__stack.addWidget(widget)
```
But let's keep this for later.

t20100 (approved):

PR #4137, opened 2024-06-14
silx.app.view.CustomPlotSelectionWindow : Added tooltip for name and display file name only

RebeloAlexis:
Name display changed from path to name. 
Added a toolTip to display the name, path and filename.
Added modification of plot label X, which changes according to dataset name

Design sans titre

### DIFF
--- a/src/silx/app/view/CustomPlotSelectionWindow.py
+++ b/src/silx/app/view/CustomPlotSelectionWindow.py
@@ -86,7 +86,11 @@ class _HashDropZones(qt.QStyledItemDelegate):
             painter.setPen(self.__highlightDropPen if isDropTarget else self.__dropPen)
             painter.drawRect(option.rect.adjusted(3, 3, -3, -3))
 
-            painter.drawText(option.rect.adjusted(3, 3, -3, -3), qt.Qt.AlignLeft | qt.Qt.AlignVCenter, "Drop a 1D dataset")
+            painter.drawText(
+                option.rect.adjusted(3, 3, -3, -3),
+                qt.Qt.AlignLeft | qt.Qt.AlignVCenter,
+                "Drop a 1D dataset",
+            )
             painter.restore()
         else:
             qt.QStyledItemDelegate.paint(self, painter, option, index)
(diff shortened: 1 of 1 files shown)


src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -226,22 +243,51 @@ def clearAll(self):
...
+        if len(attrs) > 0:
+            tooltip = _utils.htmlFromDict(attrs, title=title)
+        else:
+            tooltip = ""
+
+        return tooltip
+
+    def _getTextSplit(self, text):
+        textSplit = text.split("/")
+        return textSplit[len(textSplit) - 1]
```
t20100, line 289: `len(textSplit)` is not needed, in python negative indices are starting from the end, so this does the same:

```suggestion
        return text.split("/")[-1]
```

src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -226,22 +243,51 @@ def clearAll(self):
...
+        attrs = self._createTooltipAttributes(url)
+        title = attrs.pop("#Title", None)
+        if len(attrs) > 0:
+            tooltip = _utils.htmlFromDict(attrs, title=title)
+        else:
+            tooltip = ""
+
+        return tooltip
+
+    def _getTextSplit(self, text):
```
t20100, line 287: A more meaningful name would improve understanding the code using it, for example:

```suggestion
    def _getBasename(self, text):
```
Named after Python's [os.path.basename](https://docs.python.org/3/library/os.path.html#os.path.basename)

This would also need to be changed where it is called.

src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -226,22 +243,51 @@ def clearAll(self):
...
+    
+    def _getDefautToolTip(self, url):
+        attrs = self._createTooltipAttributes(url)
+        title = attrs.pop("#Title", None)
+        if len(attrs) > 0:
+            tooltip = _utils.htmlFromDict(attrs, title=title)
+        else:
+            tooltip = ""
+
+        return tooltip
```
t20100, line 285: As opposed to `Hdf5Item.py`, I think it would be simpler to make it in a single method

```suggestion    
    def _getToolTip(self, url):
        if url is None:
            return ""
        attrs = {
            "Name": self._getTextSplit(url.data_path()),
            "Path": url.data_path(),
            "File name": self._getTextSplit(url.file_path()),
        }
        return _utils.htmlFromDict(attrs, title="HDF5 Dataset")
```
(not tested)

and eventually renamed.

t20100 (commented):
Great! I left a few suggestions that we can discuss.

t20100 (approved):
LGTM

PR #4140, opened 2024-06-18
silx.app.view.CustomPlotSelectionWindow : Added the documentation

RebeloAlexis:
Add comments for each function, variable typing and type returns.

The comments can surely be more precise. 
And maybe some comments are missing in the functions.

### DIFF
--- a/src/silx/app/view/CustomPlotSelectionWindow.py
+++ b/src/silx/app/view/CustomPlotSelectionWindow.py
@@ -22,6 +22,7 @@
 # ###########################################################################*/
 
 """Custom plot selection window for selecting 1D datasets to plot."""
+from __future__ import annotations
 
 from silx.gui import qt, plot, icons
 import silx.gui
--- a/src/silx/app/view/CustomPlotSelectionWindow.py
+++ b/src/silx/app/view/CustomPlotSelectionWindow.py
@@ -20,19 +20,22 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 # ###########################################################################*/
-from __future__ import annotations
 
-""""""
+"""Custom plot selection window for selecting 1D datasets to plot."""
 
 from silx.gui import qt, plot, icons
+import silx.gui
+import silx.gui.plot
+import silx.gui.plot.items
 import silx.io
 import numpy
 import functools
 from silx.gui.plot.LegendSelector import LegendIcon
+import silx.io.url
 import silx.io.utils
 from silx.gui.hdf5 import _utils
 
-
+# Custom role for highlighting the drop zones
 _DROP_HIGHLIGHT_ROLE = qt.Qt.UserRole + 1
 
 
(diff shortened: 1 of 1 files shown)


src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -21,24 +21,28 @@
 # THE SOFTWARE.
 # ###########################################################################*/
 
-""""""
+"""Custom plot selection window for selecting 1D datasets to plot."""
```
t20100, line 24: So that most of the typing syntax (e.g. use of `|` with types) is supported with Python v3.8, this is needed:

```suggestion
from __future__ import annotations
"""Custom plot selection window for selecting 1D datasets to plot."""
```

src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -186,14 +205,16 @@ def _createRowItems(self, filename, curve):
 
         return fileItem, iconItem, removeItem
 
-    def fileItemExists(self, filename):
+    def fileItemExists(self, filename: str) -> bool:
+        """Return if a file item with the given filename exists in the model for the Y datasets."""
```
t20100, line 210:
```suggestion
        """Return True if a file item with the given filename exists in the Y datasets of the model."""
```

src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -21,24 +21,28 @@
 # THE SOFTWARE.
 # ###########################################################################*/
 
-""""""
+"""Custom plot selection window for selecting 1D datasets to plot."""
 
 from silx.gui import qt, plot, icons
+import silx.gui
+import silx.gui.plot
```
t20100, line 29: Those imports does not seems to be useful:

```suggestion
```

src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -122,15 +128,18 @@ def __init__(self, plot, parent=None):
...
+        """Return the parent item for the X dataset"""
         return self._xParent
 
-    def getYParent(self):
+    def getYParent(self) -> qt.QStandardItem:
+        """Return the parent item for the Y datasets"""
         return self._yParent
 
-    def _setXFile(self, url=None, curve=None):
+    def _setXFile(self, url: silx.io.url.DataUrl | None = None, curve: None = None):
```
t20100, line 142: The `curve` argument is always `None`, better to remove it (this needs to change its usage in the function body as well):
```suggestion
    def _setXFile(self, url: silx.io.url.DataUrl | None = None):
```

src/silx/app/view/CustomPlotSelectionWindow.py
```diff
@@ -166,14 +181,18 @@ def _addYFile(self, url=None, curve=None, node="X"):
...
+        """Return the row index containing the given text, or None if not found."""
         for row in range(self.rowCount()):
             item = self.item(row, 0)
             if item and item.text() == text:
                 return row
         return None
 
-    def _createRowItems(self, filename, curve):
+    def _createRowItems(
+        self, filename: str, curve: silx.gui.plot.items.Curve
```
t20100, line 194: `curve` might be `None` (e.g. for X)
```suggestion
        self, filename: str, curve: silx.gui.plot.items.Curve | None
```

t20100 (approved):
Thanks for the typing and doc, I left a few suggestions

payno:
Lokos good :muscle: 
Maybe a short description of `CustomPlotSelectionWindow` could be added if we want to be picky.

PR #4143, opened 2024-06-27
silx.utils.enum: Deprecated `Enum.from_value` method

loichuder:


doc/source/modules/utils/retry.rst
```diff
@@ -1,6 +1,6 @@
 .. currentmodule:: silx.utils
 
-:mod:`weakref`
+:mod:`retry`
```
loichuder, line 3: FIxed a copy-paste mistake

src/silx/utils/enum.py
```diff
@@ -30,6 +30,15 @@
...
 
+from silx.utils.deprecation import deprecated_warning
+
+deprecated_warning(
+    "Class",
+    "silx.utils.enum.Enum",
+    since_version="2.1.1",
+    replacement="Python built-in Enum class",
+)
+
```
t20100, line 41: This will raise deprecation warnings everywhere silx uses this module.

To keep the compatibility while deprecating it, it needs extra work, something like:
- move the implementation in `_enum.py` and import from `_enum` in this module and at every place it is used in silx
- raise warnings in `from_value`, `members`, `names` and `values` methods

loichuder: Or could I put in in the `__init__` of the class ?

loichuder:
@t20100 On second thought, I don't think it is worth deprecating the whole class just because `from_value` is obsolete. The other methods (`members`, `names` and `values`) are still fine.

So I only deprecated the `from_value` method and still added `enum` to the silx API documentation.

src/silx/utils/test/test_enum.py
```diff
@@ -47,3 +47,9 @@ class Success(Enum):
     assert Success.from_value("B") == Success.B
     with pytest.raises(ValueError):
         Success.from_value(3)
+
+    # With built-in 'from_value'
+    assert Success(1) == Success.A
+    assert Success('B') == Success.B
+    with pytest.raises(ValueError):
+        Success(3)
```
t20100, line 55: Not worth testing the implementation of Python IMO ;)

t20100 (commented):
Looks good, thanks!

[Usage of `from_value`](https://github.com/search?q=repo%3Asilx-kit%2Fsilx%20from_value(&type=code ) throughout `silx` needs to be updated so `silx` does not generates deprecation warnings by itself.

t20100 (approved):
Thanks!

t20100:
Don't bother about the CI failure, it's a random one.
I'm currently working on fixing quite a few issues related to tests and CI.

PR #4082, opened 2024-03-14
Dependencies: Added support of `numpy` v2

t20100:
Checklist:

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

This PR prepares silx for support of numpy v2 by:
- Using `np.rec` instead of `np.core.recarray`
- Using `cython` fixed-size types instead of `cimport numpy`: This leaves the constraint of using `numpy` headers to build C extensions.
- Fixing incompatibilities spotted by `ruff check --select NPY201`

I've tested that cython extensions built with numpy v1 can be imported  with numpy v2.
Full testing with numpy v2 is still needed.

I also raised the minimum supported version of numpy to the "oldest-supported-numpy" one on Python3.7.

Closes #4080

PR #4149, opened 2024-07-12
Tests and continuous integration: Fixed issues and updated CI configuration

t20100:
Checklist:

<hr>

This PR aims at fixing tests errors/seg fault occuring in CI.
It does not fix all randomly occuring errors. For example, errors due to `os.unlink` failing because a file is accessed by another process under Windows are not fixed.

Without this PR, different ways to run the tests (`run_tests.py` or `import silx.test; silx.test.run_tests()`) had different behaviours and the `pytest` config in `pyproject.toml` was not always used... leading to ignoring some of the tests!!! and not raising errors on warning.
This PR makes `run_tests.py` calls `silx.test.run_tests()` which calls `pytest` with the correct config parameters, so there is no longer different behaviors and missed tests!!!

Instead of configuring `pytest` look-up, tests has been renamed to match `pytest` default's look-up syntax.

This PR also updates and simplify the CI configuration, mainly to also run GUI tests on Windows with github actions (without the tests that require OpenGL for now).


.github/workflows/ci.yml
```diff
@@ -145,5 +148,8 @@ jobs:
               Xorg -noreset +extension GLX +extension RANDR +extension RENDER -logfile ./99.log -config ./ci/xorg.conf :99 &
               sleep 3
           fi
-          echo "RUN_TESTS_OPTIONS="${{ matrix.RUN_TESTS_OPTIONS }}
-          python run_tests.py --installed -v ${{ matrix.RUN_TESTS_OPTIONS }}
+          if [ ${{ runner.os }} == 'Windows' ]; then
+              export WITH_GL_TEST=False
+          fi
```
t20100, line 153: Disable OpenGL tests on Windows for now: some of the tests fails and this needs more investigation...

src/silx/gui/dialog/test/test_datafiledialog.py
```diff
@@ -123,18 +125,27 @@ def qWaitForPendingActions(self, dialog):
...
 
     def assertSamePath(self, path1, path2):
-        path1_ = os.path.normcase(path1)
-        path2_ = os.path.normcase(path2)
-        if path1_ != path2_:
-            # Use the unittest API to log and display error
-            self.assertEqual(path1, path2)
+        self.assertEqual(
+            os.path.normcase(os.path.realpath(path1)),
+            os.path.normcase(os.path.realpath(path2)),
```
t20100, line 130: It was causing issue on Windows

t20100:
PR #4165 is cherry-picked here

t20100:
Ready for review.
There's a few different things in this PR, I can split it into smaller PR if needed.

src/silx/utils/array_like.py
```diff
@@ -49,6 +49,9 @@
 import numpy
 import numbers
 
+from .._utils import NP_OPTIONAL_COPY
```
vallsv, line 52: Do you use it as a flag? In case of troubles, you can switch it off?

Not sure it's a very good idea. Because it is spread everywhere. Isn't it better to put it in this module, only?

vallsv (approved):

.github/workflows/ci.yml
```diff
@@ -1,149 +1,97 @@
...
         # xserver-xorg-video-dummy: For OpenGL
-        # libxkbcommon-x11-0: needed for Qt plugins
-      - name: Install X server
+        # libxkbcommon-x11-0, ..: needed for Qt plugins
+      - name: Install system packages
         if: runner.os == 'Linux'
         run: |
           sudo apt-get update
-          sudo apt-get install libegl1-mesa ocl-icd-opencl-dev intel-opencl-icd libgl1-mesa-glx xserver-xorg-video-dummy libxkbcommon-x11-0 libxkbcommon0 libxkbcommon-dev libxcb-icccm4 libxcb-image0 libxcb-shm0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-render0 libxcb-shape0 libxcb-sync1 libxcb-xfixes0 libxcb-xinerama0 libxcb-xkb1 libxcb-cursor0 libxcb1
+          sudo apt-get install ocl-icd-opencl-dev intel-opencl-icd xvfb libegl1-mesa libgl1-mesa-glx xserver-xorg-video-dummy libxkbcommon-x11-0 libxkbcommon0 libxkbcommon-dev libxcb-icccm4 libxcb-image0 libxcb-shm0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-render0 libxcb-shape0 libxcb-sync1 libxcb-xfixes0 libxcb-xinerama0 libxcb-xkb1 libxcb-cursor0 libxcb1
```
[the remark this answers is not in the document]
t20100: `silx` uses `pytest-xvfb` but it still needs `xvfb` to be available on the host

src/silx/gui/plot/tools/profile/manager.py
```diff
@@ -54,75 +55,41 @@
...
-
-        resultReady = qt.Signal(object, object)
         runnerFinished = qt.Signal(object)
 
-    def __init__(self, threadPool, item, roi):
-        """Constructor"""
+    def __init__(self, item: items.Item, roi: ProfileRoiMixIn):
         super(_RunnableComputeProfile, self).__init__()
-        self._signals = self._Signals()
-        self._signals.moveToThread(threadPool.thread())
```
vallsv, line 75: Maybe better to move it into another PR.
That such a thing which is difficult to diagnose in case of trouble

src/silx/utils/array_like.py
t20100: `NP_OPTIONAL_COPY` was added to support both numpy v1 and v2 (PR #4082).

It was done following [numpy2 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword).

The day `silx` stop supporting numpy v1 it will also help a lot to clean-up the code.

src/silx/gui/plot/tools/profile/manager.py
t20100: I removed this change from this PR and opened PR #4169 for it.

t20100:
CI seg fault with PySide6 since the fix for this issue has been moved to PR #4169

PR #4154, opened 2024-08-05
silx.gui.colors: Add "percentile" mode for autoscaling

clemisch:
It autoscales from 1st to 99th (nan)percentiles of the data. In my experience this usually gives a sensible value window for realistic data.

### DIFF
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -331,10 +331,10 @@ class Colormap(qt.QObject):
     """constant for autoscale using mean +/- 3*std(data)
     with a clamp on min/max of the data"""
 
-    PERCENTILE = "percentile"
+    PERCENTILE_1_99 = "percentile_1_99"
     """constant for autoscale using 1st and 99th percentile of data"""
 
-    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE)
+    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE_1_99)
     """Tuple of managed auto scale algorithms"""
 
     sigChanged = qt.Signal()
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -331,7 +331,10 @@ class Colormap(qt.QObject):
     """constant for autoscale using mean +/- 3*std(data)
     with a clamp on min/max of the data"""
 
-    AUTOSCALE_MODES = (MINMAX, STDDEV3)
+    PERCENTILE = "percentile"
+    """constant for autoscale using 1st and 99th percentile of data"""
+
+    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE)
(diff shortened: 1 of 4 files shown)


src/silx/gui/colors.py
```diff
@@ -331,7 +331,10 @@ class Colormap(qt.QObject):
     """constant for autoscale using mean +/- 3*std(data)
     with a clamp on min/max of the data"""
 
-    AUTOSCALE_MODES = (MINMAX, STDDEV3)
+    PERCENTILE = "percentile"
```
woutdenolf, line 334: In analogy with

```python
STDDEV3 = "stddev3"
```

How about this?

```python
PERCENTILE_1_99 = "percentile_1_99"
```

clemisch: Yeah, done!

woutdenolf (approved):

woutdenolf:
Adding PERCENTILE_1_99 to the STDDEV3 tests would be a bonus.

vallsv:
Hi and thx for the feature.

What do you mean by realistic data? Is that a specific use case or domain?
Could it be possible to have a screenshot with and without?

Also a percentile is like a mean, it's a sort on the values of millions of pixels, right?

The following this processes twice the same sort
```
numpy.nanpercentile(i, 1)
numpy.nanpercentile(i, 99)
```
Could you please check if numpy can do something faster the following way?
```
numpy.nanpercentile(i, [1, 99])
```

clemisch:
> What do you mean by realistic data? Is that a specific use case or domain?

There is no specific usecase or domain. With "realistic" I mean that it works well on noisy data, data which has artefacts outside of the region of interest, or a simulation which has very large values only in a very small ratio of pixels. Min/max scaling would scale to the extreme values and not provide good contrast on the "main" content.

In other words, you get good linear contrast for the central 98% of the data.

The extisting mean+-std method only works for values around one mean value. This doesn't apply to many imaging examples where you have some min/max but said "extreme" values on top. 

---

> Could it be possible to have a screenshot with and without?

Simulation of Fresnel propagation without noise (min/max and percentile scaling):

image

Noisy tomo reco:

image

Admittedly not the most impressive examples. I'm on a conference without good access to real data. In any case, 1-99 percentile scaling is the go-to autowindow of everyone in my x-ray imaging / tomography bubble.  

---

> Could you please check if numpy can do something faster the following way?

Yeah, numpy is indeed able to compute both percentiles simultaneously. This doesn't seem to be a recent addition so I'm not sure why I didn't use it so far. I changed to code to use it.

t20100 (approved):

t20100:
Hi @clemisch,

I've added some basic tests and reworked the code a bit, it's in this PR on your branch: https://github.com/clemisch/silx/pull/1

clemisch:
Oh dang, I forgot to add the tests! Thank you for adding them, and props for merging into my branch :astonished:

PR #4156, opened 2024-08-06
silx.io.fioh5: Improved handling of missing data entries

### LINKED ISSUE #4155: [IO] FIO files: handle missing data
Hi silx team, 

I have encountered FIO files with entries like

```
  -89.99999279994226 -89.74998879491014                1                  <no-data>      <no-data>   
  -89.74998879491014 -89.4999937899502                 2                   1                    120.9696655273438
  -89.4999937899502   -89.24998978491807               3                   1                    120.9657211303711 
```

which cannot be read in the current implementation of fioh5.py.

Suggested fix
-------------

Use ``numpy.genfromtxt`` instead of  ``numpy.loadtxt`` in fioh5.py, https://github.com/silx-kit/silx/blob/415b020023cb1afd4154eb2caa0c9efa17f42c8b/src/silx/io/fioh5.py#L260 
which more gracefully handles missing data and automatically fills missing data with ``numpy.nan``.

I am working on the implementation of the fix and will create a PR once the code is ready. (if this is ok for you)

tifuchs:
Enables the loading of fio files with missing data entries, that can not be converted into the target data type.
Mostly uses the default behaviour of [numpy.genfromtxt](https://numpy.org/doc/stable/reference/generated/numpy.genfromtxt.html).

For example: Entries like ``<no data>`` or ``nan`` in a FLOAT data column are automatically converted to ``numpy.nan`` by ``numpy.genfromtxt``. 

Not all missing values can be handled:
Missing boolean or integer values cannot be flagged as missing (i.e. ``nan``) in a bool or integer array and thus result in a ``ValueError``.

Closes #4155

### DIFF
--- a/src/silx/io/fioh5.py
+++ b/src/silx/io/fioh5.py
@@ -166,6 +166,20 @@ dtypeConverter = {
     "BOOLEAN": "?",
 }
 
+def _bytestobool (val):
+    """Convert a byte of truth to true (1) or false (0).
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.decode().lower()
+    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+        return True
+    elif val in ('n', 'no', 'f', 'false', 'off', '0'):
+        return False
+    else:
+        raise ValueError("invalid truth value %r" % (val,))
+
 
 def is_fiofile(filename):
     """Test if a file is a FIO file, by checking if three consecutive lines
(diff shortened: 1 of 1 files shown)


woutdenolf:
Fix CI failures. Could you also add a unit test for this?

src/silx/io/fioh5.py
```diff
@@ -166,6 +166,20 @@
     "BOOLEAN": "?",
 }
 
+def _bytestobool (val):
+    """Convert a byte of truth to true (1) or false (0).
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
```
t20100, line 174: No need for a complete description in the docstring IMO, the code is self-explanatory


```suggestion
    """Convert bytes of a truth value to bool.

    Raises ValueError if 'val' is not supported.
    """
```

src/silx/io/fioh5.py
```diff
@@ -166,6 +166,20 @@
...
+    val = val.decode().lower()
+    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+        return True
+    elif val in ('n', 'no', 'f', 'false', 'off', '0'):
+        return False
+    else:
+        raise ValueError("invalid truth value %r" % (val,))
```
t20100, line 181:
```suggestion
        raise ValueError("Invalid truth value %r" % val)
```

t20100 (commented):
Thanks for the PR!

As already mentioned, the continuous integration fails highlighting an bytes/str issue and it would also be great to have a test with `nan` and `no data`

src/silx/io/fioh5.py
```diff
@@ -166,6 +166,19 @@
     "BOOLEAN": "?",
 }
 
+def _bytestobool (val):
+    """Convert bytes of a truth value to bool.
+
+    Raises ValueError if 'val' is not supported.
+    """
+    val = val.decode().lower()
```
t20100, line 174:
```suggestion
    if isinstance(val, bytes):
        val = val.decode()
    val.lower()

```

Issue #4135, opened 2024-06-13
Lineplot window doesn't open on Wayland


clemisch:
When I use `silx view <file>`, look at an 2D image and do any line profile, the extra window with the line plot does not appear. No error is reported. 

When I use 

```
QT_QPA_PLATFORM=wayland silx view <file>
```

the extra window does pop up. 

What surprises me is that I'm used to `ipython3 --gui=qt5` working without any issues on Wayland, while silx appears to require the explicit QT flag. 

I'm on Fedora 40, Gnome on Wayland. silx is installed via pipx and running under Python 3.12.

t20100:
Hi,

I don't have access to a machine to try this.

To be sure, can you check which version of Qt and which binding `silx view` is loading? You can see it from the `Help -> 
About` menu.
It would also be good to check which rendering backend is used (OpenGL or matplotlib). You can see it in the `Options -> Plot rendering backend: <backend>` menu.

Also, can you run the following code from `ipython3 --gui=qt5` and do a line profile as you do with `silx view`? This reproduces what you do in `silx view`, and would allow to check if the behavior is different. 

```python
from silx import sx; import numpy as np
w = sx.imshow(np.arange(100*100).reshape(100, 100))
```

To know which platform is used by default, you can run `QT_DEBUG_PLUGINS=1 silx view`.
It's quite verbose, but checking the `loaded library ...` lines allows to find which platform is used by default (`xcb` I guess).

loichuder:
I am also on Gnome on Wayland (Ubuntu 22.04) and I cannot reproduce this, whether it is with Python 3.9 or 3.12.

I think further information is needed then: see the instructions given by @t20100

clemisch:
Thanks for looking into this! Sorry for my delayed response.<br/><br/>

> can you check which version of Qt and which binding `silx view` is loading?

Qt version 5.15.2, Qt binding PyQt5 <br/><br/>

> It would also be good to check which rendering backend is used

At the moment it's matplotlib. Behavior is unchanged for opengl (after restart and confirming it's using opengl).   <br/><br/>

> Also, can you run the following code

The lineplot also doesn't show up here.   <br/><br/>

>  To know which platform is used by default, you can run `QT_DEBUG_PLUGINS=1 silx view`

It looks to me like that's also using qt5? Only in the first line it's `PyQt5/.../libqxcb.so`.

```
$ grep "loaded library" silx_log.txt 
[...]

PR #4170, opened 2024-09-03
`silx.gui`, silx view: Updated default Qt binding to `Pyside6`; CI: Updated config

### LINKED ISSUE #4168: Changing preferred Qt binding from pyqt5 to qt6
Qt5 has reaching the end of regular support mid-2023 and only receives updates under [extended commercial support](https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders)
`silx` issues occuring with Qt5 but not Qt6 starts to be reported (see #4135).

IMO, it's timely to consider changing the default Qt binding used by `silx`, this means:

- Changing the priorities of the different Qt bindings. Current priorities:
  https://github.com/silx-kit/silx/blob/a7da2634a2a26e1b77ea3e4bdf080027da2a7c0a/src/silx/gui/qt/_qt.py#L67
- Changing the Qt binding installed with `pip install silx[full]`

Support of `PyQt5` will remain, it will just no longer be selected by default.

The following question is: `PySide6` or `PyQt6` ?
I'm +1for `PySide6`.

t20100:
Checklist:

<hr>

This PR proposes to change the Qt binding priority list to use PySide6 as the default binding.
It also uses `PySide6` to generate the Windows executable.

This is worth a major version bump for the next release IMO.

TODO:
- [x] Test Windows application
- [x] Make sure it is OK to use `manylinux_2_28`

related to #4135, closes #4168

### DIFF
--- a/.github/actions/setup-system/action.yml
+++ b/.github/actions/setup-system/action.yml
@@ -1,0 +1,33 @@
+name: 'Setup Linux'
+description: 'Install system packages and setup Intel Opencl ICD'
+
+runs:
+  using: "composite"
+  steps:
+    # Install packages:
+    # OpenCL lib
+    # xvfb to run the GUI test headless
+    # libegl1-mesa: Required by Qt xcb platform plugin
+    # libgl1-mesa-glx: For OpenGL
+    # xserver-xorg-video-dummy: For OpenGL
+    # libxkbcommon-x11-0, ..: needed for Qt plugins
+    - name: Install system packages
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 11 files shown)


t20100:
CI seg fault with PySide (see PR #4169 for a fix)

t20100:
The release workflow is updated to work with PySide6 as the default Qt binding.

Indeed:
- Qt6 requires extra system packages
- PySide6.4 (the minimum required version) requires [`manylinux_2_28`](https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux_2_28-almalinux-8-based), i.e. Debian 10+, Ubuntu 18.10+.

So with this PR, wheels available on pypi will be `manylinux_2_28` instead of `manylinux_2014`.

It also adds `--fail-on-warning` when building the documentation to avoid generating the documentation when a module cannot be imported.

t20100:
In order to avoid copy/paste between CI and release workflows, this PR adds a `setup-system` composite action.

t20100:
It should be OK to use [manylinux_2_28 (AlmaLinux 8 based)](
https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux_2_28-almalinux-8-based):

> Toolchain: GCC 14
>     x86_64 image: quay.io/pypa/manylinux_2_28_x86_64
>     aarch64 image: quay.io/pypa/manylinux_2_28_aarch64
>     ppc64le image: quay.io/pypa/manylinux_2_28_ppc64le
>     s390x image: quay.io/pypa/manylinux_2_28_s390x
> Built wheels are also expected to be compatible with other distros using glibc 2.28 or later, including:
>     Debian 10+
>     Ubuntu 18.10+
>     Fedora 29+
>     CentOS/RHEL 8+

payno (approved):

PR #4174, opened 2024-09-19
silx.gui: support imageaggregate in _plot2d

EdgarGF93:
Checklist:

<hr>

This PR enables:

- [x] ImageAggregate filters on _plot2dview
- [x] ImageAggregate filters on Nexusimageview
- [x] ImageAggregate filters on StackView


EdgarGF93:
I'm not sure if the slot `_aggregationModeChanged` and `setAggregatedImage` are well placed into into the class `Plot2View`

src/silx/gui/plot/items/image_aggregated.py
```diff
@@ -180,7 +180,7 @@ def _getLevelOfDetails(self) -> Tuple[int, int]:
         return lodx, lody
 
     @docstring(ImageDataBase)
-    def setData(self, data, copy=True):
+    def setData(self, data, copy=True, **kwargs):
```
t20100, line 183: Why is `**kwargs` needed?

src/silx/gui/plot/actions/image.py
```diff
@@ -0,0 +1,131 @@
...
+
+import logging
+from silx.gui import qt
+from silx.gui import icons
+from ..items.image_aggregated import ImageDataAggregated
+
+_logger = logging.getLogger(__name__)
+
+class AggregationModeAction(qt.QWidgetAction):
+    """Action providing few filters to the image"""
```
t20100, line 50:
```suggestion
    """Action providing filters for an aggregated image"""
```

src/silx/gui/plot/actions/image.py
```diff
@@ -0,0 +1,131 @@
...
+
+    sigAggregationModeChanged = qt.Signal()
```
t20100, line 52:
```suggestion
    sigAggregationModeChanged = qt.Signal()
    """Signal emitted when the aggregation mode has changed"""
```

src/silx/gui/data/DataViews.py
```diff
@@ -1066,13 +1068,128 @@ def createWidget(self, parent):
...
+        aggregationMode = plot.getAggregationModeAction().getAggregationMode()
+        if aggregationMode is not None and isinstance(item, ImageDataAggregated):
+            item.setAggregationMode(aggregationMode)
+        else:
+            image = item.getData(copy=False)
+            if image is None:
+                return
+            origin = item.getOrigin()
+            scale = item.getScale()
+            self.setAggregatedImage(image, origin, scale, copy=False, resetzoom=False)
```
t20100, line 1097: Since the view controls its content, the image item could always be an `ImageDataAggregated` and so this code could be simplified.

src/silx/gui/data/DataViews.py
```diff
@@ -1089,6 +1206,15 @@ def setData(self, data):
         )
         self.__resetZoomNextTime = False
 
+    def setAggregatedData(self, data):
+        data = self.normalizeData(data)
+        self.getWidget().addImage(
+            legend="data", data=data, resetzoom=self.__resetZoomNextTime
+        )
+        self.__resetZoomNextTime = False
```
t20100, line 1214: This method does not seems to be used.

It seems that updating `def setData(self, data):` to add an `ImageAggregatedItem` instead of calling `addImage` would make it simpler: Then, changing the aggregation mode would only be a matter of changing this item's aggregation mode

t20100 (commented):
Thanks for the PR!

I left a first round of feedbacks that aims at hopefully simplifying the implementation.

src/silx/gui/plot/items/image_aggregated.py
EdgarGF93: Without the kwargs, we got this error for ImageAggregate.setdata
```
ERROR:silx.app.view.main:An error occured in silx view:
ERROR:silx.app.view.main:<class 'TypeError'> ImageDataAggregated.setData() got an unexpected keyword argument 'alternative'   File "/home/edgar1993a/work/silx/src/silx/gui/data/DataViewer.py", line 299, in __setDataInView
    self.__currentView.setData(self.__displayedData)
  File "/home/edgar1993a/work/silx/src/silx/gui/data/DataViews.py", line 673, in setData
    self.__currentView.setData(data)
  File "/home/edgar1993a/work/silx/src/silx/gui/data/DataViews.py", line 1123, in setData
    plot.addImage(
  File "/home/edgar1993a/work/silx/src/silx/gui/plot/PlotWidget.py", line 1496, in addImage
    image.setData(data, alternative=pixmap, copy=copy)
```

EdgarGF93:
how about this approach?

src/silx/gui/plot/items/image_aggregated.py
t20100: The `**kwargs` should not be needed since there is no need to call `addImage` in the DataView.

src/silx/gui/data/DataViews.py
```diff
@@ -1066,12 +1069,35 @@ def createWidget(self, parent):
...
         maskToolsWidget = widget.getMaskToolsDockWidget().widget()
         maskToolsWidget.setItemMaskUpdated(True)
         return widget
+    
+    def getAggregationModeAction(self):
+        """Action toggling the aggregation mode action
+
+        :rtype: actions.PlotAction
+        """
+        return self.__aggregationModeAction
```
t20100, line 1090: It's nowadays best to use typing rather than `:rtype:` in docstring since the type serves 2 purposes: it is used to generate the documentation and also slowly starts to type the project's API:

```suggestion
    def getAggregationModeAction(self) -> AggregationModeAction:
        """Action toggling the aggregation mode action"""
        return self.__aggregationModeAction
```

src/silx/gui/data/DataViews.py
```diff
@@ -1066,12 +1069,35 @@ def createWidget(self, parent):
         widget.setDefaultColormap(self.defaultColormap())
         widget.getColormapAction().setColormapDialog(self.defaultColorDialog())
         widget.getIntensityHistogramAction().setVisible(True)
+
+        self.__aggregationModeAction = AggregationModeAction(parent=widget)
+        widget.toolBar().addAction(self.__aggregationModeAction)
+        self.__aggregationModeAction.setVisible(True)
```
t20100, line 1075: I don't think it is needed since `QAction` visibility is `True` by default:

```suggestion
```

src/silx/gui/data/DataViews.py
```diff
@@ -1084,7 +1110,17 @@ def normalizeData(self, data):
...
+        if imageItem is None:
+            imageItem = ImageDataAggregated()
+            imageItem.setAggregationMode(self.getAggregationModeAction().getAggregationMode())            
+            imageItem.setName("data")
+            imageItem.setColormap(plot.getDefaultColormap())
+            plot.addItem(imageItem)
+
+        plot.addImage(
             legend="data", data=data, resetzoom=self.__resetZoomNextTime
         )
```
t20100, line 1125: It sounds best to not call `addImage` here and instead call `ImageDataAggregated.setData` directly + `PlotWidget.resetZoom`

```suggestion
        imageItem.setData(data)
        if self.__resetZoomNextTime:
            plot.resetZoom()
```

t20100 (commented):
By avoiding calling `addImage`  there is no need for `setData` `**kwargs`

EdgarGF93:
About the `_ComplexImageView`, I'm not sure how to implement the filter without changing the `ImageComplexView` class

src/silx/gui/data/DataViews.py
```diff
@@ -1302,8 +1335,27 @@ def createWidget(self, parent):
...
+        plot = self.getWidget().getPlotWidget()
+        item = plot._getItem("image")
+
+        if item is None:
+            return
+
+        if isinstance(item, ImageDataAggregated):
+            aggregationMode = self.getAggregationModeAction().getAggregationMode()
+            item.setAggregationMode(aggregationMode)
+
```
t20100, line 1358: Let's add the `AggregationModeAction` in the widget which controls the item in the plot.
- For `_Plot2dView` above, it is the widget that controls the plot item.
- Here, the widget controlling the plot item is the `StackView`, so best to control it's plot item from this widget. Otherwise it adds implicit relations: When looking at the `StackView` code, one can wonder why an `ImageDataAggregated item is used there.

src/silx/gui/data/DataViews.py
```diff
@@ -1792,7 +1844,25 @@ def createWidget(self, parent):
...
+    def _aggregationModeChanged(self):
+        plot = self.getWidget().getPlot()
+        item = plot._getItem("image")
+
+        if item is None:
+            return
+        
+        if isinstance(item, ImageDataAggregated):
+            item.setAggregationMode(self.getAggregationModeAction().getAggregationMode())
```
t20100, line 1865: Same here, the widget controlling the plot item is `ArrayImagePlot`, so the let it handle the aggregation mode as well (this would allow to enable/disable this action according to the displayed item image/scatter)

t20100 (commented):
Let's keep the complex number for another PR!

I left a few more comments. The main point is to put the aggregation mode action in the widget that adds/updates the aggregated plot item rather in the widget that is using this widget.

EdgarGF93:
Do you mean adding the Aggregation items in the class where the `self._plot` is? (`StackView` and `ArrayImagePlot` as you said)
In the case of `_Plot2dView` the plot is what is called `widget`, but yes, `_Plot2dView` is the class creating the plot anyway

t20100:
> Do you mean adding the Aggregation items in the class where the self._plot is? (StackView and ArrayImagePlot as you said)

Yes, StackView and ArrayImagePlot which are the widgets that create and add/remove the `ImageAggregatedItem` to the plot (there is not necessarily a self._plot in there). Doing so will make those widgets self-consistent.

t20100:
I made a PR on your branch with a few updates/fixes: https://github.com/EdgarGF93/silx/pull/3
I let you review it and merge if you agree with it, else let's discuss it.

Apart from that, this looks good to me! Thanks for the work and your patience !

EdgarGF93:
Looks good to me, thanks Thomas!

t20100 (approved):

PR #4176, opened 2024-09-29
HorizontalSliderWithBrowser: slider with play button

EdgarGF93:
Checklist:

<hr>

This PR provides:

- [x] A Play Button for the HorizontalSliderWithBrowser to play a "movie" with the frames along that axis
- [x] ContextMenu with a Lineedit to change the frame per seconds, (the maximum is set by the rendering time per frame)
- [x] ContextMenu with a SpinBox to set an interval between displayed frames (not one by one)

### DIFF
--- a/src/silx/gui/widgets/FrameBrowser.py
+++ b/src/silx/gui/widgets/FrameBrowser.py
@@ -220,6 +220,9 @@ class FrameBrowser(qt.QWidget):
 
 
 class _SliderPlayWidgetAction(qt.QWidgetAction):
+
+    sigValueChanged = qt.Signal(int)
+
     def __init__(
             self,
             parent: qt.QWidget | None = None,
@@ -236,6 +239,7 @@ class _SliderPlayWidgetAction(qt.QWidgetAction):
         self._spinbox = qt.QSpinBox()
         self._spinbox.setToolTip(tooltip)
         self._spinbox.setRange(1,1000000)
+        self._spinbox.valueChanged.connect(self.sigValueChanged)
         label = qt.QLabel(label)
         label.setToolTip(tooltip)
         layout.addWidget(label)
@@ -250,12 +254,16 @@ class _SliderPlayWidgetAction(qt.QWidgetAction):
 
 
 class _PlayButtonContextMenu(qt.QMenu):
+
+    sigFrameRateChanged = qt.Signal(int)
+
     def __init__(self, parent: qt.QWidget | None = None):
         super().__init__(parent)
         self._build()
 
     def _build(self):
         self._framerateAction = _SliderPlayWidgetAction(self, label="FPS:", tooltip="Display speed in frames per second")
+        self._framerateAction.sigValueChanged.connect(self.sigFrameRateChanged)
         self._intervalAction = _SliderPlayWidgetAction(self, label="Interval:", tooltip="Jump between frames")
         self.addAction(self._framerateAction)
         self._framerateAction.setValue(10)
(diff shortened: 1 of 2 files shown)


EdgarGF93:
silxplay

src/silx/gui/widgets/FrameBrowser.py
```diff
@@ -312,3 +312,100 @@ def setValue(self, value):
...
+    def getFrameRate(self):
+        return int(self.line_edit_action.line_edit.text())
+    
+    def setInterval(self, value:int):
+        self.interval_action.spinbox.setValue(value)
+    
+    def getInterval(self):
+        return int(self.interval_action.spinbox.value())
+        
+class HorizontalSliderWithBrowserPlay(HorizontalSliderWithBrowser):
```
t20100, line 369: I would add the play/pause feature in the `HorizontalSliderWithBrowser`.
The new feature is quite consistent with what's already there.

We can consider adding methods to show/hide the Play button if the play feature needs to be disabled at some point.

src/silx/gui/widgets/FrameBrowser.py
```diff
@@ -312,3 +312,100 @@ def setValue(self, value):
...
+        self.playButton = qt.QPushButton(self)
+        self.playButton.setIcon(icons.getQIcon("camera"))
+        self.playButton.setIconSize(iconSize)
+        self.mainLayout.addWidget(self.playButton)
+        
+        self.playButton.clicked.connect(self._playStopSequence)
+        self.menu = PlayButtonContextMenu(self)
+        
+    def contextMenuEvent(self, event):
+        self.menu.exec_(self.mapToGlobal(event.pos()))
```
t20100, line 388: Providing access to the settings this way raises 2 issues:
- The context menu is for the whole horizontal slider, not just the menu.
- There is no visual hints about the availability of those settings.

I would try to use of a QToolButton here which can display an arrow (see https://doc.qt.io/qt-6/qtoolbutton.html#popupMode-prop, https://github.com/silx-kit/silx/blob/main/src/silx/gui/plot/PlotToolButtons.py)

src/silx/gui/widgets/FrameBrowser.py
```diff
@@ -312,3 +312,100 @@ def setValue(self, value):
...
+        self.line_edit_action.line_edit.setText(value)
+        
+    def getFrameRate(self):
+        return int(self.line_edit_action.line_edit.text())
+    
+    def setInterval(self, value:int):
+        self.interval_action.spinbox.setValue(value)
+    
+    def getInterval(self):
+        return int(self.interval_action.spinbox.value())
```
t20100, line 367: In terms of API, it would be best to expose this in the `HorizontalSliderWithBrowser`

src/silx/gui/widgets/FrameBrowser.py
```diff
@@ -312,3 +312,100 @@ def setValue(self, value):
...
+        if self._browser.getValue() < self._browser.getRange()[-1]:
+            self.setValue(self._browser.getValue() + interval)
+        else:
+            self._stopTimer()
+            
+    def _startTimer(self):
+        framerate = self.menu.getFrameRate()
+        waiting_time_ms = int(1 / framerate * 1e3)
+        self.__timer.start(waiting_time_ms)
+        self.playButton.setIcon(icons.getQIcon("close"))
```
t20100, line 407: A square "stop" icon would makes more sense, or keeping the button checked while playing and uncheck it at the end.

src/silx/gui/widgets/FrameBrowser.py
```diff
@@ -312,3 +312,100 @@ def setValue(self, value):
...
+        fontMetric = self.fontMetrics()
+        iconSize = qt.QSize(fontMetric.height(), fontMetric.height())
+        
+        self.playButton = qt.QPushButton(self)
+        self.playButton.setIcon(icons.getQIcon("camera"))
+        self.playButton.setIconSize(iconSize)
+        self.mainLayout.addWidget(self.playButton)
+        
+        self.playButton.clicked.connect(self._playStopSequence)
+        self.menu = PlayButtonContextMenu(self)
```
t20100, line 385: Best to keep private what can be: `self._playButton`, `self._menu`

t20100 (commented):
Thanks for the PR, it's a nice feature!

I left a few comments, mainly:
- Probably best to extend `HorizontalSliderWithBrowser` rather than using inheritance
- Improve the play settings menu to be discoverable and only available on the play button.

EdgarGF93:
Thanks Thomas, yesterday I user asked me if it would be possible to export the movie as a file to use it in presentations. What do you think?

t20100:
I made a few propositions of changes in https://github.com/EdgarGF93/silx/pull/2
I let you review it.

t20100 (approved):
Looks good to me.

Nice feature!

PR #4177, opened 2024-10-01
silx.io.h5py_utils: Fixed support of libhdf5 v1.14.4

t20100:
Checklist:

<hr>

This PR keeps `locking=None` instead of forcing it to `True` when checking if it can disable it so that it uses libhdf5 default locking mode (either `best-effort` - the default - or `True` - which can be set as default at compile time...).

libhdf5 v1.14.4 (that comes in h5py v3.12.1 wheel) fixed an issue with retrieving the file locking information from opened files in order to fix the opening of external links (that was not respecting the file locking).
It seems that beforehand, the compatibility of the ignore locking errors (the difference between `locking=True` and `locking="best-effort"`) was not tested when opening already opened files while it is tested now.

See https://github.com/HDFGroup/hdf5/blob/57948148dbcb275c9a39212eebc62e071b11a00d/release_docs/RELEASE.txt#L917-L940

It also adds typing on the modified functions.

### DIFF
--- a/src/silx/io/h5py_utils.py
+++ b/src/silx/io/h5py_utils.py
@@ -98,7 +98,8 @@ def _hdf5_file_locking(
 
     :param mode: read-only by default
     :param locking: by default it is disabled for `mode='r'`
-                    and `swmr=False` and enabled for all other modes.
+                    and `swmr=False` and enabled when supported
+                    for all other modes.
     :param swmr: try both modes when `mode='r'` and `swmr=None`
     :param libver:
     """
@@ -361,8 +362,8 @@ class File(h5py.File):
         :param filename:
         :param mode: read-only by default
         :param locking: by default it is disabled for `mode='r'`
-                        and `swmr=False` and enabled for all
-                        other modes.
+                        and `swmr=False` and enabled when supported
+                        for all other modes.
         :param enable_file_locking: deprecated
         :param swmr: try both modes when `mode='r'` and `swmr=None`
         :param libver:
--- a/src/silx/io/h5py_utils.py
+++ b/src/silx/io/h5py_utils.py
@@ -1,5 +1,5 @@
 # /*##########################################################################
-# Copyright (C) 2016-2023 European Synchrotron Radiation Facility
+# Copyright (C) 2016-2024 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/silx/io/h5py_utils.py
```diff
@@ -111,7 +118,7 @@ def _hdf5_file_locking(mode="r", locking=None, swmr=None, libver=None, **_):
             _logger.debug(
                 "Non-locking readers will fail when a writer has already locked the HDF5 file (this restriction applies to libver >= v110)"
             )
-    return locking
+    return "best-effort" if locking else False
```
t20100, line 121: This is the relevant change

t20100:
While this fixes issues and keeps `h5py_utils.File`'s `locking` argument consistent with previous version, the behavior is no longer consistent with `h5py.File`'s `locking` arg..... So not sure it is the  right way.

src/silx/io/h5py_utils.py
```diff
@@ -88,17 +96,15 @@ def _hdf5_file_locking(mode="r", locking=None, swmr=None, libver=None, **_):
...
-    :param str or None mode: read-only by default
-    :param bool or None locking: by default it is disabled for `mode='r'`
-                                 and `swmr=False` and enabled for all
-                                 other modes.
-    :param bool or None swmr: try both modes when `mode='r'` and `swmr=None`
-    :param None or str or tuple libver:
-    :returns bool:
+    :param mode: read-only by default
+    :param locking: by default it is disabled for `mode='r'`
+                    and `swmr=False` and enabled for all other modes.
```
woutdenolf, line 101: Perhaps say this:

"enabled for all other modes" -> "enabled when supported for all other modes"

src/silx/io/h5py_utils.py
```diff
@@ -341,25 +347,25 @@ class File(h5py.File):
...
-                                        and `swmr=False` and enabled for all
-                                        other modes.
-        :param bool or None enable_file_locking: deprecated
-        :param bool or None swmr: try both modes when `mode='r'` and `swmr=None`
-        :param None or str or tuple libver:
+        :param filename:
+        :param mode: read-only by default
+        :param locking: by default it is disabled for `mode='r'`
+                        and `swmr=False` and enabled for all
+                        other modes.
```
woutdenolf, line 365: same here

woutdenolf (approved):

src/silx/io/h5py_utils.py
t20100, line 101:
```suggestion
                    and `swmr=False` and enabled when supported
                    for all other modes.
```

src/silx/io/h5py_utils.py
t20100, line 365:
```suggestion
                        and `swmr=False` and enabled when supported
                        for all other modes.
```

PR #4189, opened 2024-10-31
silx.gui.plot.ImageStack: fix `_urlIndexes` initialization. 

payno:
Make sure `ImageStack._urlIndexes` is always a dict.

This can be lead to errors when calling `ImageStack.getUrls` before any url is called and better for consistency.


t20100 (approved):
CI fails for other reasons (some deprecation warnings) that can be fixed in another PR.

PR #4198, opened 2024-11-20
OpenCL: work-around for double precision GPU on Apple-silicon 

### LINKED ISSUE #4197: implement work around for buggy Apple Silicon GPU driver
Already available in:
https://github.com/silx-kit/pyFAI/pull/2341
and referenced to in:
https://github.com/silx-kit/pyFAI/pull/2339

kif:
Close #4197

Expose a not option for compile option setting properly `cl_khr_fp64`


pierrepaleo:
Looks good, works for me.

the try/except with

```python
d_data_1.fill(numpy.float32(1.0))
```

might be a first step if we want to include a check that driver works correctly. To think about.

PR #4204, opened 2025-01-21
gui matplotlib backend warning: fix conflict between y limits and autoscale

### LINKED ISSUE #4203: warning from matplotlib
Matplotlib 3.10 generate new warning: 
``` bash
Ignoring fixed y limits to fulfill fixed data aspect with adjustable data limits.
```

It can be reproduce when resizing the widget obtained by the following code:
```python
import numpy
from silx.gui import qt

app = qt.QApplication([])

widget = Plot2D()
widget.show()
data = numpy.random.rand(100).reshape(10, 10)
widget.setKeepDataAspectRatio(True)
widget.addImage(data)

app.exec_()
```

payno:
This PR is an attempt to fix a warning appearing with matplotlib 3.10

See details in #4203 

close #4203

### DIFF
--- a/src/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/src/silx/gui/plot/backends/BackendMatplotlib.py
@@ -1351,8 +1351,8 @@ class BackendMatplotlib(BackendBase.BackendBase):
     def setKeepDataAspectRatio(self, flag):
         self.ax.set_aspect(1.0 if flag else "auto")
         self.ax2.set_aspect(1.0 if flag else "auto")
-        # self.ax.set_autoscaley_on(flag)
-        # self.ax2.set_autoscaley_on(flag)
+        self.ax.set_autoscaley_on(flag)
+        self.ax2.set_autoscaley_on(flag)
 
     def setGraphGrid(self, which):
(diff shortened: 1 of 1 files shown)


payno:
What is not clear to me (does this PR creates a conflict with existing code) is the usage of  [`set_autoscaley_on` for ax2 in the initialization of the backend](https://github.com/silx-kit/silx/blob/main/src/silx/gui/plot/backends/BackendMatplotlib.py#L555)

t20100:
I still get the warning with this PR. I have a look at it.

t20100:
Each call to [`set_ylim` sets `autoscaley_on` to `False`](https://github.com/matplotlib/matplotlib/blob/dd8dbca1b1343afd5584cabc8f149a75f42de41c/lib/matplotlib/axes/_base.py#L3964)... 

Let's use [`set_ybound`](https://github.com/matplotlib/matplotlib/blob/dd8dbca1b1343afd5584cabc8f149a75f42de41c/lib/matplotlib/axes/_base.py#L3908) which does not update `autoscaley_on` and also takes care of sorting ymin, ymax according to their value and whether the axis is inverted or not.

t20100 (approved):

PR #4245, opened 2025-03-28
OpenCL: Fix context creation through PYOPENCL_CTX environment variable

pierrepaleo:
Checklist:

<hr>

This MR addresses an issue in `opencl/common.py` when using `PYOPENCL_CTX` variable.  

When using `PYOPENCL_CTX`, a new context is always created regardless of the cache content. This contrasts with the behavior obtained when not using `PYOPENCL_CTX` (contexts are re-used whenever possible).

```python
import os
from silx.opencl.processing import OpenclProcessing

op1 = OpenclProcessing()
op2 = OpenclProcessing()
print(op1.ctx is op2.ctx) # True

os.environ["PYOPENCL_CTX"] = "0:0"
op3 = OpenclProcessing()
op4 = OpenclProcessing()

print(op3.ctx is op4.ctx) # False
```

### DIFF
--- a/src/silx/opencl/common.py
+++ b/src/silx/opencl/common.py
@@ -119,7 +119,8 @@ def get_pyopencl_ctx_tuple(pyopencl_ctx_str):
         try:
             ret = int(val)
         except ValueError:
-            return default_val
+            ret = default_val
+        return ret
 
     if pyopencl_ctx_str in ["", ":"]:
         return (0, 0)
--- a/src/silx/opencl/common.py
+++ b/src/silx/opencl/common.py
@@ -110,6 +110,28 @@ NVIDIA_FLOP_PER_CORE = {
 
 AMD_FLOP_PER_CORE = 160  # Measured on a M7820 10 core, 700MHz 1120GFlops
 
+def get_pyopencl_ctx_tuple(pyopencl_ctx_str):
+    """
+    Converts a PYOPENCL_CTX environment variable into a tuple (platform, device)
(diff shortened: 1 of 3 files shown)


kif:
Hi Pierre,

Thanks for taking care of this ...

Does this PR still work when `PYOPENCL_CTX=Intel` for example ?
This is a feature from PyOpenCL `silx` should not break.
```
PYOPENCL_CTX=Intel python -c "import pyopencl;print(pyopencl.create_some_context().devices)"
[<pyopencl.Device 'AMD Ryzen Threadripper PRO 3975WX 32-Cores' on 'Intel(R) OpenCL' at 0x2fda618>]

PYOPENCL_CTX=Intel python -c "import silx.opencl;print(silx.opencl.ocl.create_context().devices)"
/usr/lib/python3/dist-packages/pyopencl/__init__.py:270: CompilerWarning: Non-empty compiler output encountered. Set the environment variable PYOPENCL_COMPILER_OUTPUT=1 to see more.
  warn("Non-empty compiler output encountered. Set the "
[<pyopencl.Device 'AMD Ryzen Threadripper PRO 3975WX 32-Cores' on 'Intel(R) OpenCL' at 0x1f8a128>]
```
Apparently this PR does not break the compatibility with PyOpenCL ... so :+1:

kif (approved):
This PR does not break anything but could probably be improved by using
`pyopencl.choose_device` and then use the cached version of the ctx ...

PR #4259, opened 2025-04-07
Maintenance: Updated syntax to python3.10

t20100:
Checklist:

<hr>

This PR updates the code syntax using `pyupgrade --py310-plus` to Python3.10 syntax.
It also converts some extra `str.format` to f-string that where not found by the tool and removes unused imports.


payno:
Maybe this is the kind of PR a 'silx non expert' could help with ?

payno (approved):
:+1: this is what I call moving forward :)

PR #4271, opened 2025-04-14
Draft: Add 'dynamic colormap' mode

### LINKED ISSUE #4263: Dynamically adjust colormap around cursor
Adjust the colormap used to render the image to a 40x40 (e.g.) pixels ROI around the current cursor position. This should be done dynamically, i.e. as the cursor moves, the colormap is adjusted. The behaviour is obtained when a keyboard key (e.g. w) is pressed as the mouse moves.

The original colormap (min/max or mean +/-3 stds) could be restored by pressing another key.

payno:
close https://github.com/silx-kit/silx/issues/4263


lesaintjerome:
- [x] Define the area (near) the mouse to be used to compute the colormap
- [x] If possible display the region near the mouse where the mode (this is implemented by setting `z=1` in the `addShape()` function.
- [x] Add a dedicated icon for this new mode
- [ ] Add documentation
- [ ] Clean the code
- [ ] Add test
- [ ] Link this mode to the ColormapDialog

lesaintjerome:
It would be more relevant to adjust the colormap over the ROI of screen pixels (rather than image pixels).

lesaintjerome:
> It would be more relevant to adjust the colormap over the ROI of screen pixels (rather than image pixels).

Shall I try to implement? Actually, I am not entirely convinced by this. If zoomed in, one image pixel might be represented by a 20x20 square of screen pixels. Adjusting over this does not make sense...

lesaintjerome:
For some reason, it seems that no check is done on the PR. There were some yesterday (and they failed, for reason that are beyond my understanding...). No more today.

lesaintjerome:
@t20100 : asked the review to you, because Henri opened the PR... We discussed with Henri: I ask a review as a first iteration.

payno:
> > It would be more relevant to adjust the colormap over the ROI of screen pixels (rather than image pixels).
> 
> Shall I try to implement? Actually, I am not entirely convinced by this. If zoomed in, one image pixel might be represented by a 20x20 square of screen pixels. Adjusting over this does not make sense...

IMO it can be done in another PR. Let's try to have the good structure and something fairly working first. Then it can be modified later.

payno:
> o check is done on the PR. There were some yesterday (and they failed, for reason that are beyond my understanding...). No more today.

Indeed this is a bit weird.

src/silx/gui/plot/test/test_plotinteraction.py
```diff
@@ -49,6 +51,31 @@ def received(self):
...
+        vmin, vmax, bb_x, bb_y = DynamicColormapMode.compute_vmin_vmax(roi,(11,11))
+        assert vmin == 23 and vmax == 460 and bb_x == (1,1,21,21) and bb_y == (1,21,21,1)
+
+        roi = numpy.arange(484).reshape((22,22))
+        vmin, vmax, bb_x, bb_y = DynamicColormapMode.compute_vmin_vmax(roi,(0,0))
+        assert vmin == 0 and vmax == 207 and bb_x == (0,0,10,10) and bb_y == (0,10,10,0)
+
+        roi = numpy.arange(484).reshape((22,22))
+        vmin, vmax, bb_x, bb_y = DynamicColormapMode.compute_vmin_vmax(roi,(21,21))
+        assert vmin == 253 and vmax == 483 and bb_x == (11,11,22,22) and bb_y == (11,22,22,11)
```
payno, line 76: new test mus be done with `pytest` instead of `unittest`.
And maybe it could be split into two tests instead of one.

So more something like
```python
def test_dynamic_colormap_interaction(qapp):
        plot = PlotWidget()
        plot.setInteractiveMode("dynamic_colormap", shape="rectangle", label="test")

        interaction = plot.getInteractiveMode()
        assert interaction["mode"] == "dynamic_colormap"
        assert isinstance(plot.interaction()._eventHandler, DynamicColormapMode)

def test_dynamic_colormap_vmin_vmax_calculation(qapp):
        roi = numpy.arange(484).reshape((22,22))
        vmin, vmax, bb_x, bb_y = DynamicColormapMode.compute_vmin_vmax(roi,(11,11))
        assert vmin == 23 and vmax == 460 and bb_x == (1,1,21,21) and bb_y == (1,21,21,1)

        roi = numpy.arange(484).reshape((22,22))
        vmin, vmax, bb_x, bb_y = DynamicColormapMode.compute_vmin_vmax(roi,(0,0))
        assert vmin == 0 and vmax == 207 and bb_x == (0,0,10,10) and bb_y == (0,10,10,0)

        roi = numpy.arange(484).reshape((22,22))
        vmin, vmax, bb_x, bb_y = DynamicColormapMode.compute_vmin_vmax(roi,(21,21))
        assert vmin == 253 and vmax == 483 and bb_x == (11,11,22,22) and bb_y == (11,22,22,11)
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1026,6 +1026,17 @@ def __init__(self, parent=None, title="Colormap Dialog"):
...
+        self._dynamicColormap = qt.QPushButton(self)
+        self._dynamicColormap.setCheckable(True)
+        self._dynamicColormap.setEnabled(False)
+        self._dynamicColormap.setText("Dynamic colormap")
+        self._dynamicColormap.setIcon(icons.getQIcon("add-shape-rectangle"))
+        self._dynamicColormap.setCheckable(True)
+        # self._dynamicColormap.toggled.connect(
+        #    self._handleDynamicColormap, type=qt.Qt.QueuedConnection
+        # )
+
```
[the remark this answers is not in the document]
lesaintjerome: OK. I remove it completely since I see no point in having a checkbox in this dialog which does nothing but what the icon in the widget toolbar does.

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1643,6 +1645,72 @@ def endDrag(self, startPos, endPos, btn):
...
+    """This mode automatically adjusts the colormap range of the image
+    based on a 20x20 ROI centered on the current cursor position.
+
+    :param plot: The Plot to which this interaction is attached
+    """
+
+    ROI_SIZE = (10, 10)  # (y,x). The ROI <<radius>>
+
+    @staticmethod
+    def compute_vmin_vmax(data, dataPos):
```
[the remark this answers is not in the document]
lesaintjerome: Will do. By this, you simply mean renaming it `_compute_vmin_vmax()` right?

src/silx/gui/dialog/ColormapDialog.py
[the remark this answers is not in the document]
payno: Yes I think removing is the best. But indeed those are covering two 'different' uses cases / concept.
1. **The QAction**
Each `PlotWindow` (like `Plot2D`) can define a set of action tunable by the users. By default not all are here (and yours might not be part of the default one - to be discussed).
But anyway everyone defining `PlotWindow` can define which action are present on the toolbar or not.
2. **The ColorbarDialog**
Well this widget is common to everyone and everyone will access to all the options for now. IMO at the end we should find your option. But again it can come in a second step. As we might also want to define the shape; the radius...

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1643,6 +1645,72 @@ def endDrag(self, startPos, endPos, btn):
...
+        # Add a blue rectangle that shows the ROI
+        self.plot.addShape(
+            bb_x,
+            bb_y,
+            legend="ColorMap reference",
+            replace=False,
+            fill=False,
+            color="blue",
+            gapcolor=None,
+            linestyle="--",
```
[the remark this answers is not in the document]
lesaintjerome: Done.

src/silx/gui/plot/actions/mode.py
```diff
@@ -137,3 +137,37 @@ def _actionTriggered(self, checked=False):
...
         if plot is not None:
             plot.setInteractiveMode("pan", source=self)
+
+
+class DynamicColormapAction(PlotAction):
+    """QAction controlling the colormap mode of a :class:`.PlotWidget`.
+
+    :param plot: :class:`.PlotWidget` instance on which to operate
+    :param parent: See :class:`QAction`
+    """
```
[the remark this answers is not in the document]
lesaintjerome: Done.

src/silx/gui/plot/test/test_plotinteraction.py
lesaintjerome: Not sure I understand the comment on `unittest` vs. `pytest`. OK for the split. DOne.

t20100:
> > It would be more relevant to adjust the colormap over the ROI of screen pixels (rather than image pixels).
> 
> Shall I try to implement? Actually, I am not entirely convinced by this. If zoomed in, one image pixel might be represented by a 20x20 square of screen pixels. Adjusting over this does not make sense...

You could have a mix of both: use a 20x20 square on screen and then make sure it is at least 20x20 pixels in the image

examples/dynamic_colormap.py
```diff
@@ -0,0 +1,60 @@
...
+
+    example = Plot2D()
+    example.addImage(x)
+    example.show()
+
+    app.exec()
+
+
+if __name__ == "__main__":
+    main()
```
t20100, line 60: There is nothing specific about dynamic colormap in this code sample,  I would remove it.

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1643,6 +1645,73 @@ def endDrag(self, startPos, endPos, btn):
             return super().endDrag(startPos, endPos, btn)
 
 
+class DynamicColormapMode(ItemsInteraction):
```
t20100, line 1648: `PlotInteraction.py` is not the place for "specific" interaction (i.e., that depends on specific items), and it's best to not add features in this module since it would benefit from a large refactoring.
I'll look for an alternative.

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1643,6 +1645,73 @@ def endDrag(self, startPos, endPos, btn):
...
+            x, y = args[:2]
+        except ValueError:
+            return
+
+        # Get data
+        result = self.plot._pickTopMost(x, y, lambda i: isinstance(i, items.ImageBase))
+        if result is None:
+            return
+        else:
+            item = result.getItem()
```
t20100, line 1688:
```suggestion

        item = result.getItem()
```

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1643,6 +1645,73 @@ def endDrag(self, startPos, endPos, btn):
...
+
+    ROI_SIZE = (10, 10)  # (y,x). The ROI <<radius>>
+    COLOR = "blue"
+    LINESTYLE = "--"
+
+    @staticmethod
+    def _compute_vmin_vmax(data: numpy.ndarray, dataPos: tuple[float, float]):
+        """Compute the min and max values of the data in a ROI centered on (x,y)"""
+        roi_size = DynamicColormapMode.ROI_SIZE
+        idx_x, idx_y = int(dataPos[0]), int(dataPos[1])
```
t20100, line 1663: This doesn't take care of origin and scale of the `Image` item

src/silx/gui/plot/PlotWidget.py
```diff
@@ -332,6 +332,13 @@ class PlotWidget(qt.QMainWindow):
     It provides the source as passed to :meth:`setInteractiveMode`.
     """
 
+    # sigDynamicColormapModeChanged = qt.Signal(object)
+    # """
+    # Signal emitted when the dynamic colormap changed
+
+    # It provides the source as passed to :meth:`setInteractiveMode`.
+    # """
+
```
t20100, line 341:
```suggestion
```

src/silx/gui/plot/PlotWidget.py
```diff
@@ -3785,6 +3792,12 @@ def keyPressEvent(self, event):
             # that even if mouse didn't move on the screen, it moved relative
             # to the plotted data.
             self.__simulateMouseMove()
+        elif key == qt.Qt.Key_W:
+            self.setInteractiveMode("dynamic_colormap")
+        elif key == qt.Qt.Key_P:
+            self.setInteractiveMode("pan")
+        elif key == qt.Qt.Key_Z:
+            self.setInteractiveMode("zoom")
```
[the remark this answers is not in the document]
t20100: Not sure we want to add this to `PlotWidget`, at least without a review of all keyboard shortcuts + it's easy to add in an application

src/silx/gui/plot/actions/mode.py
```diff
@@ -137,3 +137,39 @@ def _actionTriggered(self, checked=False):
         plot = self.plot
         if plot is not None:
             plot.setInteractiveMode("pan", source=self)
+
+
+class DynamicColormapAction(PlotAction):
+    """QAction controlling the colormap mode of a :class:`.PlotWidget`.
+    This mode adjusts the colormap based on a small region around the
```
t20100, line 144:
```suggestion
    """QAction controlling the colormap mode of a :class:`.PlotWidget`.

    This mode adjusts the colormap based on a small region around the 
```

src/silx/gui/plot/tools/toolbars.py
```diff
@@ -53,6 +53,11 @@ def __init__(self, parent=None, plot=None, title="Plot Interaction"):
         self._panModeAction = actions.mode.PanModeAction(parent=self, plot=plot)
         self.addAction(self._panModeAction)
 
+        self._dynamicColormapAction = actions.mode.DynamicColormapAction(
+            parent=self, plot=plot
+        )
+        self.addAction(self._dynamicColormapAction)
+
```
t20100, line 60: This will add it to all kind of plots (curves, scatters) while it only works with images, so it's not the place.
This could be added to the colormap dialog, though it is already quite crowded or the `ImageToolBar`... or not included in any toolbar and added by the application that wants it

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -1795,6 +1864,9 @@ def _getInteractiveMode(self):
         elif isinstance(self._eventHandler, PanAndSelect):
             return {"mode": "pan"}
 
+        elif isinstance(self._eventHandler, DynamicColormapMode):
+            return {"mode": "dynamic_colormap"}
+
```
t20100, line 1869: Let's avoid to add a "generic" interaction for a specific case

t20100 (commented):
There's a structural issue of where to implement this feature: IMO it should be implemented "on-top" of PlotWidget rather than in `PlotInteraction.py`.

There's also a question of where this feature should be provided: ColormapDialog, ImageToolBar, a QAction that is not included in plots by default ?  I would suggest to give a try to the colormap dialog since it is related to it.. even if there's quite a lot of things already.

Finally the icon does not convey the meaning of a dynamic colormap IMO.
Use https://github.com/silx-kit/silx/blob/main/src/silx/resources/gui/icons/image-select-box.png ?

src/silx/gui/plot/PlotInteraction.py
lesaintjerome: True. But it is not incompatible. In the worst case, it does not improve the local contrast. The user can choose either a specific color scale or to use that feature.

src/silx/gui/plot/PlotInteraction.py
lesaintjerome: OK. I understand that this interaction only aims 2d plots. Then, its place could be in the ImageToolbar. In the same time, it is really an interactive mode.

src/silx/gui/plot/PlotInteraction.py
t20100: The image can have an offset from origin and a scaling so the coordinates of the plot `dataPos` cannot be converted directly to indices in the image array.
As it is, the local contrast will be computed on a wrong area of the image if this image has an offset or scale.

This should be handled and it should not be an issue to do so

src/silx/gui/plot/PlotInteraction.py
t20100: Yes, it is an interaction mode so independent of the `ImageToolbar`.
I'll look at it an propose a way to structure this

src/silx/gui/plot/PlotInteraction.py
payno: You retrieve the `data` from the an item upper (on the stack).
I think this `data` is an instance of [ImageData](https://www.silx.org/doc/silx/latest/modules/gui/plot/items.html#silx.gui.plot.items.ImageData). If I m right you should have all your need there.

src/silx/gui/plot/PlotInteraction.py
payno: Not sure I get what is behind the 'generic' interaction. Do you suggest that we have some kind of interaction subset / categories ?
Because at the moment we need a specific interaction mode to avoid conflict with other modes.

src/silx/gui/plot/tools/toolbars.py
lesaintjerome: Same remark as before: could be in ImageToolbar but is truly an interactive mode.  On the top of that, this ImageToolbar does not seem to be used on the PlotWindow/ImageView that displays images. I can add the action in the Plot2D toolbar. But don't really know how to handle the interaction mode (hence the initial positioning in the PlotWidget/InteractionToolbar.

src/silx/gui/plot/PlotWidget.py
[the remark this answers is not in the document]
lesaintjerome: I understand the point, but it's really a key feature to me: to make it light... : in that respect, adding the icon in the colormap widget would not make much sense IMO. But maybe it is independent where the icon is and how to activate the interaction mode...

t20100:
Here is a minimal working prototype of how to implement it without changing the internals of `PlotWidget`:

```python
from typing import Any

import numpy as np

from silx.gui import qt
from silx.gui import icons
from silx.gui.plot import items
from silx.gui.plot.items.core import PickingResult
from silx.gui.plot.PlotWidget import PlotWidget
from silx.gui.plot.actions import PlotAction


class DynamicColormapAction(PlotAction):
    def __init__(self, plot: PlotWidget, parent: qt.QWidget | None = None):
        super().__init__(
            plot,
            icon=icons.getQIcon("image-select-box"),
            text="Dynamic colormap mode",
            tooltip="Update the colormap according to the mouse position in the plot",
            triggered=self.__triggered,
            checkable=True,
            parent=parent,
        )

    def __triggered(self, checked: bool) -> None:
        plotWidget = self.plot
        if checked:
            plotWidget.sigPlotSignal.connect(self.__plotChanged)
        else:
            plotWidget.sigPlotSignal.disconnect(self.__plotChanged)

    def __pickTopMostImageData(self, x: float, y: float) -> PickingResult | None:
        for pickingResult in self.plot.pickItems(
            x,
            y,
            condition=lambda item: isinstance(item, items.ImageDataBase)
        ):
            return pickingResult
        return None
    
    def __plotChanged(self, event: dict[str, Any]) -> None:
        if event.get("event") != "mouseClicked":
            return
        
        pickingResult = self.__pickTopMostImageData(event["xpixel"], event["ypixel"])
        if pickingResult is None:
            return
        
        image = pickingResult.getItem()
        indices = pickingResult.getIndices()

        if indices is None:
            return
        
        row, col = indices[0][0], indices[1][0]

        # TODO define ROI in pixel coords and compute range accordingly
        value = image.getData(copy=False)[row, col]
        vmin, vmax = value - 10000, value + 10000

        colormap = image.getColormap()
        colormap.setVRange(vmin, vmax)


app = qt.QApplication([])
plot = Plot2D()
dynamicColormapAction = DynamicColormapAction(plot)
plot.toolBar().insertAction(plot.getColormapAction(), dynamicColormapAction)
plot.addImage(np.arange(1024.**2).reshape(1024, 1024))
plot.show()
app.exec()
```

It's missing:
- the actually ROI-based min-max computation but that can be taken from this PR
- the W keyboard shortcut that I'm not sure where to implement and at first I propose to NOT do it in silx and think about all the keyboard shortcuts at hand. This can be added in a `Plot2D`/`PlotWidget` subclass in the application in the meantime.

t20100:
Regarding where to put this feature in silx: It would make sense to me to add it to the colormap dialog since there is already 2 related options: "Fixed scale on: [Visible Area] [Selection] .
The colormap dialog is already quite busy... but so are the toolbars..

image

t20100:
BTW, here are some alternative for the implementation:
- The `ColormapDialog` is using a [`RegionOfInterestManager`](https://www.silx.org/doc/silx/latest/modules/gui/plot/tools/index.html#silx.gui.plot.tools.roi.RegionOfInterestManager) to handle the `Selection` interaction. A similar approach could be used with [`PointROI`](https://www.silx.org/doc/silx/latest/modules/gui/plot/items.html#silx.gui.plot.items.roi.PointROI) here.
- [`PlotWidget.setInteractiveMode("select", ...)`](https://www.silx.org/doc/silx/latest/modules/gui/plot/plotwidget.html#silx.gui.plot.PlotWidget.PlotWidget.setInteractiveMode) to have a dedicated interaction mode.

src/silx/gui/plot/PlotWidget.py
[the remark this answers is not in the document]
t20100: >  But maybe it is independent where the icon is and how to activate the interaction mode...

Yes!

> it's really a key feature to me

Sure, but this doesn't mean it is a key feature for other usage of `PlotWidget` (e.g., when plotting curves). That's why I propose to start by having this shortcut in your application rather than as a default in `PlotWidget`

PR #4277, opened 2025-04-14
Build:Migrate build system to `meson-python`

### LINKED ISSUE #4215: Migrate build system to `meson-python`
Pro:
* speed of build
* today's standard for numpy/scipy

Con:
* one needs to declare every single (new) python file
* source distribution is tightly bound to git tool

Unknown:
* Mixing with pytest

kif:
Checklist:

- [X] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

close #4215


kif:
At this stage, `bootstrap.py` works

kif:
`bootstrap.py` works and compiles all extensions. 
Here is a quick benchmark for `time ./bootstrap.py -m silx`:
* Main branch:   2m11,958s
* Meson branch: 0m14,347s 
on a computer with 8 cores.

Tests are not yet OK, I am still working on it.

vallsv:
The package data is not part anymore of the project.

Here is what i get from flint for example.

```
# pyproject.toml

[tool.setuptools.package-data]
"flint.resources" = [
    "logo/*.png",
    "logo/*.svg",
    "icons/*.png",
    "icons/*.svg",
    "icons/*.mng",
    "icons/*.gif",
    "icons/*/*.png",
    "gui/*.ui",
]
```

kif:
Thanks Valentin. The idea is to remove completely setuptools.
here is a hint:
```
In [24]: silx.gui.icons.getQFile("axis")
Out[24]: <PySide6.QtCore.QFile(0x55a61af45c50) at 0x7f8ff823d640>

In [25]: silx.gui.icons.getQFile("process-working")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[25], line 1
----> 1 silx.gui.icons.getQFile("process-working")

File ~/workspace-400/silx/build/lib/python3.13/site-packages/silx/gui/icons.py:407, in getQFile(name)
    405         return qfile
    406     _logger.debug("File '%s' not found.", filename)
--> 407 raise ValueError("Not an icon name: %s" % name)

ValueError: Not an icon name: process-working
```
because the MNG fileformat is not recognized.
```
Indeed after those modifications, the file is actually recognized:
```
In [26]: silx.gui.icons._supported_formats
Out[26]: ['gif', 'svg', 'png', 'jpg']

In [27]: silx.gui.icons._supported_formats.append("mng")

In [28]: silx.gui.icons.getQFile("process-working")
Out[28]: <PySide6.QtCore.QFile(0x55a61af8f750) at 0x7f8ff38e5980>
```

**Comment by vallsv** [comment]
MNG is not a icon format, but an animated icon format.

> because the MNG fileformat is not recognized.

`getAnimatedIcon` should fall back anyway with PNG, so you probably have a problem anyway to package or read the resources for `process-working` stored as PNG.

**Comment by kif** [comment]
I just checked and MNG is not supported by pyside6 nor pyqt6. The fact that the file exists is not enough, it needs to be interpreted as well. 
How is the animated icon supposed to be created from the set of png-files ?

**Comment by vallsv** [comment]
BTW sounds like MNG and GIF are anyway always ignored in Qt5, no idea for Qt6

**Comment by vallsv** [comment]
> How is the animated icon supposed to be created from the set of png-files ?

You could check it with `MultiImageAnimatedIcon("process-working")`.

**Comment by kif** [comment]
My bad, all files were present ... except the first one :(

**Comment by vallsv** [comment]
> My bad, all files were present ... except the first one :(

I see, you'r kind of fortan guy.

**Comment by kif** [comment]
Very related topic:
https://github.com/msys2/setup-msys2/issues/133

**Comment by kif** [comment]
@mretegan I managed to get the CI working, i.e. use the MSVC compiler under windows, gcc on ubuntu and clang on macos. You can review the code ...

File: src/silx/io/specfile/specfile.pyx [hunk 1]
```diff
@@ -116,7 +116,7 @@ _logger = logging.getLogger(__name__)
 cimport cython
 from libc.stdlib cimport free
 
-cimport silx.io.specfile_wrapper as specfile_wrapper
+cimport specfile_wrapper
```
t20100, line 119: Why did the cimport needed to be changed?

src/silx/test/__init__.py
```diff
@@ -41,11 +42,8 @@
     raise
 
 
-import silx
-
-
 def run_tests(
-    module: str | None = "silx",
+    modules: Sequence[str] = ("silx",),
```
t20100, line 46:
```suggestion
    modules: Sequence[str] | None = ("silx",),
```

tools/mesonify.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/python3
+
```
t20100, line 2: A small docstring would be nice to explain what the tool does.

Can this be used to update the existing `meson.build` files?

.github/workflows/ci.yml
```diff
@@ -59,25 +59,41 @@ jobs:
...
+                  newpath.insert(0, p)
+              else:
+                  newpath.append(p)
+          print(os.linesep.join(newpath))
+          os.environ["PATH"] = ":".join(newpath)
+          
+          from build.__main__ import main
+          main(sys.argv[1:], 'python -m build')
+          EOF
+          python mybuild.py
```
t20100, line 96: We'll have to find a neater way to fix this in a future PR

.github/workflows/ci.yml
```diff
@@ -94,4 +110,4 @@ jobs:
           SILX_TEST_LOW_MEM: "False"
           SILX_OPENCL: ${{ matrix.with_opencl && 'True' || 'False' }}
         run: |
-          python -c "import silx.test, sys; sys.exit(silx.test.run_tests(verbosity=1, args=['--qt-binding=${{ matrix.QT_API }}']));"
+          python run_tests.py --installed --qt-binding=${{ matrix.QT_API }}
```
t20100, line 113: Why is run_tests.py need here?

t20100 (commented):
Thanks for the effort of migrating the build system! 💪 

I do have a major concern: Forgetting to update the `meson.build` files when adding/renaming python files is certainly going to happen (and tests won't discover that a test file is missing !). So, IMO this deserves a CI step that checks that all files under git are listed in the `meson.build`.

The debian packaging rules are referring to pyFAI, I propose that this is fixed or remove from this PR which is already quite huge.

I left a few comments, questions.

t20100:
BTW, before merging this, it's probably needed to make sure no new files were added and no change was made to `setup.*` in the mean time...

.github/workflows/ci.yml
mretegan: Is this because the tools from MVS are shadoed by the GNU tools as discussed here https://github.com/ilammy/msvc-dev-cmd/issues/25
If you run this in `cmd` instead of `bash` should solve the issue.
```yaml
shell: ${{ runner.os == 'Windows' && 'cmd' || 'bash' }}
```

meson.build
```diff
@@ -0,0 +1,55 @@
...
+endif
+
+# OpenMP can be used? several ways to disable it:
+#  -> use -Duse_openmp=disabled
+#  -> or the file SILX_WITH_OPENMP should contain False
+#  -> or OpenMP can be unsuported on your achitecture/compiler
+
+
+# How to disable OpenMP:
+# The 'SILX_WITH_OPENMP' file should contain 'False'
```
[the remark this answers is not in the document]
kif: There are several ways of disabling OpenMP:
* the environment variable way, even though meson developer dislike it: SILX_WITH_OPENMP
* a meson option: meson build . -Duse_openmp=disabled
* if a file named  SILX_WITH_OPENMP exists at the top level and contains False
* if OpenMP is unsuported by the compiler

Documented in https://github.com/kif/silx/blob/4215_meson/meson.build#L19

package/debian12/rules
```diff
@@ -1,74 +1,34 @@
...
-rwildcard=$(wildcard $1$2) $(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2))
-
-# How to recursively find all files with the same name in a given folder
-ALL_PYX := $(call rwildcard,silx/,*.pyx)
-#NOTA: No space before *
-
-# get the default python3 interpreter version
-PY3VER := $(shell py3versions -dv)
+export DH_VERBOSE=1
+export PYBUILD_NAME=pyfai
```
mretegan, line 4: Probably not what you want...

package/debian12/rules
```diff
@@ -1,74 +1,34 @@
...
-	dh_install -p silx package/desktop/silx.png usr/share/icons/hicolor/48x48/apps
-	dh_install -p silx package/desktop/silx.svg usr/share/icons/hicolor/scalable/apps
-	dh_install -p silx package/desktop/silx.xml usr/share/mime/packages
-
-	# install the qtdesigner files only for the python3 package
-	dh_install -p python3-silx qtdesigner_plugins/*.py /usr/lib/$(DEB_HOST_MULTIARCH)/qt5/plugins/designer/python
+	# move the scripts to right package
+	dh_install -p pyfai debian/python3-pyfai/usr/bin usr/
+	# remove all scripts installed by pybuild
+	rm -rf debian/python3-pyfai/usr/bin
```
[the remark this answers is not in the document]
kif: I did not check the debian12 packaging indeed. I will check

.github/workflows/ci.yml
kif: Needed not really, but it is actually much simpler ...

kif:
> BTW, before merging this, it's probably needed to make sure no new files were added and no change was made to `setup.*` in the mean time...

This should not be an issue since there would be a conflict at the git level.

pyproject.toml
mretegan, line 1: This should be a string

pyproject.toml
```diff
@@ -1,10 +1,106 @@
...
+silx = 'silx.__main__:main'
+
+
+[tool.cibuildwheel]
+# Skip 32-bit builds and PyPy
+skip = ["*-win32", "*-manylinux_i686", "pp*", "*musllinux*"]
+
 
 [tool.black]
 required-version = 25
```
mretegan, line 106: This could be a string

build-deb.sh
```diff
@@ -53,6 +53,12 @@ then
             bookworm)
                 debian_version=12
                 ;;
+	    trixie)
+		debian_version=13
+		;;
+	    sid)
+		debian_version=13
+		;;
```
mretegan, line 61: The added code uses tabs. Could you convert them to spaces?

tools/mesonify.py
mretegan: Also, it would be nice to have this file properly formated.

pyproject.toml
```diff
@@ -1,10 +1,107 @@
...
 requires = [
-    "wheel",
-    "setuptools",
-    "Cython>=0.29.31"
+    'meson-python>=0.11',
+    "meson>=1.1",
+    'ninja',
+    'wheel',
+    'oldest-supported-numpy',
+    'Cython>=0.29.31',
```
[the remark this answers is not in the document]
kif: cython3 is not yet available on debian12 ...

pyproject.toml
```diff
@@ -1,10 +1,107 @@
...
+tracker = 'https://github.com/silx-kit/silx/issues'
+
+
+[project.scripts]
+silx = 'silx.__main__:main'
+
+
+[tool.cibuildwheel]
+# Skip 32-bit builds and PyPy
+skip = ["*-win32", "*-manylinux_i686", "pp*", "*musllinux*"]
```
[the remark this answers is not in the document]
kif: `pyproject.toml` is the centralizing all configuration for all tools related to packaging ...

src/silx/app/compare/meson.build
```diff
@@ -0,0 +1,11 @@
+subdir('test')
+
+
+py.install_sources([
+    'CompareImagesWindow.py',
+    '__init__.py',
+    'main.py',
+],
+pure: false,    # Will be installed next to binaries
```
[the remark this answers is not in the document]
kif: The `pure:false` allows mixing binary extensions with pure python packages. For sub-packages without  binary extensions, this could be set to true.

src/silx/io/specfile/specfile.pyx
kif: It did not build. Anyway an absolute _C_import of a pxd file looks really odd to me. A relative import would have be better ...

tools/mesonify.py
kif: No, this tool was just used to generate empty meson tree, which requires manual tuning.
I updated it to prevent any harm (i.e. don't overwrite hand-crafted `meson.build` files).

meson.options
```diff
@@ -0,0 +1,8 @@
+option('use_openmp', 
+       type : 'feature', 
+       value : 'auto',
+       description : 'force enable/disable the usage of OpenMP')
+#option('SPECFILE_use_GNU_source', 
+#       type : 'feature', 
+#       value : 'disabled',
+#       description : 'force specfile to use the `strtod_l` from gcc')
```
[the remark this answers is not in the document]
kif: Well, this is the official way of dealing with options in `meson`. Using an environment variable should never be used according to the author of meson.

Right now, I reproduced the way of working of the previous builder (using environment variables) but did not wire it the official way. This is just the way to go. Having it defined in the right place may help the person who will implement it properly. Bu who still cares about `specfile` ? Maybe we could even remove it completely :-þ

pyproject.toml
kif: Maybe it could, but it is better to have it in a structured way to handle it programmatically from `tomli`

bootstrap.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
```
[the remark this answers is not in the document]
kif: I tried and was not convinced by the result:
```
-                python_version = f"python{sys.version_info.major}.{sys.version_info.minor}"
+                python_version = (
+                    f"python{sys.version_info.major}.{sys.version_info.minor}"
+                )
```

.github/workflows/ci.yml
mretegan: There appears to be an error when using the conditional operator. Sorry about that, I should have checked the code before.

The only solution I found is to split the step in two, like this:

```yaml
      - name: Build Wheel (Windows)
        if: runner.os == 'Windows'
        shell: cmd
        run: |
          python -m build

      - name: Build Wheel (Unix)
        if: runner.os != 'Windows'
        env:
          MACOSX_DEPLOYMENT_TARGET: "10.9"
        shell: bash
        run: |
          python -m build
```

I checked and test pass with this.

tools/check_meson.py
kif, line 1: @t20100 Here is the script checking that all python files are actually registered in a `meson.build` file

tools/check_meson.py
```diff
@@ -0,0 +1,61 @@
...
+                    mesoncache[mesonfile] = fd.read()
+        else:
+            print(f"File `{f}` is in a directory without `meson.build` file")
+            cnt += 1
+            continue
+        mesondata = mesoncache[mesonfile]
+        if fn not in mesondata:
+            print(f"File `{f}` is not listed in `{mesonfile}` file")
+            cnt += 1
+    print(f"Analyzed {len(src)} python source files in {len(meson)} submodules and found {cnt} issues to investigate")
```
t20100, line 55:
```suggestion
    print(f"Analyzed {len(src)} source files in {len(meson)} submodules and found {cnt} issues to investigate")
```

tools/check_meson.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    cnt = 0
+    src = []
+    meson = []
+
+    for line in get_all_files(root):
+        if line.endswith(".py") and line.startswith("src"):
+            src.append(line)
+        elif line.endswith("meson.build"):
+            meson.append(line)
```
t20100, line 38: This check does not only apply to Python files but to all files (.png, .svg, .cl, .c, .pyx, ....) except C/C++/Cython headers.
To check everything, all files should be included but ".h", ".hpp" and ".pxd" extensions:

```suggestion
        if line.endswith("meson.build"):
            meson.append(line)
        else:
            extension = os.path.splitext(line)[1]
            if line.startswith("src") and extension not in (".h", ".hpp", ".pxd"):
                src.append(line)
```

Actually, trying this, I found a missing file: `src/silx/math/combo.pyi` and a `README` in histogramnd that I'm not sure what to do with.

It also needs to handle C files declared in a `meson.build` in a parent folder.. I'll continue this tomorrow

t20100 (commented):
Thanks for the updates!
The check_meson script is a good way to make meson sustainable.

Some more comments:

- For me it is OK to fully switch to the way to pass build options as meson does even if it is not simple (https://mesonbuild.com/meson-python/how-to-guides/meson-args.html#examples) and to remove files and env. var.
- There are some empty meson.build files (src/silx/math/fit/functions/include/meson.build and src/silx/math/fit/functions/src/meson.build, src/silx/math/histogramnd/src/meson.build), is this needed?
- The check script still needs some updates.

BTW, following https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html , I tried using editable install and it works fine even when changing cython code. IMO we could drop the `bootstrap.py`...

kif:
> Some more comments:
> 
>     * For me it is OK to fully switch to the way to pass build options as meson does even if it is not simple (https://mesonbuild.com/meson-python/how-to-guides/meson-args.html#examples) and to remove files and env. var.

I used it only once, to provide the build-type (release or debug) and found it obfuscated.

>     * There are some empty meson.build files (src/silx/math/fit/functions/include/meson.build and src/silx/math/fit/functions/src/meson.build, src/silx/math/histogramnd/src/meson.build), is this needed?

Not really, reminders of the first version of mesonify. It also requires the suppression of their references in parent folder as well.

> BTW, following https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html , I tried using editable install and it works fine even when changing cython code. IMO we could drop the `bootstrap.py`...

If you remove `bootstrap.py`: `run_tests.py` and `build_doc.py` needs also to be modified accordingly since they both rely on variables configured by `bootstrap.py`. I would not recommend doing so ...

tools/check_meson.py
kif: Those files are never used from any other module, so they don't deserve installation, nor the c, h, cpp, hpp,  pyx, pxd, pxi, ... files. But this is my interpretation.

t20100: > Those files are never used from any other module, so they don't deserve installation

I agree for the `README` (BTW it probably shouldn't be there) but not for the rest: Resource files (icons, OpenCL kernels,...) are required to be installed and so should be checked.
The `combo.pyi` type stub as well according to [PEP561](https://peps.python.org/pep-0561/#stub-only-packages):

> For package maintainers wishing to ship stub files containing all of their type information, it is preferred that the *.pyi stubs are alongside the corresponding *.py files.

t20100:
> For me it is OK to fully switch to the way to pass build options as meson does
> I used it only once, to provide the build-type (release or debug) and found it obfuscated.

OK, but I would propose to remove the file based approach and only keep the env. var. setting so it stays as it was before.
IMO we'd better keep this kind of code as simple as possible.

> If you remove bootstrap.py: run_tests.py and build_doc.py needs also to be modified accordingly since they both rely on variables configured by bootstrap.py. I would not recommend doing so ...

I would also keep them for now.
However in the mid-term if the tools we depends on provide the same features, I don't see the point of keeping them.

tools/check_meson.py
t20100: BTW, numpy has a similar check: https://github.com/numpy/numpy/blob/main/tools/check_installed_files.py

t20100:
For pure Python sub-packages, numpy is using [install_subdir](https://mesonbuild.com/Reference-manual_functions.html#install_subdir) which looks to spare the explicit listing of all files:

https://github.com/numpy/numpy/blob/bec11a4672af731e52f1ed62e80bb23547a61619/numpy/meson.build#L366-L371

kif:
[install_subdir](https://mesonbuild.com/Reference-manual_functions.html#install_subdir) is deprecated (available since 0.45.0, deprecated since 0.60.0)

t20100:
I made a PR to update check_meson.py and fix CI according to https://github.com/silx-kit/silx/pull/4277#discussion_r2059151527:
https://github.com/kif/silx/pull/14

t20100:
install_subdir is not deprecated, only its behavior with non existing folders is deprecated: https://github.com/mesonbuild/meson/pull/9344

meson.build
```diff
@@ -0,0 +1,55 @@
...
+  openmp_var = os.read('SILX_WITH_OPENMP').strip()
+else
+  res = run_command(py, '-c', 'import os; print(os.environ["SILX_WITH_OPENMP"])',
+                    check:false)
+  if res.returncode() == 0
+    openmp_var = res.stdout().strip()
+  else
+    openmp_var = ''
+  endif
+endif
```
t20100, line 37: Since we were not providing file-based config, I would not add it:

 ```suggestion
res = run_command(py, '-c', 'import os; print(os.environ["SILX_WITH_OPENMP"])',
                  check:false)
if res.returncode() == 0
  openmp_var = res.stdout().strip()
else
  openmp_var = ''
endif
```

File: src/silx/io/specfile/meson.build [hunk 19]
```diff
@@ -0,0 +1,44 @@
...
+        specfile_use_gnu = os.read('SPECFILE_USE_GNU_SOURCE').strip()
+    else
+        res = run_command(py, '-c', 'import os; print(os.environ["SPECFILE_USE_GNU_SOURCE"])',
+                          check:false)
+        if res.returncode() == 0
+            specfile_use_gnu = res.stdout().strip()
+        else
+            specfile_use_gnu = ''
+        endif
+    endif
```
t20100, line 16: Same, we were not providing file-based config, so I would not add it:

```suggestion
    res = run_command(py, '-c', 'import os; print(os.environ["SPECFILE_USE_GNU_SOURCE"])',
                      check:false)
    if res.returncode() == 0
        specfile_use_gnu = res.stdout().strip()
    else
        specfile_use_gnu = ''
    endif
    endif
```

t20100 (approved):
Last comment to propose to remove file-based config.

Otherwise I'm +1 to merge it as it is and see if `install_subdir` can be used in another PR.

Issue #4199, opened 2024-12-17
Silx view: add units to axes and color bars when present


woutdenolf:
NeXus fields can have units (i.e. HDF5 datasets can have the attribute `"units"`)

![image](https://github.com/user-attachments/assets/e69e0bb3-760e-446f-8258-0256e1d27ccf)

It would be good to use these units on the axes for curves or images

![image](https://github.com/user-attachments/assets/a2062575-70fe-4bb7-92ea-332f1674bc9c)

or for nxdata plots (on the axes or the signal)

![image](https://github.com/user-attachments/assets/68075fde-3161-4f50-bafb-174ac0fd3284)

FYI ewoksweb shows units for NXdata (on axes and signal)

![image](https://github.com/user-attachments/assets/3fb7e22d-4dbe-45e6-99ef-8634aa8938a5)

Not for curves and images though

![image](https://github.com/user-attachments/assets/e6c8714c-e6c8-4df6-b206-55921292b9b7)

![image](https://github.com/user-attachments/assets/79c7d901-9687-4515-a661-e4dbb635948f)

loichuder:
Closed by #4282

PR #4282, opened 2025-04-14
silx.io.nxdata: add units to dataset name (signal/axes) when present

### LINKED ISSUE #2927: [plot] Display units in addition to the label of axes
When the units are provided as attribute of an axis in a NXdata, I guess it would be quite easy to display them in addition to the label of the axis.

My use-case is quite simple: in SAXS, half of the software expect the scattering vector to be in 1/nm, while the other half expects them to be in 1/Å. Those units differ by a factor 10 so that the user is easily confused between the two.

The initial idea was to provide 2 hoizontal axis, one in 1/nm, the other in 1/Å ...
Then the same story occurred on the vertical axis between absolute scattering intensities and kDa equivalent. This makes then 4 options and as a consequence 3 users out of 4 will be upset by the default display (and are probably not interested in learning how to switch the display).

This is why I believe this issue can easily be addressed by _properly_ labeling the various axis not only with the name of the axe but also the units (this is regardless to the log-scale)

loichuder:
Checklist:

Fix #2927 #4199 

The dataset name (signal or axes) is by order of priority:
- `@long_name`
- `"<dataset_name> (@units)"`
- `"<dataset_name>"`

image

### DIFF
--- a/src/silx/io/nxdata/_utils.py
+++ b/src/silx/io/nxdata/_utils.py
@@ -94,9 +94,7 @@ def get_attr_as_unicode(
         return copy.deepcopy(attr)
 
 
-def get_uncertainties_names(
-    group: h5py.Group, signal_name: str
-) -> list[str] | None:
+def get_uncertainties_names(group: h5py.Group, signal_name: str) -> list[str] | None:
     # Test consistency of @uncertainties
     uncertainties_names = get_attr_as_unicode(group, "uncertainties")
     if uncertainties_names is None:
@@ -201,3 +199,17 @@ def validate_number_of_axes(
                 + "must define %d or %d axes." % (ndims, INTERPDIM[interpretation])
             )
     return issues
+
+
+def get_dataset_name(nxdata_group: h5py.Group, dataset_name: str | None) -> str | None:
+    if dataset_name is None:
+        return None
+
+    dataset = nxdata_group[dataset_name]
+    if "long_name" in dataset.attrs:
+        return f"{get_attr_as_unicode(dataset, 'long_name')}"
+
+    if "units" in dataset.attrs:
+        return f"{dataset_name} ({get_attr_as_unicode(dataset, 'units')})"
+
+    return dataset_name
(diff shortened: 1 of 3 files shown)


src/silx/io/nxdata/_utils.py
```diff
@@ -201,3 +199,17 @@ def validate_number_of_axes(
                 + "must define %d or %d axes." % (ndims, INTERPDIM[interpretation])
             )
     return issues
+
+
+def get_dataset_name(nxdata_group: h5py.Group, dataset_name: str | None) -> str | None:
```
t20100, line 204: `get_dataset_name` function with a `dataset_name` argument sounds a bit unclear to me at first.
Maybe use a more explicit function name, `get_dataset_long_name`, `get_dataset_display_name`, ...? 🤷 

Also, sounds clearer to have:
`def get_dataset_name(dataset: h5py.Dataset) -> str:`
but it's at the expense of checking dataset_name is `None` outside of this function.. so not sure it worth it

loichuder: I'll sleep on it and try to find a better name for the function or the argument

axelboc: We use the term `label` in H5Web: https://github.com/silx-kit/h5web/blob/8c0f462223c5b0a0df21d1a02c99dc619a68df54/packages/app/src/vis-packs/nexus/utils.ts#L279

t20100: Indeed, this is what is used for in the end.

loichuder: Well, if we go for `label`, ideally, it would have to be consistent. This means changing the names of the attributes `signal_name`, `axes_names` of the `NXdata` class

axelboc: I'm more arguing that you need both a base "name" (the name of the dataset in the HDF5 file) and a display "label" (long name with units), as you may need the refer back to the base name in some places (tooltip, etc.) Same with axes.

Of course, this does have broader implications, as the term `label` seems to already be used in a few places (e.g. `_NXdataStackView.setData()`, `ArrayStackPlot`, etc.) But I don't necessarily think this normalisation is critical for the work at hand.

PR #4284, opened 2025-04-14
Draft: Improve combo

vallsv:
This PR is an attempt to solve visibility of cython function from mypy

### Sample code
```python
import numpy

arr = numpy.array([0, 1, 2, 3, 4, 5, 6])
r = min_max(arr)
print(r)


print(r.minimum + 1.0)

print(r.min_positive + 1.0)  # <- Error
```

### Result

```
$ mypy test_combo.py 
test_combo.py:11: error: Unsupported operand types for + ("None" and "float")  [operator]
test_combo.py:11: note: Left operand is of type "float | None"
Found 1 error in 1 file (checked 1 source file)
```

That's more or less what we expect.

Checklist:

<hr>

### DIFF
--- a/src/silx/image/marchingsquares/_mergeimpl.pyx
+++ b/src/silx/image/marchingsquares/_mergeimpl.pyx
@@ -669,7 +669,8 @@ cdef class _MarchingSquaresContours(_MarchingSquaresAlgorithm):
     @cython.cdivision(True)
     cdef void insert_segment(
         self, TileContext *context,
-        int x, int y,
+        int x,
+        int y,
         uint8_t begin_edge,
         uint8_t end_edge,
         double level
@@ -970,7 +971,8 @@ cdef class _MarchingSquaresPixels(_MarchingSquaresAlgorithm):
     @cython.cdivision(True)
     cdef void insert_segment(
         self, TileContext *context,
-        int x, int y,
+        int x,
+        int y,
         uint8_t begin_edge,
         uint8_t end_edge,
         double level
@@ -1153,7 +1155,8 @@ cdef class MarchingSquaresMergeImpl(object):
 
     def __init__(
         self,
-        image, mask=None,
+        image,
+        mask=None,
         group_size=256,
         use_minmax_cache=False
     ):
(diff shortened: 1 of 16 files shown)


vallsv:
Actually i found that projects https://github.com/Vizonex/CyStub, but i will not have time to integrate it in the build.

vallsv:
So in the end none of this projects was able to to generate something close to something good enough.

Maybe because we use an old syntax?

I will push what i have, but i am afraid only a small part of it could be merged.

src/silx/image/marchingsquares/_mergeimpl.pyx
```diff
@@ -1127,10 +1151,12 @@ cdef class MarchingSquaresMergeImpl(object):
     cdef _MarchingSquaresContours _contours_algo
     cdef _MarchingSquaresPixels _pixels_algo
 
-    def __init__(self,
-                 image, mask=None,
-                 group_size=256,
-                 use_minmax_cache=False):
+    def __init__(
+        self,
+        image, mask=None,
```
t20100, line 1156:
```suggestion
        image,
        mask=None,
```

src/silx/math/combo.pyi
```diff
@@ -0,0 +1,62 @@
...
+        """
+        Index of the strictly positive minimum value.
+
+        It is None if no value is strictly positive.
+        It is the index of the first occurrence.
+        """
+        ...
+
+    def __getitem__(self, key: int):
+        ...
```
t20100, line 54: Why is this needed? no return type?

src/silx/math/combo.pyx
```diff
@@ -74,34 +80,46 @@ class _MinMaxResult(object):
...
+        Index of the strictly positive minimum value.
 
         It is None if no value is strictly positive.
-        It is the index of the first occurrence.""")
+        It is the index of the first occurrence.
+        """
+        return self._argmin_positive
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: int):
```
t20100, line 122: missing return type

src/silx/math/combo.pyi
vallsv: For compatibility with the previous tuple version, i think

So return type is kind of pointless because it's the type of the arg1 or arg2. To make is fine you have to created overload functions, for something which is kind of deprecated and probably unused.

Not very useful

PR #4290, opened 2025-04-14
Draft: Fix 4120

payno:
Checklist:

<hr>


src/silx/gui/dialog/GroupDialog.py
```diff
@@ -164,6 +164,26 @@ def getSelectedDataUrl(self):
         """
         return self._selectedUrl
 
+    def setSelectedUrl(self, url: DataUrl):
```
payno: Today the url `data_path` can be anything: a path to a dataset or a group.
On the `DatasetDialog` especially we can select a group and not necessarily a `h5py.Dataset`. But this is the original implement of the class. And to me this can make sense.

src/silx/gui/dialog/GroupDialog.py
```diff
@@ -164,6 +164,26 @@ def getSelectedDataUrl(self):
...
+        """
+        Make the given url the selected URL. Raise an error if doesn't exists
+
+        :raises: ValueError if the url cannot be selected (file or data path not existing)
+        """
+        try:
+            self.addFile(path=url.file_path())
+        except OSError:
+            raise ValueError(f"File {url.file_path()} doesn't exists")
```
payno, line 176: If the file doesn't exists `addFile` will raise an `OSError`. Having the different use cases where a Url cannot be set make sense to me. But having an `OSError` would also make sense IMO

src/silx/gui/hdf5/Hdf5TreeView.py
```diff
@@ -277,3 +279,52 @@ def mousePressEvent(self, event):
         if event.button() != qt.Qt.LeftButton:
             qindex = self.indexAt(event.pos())
             self.clicked.emit(qindex)
+
+    def findHdf5Object(self, url: DataUrl) -> Hdf5Item | None:
```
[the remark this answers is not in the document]
t20100: Maybe best in the model? 🤷 
Anyway, sounds good to have an easy way to do so if Hdf5Item are exposed.

src/silx/gui/dialog/GroupDialog.py
payno: No the error will be raised if we cannot add the file cannot be added (so if doesn't exists most likely)

src/silx/gui/dialog/GroupDialog.py
```diff
@@ -164,6 +164,26 @@ def getSelectedDataUrl(self):
...
         return self._selectedUrl
 
+    def setSelectedDataUrl(self, url: DataUrl):
+        """
+        Make the given url the selected URL. Raise an error if doesn't exists
+
+        :raises: ValueError if the url cannot be selected (file or data path not existing)
+        """
+        try:
+            self.addFile(path=url.file_path())
```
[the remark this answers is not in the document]
payno: should be handled by fb9984b5c6e44a7840901b050f263d72f51b5096
Ignore request and add an info log message

src/silx/gui/dialog/GroupDialog.py
```diff
@@ -164,6 +164,26 @@ def getSelectedDataUrl(self):
...
+        except OSError:
+            raise ValueError(f"File {url.file_path()} doesn't exists")
+
+        h5Object = self._tree.findHdf5Object(url=url)
+
+        if h5Object is None:
+            raise ValueError(
```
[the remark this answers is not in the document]
payno: Fixed by b506949696f7b2f69ccf65300ae09f6364e83909

src/silx/gui/hdf5/Hdf5TreeView.py
```diff
@@ -277,3 +279,52 @@ def mousePressEvent(self, event):
...
+        """Return the Hdf5Object matching the url if exists in the model. Else None"""
+        model = self.findHdf5TreeModel()
+
+        # 1.0 find file name
+        file_name = url.file_path()
+
+        def find_node(start_index: qt.QModelIndex, name: str) -> qt.QModelIndex | None:
```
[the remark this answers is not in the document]
payno: I will change it. I am quiet a user of those :(

src/silx/gui/hdf5/Hdf5TreeView.py
```diff
@@ -277,3 +279,52 @@ def mousePressEvent(self, event):
...
+            start_index=model.index(0, 0), name=os.path.basename(file_name)
+        )
+        if start_index is None:
+            return None
+        node = model.nodeFromIndex(start_index)
+
+        # 2.0 find data path node
+        node_names = filter(None, url.data_path().split("/"))
+
+        def find_children(
```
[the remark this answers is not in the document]
payno: I will change it. I am quiet a fan of those (it gives a context)

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,56 @@ def parent(self, child):
...
+        return matching_items[0]
+
+    def _findChildren(
+        self, parent_node: Hdf5Item, child_node_name: str
+    ) -> Hdf5Item | None:
+        # TODO: FIXME: we should be able to use the generic 'model.match' but Hdf5Item is not inheriting from the default qt.QAbstractView
+        "cannot use the default 'search' function as the item is not inheriting from the default QTreeItem..."
+        for i in range(parent_node.childCount()):
+            if parent_node.child(i).basename == child_node_name:
+                return parent_node.child(i)
```
payno: As the `Hdf5Item` is not an instance of `qt.QAbstractView` the model cannot search on them. So we cannot use the `model.match` function.

src/silx/gui/hdf5/test/test_hdf5.py
```diff
@@ -457,6 +457,10 @@ def testInsert(self):
         self.assertEqual(self.listener.callCount(), 0)
 
     def testLoaded(self):
+        # the h5 file is already added. If already there then we need to remove it to add it back
+        self.model.clear()
+        self.listener.clear()
+
```
payno: We need to clear the model and the listener here because on the test class `setUp` the file is already added.

But now when calling `insertFile` if the file is already there we simply return and the `sigH5pyObjectLoaded` signal will not be emitted... the file is already loaded. From what I can this is shouldn't be an issue. Maybe better if @vallsv thinks of some cases where it can bring issues.

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,56 @@ def parent(self, child):
...
+        self, start_index: qt.QModelIndex, name: str
+    ) -> qt.QModelIndex | None:
+        matching_items = self.match(
+            start_index,
+            qt.Qt.DisplayRole,
+            name,
+        )
+        if len(matching_items) == 0:
+            return None
+        if len(matching_items) > 0:
```
t20100:
```suggestion
        if len(matching_items) > 1:
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,56 @@ def parent(self, child):
...
+                f"More than one item found matching {name}. Pick the first one"
+            )
+
+        return matching_items[0]
+
+    def _findChildren(
+        self, parent_node: Hdf5Item, child_node_name: str
+    ) -> Hdf5Item | None:
+        # TODO: FIXME: we should be able to use the generic 'model.match' but Hdf5Item is not inheriting from the default qt.QAbstractView
+        "cannot use the default 'search' function as the item is not inheriting from the default QTreeItem..."
```
t20100: This string seems a duplicate of the previous comment.
If kept, best to convert to a comment
```suggestion
        # cannot use the default 'search' function as the item is not inheriting from the default QTreeItem...
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,11 +771,34 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
     def __releaseRunner(self, runner):
         self.__runnerSet.remove(runner)
 
+    def _getFiles(self) -> tuple[str]:
```
t20100:
```suggestion
    def _getFiles(self) -> tuple[str, ...]:
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,11 +771,34 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
...
+        """Return the list of files open in the model"""
+        files = []
+        for index in range(self.rowCount()):
+            modelIndex = self.index(row=index, column=0)
```
t20100:
```suggestion
        for rowIndex in range(self.rowCount()):
            modelIndex = self.index(row=rowIndex, column=0)
```

🤷

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,11 +771,34 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
...
+            files.append(obj.file.filename)
+        return tuple(files)
+
+    def _clean_childless_nodes(self):
+        """
+        Remove any childless nodes
+        Use case: remove file nodes without at least one dataset
+        """
+        for index in range(self.rowCount()):
+            modelIndex = self.index(row=index, column=0)
```
t20100:
```suggestion
        for rowIndex in range(self.rowCount()):
            modelIndex = self.index(row=rowIndex, column=0)
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,11 +771,34 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
...
+            obj = self.data(modelIndex, Hdf5TreeModel.H5PY_OBJECT_ROLE)
+            files.append(obj.file.filename)
+        return tuple(files)
+
+    def _clean_childless_nodes(self):
+        """
+        Remove any childless nodes
+        Use case: remove file nodes without at least one dataset
```
t20100:
```suggestion
        """Remove any childless nodes

        Use case: remove file nodes without at least one dataset
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,56 @@ def parent(self, child):
     def nodeFromIndex(self, index):
         return index.internalPointer() if index.isValid() else self.__root
 
+    def _find_node(
+        self, start_index: qt.QModelIndex, name: str
```
t20100: Best to stick to CamelCase for consistency

```suggestion
    def _findNode(
        self, startIndex: qt.QModelIndex, name: str
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,11 +771,34 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
...
+                self.removeIndex(modelIndex)
+
     def insertFile(self, filename, row=-1):
         """Load a HDF5 file into the data model.
 
         :param filename: file path.
         """
+        if filename in self._getFiles():
+            _logger.info("Skip file insertion. File '%s' already exists", filename)
+            return
```
t20100: Best to not change the general behavior and check in the dialog if the file is there before adding it

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,11 +771,34 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
     def __releaseRunner(self, runner):
         self.__runnerSet.remove(runner)
 
+    def _getFiles(self) -> tuple[str]:
+        """Return the list of files open in the model"""
+        files = []
+        for index in range(self.rowCount()):
+            modelIndex = self.index(row=index, column=0)
+            obj = self.data(modelIndex, Hdf5TreeModel.H5PY_OBJECT_ROLE)
```
t20100:
```suggestion
            obj = self.data(modelIndex, self.H5PY_OBJECT_ROLE)
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,56 @@ def parent(self, child):
...
+    def _findChildren(
+        self, parent_node: Hdf5Item, child_node_name: str
+    ) -> Hdf5Item | None:
+        # TODO: FIXME: we should be able to use the generic 'model.match' but Hdf5Item is not inheriting from the default qt.QAbstractView
+        "cannot use the default 'search' function as the item is not inheriting from the default QTreeItem..."
+        for i in range(parent_node.childCount()):
+            if parent_node.child(i).basename == child_node_name:
+                return parent_node.child(i)
+
+    def findHdf5Object(self, url: DataUrl) -> Hdf5Item | None:
```
t20100:
```suggestion
    def findHdf5Item(self, url: DataUrl) -> Hdf5Item | None:
```

src/silx/gui/hdf5/Hdf5TreeModel.py
payno: :+1: 
Indeed this function has a `hits` attribute. It will return once finding this amount of hits. By default hits is 1.
I will remove the condition and set hits to 1 to ease understanding

src/silx/gui/hdf5/Hdf5TreeModel.py
payno: A discussed, ... will check if the file already exists instead of modifying the behavior of `insertFile`.

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -720,6 +766,26 @@ def insertFileAsync(self, filename, row=-1, synchronizingNode=None):
...
+    def _getFiles(self) -> tuple[str, ...]:
+        """Return the list of files open in the model"""
+        files = []
+        for rowIndex in range(self.rowCount()):
+            modelIndex = self.index(row=rowIndex, column=0)
+            obj = self.data(modelIndex, self.H5PY_OBJECT_ROLE)
+            files.append(obj.file.filename)
+        return tuple(files)
+
+    def _clean_childless_nodes(self):
```
t20100: CamelCase:

```suggestion
    def _cleanChildlessNodes(self):
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,51 @@ def parent(self, child):
...
+        matching_items = self.match(
+            startIndex,
+            qt.Qt.DisplayRole,
+            name,
+            hits=1,
+        )
+        if len(matching_items) == 0:
+            return None
+
+        return matching_items[0]
```
t20100:
```suggestion
        matchingItems = self.match(
            startIndex,
            qt.Qt.DisplayRole,
            name,
            hits=1,
        )
        if len(matchingItems) == 0:
            return None

        return matchingItems[0]
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,51 @@ def parent(self, child):
...
+
+    def _findChildren(
+        self, parent_node: Hdf5Item, child_node_name: str
+    ) -> Hdf5Item | None:
+        # TODO: FIXME: we should be able to use the generic 'model.match' but Hdf5Item is not inheriting from the default qt.QAbstractView
+        # cannot use the default 'search' function as the item is not inheriting from the default QTreeItem...
+        for i in range(parent_node.childCount()):
+            if parent_node.child(i).basename == child_node_name:
+                return parent_node.child(i)
```
t20100:
```suggestion
    def _findChildren(
        self, parentNode: Hdf5Item, childName: str
    ) -> Hdf5Item | None:
        # TODO: FIXME: we should be able to use the generic 'model.match' but Hdf5Item is not inheriting from the default qt.QAbstractView
        # cannot use the default 'search' function as the item is not inheriting from the default QTreeItem...
        for i in range(parentNode.childCount()):
            if parentNode.child(i).basename == childName:
                return parentNode.child(i)
```

src/silx/gui/hdf5/Hdf5TreeModel.py
```diff
@@ -565,6 +566,51 @@ def parent(self, child):
...
+        node = self.nodeFromIndex(startIndex)
+
+        # 2.0 find data path node
+        nodeNames = filter(None, url.data_path().split("/"))
+
+        for nodeName in nodeNames:
+            # find file name
+            if node is None:
+                return None
+            node = self._findChildren(parent_node=node, child_node_name=nodeName)
```
t20100:
```suggestion
            node = self._findChildren(parentNode=node, childName=nodeName)
```

src/silx/gui/dialog/GroupDialog.py
```diff
@@ -164,6 +164,30 @@ def getSelectedDataUrl(self):
...
+                self.addFile(path=url.file_path())
+            except OSError:
+                raise ValueError(f"File {url.file_path()} doesn't exists")
+
+        model = self._tree.findHdf5TreeModel()
+        h5Object = model.findHdf5Item(url=url)
+
+        if h5Object is None:
+            # clean orphan nodes (in the case the file was added but the creation of the h5object failed)
+            model._clean_childless_nodes()
```
t20100:
```suggestion
            model._cleanChildlessNodes()
```

src/silx/gui/dialog/GroupDialog.py
```diff
@@ -164,6 +164,30 @@ def getSelectedDataUrl(self):
         """
         return self._selectedUrl
 
+    def setSelectedDataUrl(self, url: DataUrl):
+        """
+        Make the given url the selected URL. Raise an error if doesn't exists
```
t20100: Duplicates of the :raise:

```suggestion
        Make the given url the selected URL.
```

:shrug:

PR #4296, opened 2025-04-15
silx.gui.widgets: provide more 'overlay' widgets.

### LINKED ISSUE #4216: Add a widget to display a message from the plot until no images loaded
It could be convenient to have a simple widget allowing to display a message (instruction for example)

Image

Here is the piece used for the screenshot and that could be used to start this:

```python

class _PlotMsg(qt.QLabel):
    def __init__(self, text: str, parent: PlotWidget):
        super().__init__(text, parent)
        self._registerParent(parent=parent)
    
    def _listenedWidget(self, parent: qt.QWidget) -> qt.QWidget:
        """Returns widget to register event filter to according to parent"""
        if isinstance(parent, PlotWidget):
            return parent.getWidgetHandle()
        return parent

    def _backendChanged(self):
        self._listenedWidget(self.parent()).installEventFilter(self)
        self._resizeLater()

    def _registerParent(self, parent: qt.QWidget | None):
        if parent is None:
            return
        self._listenedWidget(parent).installEventFilter(self)
        if isinstance(parent, PlotWidget):
            parent.sigBackendChanged.connect(self._backendChanged)
        self._resize()

    def _unregisterParent(self, parent: qt.QWidget | None):
        if parent is None:
            return
        if isinstance(parent, PlotWidget):
            parent.sigBackendChanged.disconnect(self._backendChanged)
        self._listenedWidget(parent).removeEventFilter(self)

    def setParent(self, parent: qt.QWidget):
        self._unregisterParent(self.parent())
        super().setParent(parent)
        self._registerParent(parent)

    def showEvent(self, event: qt.QShowEvent):
        super().showEvent(event)
        self.setVisible(True)

    def hideEvent(self, event: qt.QHideEvent):
        super().hideEvent(event)
        self.setVisible(False)

    def _resize(self):
        if not qt_inspect.isValid(self):
            return  # For _resizeLater in case the widget has been deleted

        parent = self.parent()
        if parent is None:
            return

        size = self.sizeHint()
        if isinstance(parent, PlotWidget):
            offset = parent.getWidgetHandle().mapTo(parent, qt.QPoint(0, 0))
            left, top, width, height = parent.getPlotBoundsInPixels()
            rect = qt.QRect(
                qt.QPoint(
                    int(offset.x() + left + width / 2 - size.width() / 2),
                    int(offset.y() + top + height / 2 - size.height() / 2),
                ),
                size,
            )
        else:
            position = parent.size()
            position = (position - size) / 2
            rect = qt.QRect(qt.QPoint(position.width(), position.height()), size)
        self.setGeometry(rect)
        self.raise_()

    def _resizeLater(self):
        qt.QTimer.singleShot(0, self._resize)

    def eventFilter(self, watched: qt.QWidget, event: qt.QEvent):
        if event.type() == qt.QEvent.Resize:
            self._resize()
            self._resizeLater()  # Defer resize for the receiver to have handled it
        return super().eventFilter(watched, event)
```

If can be called like
``` python
plot = Plot2D()
msg = "click 'load' to load the sinogram of the defined slice"
startUpMsg = _PlotMsg(msg, plot)
plot.show()
```

**warning**: this code is sharing 90% of the code with `silx.gui.widgets.WaitingOverlay`. The idea would be to have a base class or some inheritance there. Maybe it can be extended to other interesting use cases

payno:
* move all 'generic' code from `WaitingOverlay` to new class `_OverlayMixIn`
* Add two more use cases for 'overlay' widget: button (`ButtonOverlay`) and label (`LabelOverlay`)
* close #4216


src/silx/gui/widgets/WaitingPushButton.py
```diff
@@ -227,6 +227,10 @@ def isWaiting(self):
         :rtype: bool"""
         return self.__waiting
 
+    def close(self):
+        self.stopWaiting()
+        return super().close()
```
payno, line 232: I needed this to make test pass on ubuntu. Probably safer to have it on the class directly.

src/silx/gui/widgets/ButtonOverlay.py
```diff
@@ -0,0 +1,11 @@
+from silx.gui import qt
+
+from .OverlayMixIn import _OverlayMixIn
```
t20100, line 3: If the mix-in is private, then it's best if the module is private rather than the class:
```suggestion
from ._OverlayMixIn import OverlayMixIn as _OverlayMixIn
```

src/silx/gui/widgets/LabelOverlay.py
```diff
@@ -0,0 +1,11 @@
+from silx.gui import qt
+
+from .OverlayMixIn import _OverlayMixIn
```
t20100, line 3:
```suggestion
from ._OverlayMixIn import OverlayMixIn as _OverlayMixIn
```

src/silx/gui/widgets/WaitingOverlay.py
```diff
@@ -1,109 +1,26 @@
 from silx.gui.widgets.WaitingPushButton import WaitingPushButton
 from silx.gui import qt
-from silx.gui.qt import inspect as qt_inspect
-from silx.gui.plot import PlotWidget
+from .OverlayMixIn import _OverlayMixIn
```
t20100, line 3:
```suggestion
from ._OverlayMixIn import OverlayMixIn as _OverlayMixIn
```

src/silx/gui/widgets/WaitingOverlay.py
```diff
@@ -1,109 +1,26 @@
...
 
-    def text(self) -> str:
-        """Returns displayed text"""
-        return self._waitingButton.text()
-
     def setText(self, text: str):
         """Set displayed text"""
-        self._waitingButton.setText(text)
+        super().setText(text)
         self._resize()
```
t20100, line 26: Is it needed to keep resize here?

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -0,0 +1,83 @@
...
+        self._registerParent(parent)
+
+    def showEvent(self, event: qt.QShowEvent):
+        super().showEvent(event)
+        self.setVisible(True)
+
+    def hideEvent(self, event: qt.QHideEvent):
+        super().hideEvent(event)
+        self.setVisible(False)
+
```
t20100, line 49: This should not be needed:
```suggestion
```

t20100:
Would be good to check if it's in the doc and update it

examples/overlayWidgets.py
```diff
@@ -61,7 +55,7 @@ def __init__(self, parent=None):
 
     def _generateRandomData(self):
         self.setData(numpy.random.random(1000 * 500).reshape((1000, 500)))
-        self._button.setEnabled(True)
+        self._button.hide()
```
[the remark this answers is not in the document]
t20100: OK.
I missed the renaming, you then have to update the reference this the sample code in the doc.

src/silx/gui/widgets/WaitingOverlay.py
payno: Yes, else we have display issues like:

image

instead of 

image

t20100 (approved):
Thanks!

PR #4298, opened 2025-04-15
CI: use ubuntu-ci for linux test instead of ubuntu-20.04

### LINKED ISSUE #4206: Migrate CI and release to ubuntu-latest (24.04)
`ubuntu-latest` has been replaced by `ubuntu-22.04` to make the 2.2.0 release... but let's not keep this and switch to 24.04

payno:
Move CI to use ubuntu-latest

~~close #4206~~

### DIFF
--- a/.github/actions/setup-system/action.yml
+++ b/.github/actions/setup-system/action.yml
@@ -16,7 +16,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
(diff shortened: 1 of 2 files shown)


t20100:
What was proposed in #4206 was to change ubuntu-22.04 by 24.04, not 20.04 by 24.04 ;)
IMO it's good to keep 20.04 for a few more months.

BTW, the release workflow also needs to be updated (and the commented targets in ci.yml)

payno:
ok my bad sorry. But I just had error messages saying:
```
This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
```
so I guess we should also drop this one ?

payno:
And the 22.04 have been dropped recently (for the code camp). Those are the one you want to reactivate ?

PR #4300, opened 2025-04-15
ProfileManager: remove workaround for Qt5 < 5.14

payno:
I was not able to reproduce blinking with 'recent' PyQt5 version (5.15).

PyQt5 5.14 is form december 2019 and the most recent version of python it handle is 3.8... we are now supporting 3.10 or more recent. So let's go for the clean.

No issue met with PyQt6


t20100 (approved):
Sounds reasonable to me.

Can you check that the doc clearly mention the minimum supported version of pyqt5, and update it if needed?

PR #4301, opened 2025-04-15
RegionOfInterestManager: make sure drawnROI is dropped

### LINKED ISSUE #4229: ROI manager clear+setInteractiveMode segfault with py3.11
Feedback from Flint CI: https://gitlab.esrf.fr/bliss/bliss/-/issues/4660

The following sequence segfault.

```
roi_manager.clear()
roi_manager.setInteractiveMode(prevMode)
```

The `clear()` call drop the ROIs but do not invalidate `self._drawnROI`.

As result when `setInteractiveMode()` is called, it tries to drop `self._drawnROI` which is not a valid ROI anymore (not the right parent, not part of the ROIs).

It raise an exception `raise ValueError("RegionOfInterest does not belong to this instance")`.

Sounds like it was managed somehow with other python 3.9 and 3.10.

But with python 3.11 it segfault.

payno:
close #4229

### DIFF
--- a/src/silx/gui/plot/tools/roi.py
+++ b/src/silx/gui/plot/tools/roi.py
@@ -661,13 +661,16 @@ class RegionOfInterestManager(qt.QObject):
         :rtype: bool
         """
         if self.getRois():  # Something to reset
-            for roi in self._rois:
-                roi.sigRegionChanged.disconnect(self._regionOfInterestChanged)
-                roi.setParent(None)
-            self._rois = []
-            self._roisUpdated()
-            return True
-
+            try:
+                for roi in self._rois:
+                    roi.sigRegionChanged.disconnect(self._regionOfInterestChanged)
+                    roi.setParent(None)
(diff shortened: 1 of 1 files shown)


vallsv:
Do you think the try/finally is mandatory? I don't remember why i put that.

Else i would just put the clean up between this 2 lines, because `_roisUpdated` is a user callback.
```
                self._rois = []
                                              <--- here
                self._roisUpdated()
```

vallsv (approved):

payno:
> Do you think the try/finally is mandatory? I don't remember why i put that.
> 
> Else i would just put the clean up between this 2 lines, because `_roisUpdated` is a user callback.
> 
> ```
>                 self._rois = []
>                                               <--- here
>                 self._roisUpdated()
> ```

not at all. But if we have it for sure it will work. I think you were more aware about the issue and the time and safer way to handle it. So let's go for it.

PR #4302, opened 2025-04-15
silx.math.colormap: Added static typing

vallsv:
This MR static type the module for internal and external use.

For external use i think it is fine.

For internal use, there is still typing problem
- because of the use of mixin without abstract functions (i.e..a call to `apply()`)
- no interface for `_colormap`
- And some types could be more strict by handling side cases as exception (i.e. remove `tuple[None, None]`)

Checklist:

<hr>

### DIFF
--- a/src/silx/math/colormap.py
+++ b/src/silx/math/colormap.py
@@ -30,7 +30,7 @@ __date__ = "25/08/2021"
 
 import collections
 import numbers
-from typing import NamedTuple
+from typing import NamedTuple, Literal
 import warnings
 import numpy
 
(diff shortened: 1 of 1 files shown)


src/silx/math/colormap.py
```diff
@@ -489,18 +503,18 @@ def apply_colormap(
 
 class NormalizeResult(NamedTuple):
     data: numpy.ndarray
-    vmin: numbers.Number
-    vmax: numbers.Number
+    vmin: float
+    vmax: float
```
vallsv, line 507: Note that now the python base types are supposed to support `int < float`

The other problem could be with types like `numpy.float32`, but in this case we should use `numbers.Number` everywhere instead of `float`, or maybe `mypy` will just allow that at some point, if it's not already the case.

t20100 (approved):
I've added a commit with some docstring updates: https://github.com/silx-kit/silx/pull/4302/commits/8f3138362112956392050edaddae1402c855cdee (simpler than the suggestions).

LGTM

vallsv:
Note that they is ways to type the kind of ndarrays, that could be useful for the future.

PR #4305, opened 2025-04-16
silx.gui.colors: Added typing

vallsv:
To make it work with Flint + most of internal typing

Checklist:

<hr>

### DIFF
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -1037,7 +1037,7 @@ class Colormap(qt.QObject):
         return data
 
 
-_PREFERRED_COLORMAPS = None
+_PREFERRED_COLORMAPS: list[str] | None = None
 """
 Tuple of preferred colormap names accessed with :meth:`preferredColormaps`.
 """
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -887,10 +887,12 @@ class Colormap(qt.QObject):
         vmin, vmax = self.getColormapRange(reference)
 
         if isinstance(data, _Colormappable):  # Use item's data
-            data = data.getColormappedData(copy=False)
+            array = data.getColormappedData(copy=False)
+        else:
+            array = data
 
         return _colormap.cmap(
-            data, self._colors, vmin, vmax, self._getNormalizer(), self.__nanColor
+            array, self._colors, vmin, vmax, self._getNormalizer(), self.__nanColor
         )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


src/silx/gui/colors.py
```diff
@@ -275,7 +277,7 @@ def _getColormapAutoscaleRange(
         """
         raise NotImplementedError("This method must be implemented in subclass")
 
-    def getColormappedData(copy: bool = False) -> numpy.ndarray | None:
+    def getColormappedData(self, copy: bool = False) -> numpy.ndarray | None:
```
t20100, line 280: 😱

src/silx/gui/colors.py
```diff
@@ -828,6 +841,8 @@ def _setFromDict(self, dic: dict):
         if name is not None:
             self.setName(name)
         else:
+            # if name is None => colors is not None
+            assert colors is not None
```
t20100, line 845: Best to stop using asserts in code and raise appropriate exception:
```suggestion
            if colors is None:
                raise ValueError("name and colors cannot be both None")
```

src/silx/gui/colors.py
```diff
@@ -1059,6 +1076,10 @@ def preferredColormaps() -> tuple[str, ...]:
     if _PREFERRED_COLORMAPS is None:
         # Initialize preferred colormaps
         setPreferredColormaps(_DEFAULT_PREFERRED_COLORMAPS)
+
+    # setPreferredColormaps initialize _PREFERRED_COLORMAPS
+    assert _PREFERRED_COLORMAPS is not None
```
t20100, line 1081: There is already a `ValueError` in `setPreferredColormaps`, so I guess this is for type linter, otherwise it is not needed

```suggestion
    if _PREFERRED_COLORMAPS is None:
        raise RuntimeError("_PREFERRED_COLORMAPS should not be None")
```

t20100 (approved):
Best to change asserts to raise
otherwise LGTM

src/silx/gui/colors.py
vallsv: I disagree, assert is supposed to be an assert. No more no less than what it is in C or C++.

That's an extra check which is not mandatory.

src/silx/gui/colors.py
vallsv: Yes, that's for the type checking.

Actually there is no better solution, except reworking the code.

src/silx/gui/colors.py
t20100: Indeed, there is a `raise ValueError` above in the code

t20100 (approved):

PR #4306, opened 2025-04-17
yaxis arcsinh with matplotlib

EdgarGF93:
Checklist:

<hr>

### DIFF
--- a/src/silx/gui/plot/PlotWindow.py
+++ b/src/silx/gui/plot/PlotWindow.py
@@ -183,6 +183,12 @@ class PlotWindow(PlotWidget):
         self.yAxisLogarithmicAction.setVisible(logScale)
         self.addAction(self.yAxisLogarithmicAction)
 
+        self.yAxisArcsinhAction = self.group.addAction(
+            actions.control.YAxisArcsinhAction(self, parent=self)
+        )
+        self.yAxisArcsinhAction.setVisible(logScale)
+        self.addAction(self.yAxisArcsinhAction)
+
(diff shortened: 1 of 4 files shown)


vallsv:
Nice.

This could also be a good opportunity to rework the backend API with normalization as actually, to set linear, we have to set `setLog(False)`.

src/silx/gui/plot/PlotWindow.py
```diff
@@ -183,6 +183,12 @@ def __init__(
         self.yAxisLogarithmicAction.setVisible(logScale)
         self.addAction(self.yAxisLogarithmicAction)
 
+        self.yAxisArcsinhAction = self.group.addAction(
+            actions.control.YAxisArcsinhAction(self, parent=self)
+        )
+        self.yAxisArcsinhAction.setVisible(logScale)
+        self.addAction(self.yAxisArcsinhAction)
```
t20100, line 190: Adding a new axis scaling changes its control from a boolean to an enum (it should have been so since the beginning).
It would make sense to me to replace `yAxisLogarithmicAction` with a toolbutton with a drop-down.

t20100 (commented):
Let's OpenGL support for later.

I also propose to do this PR on adding arcsinh to PlotWidget with matplotlib and keep the QAction for a second one. 

Could you add arcsinh scale for x axis as well for consistency?

payno:
@EdgarGF93 would you have time to have a look at @t20100 review and update the (matplotlib backend) code or do you step down in favor of somone else ?

PR #4311, opened 2025-04-17
silx.sx/silx.test/silx.utils: Improved code to pass flake8, black, ruff tests

tj-eldridge:
Added # noqa: F401 to ignore valid imported modules which aren't used
Changed lines with length longer than 88 characters (according to black) to properly wrap
Fixed a typo in silx.utils.ExternalResources which may have caused bugs elsewhere
Refactored ambiguous variable "l" within silx.utils with a more descriptive name

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -60,8 +60,3 @@ silx.resources =
     opencl/codec/*.cl
     gui/colormaps/*.npy
 silx.examples = *.png
-
-[flake8]
-max-line-length = 88
-extend-ignore = E203, W503
-exclude = .venv,build,dist
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,4 +63,5 @@ silx.examples = *.png
 
 [flake8]
 max-line-length = 88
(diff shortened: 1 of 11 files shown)


setup.cfg
```diff
@@ -60,3 +60,8 @@ silx.resources =
     opencl/codec/*.cl
     gui/colormaps/*.npy
 silx.examples = *.png
+
+[flake8]
+max-line-length = 88
+extend-ignore = E203, W503
+exclude = .venv,build,dist
```
t20100, line 67: There's already a `.flake8` file to configure flake8 so this should not be needed.
Could you revert this?

On top of that, there is also an opened Pull Request to change the build system from setuptools to meson, so the `setup.cfg` file will be removed.

t20100 (commented):
Thanks for your PR!

I left a small comment regarding `setup.cfg`

t20100 (approved):
Thanks for the contribution!

Issue #4123, opened 2024-05-17
Add a 'auto' imagej like option


payno:
IT would be interesting to have (yet) another option for defining the colormap min/max values.

Closer to what imageJ is doing:

![image](https://github.com/silx-kit/silx/assets/12907796/12018315-67dd-4aa2-8621-ecd764a33462)

note: Suggested by Elodie

t20100:
Can you provide a link to the relevant doc, the screen shot is not readable

payno:
Yes sorry. Here is the link: https://imagej.net/ij/docs/guide/146-28.html

payno:
Indeed this is the 'percentile' colormap but with different values. We can add a 'saturation' spin box to the colormap to have this 'mode'

payno:
done in https://github.com/silx-kit/silx/pull/4379

PR #4312, opened 2025-04-17
colormap: Improve percentile mode

payno:
# Question

1. usage of '__cacheColormapRange' might be reconsidered. This was targetting another usage but might add complexity. If we keep it this was we must at least have a limited cache (10 elements ? 100 ?)
2. naming: I think @t20100 was not much in favor of using 'saturation'. Maybe percentile can be considered - more generic - or something else. Ideas welcome
3. The GUI side can be improved. But I don't have strong opinion on it. Should we allow a QLineEdit / QSpinBox to let the user see and define manually the saturation value ?

# video

[simple video demo](https://github.com/user-attachments/assets/4a935964-c76a-46af-981e-6e7df5665951)

### DIFF
--- a/examples/plotWidget.py
+++ b/examples/plotWidget.py
@@ -41,7 +41,7 @@ import numpy
 
 from silx.gui import qt
 
-from silx.gui.plot import PlotWidget
+from silx.gui.plot import Plot2D
 from silx.gui.plot import tools  # QToolbars to use with PlotWidget
 from silx.gui.plot import actions  # QAction to use with PlotWidget
 from silx.gui.plot import PlotToolButtons  # QToolButton to use with PlotWidget
@@ -57,7 +57,7 @@ class MyPlotWindow(qt.QMainWindow):
         super().__init__(parent)
 
         # Create a PlotWidget
-        self._plot = PlotWidget(parent=self)
+        self._plot = Plot2D(parent=self)
 
         # Create a colorbar linked with the PlotWidget
         colorBar = ColorBarWidget(parent=self, plot=self._plot)
--- a/examples/plotWidget.py
+++ b/examples/plotWidget.py
@@ -41,7 +41,7 @@ import numpy
 
 from silx.gui import qt
 
-from silx.gui.plot import PlotWindow
+from silx.gui.plot import PlotWidget
 from silx.gui.plot import tools  # QToolbars to use with PlotWidget
 from silx.gui.plot import actions  # QAction to use with PlotWidget
 from silx.gui.plot import PlotToolButtons  # QToolButton to use with PlotWidget
@@ -57,7 +57,7 @@ class MyPlotWindow(qt.QMainWindow):
         super().__init__(parent)
 
         # Create a PlotWidget
-        self._plot = PlotWindow(parent=self)
+        self._plot = PlotWidget(parent=self)
 
         # Create a colorbar linked with the PlotWidget
         colorBar = ColorBarWidget(parent=self, plot=self._plot)
--- a/examples/plotWidget.py
+++ b/examples/plotWidget.py
@@ -41,7 +41,7 @@ import numpy
 
 from silx.gui import qt
 
-from silx.gui.plot import Plot2D
+from silx.gui.plot import PlotWindow
 from silx.gui.plot import tools  # QToolbars to use with PlotWidget
 from silx.gui.plot import actions  # QAction to use with PlotWidget
 from silx.gui.plot import PlotToolButtons  # QToolButton to use with PlotWidget
@@ -57,7 +57,7 @@ class MyPlotWindow(qt.QMainWindow):
         super().__init__(parent)
 
         # Create a PlotWidget
-        self._plot = Plot2D(parent=self)
+        self._plot = PlotWindow(parent=self)
 
         # Create a colorbar linked with the PlotWidget
         colorBar = ColorBarWidget(parent=self, plot=self._plot)
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -333,10 +333,10 @@ class Colormap(qt.QObject):
     """constant for autoscale using mean +/- 3*std(data)
     with a clamp on min/max of the data"""
 
-    PERCENTILE_1_99 = "percentile_1_99"
-    """constant for autoscale using 1st and 99th percentile of data"""
+    PERCENTILE = "percentile"
+    """constant for autoscale using n'st and m'th percentile of data"""
 
-    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE_1_99)
+    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE)
     """Tuple of managed auto scale algorithms"""
 
     sigChanged = qt.Signal()
@@ -557,11 +557,11 @@ class Colormap(qt.QObject):
         return self.__gamma
 
     def getAutoscaleMode(self) -> str:
-        """Return the autoscale mode of the colormap ('minmax' or 'stddev3')"""
+        """Return the autoscale mode of the colormap. Possible values are ('minmax', 'stddev3', 'percentile')"""
         return self._autoscaleMode
 
     def setAutoscaleMode(self, mode: str):
-        """Set the autoscale mode: either 'minmax' or 'stddev3'
+        """Set the autoscale mode: either 'minmax', 'stddev3', 'percentile' or 'contrast_enhancer'
 
         :param mode: the mode to set
         """
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -344,6 +344,8 @@ class Colormap(qt.QObject):
 
     _DEFAULT_NAN_COLOR = 255, 255, 255, 0
 
+    DEFAULT_SATURATION: int = 2
+
     def __init__(
         self,
         name: str | None = None,
@@ -352,7 +354,6 @@ class Colormap(qt.QObject):
         vmin: float | None = None,
         vmax: float | None = None,
         autoscaleMode: str = MINMAX,
-        saturation: float = 0.0,
     ):
         qt.QObject.__init__(self)
         self._editable = True
@@ -392,11 +393,7 @@ class Colormap(qt.QObject):
         self._vmax = float(vmax) if vmax is not None else None
         self.__warnBadVmin = True
         self.__warnBadVmax = True
-        self._saturation: float = saturation
-
-    @property
-    def getSaturation(self) -> float:
-        return self._saturation
+        self._saturation: int = self._DEFAULT_NAN_COLOR
 
     def setFromColormap(self, other: Colormap):
         """Set this colormap using information from the `other` colormap.
@@ -546,14 +543,18 @@ class Colormap(qt.QObject):
             self.__warnBadVmax = True
             self.sigChanged.emit()
 
-    def getSaturation(self) -> float:
+    def getSaturation(self) -> int:
         """Colormap saturation in (0, 100)"""
         return self._saturation
 
-    def setSaturation(self, saturation: float):
-        if not (0.0 <= saturation <= 100.0):
+    def setSaturation(self, saturation: int):
+        if not isinstance(saturation, int):
+            raise TypeError(
+                f"saturation is expected to be an int. Got {type(saturation)}"
+            )
+        if not (0 <= saturation <= 100):
             raise ValueError(
-                f"saturation should be a float in [0.0, 100.0]. Got {saturation}"
+                f"saturation should be a float in [0, 100]. Got {saturation}"
             )
         if saturation != self._saturation:
             self._saturation = saturation
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -349,6 +349,9 @@ class Colormap(qt.QObject):
     _DEFAULT_NAN_COLOR = 255, 255, 255, 0
 
     _DEFAULT_SATURATION: int = 2
+    """Saturation value to be used with percentile mode.
+    We convert saturation to percentile by taking the percentile [saturation / 2, 100 - saturation / 2]
+    For example if saturation is 2 then we will take 1st and 99th percentiles."""
 
     def __init__(
         self,
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -421,7 +421,9 @@ class Colormap(qt.QObject):
                 assert colors is not None
                 self.setColormapLUT(colors)
             self.setNaNColor(other.getNaNColor())
-            self.setSaturationAutoscaleParameter(other.getSaturationAutoscaleParameter())
+            self.setSaturationAutoscaleParameter(
+                other.getSaturationAutoscaleParameter()
+            )
             self.setNormalization(other.getNormalization())
             self.setGammaNormalizationParameter(other.getGammaNormalizationParameter())
             self.setAutoscaleMode(other.getAutoscaleMode())
@@ -708,7 +710,9 @@ class Colormap(qt.QObject):
         :return: (vmin, vmax) range
         """
         return self._getNormalizer().autoscale(
-            data, mode=self.getAutoscaleMode(), saturation=self.getSaturationAutoscaleParameter()
+            data,
+            mode=self.getAutoscaleMode(),
+            saturation=self.getSaturationAutoscaleParameter(),
         )
 
     def getColormapRange(
@@ -751,7 +755,9 @@ class Colormap(qt.QObject):
                 max_ = normalizer.DEFAULT_RANGE[1] if max_ is None else max_
             else:
                 min_, max_ = normalizer.autoscale(
-                    data, mode=self.getAutoscaleMode(), saturation=self.getSaturationAutoscaleParameter()
+                    data,
+                    mode=self.getAutoscaleMode(),
+                    saturation=self.getSaturationAutoscaleParameter(),
                 )
 
         # Handle autoscale
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -400,7 +400,7 @@ class Colormap(qt.QObject):
         self._vmax = float(vmax) if vmax is not None else None
         self.__warnBadVmin = True
         self.__warnBadVmax = True
-        self._saturation: int = self.DEFAULT_SATURATION
+        self._saturation: int = self._DEFAULT_SATURATION
 
     def setFromColormap(self, other: Colormap):
         """Set this colormap using information from the `other` colormap.
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -421,7 +421,7 @@ class Colormap(qt.QObject):
                 assert colors is not None
                 self.setColormapLUT(colors)
             self.setNaNColor(other.getNaNColor())
-            self.setSaturation(other.getSaturation())
+            self.setSaturationAutoscaleParameter(other.getSaturationAutoscaleParameter())
             self.setNormalization(other.getNormalization())
             self.setGammaNormalizationParameter(other.getGammaNormalizationParameter())
             self.setAutoscaleMode(other.getAutoscaleMode())
@@ -550,11 +550,18 @@ class Colormap(qt.QObject):
             self.__warnBadVmax = True
             self.sigChanged.emit()
 
-    def getSaturation(self) -> int:
+    def getSaturationAutoscaleParameter(self) -> int:
         """Colormap saturation in (0, 100)"""
         return self._saturation
 
-    def setSaturation(self, saturation: int):
+    def setSaturationAutoscaleParameter(self, saturation: int):
+        """Set the 'saturation' parameter.
+
+        Only used for autoscale 'percentile' mode
+
+        :raise ValueError: If saturation is not valid
+        :raise TypeError: If saturation is not an integer
+        """
         if not isinstance(saturation, int):
             raise TypeError(
                 f"saturation is expected to be an int. Got {type(saturation)}"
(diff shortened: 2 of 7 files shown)


src/silx/gui/colors.py
```diff
@@ -350,6 +350,7 @@ def __init__(
         vmin: float | None = None,
         vmax: float | None = None,
         autoscaleMode: str = MINMAX,
+        saturation: float = 0.0,
```
payno, line 353: saturation is in percentage here. Maybe `saturation_percentage`

src/silx/gui/plot/items/core.py
```diff
@@ -658,9 +658,21 @@ def _setColormappedData(
         # Fill-up colormap range cache if values are provided
         if max_ is not None and numpy.isfinite(max_):
             if min_ is not None and numpy.isfinite(min_):
-                self.__cacheColormapRange[Colormap.LINEAR, Colormap.MINMAX] = min_, max_
+                if (Colormap.LINEAR, Colormap.MINMAX) not in self.__cacheColormapRange:
+                    self.__cacheColormapRange = {}
```
payno, line 662: This really the bad thing about the PR. We now need an extra level for the saturation.
Having `self.__cacheColormapRange[normalization, autoscalemode, saturation] ` would also make sense.
But in both cases the cache could quickly become heavy (we can limit it of course to n elements). But the calculation is also taking a bit of time...

src/silx/gui/plot/items/core.py
```diff
@@ -608,7 +608,7 @@ def __init__(self):
         self._colormap = Colormap()
         self._colormap.sigChanged.connect(self._colormapChanged)
         self.__data = None
-        self.__cacheColormapRange = {}  # Store {normalization: range}
+        self.__cacheColormapRange = {}  # Store {normalization: {saturation: range}}
```
payno, line 611: TODO: update doc

src/silx/gui/test/test_colors.py
```diff
@@ -657,21 +657,21 @@ def testAutoscaleRange(self):
...
-                Colormap.PERCENTILE_1_99,
+                Colormap.PERCENTILE,
                 numpy.array([10, 50, 100, float("inf")]),
                 (10.8, 99.0),
             ),
         ]
         for norm, mode, array, expectedRange in data:
             with self.subTest(norm=norm, mode=mode, array=array):
-                colormap = Colormap()
+                colormap = Colormap(saturation=2.0)
```
payno, line 674: Something I don't like either. Having to provide the saturation in the constructor. I could also define a default value and provide this default value in the constructor.
But having it here can let the user think this is used for all the modes and not only for the 'percentile' mode.

vallsv: > But having it here can let the user think this is used for all the modes...

I think you just have to call it `autoscalePercentilSaturation` and it's done.

The saturation used for the percentil autoscale.

You have the exact same problem for the get/set anyway.

But i don't see why you use a default value with `0.0`. I guess this should be `1.0` or `99.0`, else it's like having a min/max no?

vallsv: Also maybe you should use `sigma` instead of `saturation`.

I feel like `saturation` is a user feeling. We could use this label in the colormap, but here we deal a normalized math function.

src/silx/gui/plot/items/core.py
```diff
@@ -655,17 +655,17 @@ def _setColormappedData(
...
 
+        colormap = self.getColormap()
         # Fill-up colormap range cache if values are provided
         if max_ is not None and numpy.isfinite(max_):
             if min_ is not None and numpy.isfinite(min_):
-                self.__cacheColormapRange[Colormap.LINEAR, Colormap.MINMAX] = min_, max_
+                self.__cacheColormapRange[Colormap.LINEAR, Colormap.MINMAX, colormap.getSaturation()] = (min_, max_)
             if minPositive is not None and numpy.isfinite(minPositive):
-                self.__cacheColormapRange[Colormap.LOGARITHM, Colormap.MINMAX] = (
+                self.__cacheColormapRange[Colormap.LOGARITHM, Colormap.MINMAX, colormap.getSaturation()] = (
```
vallsv, line 664: If it is about to cache min and max for `MINMAX` autoscale, you should not index it with `getSaturation()`. You better to put `None` or `0.0` -> Changing the saturation will not change the cached `MINMAX`.

src/silx/gui/plot/items/core.py
```diff
@@ -698,7 +698,7 @@ def _getColormapAutoscaleRange(self, colormap=None):
 
         normalization = colormap.getNormalization()
         autoscaleMode = colormap.getAutoscaleMode()
-        key = normalization, autoscaleMode
+        key = normalization, autoscaleMode, colormap.getSaturation()
```
vallsv, line 701: Same, `getSaturation()` only have a meaning here if the mode is `PERCENTILE`.

vallsv:
Looks really interesting.
I am curious to see it live.

src/silx/gui/plot/items/core.py
vallsv: Yes it's true,you have to use a LRU cache.

If it's not already there you can add `lru-dict` to the dependencies.

Also, in this case i think you better to have 2 caches.

- One with the fixed values (the actual one)
- And the one with the infinite amount of values which have to be the LRU cache.

This way the fixed `MINMAX` values will not be dropped. I think that's important at some places. I can be wrong.

Another solution could be to use integer only for the saturation (do you really care about the difference between 5.0 and 5.1?)

payno: IMO we can leave with integers. This is the API provided by the GUI

src/silx/gui/test/test_colors.py
payno: Yes, but maybe this saturation could be used somewhere else. After some reflection IMO the best would be not to have it in the constructor paramters and have a clear details on the getter and setter so the users knows.
This was also why the default value is 0.0 instead of 2.0. If the value is two then we will have percentiles 1.0 and 99.0 (as it was originally.). But having it to 2.0 was weird IMO in the constructor. This is why removing it from the constructor seems better.

We could also define the 'saturation' as percentiles and have (1.0, 99.0). More generic but also bring complexity. I don't think anyone want to define 'imbalanced' percentiles at the moment (like 1, 80)

About the naming I think @t20100 had a close feeling. Maybe we can just go for `percentiles_sigma` or something else.  Not a bit fan of `sigma`

payno:
> Also another question. Does it really do what we need?
> 
> I mean the imagej feature was a static computation right?

Good question, indeed imajej from what I see saw to be a static computation but with a default value. Nevertheless the code let me think that the option of fine tuning this value is here.

> 
> Does it make more sens to parameterize the percentile than the standard deviation?

This I don't know. I would say both make sense but I don't know what are the usual 'habits' there.

examples/plotWidget.py
```diff
@@ -41,7 +41,7 @@
 
 from silx.gui import qt
 
-from silx.gui.plot import PlotWidget
```
t20100, line 44: Why switching to `PlotWindow` while the name of the example is `plotWidget`?

src/silx/gui/colors.py
```diff
@@ -331,17 +331,19 @@ class Colormap(qt.QObject):
     """constant for autoscale using mean +/- 3*std(data)
     with a clamp on min/max of the data"""
 
-    PERCENTILE_1_99 = "percentile_1_99"
-    """constant for autoscale using 1st and 99th percentile of data"""
+    PERCENTILE = "percentile"
```
t20100, line 334: That's an API break, we need to keep `PERCENTILE_1_99` for backward compatibility and add a new one here.

src/silx/gui/colors.py
```diff
@@ -331,17 +331,19 @@ class Colormap(qt.QObject):
...
-    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE_1_99)
+    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE)
     """Tuple of managed auto scale algorithms"""
 
     sigChanged = qt.Signal()
     """Signal emitted when the colormap has changed."""
 
     _DEFAULT_NAN_COLOR = 255, 255, 255, 0
 
+    DEFAULT_SATURATION: int = 2
```
t20100, line 345: Would be good to have a docstring here.

Can this be private `_DEFAULT_SATURATION`?

src/silx/gui/colors.py
```diff
@@ -389,6 +391,7 @@ def __init__(
         self._vmax = float(vmax) if vmax is not None else None
         self.__warnBadVmin = True
         self.__warnBadVmax = True
+        self._saturation: int = self._DEFAULT_NAN_COLOR
```
t20100, line 394: looks like a typo:
```suggestion
        self._saturation: int = self.DEFAULT_SATURATION
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -974,6 +974,12 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         self._autoButtons = _AutoScaleButton(self)
         self._autoButtons.autoRangeChanged.connect(self._autoRangeButtonsUpdated)
 
+        self._saturation = qt.QSlider(qt.Qt.Horizontal, self)
```
t20100, line 977: `_saturation` -> `_saturationSlider` ?

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1723,9 +1736,28 @@ def _autoscaleModeUpdated(self):
...
+                    colormap.setSaturation(Colormap.DEFAULT_SATURATION)
                 colormap.setAutoscaleMode(mode)
 
+        self._updateSaturationVisibility()
+        self._updateWidgetRange()
+
+    def _updateSaturationVisibility(self):
+        activate_saturation = self._autoScaleCombo.currentText() == "Percentile"
+        self._saturation.setVisible(activate_saturation)
+        self._saturationLabel.setVisible(activate_saturation)
```
t20100, line 1753: IMO best to enable/disable rather than show/hide widgets.
BTW, you can use `setEnabled` on the layout, so that can be controlled once for both the label and slide.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -974,6 +974,12 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         self._autoButtons = _AutoScaleButton(self)
         self._autoButtons.autoRangeChanged.connect(self._autoRangeButtonsUpdated)
 
+        self._saturation = qt.QSlider(qt.Qt.Horizontal, self)
+        self._saturation.setTickPosition(qt.QSlider.TicksBelow)
+        self._saturation.setRange(0, 100)
+        self._saturation.setValue(2)  # 2 <=> 1-99 percentile mode
+        self._saturation.valueChanged.connect(self._saturationChanged)
```
t20100, line 981: Probably good to disable tracking here, or it's going to lag

https://doc.qt.io/qt-6/qabstractslider.html#tracking-prop

src/silx/math/colormap.py
```diff
@@ -515,11 +522,12 @@ class NormalizeResult(NamedTuple):
...
-    vmin: float | None = None,
-    vmax: float | None = None,
-    gamma: float = 1.0,
-) -> NormalizeResult:
+    autoscale: str = "minmax",
+    vmin=None,
+    vmax=None,
+    gamma=1.0,
+    saturation: int = 0,
+):
```
t20100, line 530: Typing has been added in the mean time, can you take it into account

src/silx/math/colormap.py
```diff
@@ -515,11 +522,12 @@ class NormalizeResult(NamedTuple):
...
-    autoscale: AutoScaleModeType = "minmax",
-    vmin: float | None = None,
-    vmax: float | None = None,
-    gamma: float = 1.0,
-) -> NormalizeResult:
+    autoscale: str = "minmax",
+    vmin=None,
+    vmax=None,
+    gamma=1.0,
+    saturation: int = 0,
```
t20100, line 529: Saturation could be a float, right?
I'm wondering if we would need a "generic" way to provide additional args (gamma & saturation).
Or maybe force them to be keyword only arguments

src/silx/math/colormap.py
```diff
@@ -473,10 +476,11 @@ def apply_colormap(
...
-    autoscale: AutoScaleModeType = "minmax",
-    vmin: float | None = None,
-    vmax: float | None = None,
-    gamma: float = 1.0,
+    autoscale: str = "minmax",
+    vmin=None,
+    vmax=None,
+    gamma=1.0,
+    saturation: int = 0,
 ):
```
t20100, line 484: Can you take added typing into account?

src/silx/math/colormap.py
```diff
@@ -449,16 +456,12 @@ def _get_normalizer(norm: str, gamma: float) -> _NormalizationMixIn:
...
 
 
-def _get_range(
-    normalizer,
-    data: numpy.ndarray,
-    autoscale: AutoScaleModeType,
-    vmin: float | None,
-    vmax: float | None,
-) -> tuple[float, float]:
+def _get_range(normalizer, data, autoscale, vmin, vmax, saturation: float = 0.0):
```
t20100, line 459: Can you take added typing into account?

src/silx/math/colormap.py
```diff
@@ -235,13 +235,12 @@ def is_valid(
         else:
             return True
 
-    def autoscale(
-        self, data: numpy.ndarray | None, mode: AutoScaleModeType
-    ) -> tuple[float, float]:
+    def autoscale(self, data, mode, saturation: float = 0.0):
```
t20100, line 238: missing typing

src/silx/math/colormap.py
```diff
@@ -324,20 +322,29 @@ def autoscale_mean3std(
...
+        self, data: numpy.ndarray, saturation: int = 2
     ) -> tuple[float, float] | tuple[None, None]:
-        """Autoscale using [1st, 99th] percentiles
+        """Autoscale using percentiles
 
         :param data: The data to process
+        :param saturation: in [0, 100] 'saturation' of the image that will impact percentiles as [saturation/2.0, 100-saturation/2.0]
         :returns: (vmin, vmax)
         """
+        if not isinstance(saturation, int):
```
t20100, line 334: can't it be a float?

src/silx/gui/plot3d/_model/items.py
```diff
@@ -965,6 +965,8 @@ def __init__(self, item):
         self.addRow(_ColormapBoundRow(item, name="Min.", index=0))
         self.addRow(_ColormapBoundRow(item, name="Max.", index=1))
 
+        # TODO: add option of the saturation for 3D
+
```
t20100, line 969: Best to open an issue for it IMO

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1723,9 +1736,28 @@ def _autoscaleModeUpdated(self):
...
+                if activate_saturation:
+                    colormap.setSaturation(self._saturation.value())
+                else:
+                    colormap.setSaturation(Colormap.DEFAULT_SATURATION)
                 colormap.setAutoscaleMode(mode)
 
+        self._updateSaturationVisibility()
+        self._updateWidgetRange()
+
+    def _updateSaturationVisibility(self):
```
t20100, line 1750: Why is this called at multiple places, can't it be connected directly to self._autoScaleCombo's currentTextChanged: https://doc.qt.io/qt-6/qcombobox.html#currentTextChanged

src/silx/gui/colors.py
```diff
@@ -534,6 +538,23 @@ def setNormalization(self, norm: str):
             self.__warnBadVmax = True
             self.sigChanged.emit()
 
+    def getSaturation(self) -> int:
+        """Colormap saturation in (0, 100)"""
+        return self._saturation
+
+    def setSaturation(self, saturation: int):
```
t20100, line 545: Maybe add a docstring.
`saturation` name is not explicitly linked to the percentile mode.

What about an explicit name? e.g. `setPercentileAutoscaleSaturation` or `setPercentileAutoscaleParameter` like `setGammaNormalizationParameter`

src/silx/gui/dialog/ColormapDialog.py
t20100: What about using a QSpinBox instead?
Issue with silder is that the value is not visible.

A tooltip would be good as well since it is not necessarily obvious

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1063,6 +1070,11 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         layoutScale.addWidget(self._autoScaleCombo)
         layoutScale.addStretch()
 
+        layoutSaturation = qt.QFormLayout()
```
t20100, line 1073: How does this layout stretch horizontally?
What about combining this with `layoutScale` to avoid extra margins?

src/silx/gui/test/test_colors.py
vallsv: > this saturation could be used somewhere else

I think it's a very bad idea. That's nothing generic, it's a specific param which have a specific meaning in a specific case.

Or you have to define now what it means in a generic way.

The only possibility i see is to have something shared with the `stddev`. But you have to find a quantity which match. I don't think it's obvious, and if it's not obvious, it will not have overlap in term of user interaction.

contructor param/getter/setter is the same. If you put a getter/setter here you have to put a constructor param. That's the way we use this object. And mostly nobody creates this object.

src/silx/gui/dialog/ColormapDialog.py
vallsv: To me it's a detail which can be improved later.

But for a user point of view that's something you want to play with manually.

You can't play with a spinbox, it's boring.

t20100: Yes, but you also want to see the value somehow

examples/plotWidget.py
payno: My bad, this is a leftovers from some test :(

src/silx/gui/dialog/ColormapDialog.py
payno: I think I agree with both of you. So I think I will go for slider + ? QSpinBox ? or QLineEdit ...

src/silx/gui/dialog/ColormapDialog.py
payno: :ok_hand: ;  for enabling / disabling from the layout it was not working on my side. I guess I am missing something byut what...

src/silx/gui/dialog/ColormapDialog.py
payno: Agree but then to come back on the previous discussion the slider is also loosing of his interest...

src/silx/math/colormap.py
payno: No more as we have switch it to int to have simpler handling of the `__cacheColormapRange`.

For the second part to be honest I don't have a strong opinion.

src/silx/math/colormap.py
payno: I really don't get how I could have miss this one. Did you did some rebase at one point ?

vallsv: In between we have merged stuff in the main branch.

You just have to rebase i think. You will have conflicts.

payno:
> An idea, it's maybe good here to differentiate what the user controls (a single value with the slider) and what the API provides: a 2-tuple of the percentile to use (low, high). I don't expect passing one value or a 2-tuple to change the complexity of the code and it provides a full control over the percentile normalization parameters.

I agree. This is also related to the naming. I think we need to choose between:

* Option A: 'Keep saturation' as an integer (and as API parameter).
    * Pros:
        * It keeps coherence between the GUI and the underlying code.
        * It avoids to handle more complex use cases (one int vs two floats).
    * Cons
        * we are not fully benefiting from the percentile (100 values vs an infinity)
* Option B: moving from saturation to 'percentile'. Then it becomes more generic. But it add complexity: 
    * Cons:
        * add complexity
            * colormap vrange cache needs to become a lru_cache
            * there is some incoherence between the GUI (expecting a single int) and the code (expecting a tuple of float)...
                * users cannot define a tuple of float. And I don't really see how to do this from a friendly way. We can use the [Rangeslider](https://www.silx.org/doc/silx/latest/modules/gui/widgets/rangeslider.html) but I don't think this will be as convenient as the slider.
                * More important: I am not sure there is a use case (1) for handling so much fine tuning
    * Pros
        * more generic

At this stage, the decision likely depends on our priorities: generality versus simplicity. Ultimately, the choice should be driven by the specific use case we are trying to address (1).

src/silx/gui/colors.py
payno: For now I renamed it 'setSaturationAutoscaleParameter'.
Because at this stage this is an int when percentil is expected to be a tuple of float.

Might be renamed again if at the end we go more for 'percentile' (tuple of float) vs 'saturation' (single int)

vallsv:
I think we have to have a meeting, else you will wast too much time.

As the technical aspect looks to be working, i am interested in improving the UI on top.

Still, I think it's an interesting feature, but i am not sure it is about to answer to #4123

t20100:
Agreed, best to discuss it together.

I would currently go for a 2-tuple in the low level API and keep the saturation slider in the GUI.
For me, the `__cacheColormapRange` is an implementation detail, it should not impact what we do, we'll adapt it as needed.

payno:
After a discussion all together we agreed on the fact on simply adding a 'unique' mode that mimics the percentiles value of imagej (saturation of 0.35 -> percentiles 17.5 - 82.5)

payno:
Close to the benefit of https://github.com/silx-kit/silx/pull/4341

PR #4319, opened 2025-04-29
Tests: replace `--low-mem` arg by `--no-high-mem` and skip high memory tests by default

### LINKED ISSUE #4285: Tests: invert the `--low-mem` option logic
When running `pytest`, some collected tests take a **huge** amount of memory (ex: `test_histo_big_array` in `test_histogramnd_vs_np.py`. Enough to make the process crash.

These can be disabled by passing the `--low-mem` flag but I could not find any mention of this in the CONTRIBUTING guide ? I would therefore suggest to invert the logic of this flag. **This means not running the tests that requires high memory unless explicitly asked.**

loichuder:
Checklist:

<hr>

Fix #4285 

Argument name and associated env variables were changed:
- `--low-mem` becomes `--no-high-mem`
- `SILX_TEST_LOW_MEM` becomes `WITH_HIGH_MEM_TEST` 

I tried to be consistent with other options: `--no-opengl` /`WITH_GL_TEST`, `--no-gui/`WITH_QT_TEST` but there is also `--no-opencl`/`SILX_OPENCL`. 

Should I keep the `SILX` prefix for the env variable ?

The logic was also inverted so that high memory tests are skipped unless the env variable is set.

### DIFF
--- a/src/silx/conftest.py
+++ b/src/silx/conftest.py
@@ -55,11 +55,11 @@ def pytest_addoption(parser):
         help="Disable the test of the OpenCL part",
     )
     parser.addoption(
-        "--low-mem",
-        dest="low_mem",
-        default=False,
-        action="store_true",
-        help="Disable test with large memory consumption (>100Mbyte",
+        "--no-high-mem",
+        dest="high_mem",
+        default=True,
+        action="store_false",
+        help="Disable tests with large memory consumption (>100Mbytes)",
     )
 
 
@@ -112,10 +112,10 @@ def use_opencl(test_options):
 def use_large_memory(test_options):
     """Fixture to flag test using a large memory consumption.
 
-    This can be skipped with `--low-mem`.
+    This can be skipped with `--no-high-mem`.
     """
-    if test_options.TEST_LOW_MEM:
-        pytest.skip(test_options.TEST_LOW_MEM_REASON, allow_module_level=True)
+    if not test_options.WITH_HIGH_MEM_TEST:
+        pytest.skip(test_options.WITH_HIGH_MEM_TEST_REASON, allow_module_level=True)
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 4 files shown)


pierrepaleo:
Looks good, thanks. Tested on my machine and it behaves as expected :+1: 

The env variable name `WITH_HIGH_MEM_TEST` is probably fine. The others are `WITH_QT_TEST`, `SILX_OPENCL`, and `WITH_GL_TEST` ; which is almost consistent.

src/silx/test/utils.py
```diff
@@ -118,12 +118,12 @@ def configure(self, parsed_options=None):
...
-            self.TEST_LOW_MEM_REASON = "Skipped by command line"
-        elif os.environ.get("SILX_TEST_LOW_MEM", "True") == "False":
-            self.TEST_LOW_MEM = True
-            self.TEST_LOW_MEM_REASON = "Skipped by SILX_TEST_LOW_MEM env var"
+        if parsed_options is not None and not parsed_options.high_mem:
+            self.WITH_HIGH_MEM_TEST = False
+            self.WITH_HIGH_MEM_TEST_REASON = "Skipped by command line"
+        elif os.environ.get("WITH_HIGH_MEM_TEST") == "False":
+            self.WITH_HIGH_MEM_TEST = False
+            self.WITH_HIGH_MEM_TEST_REASON = "Skipped by WITH_HIGH_MEM_TEST env var"
```
t20100, line 126: I may miss something here, but it looks you can never set `self.WITH_HIGH_MEM_TEST` to `True`.

src/silx/conftest.py
```diff
@@ -55,11 +55,11 @@ def pytest_addoption(parser):
...
-        "--low-mem",
-        dest="low_mem",
-        default=False,
-        action="store_true",
-        help="Disable test with large memory consumption (>100Mbyte",
+        "--no-high-mem",
+        dest="high_mem",
+        default=True,
+        action="store_false",
+        help="Disable tests with large memory consumption (>100Mbytes)",
```
t20100, line 62: I may miss something, but shouldn't this be revert and have a `--high-mem` options that `store_true` instead of a `False` default?

PR #4320, opened 2025-05-15
Build: Updated meson build system

t20100:
Checklist:

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

This PR is a follow-up of #4277 to:
- Use `install_subdir` instead of explicit list wherever possible [as in numpy](https://github.com/numpy/numpy/blob/d936ca00751d55349d7304d67dfa9b40455e89b7/numpy/meson.build#L365-L370) and update `check_meson.py` accordingly.
- Add `silx.examples` which was missed in #4277

PR #4323, opened 2025-05-16
`silx.gui.widgets`: Added `FilenameCompleter` that provides autocompletion for file paths

loichuder:
Checklist:

Needs #4320 to integrate the new file in the meson build

### DIFF
--- a/src/silx/gui/widgets/FilenameCompleter.py
+++ b/src/silx/gui/widgets/FilenameCompleter.py
@@ -1,0 +1,22 @@
+from .. import qt
+
+
+class FilenameCompleter(qt.QCompleter):
+    """
+    A QCompleter that provides autocompletion for file paths.
+    """
+
+    def __init__(
+        self, parent: qt.QWidget | None = None, root: str | None = None
+    ) -> None:
+        super().__init__(parent=parent)
+
+        completer = qt.QCompleter()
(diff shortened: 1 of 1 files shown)


src/silx/gui/widgets/FilenameCompleter.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(
+        self, parent: qt.QWidget | None = None, root: str | None = None
+    ) -> None:
+        super().__init__(parent=parent)
+
+        completer = qt.QCompleter()
+        model = qt.QFileSystemModel(completer)
+        model.setOption(qt.QFileSystemModel.Option.DontWatchForChanges, True)
+        if root is None:
+            model.setRootPath(Path(__file__).root)
```
t20100, line 20: Maybe good to check how it behaves on Windows

t20100 (approved):

src/silx/gui/widgets/FilenameCompleter.py
loichuder: Yes. [anchor](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.anchor) might be a better default

loichuder: Tested on Windows and `/` is actually a sensible working default

PR #4324, opened 2025-05-16
silx.resources: Fixed not existing resource file access

t20100:
Checklist:

<hr>

This PR fixes an issue spotted with meson editable install (but actually not related to it): silx was calling `importlib.resources.as_file` on a non-existing resource file and this raises an exception in this case.
It works with the default resource management from Python, but we are miss-using the `as_file`.

For later, we should consider relying on `importlib.resources` directly now that it's standard rather than wrapping it in `silx.resources`.

### DIFF
--- a/src/silx/resources/__init__.py
+++ b/src/silx/resources/__init__.py
@@ -67,6 +67,7 @@ import os
 import sys
 from typing import NamedTuple
 
+import importlib
 import importlib.resources as importlib_resources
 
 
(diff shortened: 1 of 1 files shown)


src/silx/resources/__init__.py
```diff
@@ -237,9 +237,16 @@ def _get_resource_filename(package: str, resource: str) -> str:
...
     # Caching prevents extracting the resource twice
-    file_context = importlib_resources.as_file(
-        importlib_resources.files(package) / resource
-    )
+    traversable = importlib_resources.files(package).joinpath(resource)
+    if not traversable.is_file() and not traversable.is_dir():
+        package_dir_context = importlib_resources.as_file(
+            importlib_resources.files(package)
+        )
+        path = _file_manager.enter_context(package_dir_context)
```
t20100: This needs to be reworked to avoid extracting all resources for a non-existing file path..

t20100: Changed the way to handle not existing resources

PR #4330, opened 2025-05-19
silx.test.run_tests: Updated usage of pytest

t20100:
Checklist:

<hr>

Merge PR #4320 first!

This PR reworks the config of pytest regarding warnings.
It moves the configuration of warning raising an error/being ignored from `silx.test.run_tests` to the project's `conftest.py` by using the [`pytest_collection_modifyitems` hook](https://docs.pytest.org/en/latest/reference/reference.html#std-hook-pytest_collection_modifyitems).
This allows using the same rules no matter how the tests are run (`pytest` or `silx.test.run_tests`).

Relevant commits:
https://github.com/silx-kit/silx/commit/1c4062b42d5aaf8b3b51cdee6039158daba48e95
https://github.com/silx-kit/silx/commit/382b690e96a0280d2fa52b925c3fe56c87822eae
https://github.com/silx-kit/silx/commit/312a83bbb0efb37fd1395d4e4b08fa93cb0d9b9e


t20100:
This includes a fix from #4343 (thanks @payno for the investigation!) and PR #4324.
It fixes running tests locally and looks to fix CI as well

src/silx/gui/_glutils/gl.py
```diff
@@ -184,6 +186,14 @@ def glGetActiveAttrib(program, index):
...
 
 
+def glGetActiveUniform(program, index):
+    """Wrap PyOpenGL glGetActiveUniform"""
+    name, size, type_ = _GL.glGetActiveUniform(program, index)
+    if isinstance(name, np.ndarray):
+        name = name.tobytes().rstrip(b"\000")
+    return name, size, type_
+
+
```
t20100, line 196: There was already a wrapper for `glGetActiveAttrib` (I don't recall why...)
Let's have the workaround for `glGetActiveUniform` at a single place.

src/silx/gui/plot/backends/glutils/GLPlotCurve.py
```diff
@@ -1218,7 +1218,10 @@ def __init__(
         # Handle data offset
         if xData.itemsize > 4 or yData.itemsize > 4:  # Use normalization
             # offset data, do not offset error as it is relative
-            self.offset = self.xMin, self.yMin
+            self.offset = (
+                self.xMin if numpy.isfinite(self.xMin) else 0.0,
+                self.yMin if numpy.isfinite(self.yMin) else 0.0,
+            )
```
t20100, line 1224: Avoid non-finite here to avoid numpy.dot warnings later

src/silx/resources/__init__.py
```diff
@@ -237,9 +238,13 @@ def _get_resource_filename(package: str, resource: str) -> str:
...
     # Caching prevents extracting the resource twice
-    file_context = importlib_resources.as_file(
-        importlib_resources.files(package) / resource
-    )
+    traversable = importlib_resources.files(package).joinpath(resource)
+    if not traversable.is_file() and not traversable.is_dir():
+        module = importlib.import_module(package)
+        return os.path.join(os.path.dirname(module.__file__), resource)
+
+    file_context = importlib_resources.as_file(traversable)
```
t20100, line 246: This is to work with meson "editable" mode

src/silx/test/__init__.py
```diff
@@ -43,48 +43,45 @@
...
             imported_module = importlib.import_module(module)
-            list_path.append(
-                imported_module.__path__[0]
-                if hasattr(imported_module, "__path__")
-                else imported_module.__file__
-                )
+            if getattr(imported_module, "__path__", None):
+                module_path = imported_module.__path__[0]
+            elif os.path.basename(imported_module.__file__) == "__init__.py":
+                module_path = os.path.dirname(imported_module.__file__)
```
t20100, line 78: Add special case for a package with no `__path__`

src/silx/gui/plot/backends/glutils/GLPlotCurve.py
payno: Do we expect to have infinite at some point ? Maybe in this case we could display a warning (if this is not too much costly to do)
Not a big fan but to be honest I don't see how we could do something else.
A short note on how it could have `inf` values would be great :smile:

t20100: There should be no `inf`... and anyway it's only a warning. It makes the CI fails but in normal condition it would keep working

PR #4341, opened 2025-06-20
DRAFT: 4123 add new percentiles mode to colormap

payno:
Checklist:

- [x] Colormap: Add 17.5-82.5 percentile mode

This PR adds a 'imageJ auto' like mode for the colormap. Which is using 17.2-82.5 percentile


src/silx/gui/colors.py
```diff
@@ -336,7 +336,10 @@ class Colormap(qt.QObject):
     PERCENTILE_1_99 = "percentile_1_99"
     """constant for autoscale using 1st and 99th percentile of data"""
 
-    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE_1_99)
+    PERCENTILE_17_5_82_5 = "percentile_17.5_82.5"
+    """constant for autoscale using 17.5 and 82.5 percentile of data"""
+
+    AUTOSCALE_MODES = (MINMAX, STDDEV3, PERCENTILE_1_99, PERCENTILE_17_5_82_5)
```
vallsv, line 342: I think the idea was to have a configurable `PERCENTILE`. Your PR was doing that before, or maybe i am wrong.

But with only 3 setup reachable from the GUI.

Or i misunderstand? Maybe double check with @t20100?

t20100: > I think the idea was to have a configurable PERCENTILE. But with only 3 setup reachable from the GUI.

This is what I understood from our last discussion as well.

PR #4352, opened 2025-07-15
DRAFT: silx.gui.plot.otems.ColormapMixIn: replace '__cacheColormapRange' by a lru_cache ('_colormapRange').

payno:
This PR is a suggestion to replace the existing `__cacheColormapRange` by a lru_cache


src/silx/gui/plot/items/core.py
```diff
@@ -608,7 +609,31 @@ def __init__(self):
...
         self._colormap.sigChanged.connect(self._colormapChanged)
         self.__data = None
-        self.__cacheColormapRange = {}  # Store {normalization: range}
+
+    @lru_cache(maxsize=100)
+    def _colormapRange(
+        self,
+        normalization: Colormap.LINEAR | Colormap.LOGARITHM | Colormap.SQRT | Colormap.GAMMA | Colormap.ARCSINH,
+        autoscaleMode: Colormap.MINMAX | Colormap.STDDEV3 | Colormap.PERCENTILE_1_99,
+        minMax: tuple[float] | None,
```
payno, line 618: This is the first thing I am not so much fan. Is it clear that this is 'setting' the min max value to the cache ?

src/silx/gui/plot/items/core.py
```diff
@@ -608,7 +609,31 @@ def __init__(self):
...
+            return minMax
+
+        data = self.getColormappedData(copy=False)
+        if data is None:
+            return None, None
+
+        colormap = Colormap(
+            normalization=normalization,
+            autoscaleMode=autoscaleMode,
+        )
```
payno, line 635: This is the second thing I am not so fan about.
I would lilke to use directly `self.getColormap()` but when calling `_getColormapAutoscaleRange` the colormap can be a third one. Like [here](https://github.com/silx-kit/silx/blob/main/src/silx/gui/dialog/ColormapDialog.py#L479-484)

src/silx/gui/plot/items/core.py
t20100: This is not the `@lru_cache` pattern so it's indeed not clear... and since `minMax` is part of the key of the cache, here pre-filling the cache has no effect since next call will pass `minMax=None` which will compute the range

src/silx/gui/plot/items/core.py
```diff
@@ -653,16 +678,21 @@ def _setColormappedData(
...
+                    minMax=(min_, max_),
+                )
             if minPositive is not None and numpy.isfinite(minPositive):
-                self.__cacheColormapRange[Colormap.LOGARITHM, Colormap.MINMAX] = (
-                    minPositive,
-                    max_,
+                self._colormapRange(
+                    normalization=Colormap.LOGARITHM,
+                    autoscaleMode=Colormap.MINMAX,
+                    minMax=(minPositive, max_),
```
t20100, line 695: `min_`, `minPositive`, `max_` are passed at only one place: https://github.com/silx-kit/silx/blob/c179b58509dd304f033f1566876da44f1bb8b7b6/src/silx/gui/plot3d/items/volume.py#L96
but that's to spare computing min/max over a 3D volume.
Otherwise it would be straight forward, but with this since there is no way to put values in the cache, I don't see simple way around it.

PR #4355, opened 2025-07-17
DRAFT: OverlayMixIn: allow users to provide an alignment and an offset

payno:
# Details

This PR add parameters to the OverlayMixIn in order to let the user decide about the alignment and a possible offset

# Note

https://github.com/silx-kit/silx/pull/4354 to be merged first

# screenshot

Some demo results

### qt.Qt.AlignRight | qt.Qt.AlignTop
Screenshot from 2025-07-17 15-41-31

### qt.Qt.AlignCenter,
Screenshot from 2025-07-17 15-41-26

### qt.Qt.AlignBottom | qt.Qt.AlignLeft

Screenshot from 2025-07-17 15-41-21

### qt.Qt.AlignVCenter | qt.Qt.AlignLeft
Screenshot from 2025-07-17 15-41-15


src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -0,0 +1,122 @@
...
+    def __init__(
+        self,
+        parent,
+        alignment: qt.Qt.AlignmentFlag = qt.Qt.AlignCenter,
+        alignment_offsets: tuple[int] = (0, 0),
+    ):
+        """
+        :param parent: parent widget
+        :param alignment: alignment of the overlay.
+        :param alignment_offsets: alignment offset as (horizontal offset, vertical offset). Values can be positive or negative. It will offset the alignment of this value
```
payno: I thinks with those two parameters we should be able to cover almost all use cases

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -0,0 +1,122 @@
...
+        self._unregisterParent(self.parent())
+        super().setParent(parent)
+        self._registerParent(parent)
+
+    def _getGeometryTopLeft(self) -> tuple[int]:
+        """Return the top left corner of the geometry to set up the geometry"""
+
+        parent = self.parent()
+
+        overlay_size: qt.QSize = self.sizeHint()
```
payno: This is the only part I am not such a big fan. But I think we already discussed about it.
This is updated at run time by the [WaitingPushButton](https://github.com/silx-kit/silx/blob/main/src/silx/gui/widgets/WaitingPushButton.py#L65)

Might be a bit unexpected.

src/silx/gui/widgets/test/test_overlay_widgets.py
```diff
@@ -32,14 +32,25 @@ def test_show(qapp, qapp_utils, widget_parent):
...
+        qt.Qt.AlignLeft,
+        qt.Qt.AlignRight | qt.Qt.AlignTop,
+        qt.Qt.AlignCenter,
+        qt.Qt.AlignLeft | qt.Qt.AlignBottom,
+    ),
+)
 @pytest.mark.parametrize("widget_parent", (Plot2D, qt.QFrame))
 @pytest.mark.parametrize("constructor", (LabelOverlay, ButtonOverlay))
-def test_overlay_widgets(qapp, qapp_utils, widget_parent, constructor):
+def test_overlay_widgets(qapp, qapp_utils, widget_parent, constructor, alignment):
```
payno, line 46: There is no much test. I checked different uses cases visually (using a `QWidget` and a `Plot2D`) but no much.
I wonder if testing the output of `_getGeometryTopLeft` could be useful but not so sure because the complexity is limited.

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -12,9 +12,28 @@ class OverlayMixIn:
     .. warning:: Any class inheriting from this mixin must also inherit from a QWidget.
     """
 
-    def __init__(self, parent):
+    def __init__(
+        self,
+        parent,
+        alignment: qt.Qt.AlignmentFlag = qt.Qt.AlignCenter,
+        alignment_offsets: tuple[int] = (0, 0),
```
t20100: To be consistent with Qt, it's best to have setters for features rather than passing too many arguments in the `__init__`, so best to remove `alignment` and `alignment_offsets` arguments from here and provide `get|setAlignment` and `get|setAlignmentOffsets`

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -12,9 +12,28 @@ class OverlayMixIn:
...
+        self._alignment: qt.Qt.AlignmentFlag = alignment
+        self._alignment_offsets: tuple[int] = alignment_offsets
         self._registerParent(parent=parent)
 
+    def getAlignment(self) -> qt.Qt.AlignmentFlag:
+        return self._alignment
+
+    def setAlignment(self, alignment: qt.Qt.AlignmentFlag):
+        self._alignment = alignment
+        self.repaint()
```
t20100: `repaint` is synchronous, it's best to ask for an asynchronous repaint with:
```suggestion
        self.update()
```

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -45,6 +64,46 @@ def setParent(self, parent: qt.QWidget):
         super().setParent(parent)
         self._registerParent(parent)
 
+    def _getGeometryTopLeft(self) -> tuple[int]:
```
t20100:
```suggestion
    def _getGeometryTopLeft(self) -> qt.QPoint:
```

src/silx/gui/widgets/test/test_overlay_widgets.py
t20100: It's OK for me as it is: At least it checks that it doesn't raise an exception.

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -45,6 +64,46 @@ def setParent(self, parent: qt.QWidget):
...
+        """Return the top left corner of the geometry to set up the geometry"""
+
+        parent = self.parent()
+
+        overlay_size: qt.QSize = self.sizeHint()
```
t20100: Can you use camelCase variables for consistency?

src/silx/gui/widgets/OverlayMixIn.py
t20100: Also IMO this could compute the geometry QRect instead of just the point

 ```python
def _getGeometry(self) -> qt.QRect | None:
```

File: src/silx/gui/widgets/OverlayMixIn.py [hunk 8]
```diff
@@ -45,6 +64,46 @@ def setParent(self, parent: qt.QWidget):
         super().setParent(parent)
         self._registerParent(parent)
 
+    def _getGeometryTopLeft(self) -> tuple[int]:
+        """Return the top left corner of the geometry to set up the geometry"""
+
+        parent = self.parent()
```
t20100, line 71: Need to check if it is `None`

src/silx/gui/widgets/OverlayMixIn.py
t20100: BTW, is typing info needed? It can be inferred from `sizeHint()`, no?

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -45,6 +64,46 @@ def setParent(self, parent: qt.QWidget):
...
+        if isinstance(parent, PlotWidget):
+            offset = parent.getWidgetHandle().mapTo(parent, qt.QPoint(0, 0))
+            canvas_left, canvas_top, canvas_width, canvas_height = (
+                parent.getPlotBoundsInPixels()
+            )
+            canvas_left += offset.x()
+            canvas_top += offset.y()
+        else:
+            canvas_width = parent.size().width()
+            canvas_top = canvas_height = parent.size().height()
```
t20100:
```suggestion
            canvas_height = parent.size().height()
```

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -45,6 +64,46 @@ def setParent(self, parent: qt.QWidget):
...
+            canvas_left = 0
+            canvas_top = 0
+
+        # calculate left position
+        if self._alignment & qt.Qt.AlignTop:
```
t20100, line 90: Does this works with PyQt5, PyQt6 and PySide6? That's the kind of stuff that changed AFAIR

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -54,20 +113,8 @@ def _resize(self):
...
-                    int(offset.x() + left + width / 2 - size.width() / 2),
-                    int(offset.y() + top + height / 2 - size.height() / 2),
-                ),
-                size,
-            )
-        else:
-            position = parent.size()
-            position = (position - size) / 2
-            rect = qt.QRect(qt.QPoint(position.width(), position.height()), size)
+        rect = qt.QRect(self._getGeometryTopLeft(), size)
```
t20100: I propose to move:
```python
        parent = self.parent()
        if parent is None:
            return

        size = self.sizeHint()
        rect = qt.QRect(self._getGeometryTopLeft(), size)
```
into `_getGeometryTopLeft` and rename it `_getGeometry`

src/silx/gui/widgets/OverlayMixIn.py
t20100: I agree

t20100 (commented):
Good generalization!
I left some comments to make it more qt-ish :smile:

src/silx/gui/widgets/OverlayMixIn.py
t20100: I agree!

src/silx/gui/widgets/OverlayMixIn.py
vallsv: BTW here the right type is `tuple[int, int]`

src/silx/gui/widgets/OverlayMixIn.py
payno: Ah sh*** I will test the different version

payno: It seems to work the same for PyQt5 and PySide6 at least...

payno: Values seems to be the same and are the same and operations are done in python. I don't really see what could break here between Qt5 and Qt6.

t20100: Great, then!
Thanks for checking!

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -12,9 +12,27 @@ class OverlayMixIn:
     .. warning:: Any class inheriting from this mixin must also inherit from a QWidget.
     """
 
-    def __init__(self, parent):
+    def __init__(
+        self,
+        parent,
```
t20100: Let's add typing while at it:
```suggestion
        parent: qt.QWidget | None = None,
```

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -12,9 +12,27 @@ class OverlayMixIn:
...
+
+    def setAlignment(self, alignment: qt.Qt.AlignmentFlag):
+        self._alignment = alignment
+        self.update()
+
+    def getAlignmentOffsets(self) -> tuple[int, int]:
+        return self._alignmentOffsets
+
+    def setAlignmentOffsets(self, offsets: tuple[int, int]):
+        self._alignmentOffsets = offsets
```
t20100: Would be good to add docstrings since it is now public and `AlignmentOffsets` deserves a bit of doc IMO

src/silx/gui/widgets/OverlayMixIn.py
```diff
@@ -45,29 +63,60 @@ def setParent(self, parent: qt.QWidget):
         super().setParent(parent)
         self._registerParent(parent)
 
-    def _resize(self):
-        if not qt_inspect.isValid(self):
-            return  # For _resizeLater in case the widget has been deleted
+    def _getGeometry(self) -> qt.QRect | None:
+        """Return the top left corner of the geometry to set up the geometry"""
```
t20100: The docstring no longer describes what is returned, I would just remove it:

```suggestion
```

src/silx/gui/widgets/OverlayMixIn.py
t20100: Also when `Alignment` or `AlignmentOffsets` is changed, the widget needs to be refreshed and he geometry needs to be updated.
Calling:
```python
        self._resize()
        self.update()
````
probably does the job.

t20100 (commented):
Would be good to add to the doc!

I spotted an issue: when alignment/alignment offset is updated, this needs to be taken into account

payno:
> Would be good to add to the doc!
> 
> I spotted an issue: when alignment/alignment offset is updated, this needs to be taken into account

What specifically do you want to add to the doc ?

t20100:
> What specifically do you want to add to the doc ?

The OverlayMixIn class

payno:
It was added in the last PR when moved to the public API

   Henri Payno

Le mar. 22 juil. 2025, 09:58, Thomas VINCENT ***@***.***> a
écrit :

payno:
@t20100 ready for another review when you have some time.

t20100 (approved):
LGTM!

PR #4359, opened 2025-07-22
silx.gui.plot, silx.gui.plot3d: Fixed colormap out-of-bound color with OpenGL backend

### LINKED ISSUE #4356: Wrong color for pixels with value above or equal high colormap limit with `gl` backend
## Description

When displaying an image with the `gl` backend , the pixels with value above or equal high colormap limit get colored with the color corresponding to the **low** colormap instead of the color corresponding to the **high** colormap.

Ex: in this plot below (see code below to reproduce) with the `Gray` colormap, the pixel with the value `99` is colored with the low color value (black) instead of the high color value (white) with the `gl` backend (left) but not with `matplotlib` (right)

Image

The effect is consistent when the colormap limits are changed (here 0-75):

Image

## Context

```
Package                       Version   
----------------------------- --------------- 
AnyQt                         0.2.0
PyOpenGL                      3.1.9
PyQt6                         6.8.1
PyQt6-Qt6                     6.8.2
PyQt6_sip                     13.10.0
silx                          2.2.2
``` 

## Reproducing script

```python
import numpy

from silx.gui import qt

A = numpy.arange(0, 100).reshape((10, 10))

app = qt.QApplication([])
w = qt.QWidget()
layout = qt.QHBoxLayout(w)
for backend in ("gl", "matplotlib"):
    plt = Plot2D(backend=backend)
    plt.addImage(A)
    plt.setGraphTitle(backend)
    layout.addWidget(plt)
w.show()
app.exec()
```

t20100:
Fix #4356 

Checklist:

<hr>

This PR changes the texture wrap mode from clamp-to-edge to mirrored-repeat for the texture storing the colormap.
This fixes rendering issues on some computers (closes  #4356) and it should not impact rendering otherwise.

### DIFF
--- a/src/silx/gui/plot/backends/glutils/GLPlotImage.py
+++ b/src/silx/gui/plot/backends/glutils/GLPlotImage.py
@@ -342,7 +342,7 @@ class GLPlotColormap(_GLPlotData2D):
                 texUnit=self._CMAP_TEX_UNIT,
                 minFilter=gl.GL_NEAREST,
                 magFilter=gl.GL_NEAREST,
-                wrap=(gl.GL_CLAMP_TO_EDGE, gl.GL_CLAMP_TO_EDGE),
+                wrap=gl.GL_MIRRORED_REPEAT,
             )
             self._cmap_texture.prepare()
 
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/backends/glutils/GLPlotImage.py
```diff
@@ -342,7 +342,7 @@ def prepare(self):
                 texUnit=self._CMAP_TEX_UNIT,
                 minFilter=gl.GL_NEAREST,
                 magFilter=gl.GL_NEAREST,
-                wrap=(gl.GL_CLAMP_TO_EDGE, gl.GL_CLAMP_TO_EDGE),
+                wrap=gl.GL_MIRRORED_REPEAT,
```
t20100, line 345: `Texture.__init__` accept a single `wrap` instead of a tuple if it's the same for both dimensions

PR #4361, opened 2025-08-01
silx.gui.plot: Improved StatsWidget display slightly, especially for 1D plots

flowln:
This PR does a few small improvements to the way we display statistics in `StatsWidget`, especially in the case of 1D plots. I've made the smallest amount of changes possible, but any suggestions of better places to do changes are welcome!

Before PR, on a sample 1D plot:

image

After PR, on the same sample and plot:

image

### DIFF
--- a/src/silx/gui/plot/StatsWidget.py
+++ b/src/silx/gui/plot/StatsWidget.py
@@ -819,6 +819,7 @@ class StatsTable(_StatsWidgetBase, TableWidget):
             for column, tableItem in enumerate(tableItems):
                 tableItem.setData(qt.Qt.UserRole, _Container(item))
                 tableItem.setFlags(qt.Qt.ItemIsEnabled | qt.Qt.ItemIsSelectable)
+                tableItem.setTextAlignment(qt.Qt.AlignHCenter | qt.Qt.AlignVCenter)
                 self.setItem(row, column, tableItem)
 
             # Update table items content


src/silx/gui/plot/stats/statshandler.py
```diff
@@ -217,6 +217,8 @@ def calculate(
             data_changed=data_changed,
             roi_changed=roi_changed,
         )
-        for resName, resValue in list(res.items()):
-            res[resName] = self.format(resName, res[resName])
+        for resName, resValue in res.items():
+            if isinstance(resValue, tuple) and len(resValue) == 1:
+                resValue = resValue[0]
```
payno, line 222: The problem with this approach is that we won't be able to display tuples as results of stats. But we need it.
For example with 2D or 3D data, we expect the StatCoordMin method to return a tuple of values—and that’s precisely what we want to show to the user.

image

The simplest alternative is probably to create you own `StatCoordMin` that would return a scalar instead of a tuple like

```python
class MyStatCoordMin(StatCoordMin):
    def calculate(self, context):
        result = super().calculate(context)
        return result[0]
```

src/silx/gui/plot/StatsWidget.py
```diff
@@ -817,6 +819,7 @@ def _addItem(self, item):
             for column, tableItem in enumerate(tableItems):
                 tableItem.setData(qt.Qt.UserRole, _Container(item))
                 tableItem.setFlags(qt.Qt.ItemIsEnabled | qt.Qt.ItemIsSelectable)
+                tableItem.setTextAlignment(qt.Qt.AlignHCenter | qt.Qt.AlignVCenter)
```
payno, line 822: :+1:

src/silx/gui/plot/StatsWidget.py
```diff
@@ -633,6 +633,8 @@ def setStats(self, statsHandler):
         horizontalHeader = self.horizontalHeader()
         horizontalHeader.setSectionResizeMode(qt.QHeaderView.ResizeToContents)
 
+        horizontalHeader.setStretchLastSection(True)
```
payno, line 636: :+1:

src/silx/gui/plot/stats/statshandler.py
flowln: That's why I added the check `len(resValue) == 1`, so that it is specific to 1D values. Though in 2D+ cases, perhaps the best approach would be to change the `format` method to also check if the value is a tuple or a list even when `self.formatters[name] is None`, so we can call `str` on each individual element anyway, and get rid of the `np.floatXY` text.

payno: The `StatsHandler.calculate` function only delegates the calculation of the result.
Similarly, `StatsHandler.format` simply delegates to the formatter associated with the Stats object. This way we maintain a certain separation of concerns. 

Given the current design, I think there are two approaches to address your use case:

- Use the default formatter and ensure your stats return a scalar value, as demonstrated in the `MyStatCoordMin` example.

- Provide a custom formatter as you suggested. This approach might be more involved. An example is given in [examples/plotStats.py](https://github.com/silx-kit/silx/blob/main/examples/plotStats.py). And you can define the stats to be computed with it (formatter)[https://docs.python.org/3.9/library/formatter.html].
For example if I resume the example with` StatCoordMin` you would have something like
```python
    plot = Plot2D()

    stats = [
        (StatCoordMin(), "{0:.2f}"),
    ]

    plot.getStatsWidget().setStats(stats)
```

PR #4371, opened 2025-09-08
Colormap upgrade cache

payno:
- silx.gui.plot: ColormapMixIn: replace usage of a dict for the '__cacheColormapRange' by a LRU cache.


src/silx/gui/plot/items/core.py
```diff
@@ -604,11 +605,13 @@ def drag(self, from_, to):
 class ColormapMixIn(_Colormappable, ItemMixInBase):
     """Mix-in class for items with colormap"""
 
+    COLORMAP_CACHE_SIZE = 25
```
payno: The size is pretty arbitrary. Maybe it can be worth to increase it a bit.

src/silx/gui/plot/items/core.py
```diff
@@ -653,7 +656,7 @@ def _setColormappedData(
         self.__data = (
             None if data is None else numpy.array(data, copy=copy or NP_OPTIONAL_COPY)
         )
-        self.__cacheColormapRange = {}  # Reset cache
+        self.__cacheColormapRange.clear()
```
payno, line 661: I think using clear is clearer... without bad wordplay.

src/silx/utils/test/test_cache.py
```diff
@@ -0,0 +1,17 @@
+from ..cache import LRUCache
+
+def test_LRU_CACHE():
```
payno: I was not that inspired for the testing part...

src/silx/utils/cache.py
```diff
@@ -0,0 +1,57 @@
+from __future__ import annotations
+
+
+from collections import OrderedDict
+
+
+class LRUCache:
```
payno: Originally I wanted to inherit directly from `OrderedDict`. See [this commit](https://github.com/silx-kit/silx/pull/4371/commits/870dcddecfd51c99343c51abfa5c8e0d445749fa).
But it turned out that this implementation fails with python 3.10. Because the __getitem__ API is used when calling `move_to_end`. So the appropriate way seems to have composition instead of inheritance. This forces me to expose the API...

src/silx/utils/cache.py
```diff
@@ -0,0 +1,57 @@
...
+    def __setitem__(self, key, value):
+        self._cache[key] = value
+
+        self._cache.move_to_end(key)
+
+        if len(self) > self._maxsize:
+            self._cache.popitem(last=False)
+
+    def __getitem__(self, key):
+        value = self._cache[key]
```
payno: will raise a key error if doesn't exist.

src/silx/gui/plot/items/core.py
t20100: Clearer indeed!

src/silx/utils/cache.py
t20100: Let's not add the `LRUCache` class to the public API for a single usage inside the library.
Then, best to implement only the method that gets used.
Also, since it is used only in one class and `get` is called at a single place, you could call `move_to_end` directly from the code if `vRange` is not None here (with a comment):
https://github.com/silx-kit/silx/blob/bff0f83a1f10cb33f0f8ff975016704f85144c59/src/silx/gui/plot/items/core.py#L702-L705

src/silx/gui/plot/items/core.py
t20100: Sounds good to me as it is, but you can increase it a bit if you want:
If you propose a slider with integer steps for the percentile, having `COLORMAP_CACHE_SIZE = 128` would store all possible values.

payno: Yes good point, lets go for 128.

src/silx/utils/cache.py
payno: :ok_hand: I moved the cache to `silx.gui.plot.items._cache`. After discussion we will keep the class for now as it also handle the limitation of the size and not only keeping the last element access at the end.

PR #4379, opened 2025-09-08
colormap: allow users to provide percentile values for 'PERCENTILE' mode

payno:
- silx.gui.plot.colormap: generalize the percentile mode

# Details

https://github.com/silx-kit/silx/pull/4371 to be merged first.

# Feedback

Still not a big fan of the integration in the dialog. I don't think users can see the link between percentile and saturation.
I might also hide the slider until we are not on the 'percentile' mode.

image

# Demo

[output.webm](https://github.com/user-attachments/assets/0fdc8412-09db-4de9-8b47-8042eec8a313)

### DIFF
--- a/src/silx/gui/colors.py
+++ b/src/silx/gui/colors.py
@@ -582,7 +582,7 @@ class Colormap(qt.QObject):
                 since_version="3.0",
             )
             mode = self.PERCENTILE
-            self.setAutoscalePercentiles(self._DEFAULT_PERCENTILES)
+            self.setAutoscalePercentiles((1.0, 99.0))
         assert mode in self.AUTOSCALE_MODES
         if mode != self._autoscaleMode:
             self._autoscaleMode = mode


examples/plotWidget.py
```diff
@@ -41,7 +41,7 @@
 
 from silx.gui import qt
 
-from silx.gui.plot import PlotWidget
+from silx.gui.plot import PlotWidget, Plot2D
```
t20100:
```suggestion
from silx.gui.plot import PlotWidget
```

examples/plotWidget.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, parent=None):
         super().__init__(parent)
 
         # Create a PlotWidget
-        self._plot = PlotWidget(parent=self)
+        self._plot = Plot2D(parent=self)
```
t20100:
```suggestion
        self._plot = PlotWidget(parent=self)
```

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
         return self.__gamma
 
     def getAutoscaleMode(self) -> str:
-        """Return the autoscale mode of the colormap ('minmax' or 'stddev3')"""
+        """Return the autoscale mode of the colormap. Possible values are ('minmax', 'stddev3', 'percentile')"""
```
t20100:
```suggestion
        """Return the autoscale mode of the colormap."""
```

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
         return self.__gamma
 
     def getAutoscaleMode(self) -> str:
```
t20100:
```suggestion
    def getAutoscaleMode(self) -> Literal["minmax", "stddev3", "percentile"]:
```

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
...
         if self.isEditable() is False:
             raise NotEditableError("Colormap is not editable")
+        if mode == self.PERCENTILE_1_99:
+            deprecated_warning(
+                type_="Mode",
+                name="mode",
+                replacement="percentile",
+                since_version="3.0",
+            )
+            mode = self.PERCENTILE
```
t20100, line 584:
```suggestion
            mode = self.PERCENTILE
            self.setAutoscalePercentile((1.0, 99.0))
```

src/silx/gui/colors.py
```diff
@@ -387,6 +396,7 @@ def __init__(
 
         self._normalization = str(normalization)
         self._autoscaleMode = str(autoscaleMode)
+        self._percentiles_autoscale_values = self._DEFAULT_PERCENTILES
```
t20100: CamelCase:
```suggestion
        self._percentilesAutoscaleValues = self._DEFAULT_PERCENTILES
```

Or simpler: `self._percentiles` or `self._percentileRange` 🤷

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
         return self.__gamma
 
     def getAutoscaleMode(self) -> str:
-        """Return the autoscale mode of the colormap ('minmax' or 'stddev3')"""
+        """Return the autoscale mode of the colormap. Possible values are ('minmax', 'stddev3', 'percentile')"""
         return self._autoscaleMode
 
     def setAutoscaleMode(self, mode: str):
```
t20100:
```suggestion
    def setAutoscaleMode(self, mode: Literal["minmax", "stddev3", "percentile"]):
```

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
...
-        """Set the autoscale mode: either 'minmax' or 'stddev3'
+        """Set the autoscale mode: either 'minmax', 'stddev3' or 'percentile'.
```
t20100:
```suggestion
        """Set the autoscale mode.
```

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
...
+
+        .. warning:: 'percentile_1_99' is deprecated
```
t20100: IMO this is not needed, just remove it from the doc:
```suggestion
```

src/silx/gui/colors.py
```diff
@@ -557,21 +567,38 @@ def getGammaNormalizationParameter(self) -> float:
...
 
         :param mode: the mode to set
         """
         if self.isEditable() is False:
             raise NotEditableError("Colormap is not editable")
+        if mode == self.PERCENTILE_1_99:
+            deprecated_warning(
```
t20100, line 578: 👍

src/silx/gui/colors.py
```diff
@@ -695,6 +726,19 @@ def getColormapRange(
         if vmin is not None and vmax is not None:
             return vmin, vmax
 
+        if vmin is None or vmax is None:  # Handle autoscale
```
t20100: This is always `True` and the added code below is a copy of what is below with `percentile` arg added. Needs to be cleaned-up to avoid doing twice the same thing

.github/workflows/ci.yml
```diff
@@ -105,6 +105,10 @@ jobs:
           python ./ci/info_platform.py
           pip list
 
+      - name: Linting
+        run |
+          black --check .
+
```
t20100:
```suggestion
```

src/silx/gui/colors.py
```diff
@@ -285,6 +286,9 @@ def getColormappedData(self, copy: bool = False) -> numpy.ndarray | None:
         raise NotImplementedError("This method must be implemented in subclass")
 
 
+from functools import lru_cache
+
+
```
t20100:
```suggestion
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -83,6 +83,7 @@
 from silx.gui.qt import inspect as qtinspect
 from silx.gui.widgets.ColormapNameComboBox import ColormapNameComboBox
 from silx.gui.widgets.FormGridLayout import FormGridLayout
+from silx.gui.widgets.SliderWithSinBox import SliderWithSpinBox
```
t20100: Typo in the module name:
```suggestion
from silx.gui.widgets.SliderWithSpinBox import SliderWithSpinBox
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -233,9 +234,9 @@ class _AutoscaleModeComboBox(qt.QComboBox):
     DATA = {
         Colormap.MINMAX: ("Min/max", "Use the data min/max"),
         Colormap.STDDEV3: ("Mean±3std", "Use the data mean ± 3 × standard deviation"),
-        Colormap.PERCENTILE_1_99: (
-            "Percentile 1-99",
-            "Use 1st to 99th percentile of data",
+        Colormap.PERCENTILE: (
+            "Percentile",
+            "Use n'st to m'th percentile of data",
```
t20100:
```suggestion
            "Use n'st to (100-n)'th percentile of data",
```

src/silx/gui/widgets/SliderWithSinBox.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # connect signal / slot
+        self._slider.valueChanged.connect(self.setValue)
+        self._spinBox.valueChanged.connect(self.setValue)
+
+    def setValue(self, value: float):
+        with blockSignals(self._slider, self._spinBox):
+            self._slider.setValue(value)
+            self._spinBox.setValue(value)
+        self.valueChanged.emit(self.value())
```
t20100: Best to catch unchanged value early than to block signals
```suggestion
    def setValue(self, value: float):
        if value == self.value():
            return
        self._slider.setValue(value)
        self._spinBox.setValue(value)
        self.valueChanged.emit(self.value())
```

src/silx/gui/widgets/SliderWithSinBox.py
```diff
@@ -0,0 +1,48 @@
...
+from silx.gui import qt
+from silx.gui.utils import blockSignals
+
+
+class SliderWithSpinBox(qt.QWidget):
+    """
+    Widget with a slider and a spin box for a float
+    """
+
+    valueChanged = qt.Signal(int)
```
t20100:
```suggestion
    valueChanged = qt.Signal(float)
```

src/silx/gui/widgets/SliderWithSinBox.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+
+from silx.gui import qt
+from silx.gui.utils import blockSignals
+
+
+class SliderWithSpinBox(qt.QWidget):
```
t20100: Do we need to add this to the public API?

src/silx/image/test/test_contrast_enhancer.py
```diff
@@ -0,0 +1,29 @@
+import numpy
+import pytest
+from silx.image.ContrastEnhancer import _ContrastEnhancer
```
t20100: This does not exists

src/silx/math/colormap.py
```diff
@@ -235,18 +236,31 @@ def is_valid(
             return True
 
     def autoscale(
-        self, data: numpy.ndarray | None, mode: AutoScaleModeType
+        self,
+        data: numpy.ndarray | None,
+        mode: AutoScaleModeType,
```
t20100, line 241: `AutoScaleModeType` needs to be updated.. and best to use the same everywhere:
```
AutoScaleModeType = Literal["minmax", "stddev3", "percentile_1_99"]
```

Probably good checking there is no more `percentile_1_99` in the project

src/silx/math/colormap.py
```diff
@@ -235,18 +236,31 @@ def is_valid(
...
             return self.DEFAULT_RANGE
 
+        if mode == "percentile_1_99":
+            deprecated_warning(
+                type_="Mode",
+                name="mode",
+                replacement="percentile",
+                since_version="3.0",
+            )
+            mode = "percentile"
```
t20100, line 262: Need to set percentiles to (1, 99)

src/silx/math/colormap.py
```diff
@@ -323,11 +337,26 @@ def autoscale_mean3std(
...
     ) -> tuple[float, float] | tuple[None, None]:
         """Autoscale using [1st, 99th] percentiles
 
+        :param data: The data to process
+        :returns: (vmin, vmax)
+        """
+        return self.autoscale_percentile(data, percentile=(1, 99))
+
+    def autoscale_percentile(
+        self, data: numpy.ndarray, percentile: tuple[float, float]
```
t20100:
```suggestion
        self, data: numpy.ndarray, percentiles: tuple[float, float]
```

🤷

src/silx/math/colormap.py
```diff
@@ -453,10 +482,13 @@ def _get_range(
     autoscale: AutoScaleModeType,
     vmin: float | None,
     vmax: float | None,
+    percentile: tuple[float, float] | None = None,
```
t20100:
```suggestion
    percentiles: tuple[float, float] | None = None,
```
🤷

src/silx/math/colormap.py
```diff
@@ -475,6 +507,7 @@ def apply_colormap(
     vmin: float | None = None,
     vmax: float | None = None,
     gamma: float = 1.0,
+    percentile: tuple[float, float] | None = None,
```
t20100:
```suggestion
    percentiles: tuple[float, float] | None = None,
```

src/silx/gui/widgets/_SliderWithSpinBox.py
```diff
@@ -0,0 +1,48 @@
...
+        self.setRange(0, 100)
+
+        # connect signal / slot
+        self._slider.valueChanged.connect(self.setValue)
+        self._spinBox.valueChanged.connect(self.setValue)
+
+    def setValue(self, value: float):
+        with blockSignals(self._slider, self._spinBox):
+            self._slider.setValue(value)
+            self._spinBox.setValue(value)
```
t20100: IMO best to avoid `blockSignals` whenever possible:
```suggestion
    def setValue(self, value: float):
        if value == self._slider.value():
            return
        self._slider.setValue(value)
        self._spinBox.setValue(value)
```

src/silx/gui/widgets/_SliderWithSpinBox.py
```diff
@@ -0,0 +1,48 @@
...
+        self.valueChanged.emit(self.value())
+
+    def value(self) -> float:
+        return self._slider.value()
+
+    def setRange(self, min: float, max: float):
```
t20100:
```suggestion
    def value(self) -> int:
        return self._slider.value()

    def setRange(self, min: int, max: int):
```

src/silx/gui/widgets/_SliderWithSpinBox.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+
+from silx.gui import qt
+from silx.gui.utils import blockSignals
+
+
+class SliderWithSpinBox(qt.QWidget):
```
t20100: Since this widget is only used in `ColormapDialog`, let's move this module to `silx.gui.dialog`, there is no reason to have it in `silx.gui.widgets`.

src/silx/gui/plot3d/_model/items.py
```diff
@@ -965,6 +965,8 @@ def __init__(self, item):
         self.addRow(_ColormapBoundRow(item, name="Min.", index=0))
         self.addRow(_ColormapBoundRow(item, name="Max.", index=1))
 
+        # TODO: add option of the saturation for 3D
+
```
t20100: IMO best to open an issue, we don't look often at `# TODO`

src/silx/gui/colors.py
```diff
@@ -556,22 +563,38 @@ def getGammaNormalizationParameter(self) -> float:
...
         assert mode in self.AUTOSCALE_MODES
         if mode != self._autoscaleMode:
             self._autoscaleMode = mode
             self.sigChanged.emit()
 
+    def setAutoscalePercentile(self, value: tuple[float, float]):
+        self._percentiles = value
+        self.sigChanged.emit()
+
+    def getAutoscalePercentile(self) -> tuple[float, float]:
```
t20100: Worth a docstring IMO to explain that the 2-tuple is a (min, max) range with values expected in [0, 100]

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -974,6 +977,23 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         self._autoButtons = _AutoScaleButton(self)
         self._autoButtons.autoRangeChanged.connect(self._autoRangeButtonsUpdated)
 
+        # used percentile
+        self._centralPercentileWidget = SliderWithSpinBox(self)
```
t20100: It sounds that converting `SliderWithSpinBox` to a specific widget, e.g., `ColormapPercentileWidget` would allow to move the code handling the percentile from `ColormapDialog` to this widget:
All the init code for sure, but also the conversion code to/from 1 "saturation" value to a percentile range.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1638,6 +1666,12 @@ def _applyColormap(self):
             with utils.blockSignals(self._autoButtons):
                 self._autoButtons.setEnabled(colormap.isEditable())
                 self._autoButtons.setAutoRangeFromColormap(colormap)
+            with utils.blockSignals(self._centralPercentileWidget):
+                self._centralPercentileWidget.setValue(
+                    from_lateral_percentile_range_to_central_percentile(
```
t20100: A `ColormapPercentileWidget` could have a `setPercentileRange` for this.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1723,11 +1756,38 @@ def _autoscaleModeUpdated(self):
...
+                self._autoScaleCombo.currentText()
+                == _AutoscaleModeComboBox.DATA[Colormap.PERCENTILE][0]
+            )
             with self._colormapChange:
+                if enableCentralPercentile:
+                    colormap.setAutoscalePercentile(
+                        from_central_percentile_to_lateral_percentile_range(
+                            self._centralPercentileWidget.value()
+                        )
+                    )
```
t20100: The percentile could be set for all modes, it will not be used if not in the right mode, this makes it simpler:

```suggestion
                colormap.setAutoscalePercentile(
                    from_central_percentile_to_lateral_percentile_range(
                        self._centralPercentileWidget.value()
                    )
                )
```

BTW, a dedicated percentile widget can have a `getPercentileRange()` method

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1723,11 +1756,38 @@ def _autoscaleModeUpdated(self):
...
                 colormap.setAutoscaleMode(mode)
 
         self._updateWidgetRange()
 
+    def _updateCentralPercentileVisibility(self):
```
t20100:
```suggestion
    def _updateCentralPercentileEnabled(self):
```
to match what it actually does

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1920,3 +1980,26 @@ def keyPressEvent(self, event):
...
+def from_lateral_percentile_range_to_central_percentile(
+    lateral_percentile_range: tuple[float, float],
+) -> float:
+    """
+    Example: if we use lateral percentile (1st, 99th) we use 98% of the percentiles. This is the central percentile
+    """
+    return 100 - (lateral_percentile_range[0] + (100 - lateral_percentile_range[1]))
+
+
+def from_central_percentile_to_lateral_percentile_range(
```
t20100: CamelCase and private please and maybe shorter names (`saturation_to_percentiles` or `saturation_to_percentile_range`) 😅 
But IMO those functions would be best hidden in a percentile widget

t20100 (commented):
`ColormapDialog` is already a lengthy piece of code. IMO it's best to off-load code when we can, so I propose to turn the `SliderWithSpinBox` into a specific `PercentileWidget` containing all the code regarding widget init/sync and 1-value <-> 2-tuple conversion

src/silx/gui/dialog/ColormapDialog.py
payno: I renamed it to  `̀fromSaturationToPercentiles` But for me this is not really saturation. But anyway I have nothing better.
So I also renamed the `_centralPercentileWidget` and related widget / function.

src/silx/gui/widgets/_SliderWithSpinBox.py
payno: I don't think this would work. When the slider is changed it won't update the spin box value.

I don't see how to have a good sync of two widgets displaying and editing the same value not passing by those kind of `blockSignals`

src/silx/gui/dialog/_ColormapPercentileWidget.py
```diff
@@ -0,0 +1,82 @@
...
+        self._slider.valueChanged.connect(self.setSaturationValue)
+        self._spinBox.valueChanged.connect(self.setSaturationValue)
+
+    def setSaturationValue(self, value: float):
+        with blockSignals(self._slider, self._spinBox):
+            self._slider.setValue(value)
+            self._spinBox.setValue(value)
+        self.percentilesChanged.emit(self.getPercentilesRange())
+
+    def getSaturationValue(self) -> int:
```
payno: I hesitate between calling it `saturationValue` or `getSaturationValue` or `value`.

* value was not making the thing  much clear
* `saturationValue` would have make sense because it doesn't require processing. But if we want to change the internal mecanism having `getSaturationValue` and `getPercentilesRange` seemed more wise. And simpler from a user point of view...

src/silx/math/colormap.py
```diff
@@ -235,18 +236,32 @@ def is_valid(
...
 
+        if mode == "percentile_1_99":
+            deprecated_warning(
+                type_="Mode",
+                name="mode",
+                replacement="percentile",
+                since_version="3.0",
+            )
+            mode = "percentile"
+            percentiles = (1.0, 99.0)
```
payno, line 263: I used value directly and not the `_DEFAULT_PERCENTILES` because those default might change ... :innocent:

src/silx/gui/colors.py
```diff
@@ -556,22 +563,44 @@ def getGammaNormalizationParameter(self) -> float:
...
             self.sigChanged.emit()
 
+    def setAutoscalePercentiles(self, value: tuple[float, float]):
+        self._percentiles = value
+        self.sigChanged.emit()
+
+    def getAutoscalePercentiles(self) -> tuple[float, float]:
+        """
+        Return the (min, max) percentiles used for autoscaling in 'percentile' mode.
+        'min' and 'max' are between 0 and 100 inclusive.
```
t20100:
```suggestion
        'min' and 'max' are between 0 and 100 included.
```

?

src/silx/gui/dialog/_ColormapPercentileWidget.py
t20100: `get|setSaturationValue` sounds OK (could also be `get|setSaturation`).

BTW `get|setSaturationValue` are only used in `ColormapPercentilesWidget` so they can be private.
What is exposed is the range, which makes sense since it is what is used in the Colormap: Outside of this widget, nothing should care about "saturation".

If using `value`, it should be percentile range, IMO, since this is what should be exposed.

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -974,6 +977,14 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         self._autoButtons = _AutoScaleButton(self)
         self._autoButtons.autoRangeChanged.connect(self._autoRangeButtonsUpdated)
 
+        # used percentile
+        self._percentilesWidget = ColormapPercentilesWidget(self)
+        self._percentilesWidget.setTickPosition(qt.QSlider.TicksBelow)
+
+        self._percentilesWidget.setPercentilesRange(Colormap._DEFAULT_PERCENTILES)
+        self._percentilesWidget.setTracking(False)
```
t20100: This can be moved in `ColormapPercentilesWidget.__init__`

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1724,10 +1746,28 @@ def _autoscaleModeUpdated(self):
         colormap = self.getColormap()
         if colormap is not None:
             with self._colormapChange:
+                colormap.setAutoscalePercentiles(
+                    self._percentilesWidget.getPercentilesRange()
+                )
```
t20100: This should not be needed, since it is set all the time in `_percentilesChanged`:

```suggestion
```

src/silx/gui/dialog/ColormapDialog.py
```diff
@@ -1037,6 +1048,7 @@ def __init__(self, parent=None, title="Colormap Dialog"):
         button.setDefault(True)
         button = self._buttonsNonModal.button(qt.QDialogButtonBox.Reset)
         button.clicked.connect(self.resetColormap)
+        button.clicked.connect(self._updatePercentilesWidgetEnabled)
```
t20100: Why is this needed?

src/silx/gui/dialog/_ColormapPercentileWidget.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # connect signal / slot
+        self._slider.valueChanged.connect(self.setSaturationValue)
+        self._spinBox.valueChanged.connect(self.setSaturationValue)
+
+    def setSaturationValue(self, value: int):
+        with blockSignals(self._slider, self._spinBox):
+            self._slider.setValue(value)
+            self._spinBox.setValue(value)
+        self.percentilesChanged.emit(self.getPercentilesRange())
```
t20100: As it is, the signal is emitted even if the value is not updated: This should be avoided to avoid infinite signal loops.
Also I think `blockSignals is not needed:

```suggestion
        previousValue = self.getSaturationValue()
        self._slider.setValue(value)
        self._spinBox.setValue(value)
        if value != previousValue:
            self.percentilesChanged.emit(self.getPercentilesRange())
```

previous /current check could also be done on percentile range :shrug:

t20100 (approved):
Sounds clearer to me with `_ColormapPercentileWidget`!

I left a few comments, the main one being to avoid emitting `percentilesChanged` when the value has not changed.
Otherwise LGTM!

src/silx/gui/dialog/_ColormapPercentileWidget.py
payno: :ok_hand: 
For me we are forced to store the last value used for saturation.
Because `getSaturationValue` uses the `slider.value` which might have already changed because users can modify the value of either the slider **or** the spin-box.

So would something like that fine with you @t20100 
```
        ...
        self._saturationLastValue = None
        ...

    def _setSaturation(self, value: int):
        self._slider.setValue(value)
        self._spinBox.setValue(value)
        if self._saturationLastValue != value:
            self._saturationLastValue = value
            self.percentilesChanged.emit(self.getPercentilesRange())
```

t20100: Yes, makes sense!

src/silx/gui/dialog/ColormapDialog.py
payno: Removed

PR #4381, opened 2025-09-08
Project: Fixed flake8 warnings outside of `src/`

t20100:
Checklist:

<hr>

There was a `target` undeclared variable in `bootstrap.py` which makes the "scripts" option not working. I updated the code to what seems reasonable however it still does not work... Anyway IMO `bootstrap.py` is no longer needed since we can now pip install with `-e` thanks to meson, so I leave it as it is.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -26,14 +26,6 @@ logging.basicConfig()
 logger = logging.getLogger("bootstrap")
 
 
-def is_debug_python():
-    """Returns true if the Python interpreter is in debug mode."""
-    if sysconfig.get_config_var("Py_DEBUG"):
-        return True
-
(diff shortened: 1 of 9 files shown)


bootstrap.py
```diff
@@ -26,14 +26,6 @@
 logger = logging.getLogger("bootstrap")
 
 
-def is_debug_python():
```
t20100, line 29: Never called

bootstrap.py
```diff
@@ -244,9 +236,9 @@ def main(argv):
         res = find_executable(script)
         kind = res[0]
         if kind == "path":
-            run_file(target, argv)
```
t20100, line 247: `target` is undeclared

PR #4383, opened 2025-09-08
silx: Fix flake8 in `src/silx/*.py`

t20100:
Checklist:

<hr>

### DIFF
--- a/src/silx/conftest.py
+++ b/src/silx/conftest.py
@@ -194,7 +194,7 @@ def qWidgetFactory(qapp, qapp_utils):
         widget.setAttribute(qt.Qt.WA_DeleteOnClose)
         widget.show()
         qapp_utils.qWaitForWindowExposed(widget)
-        widgets.append(widget)  # noqa: F821
+        widgets.append(widget)
 
         return widget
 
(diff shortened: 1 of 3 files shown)


src/silx/conftest.py
```diff
@@ -194,7 +194,7 @@ def createWidget(cls, *args, **kwargs):
         widget.setAttribute(qt.Qt.WA_DeleteOnClose)
         widget.show()
         qapp_utils.qWaitForWindowExposed(widget)
-        widgets.append(widget)
+        widgets.append(widget)  # noqa: F821
```
t20100: This should not be needed since `widgets` is defined higher up, but `flake8` raises an error: need to investigate

src/silx/conftest.py
```diff
@@ -216,8 +216,6 @@ def createWidget(cls, *args, **kwargs):
     validWidgets = [widget for widget in widgets if isValid(widget)]
     assert not validWidgets, f"Some widgets were not destroyed: {validWidgets}"
 
-    del widgets
-
```
t20100: This is useless, the variable gets deleted at the end of the function anyway.
This was causing `flake8` to consider the variable not available in `createWidget`

woutdenolf (approved):

PR #4384, opened 2025-09-08
silx.gui.plot.PlotWidget: Fix _getItem (and derived methods) typing

### LINKED ISSUE #4338: Typing is wrong in `PlotWidget._getItem` and functions using it
It tells that the return type is `Item` while it could return `None`.

It should be `Item | None`

loichuder:
Checklist:

Fix #4338

### DIFF
--- a/src/silx/gui/plot/PlotWidget.py
+++ b/src/silx/gui/plot/PlotWidget.py
@@ -1981,7 +1981,7 @@ class PlotWidget(qt.QMainWindow):
 
     # Hide
 
-    def isCurveHidden(self, legend):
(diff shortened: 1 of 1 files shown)


t20100 (approved):
LGTM!

Maybe good to simplify the corresponding docstrings, now that there's typing, it seems quite a bit of the docstrings are duplicating the typing information:

- `:param str legend:` -> `:param legend:`
- `:return: None or :class:`.items.Curve` object` no longer needed if returned type is `items.Curve | None`

But maybe worth doing in a separate PR to check docstring in the whole class at once.

PR #4385, opened 2025-09-08
silx.gui.plot3d: Fixed flake8 warnings

t20100:
Checklist:

<hr>

### DIFF
--- a/src/silx/gui/_glutils/test/test_gl.py
+++ b/src/silx/gui/_glutils/test/test_gl.py
@@ -26,11 +26,11 @@ from .. import gl
 
 
 def test_version_bytes(mocker):
-    mocker.patch("silx.gui._glutils.gl.glGetString", return_value=b"3.0 Mock")
+    mocker.patch("OpenGL.GL.glGetString", return_value=b"3.0 Mock")
     assert gl.getVersion() == (3, 0)
 
 
 def test_version_str(mocker):
     """In case glGetString returns str"""
-    mocker.patch("silx.gui._glutils.gl.glGetString", return_value="3.0 Mock")
+    mocker.patch("OpenGL.GL.glGetString", return_value="3.0 Mock")
     assert gl.getVersion() == (3, 0)


src/silx/gui/_glutils/gl.py
```diff
@@ -56,26 +55,24 @@
     OpenGL.FormatHandler.by_name("ctypesparameter").check.append("_ctypes.CArgObject")
 
 
-import OpenGL.GL as _GL
```
t20100, line 59: All those imports needs to be done after setting some variables in `OpenGL` top module...

src/silx/gui/_glutils/gl.py
```diff
@@ -96,7 +93,7 @@ def getVersion() -> tuple:
         ValueError: If the version returned by the driver is not supported
     """
     try:
-        desc = glGetString(GL_VERSION)
+        desc = _GL.glGetString(_GL.GL_VERSION)
```
t20100, line 96: Replacing all functions/variables used from `*` imports

src/silx/gui/plot3d/test/test_scenewidget.py
```diff
@@ -57,7 +57,7 @@ def tearDown(self):
 
     def testFogEffect(self):
         """Test fog effect on scene primitive"""
-        image = self.widget.addImage(numpy.arange(100).reshape(10, 10))
```
t20100, line 60: unused

src/silx/gui/_glutils/gl.py
woutdenolf:
```python
>>> from OpenGL.GL import glGetString,GL_VERSION
>>> print(glGetString(GL_VERSION))
None
```

woutdenolf: Seems like you need an opengl context before you can get the version

```python
import glfw
from OpenGL.GL import glGetString, GL_VERSION

assert glGetString(GL_VERSION) is None

if not glfw.init():
    raise Exception("GLFW can't be initialized")

window = glfw.create_window(640, 480, "OpenGL Context", None, None)
glfw.make_context_current(window)

assert glGetString(GL_VERSION).decode()

glfw.terminate()
```

t20100: Yes, the tests mock `glGetString` but that was no longer the right one

PR #4386, opened 2025-09-08
silx.opencl: Fixed flake8 warnings

woutdenolf:
Checklist:

<hr>

### DIFF
--- a/src/silx/opencl/atomic.py
+++ b/src/silx/opencl/atomic.py
@@ -42,7 +42,7 @@ import pyopencl.array as cla
 def check_atomic32(device):
     try:
         ctx = pyopencl.Context(devices=[device])
-    except:
+    except Exception:
         return False, f"Unable to create context on {device}"
     else:
         queue = pyopencl.CommandQueue(ctx)
@@ -65,7 +65,7 @@ int res = atom_inc(ary);
 def check_atomic64(device):
     try:
         ctx = pyopencl.Context(devices=[device])
-    except:
+    except Exception:
         return False, f"Unable to create context on {device}"
     else:
         queue = pyopencl.CommandQueue(ctx)
--- a/src/silx/opencl/codec/byte_offset.py
+++ b/src/silx/opencl/codec/byte_offset.py
@@ -41,7 +41,8 @@ __status__ = "production"
 import functools
 import os
 import numpy
-from ..common import ocl, pyopencl
+from ..common import ocl  # noqa F401 Initialize OpenCL
+from ..common import pyopencl
 from ..processing import BufferDescription, EventDescription, OpenclProcessing
 
 import logging
--- a/src/silx/opencl/common.py
+++ b/src/silx/opencl/common.py
@@ -67,10 +67,12 @@ else:
             pyopencl = None
 
 if pyopencl is not None:
-    import pyopencl.array as array
+    import pyopencl.array as array  # noqa F401
 
     mf = pyopencl.mem_flags
 else:
+    array = None
+
     # Define default mem flags
     class mf:
         WRITE_ONLY = 1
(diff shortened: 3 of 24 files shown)


src/silx/opencl/atomic.py
```diff
@@ -42,7 +42,7 @@
 def check_atomic32(device):
     try:
         ctx = pyopencl.Context(devices=[device])
-    except:
+    except Exception:
```
woutdenolf, line 45: Note: a bare except captures CTRL-C for example.

Capture `BaseException` if you really need a cleanup regardless of the exception type. But always re-raise after. Library code should never absorb base exceptions like `KeyboardInterrupt` and `SystemError`.

I'm sure you know but just in case.

src/silx/opencl/codec/byte_offset.py
```diff
@@ -41,7 +41,8 @@
 import functools
 import os
 import numpy
-from ..common import ocl, pyopencl
+from ..common import ocl  # noqa F401 Initialize OpenCL
```
woutdenolf, line 44: I keep all unused imports of `ocl`. I'm sure there is some monkey-business going on there.

src/silx/opencl/common.py
```diff
@@ -135,17 +137,20 @@ def _convert_to_int(val, default_val=0):
...
-            device_id = device_instance.platform.get_devices().index(device_instance)
-            platform_id = pyopencl.get_platforms().index(device.platform)
-            if cache is not None:
-                cache[(platform_id, device_id)] = ctx
+            device = ctx.devices[0]
+        platform = device.platform
+        device_id = platform.get_devices().index(device)
+        platform_id = pyopencl.get_platforms().index(platform)
+        if cache is not None:
+            cache[(platform_id, device_id)] = ctx
```
woutdenolf: Bug introduced in https://github.com/silx-kit/silx/pull/4245.

@pierrepaleo Is this what you intended?


```python
import pyopencl

if hasattr(pyopencl, "choose_devices"):
    # pyopencl >=2024.2
    devices = pyopencl.choose_devices(interactive=False)
    device = devices[0]
    ctx = pyopencl.Context([device])
else:
    # pyopencl <=2024.1
    ctx = pyopencl.create_some_context(interactive=False)
    device = ctx.devices[0]
platform = device.platform
device_id = platform.get_devices().index(device)
platform_id = pyopencl.get_platforms().index(platform)
```

I tried with several pyopencl version (old and new).

I'm also assuming the caching was mistakenly disabled for `pyopencl >=2024.2`?

src/silx/opencl/atomic.py
kif: You are perfectly right ...

src/silx/opencl/sift/plan.py
```diff
@@ -58,11 +58,11 @@
...
 import math
 import logging
-import gc
 import numpy
 from .param import par
-from silx.opencl import ocl, pyopencl, kernel_workgroup_size
-from silx.opencl.utils import get_opencl_code, nextpower
+from ..common import ocl  # noqa F401 Initialize OpenCL
+from silx.opencl import pyopencl, kernel_workgroup_size  # noqa F401
+from silx.opencl.utils import get_opencl_code, nextpower  # noqa F401
```
kif: Wouldn't it make sense to use systematically relative imports (if we correct some code ...)

src/silx/opencl/common.py
pierrepaleo: You are correct, it seems that the cache is disabled for `pyopencl >=2024.2` . Thanks for spotting this.

@kif can you confirm ?

kif:
I submitted a PR to this branch with a fix ...

src/silx/opencl/sift/plan.py
woutdenolf: Sure. I will go through `silx.opencl` and make all imports relative.

woutdenolf: Done.

src/silx/opencl/common.py
kif: submitted a PR to fix it.

PR #4387, opened 2025-09-09
[OpenCL] ensure 1 and exactly 1 ctx gets cached

kif:
Simplification, and should work all the time.


woutdenolf (approved):
Looks good to me.

PR #4390, opened 2025-09-09
silx.math.combo: Added `mean_std` function for single pass mean+standard deviation computation

### LINKED ISSUE #4366: colormap: Use optional mean from std
numpy 1.26 feature `std` with an optional pre calculated `mean`.

I haven't try but this could speed up the the colormap statistics by 10-25%.

3 pass (min-max, mean, std), instead of 4 (min-max, mean, stdx2).

See https://numpy.org/doc/2.1/reference/generated/numpy.std.html

kif:
close #4366

Checklist:

- [X] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

Implement variance calculation as described in:
https://ds.ifi.uni-heidelberg.de/files/Team/eschubert/publications/SSDBM18-covariance-authorcopy.pdf
Faster since single pass, usually close to 3x vs numpy on large arrays (16Mpix) but not as numerically stable.

### DIFF
--- a/src/silx/math/combo.pyi
+++ b/src/silx/math/combo.pyi
@@ -63,7 +63,7 @@ def mean_std(
     mask: numpy.ndarray | None = None,
     dummy: float = numpy.nan,
     delta_dummy: float = 0.0,
-) -> Tuple[float, float]:
+) -> tuple[float, float]:
     """Computes mean and estimation of std in a single pass.
 
     Based on formula #12, #13 and #28 from :


src/silx/math/combo.pyx
```diff
@@ -27,12 +27,12 @@ For now it provides min/max (and optionally positive min) and indices
...
 """
 
-__authors__ = ["T. Vincent"]
+__authors__ = ["T. Vincent", "Jérôme Kieffer"]
 __license__ = "MIT"
-__date__ = "24/04/2018"
+__date__ = "09/09/2025"
 
 cimport cython
-from .math_compatibility cimport isnan, isfinite, INFINITY
```
t20100, line 35: `math_compatibility.pxd` is no longer needed on Windows?
Can be removed (in another PR)?

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def mean_std(data,
+             ddof=0,
+             mask=None,
+             dummy=None,
+             delta_dummy=0):
+    """Computes mean and estimation of std in a single pass.
+    Based on formula #12, #13 and #28 from :
```
t20100:
```suggestion
    """Computes mean and estimation of std in a single pass.

    Based on formula #12, #13 and #28 from :
```

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+
+@cython.embedsignature(True)
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def mean_std(data,
+             ddof=0,
+             mask=None,
+             dummy=None,
+             delta_dummy=0):
```
t20100: Can you add the typed signature in `combo.pyi`?
Can this also be typed here?

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+
+    :param data: Array-like dataset
+    :param int ddof:
+       Means Delta Degrees of Freedom.
+       The divisor used in calculations is data.size - ddof.
+       Default: 0 (as in numpy.std).
+    :param mask: array with 0 for valid values, same size as data
+    :param dummy: dynamic mask for value=dummy
+    :param delta_dummy: dynamic mask for abs(value-dummy)<=delta_dummy
+    :returns: An object with mean and std attributes
```
t20100:
```suggestion
    :returns: A tuple: (mean, std)
```

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+    else:
+        do_mask = True
+        cmask = numpy.ascontiguousarray(mask, dtype=numpy.int8).ravel()
+    if dummy is None:
+        do_dummy = False
+        cdummy = delta_dummy = 0.0
+    else:
+        do_dummy = True
+        cdummy = float(dummy)
+        cdelta_dummy = float(delta_dummy)
```
t20100: Can't this be ensure by typing `dummy` and `delta_dummy` in the function signature?

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+                new_cnt = cnt + 1.0
+                if cnt:
+                    delta = X-cnt*value
+                    XX += delta*delta/(cnt*new_cnt)
+                X += value
+                cnt = new_cnt
+
+    mean = X / cnt
+    if length <= ddof:
+        variance = float('nan')
```
t20100: unused:
```suggestion
```

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+    :returns: An object with mean and std attributes
+    :raises: ValueError if data is empty"""
+
+    cdef:
+        unsigned int length, index
+        bint do_mask, do_dummy
+        double value, delta, X, XX, cnt, new_cnt
+        double mean, variance, standard_deviation
+        double cdummy, cdelta_dummy
+        char[::1] cmask
```
t20100, line 403: Shouldn't this be initialised to `None` or something like it?

src/silx/math/combo.pyx
```diff
@@ -345,3 +345,96 @@ def min_max(data not None, bint min_positive=False, bint finite=False) -> _MinMa
...
+    X = 0.0
+    XX = 0.0
+    cnt = 0.0
+
+    with nogil:
+        for index in range(length):
+            value = cdata[index]
+            if _is_valid(value, cmask[index] if do_mask else 0, do_dummy, cdummy, cdelta_dummy):
+                new_cnt = cnt + 1.0
+                if cnt:
```
t20100: Could be more explicit:
```suggestion
                if cnt != 0.0:
```

t20100 (commented):
Looks good.
Can you add typing info?

src/silx/math/combo.pyx
kif: I don't use it in pyFAI nor fabio and it works great for ages ... So yes, it can probably be removed.

kif:
> Looks good. Can you add typing info?

I am not very confident with mixing python typing on top of cython typing ...

src/silx/math/combo.pyx
kif: For `ddof`, and `delta_dummy`, yes.
For `dummy`, it is not possible, unless one sets the default value to `NaN` or any infinite value.
Mask and data cannot be too strongly types, one would need a python function and a cython function, making the code much longer.

Maybe typing `dummy` would be a not so bad idea. What do you think ?

src/silx/math/combo.pyx
kif: Not compatible with `dummy=None`

src/silx/math/combo.pyx
kif: Not if not used. It is like declaring a variable and not using it.

src/silx/math/combo.pyx
t20100: -> #4398

src/silx/math/combo.pyx
t20100: I would type as much as possible so cython takes care of type checking/coercion.
`dummy=NaN` by default works for me.

src/silx/math/combo.pyi
```diff
@@ -56,3 +56,33 @@ def min_max(
...
     """
     ...
+
+def mean_std(
+    data: numpy.ndarray,
+    ddof: float = 0.0,
+    mask: numpy.ndarray | None = None,
+    dummy: float = numpy.nan,
+    delta_dummy: float = 0.0,
+) -> Tuple[float, float]:
```
t20100:
```suggestion
) -> tuple[float, float]:
```

t20100:
> I am not very confident with mixing python typing on top of cython typing ...

Yes me too, I'd prefer a unique function signature if there's a way to do it.

PR #4393, opened 2025-09-09
typing/docstring updates

t20100:
Checklist:

<hr>

### DIFF
--- a/src/silx/gui/plot/PlotWidget.py
+++ b/src/silx/gui/plot/PlotWidget.py
@@ -34,7 +34,7 @@ __date__ = "21/12/2018"
 
 import logging
 from collections import namedtuple
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 from contextlib import contextmanager
 import datetime as dt
 import itertools
@@ -378,7 +378,7 @@ class PlotWidget(qt.QMainWindow):
     def __init__(
         self,
         parent: qt.Qt.Widget | None = None,
-        backend: str | BackendBase | None = None,
+        backend: str | BackendBase | Iterable[str | BackendBase] | None = None,
     ):
         self._autoreplot = False
         self._dirty = False
@@ -490,7 +490,7 @@ class PlotWidget(qt.QMainWindow):
         self.__selection = None
 
     def __getBackendClass(
-        self, backend: str | BackendBase | list[str | BackendBase]
+        self, backend: str | BackendBase | Iterable[str | BackendBase] | None
     ) -> BackendBase:
         """Returns backend class corresponding to backend.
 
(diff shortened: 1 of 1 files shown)


src/silx/gui/plot/PlotWidget.py
```diff
@@ -378,7 +378,7 @@ class PlotWidget(qt.QMainWindow):
     def __init__(
         self,
         parent: qt.Qt.Widget | None = None,
-        backend: str | BackendBase | None = None,
+        backend: str | BackendBase | Iterable[str | BackendBase] | None = None,
```
t20100: `backend` can also be an Iterable at all the places

src/silx/gui/plot/PlotWidget.py
```diff
@@ -660,14 +662,12 @@ def _getDirtyPlot(self) -> bool | Literal["overlay"]:
...
         It can be accessed by backend to check the dirty state.
-
-        :return: False, True, 'overlay'
         """
         return self._dirty
 
     # Default Qt context menu
 
-    def contextMenuEvent(self, event: qt.Qt.QContextEvent | None):
+    def contextMenuEvent(self, event: qt.Qt.QContextEvent):
```
t20100, line 672: The code of this method do not expect `None`

src/silx/gui/plot/PlotWidget.py
```diff
@@ -2298,15 +2298,13 @@ def setActiveCurveSelectionMode(self, mode: Literal["legacy", "atmostone", "none
...
+        It can be "atmostone" (at most one), "legacy" or "none".
 
         :rtype: str
         """
         return self._activeCurveSelectionMode
 
-    def getActiveImage(
-        self, just_legend: bool = False
-    ) -> items.ImageData | items.ImageRgba | str | None:
+    def getActiveImage(self, just_legend: bool = False) -> items.ImageBase | str | None:
```
t20100, line 2309: Used the same type as returned by `addImage`, it can also be an ImageComplex for example

src/silx/gui/plot/PlotWidget.py
loichuder: Do you think it is worth creating a custom type?

src/silx/gui/plot/PlotWidget.py
loichuder: Well, this is not what the docstring was telling :smile:

src/silx/gui/plot/PlotWidget.py
t20100: I asked myself the same question :sweat_smile:, that's still readable but that would spare the copy/paste so maybe.

src/silx/gui/plot/PlotWidget.py
t20100: :sweat_smile: 

That's also inline with: `def _getActiveItem(...) -> items.Curve | items.Scatter | items.ImageBase | None:`

t20100:
I've added a type for backend args: https://github.com/silx-kit/silx/pull/4393/commits/e48a04f8bcfe6984582dfd2a12a83b3230ef27df

If this is OK for you, I let you merge this PR.

PR #4402, opened 2025-09-10
silx.gui: Optimize ROI statistics

woutdenolf:
Checklist:

<hr>

This is the basic class that checks whether a position is inside a ROI

```python
class _RegionOfInterestBase(qt.QObject):

    def contains(self, position: tuple[float, float]) -> bool:
        """Returns True if the `position` is in this ROI.

        :param position: position to check
        :return: True if the value / point is consider to be in the region of
                 interest.
        :rtype: bool
        """
```

In this MR we add a vectorized equivalent for `contains` and use it in the stats to optimize it (much too slow for images of 3kx3k for example)

```python
    def contains_multi(self, positions: ArrayLike) -> numpy.ndarray:
        """Returns a boolean array with True when `position[i]` is in this ROI.
        :param positions: array-like of shape (N, 2)
        :return: boolean array of shape (N,), True if the point is inside the ROI
        """
```

The `contains_multi` method has a basic implementation by looping over the positions and calling `contains`. Then in the derived classes we override `contains_multi` for optimization.

```
_RegionOfInterestBase (qt.QObject)
├─ RegionOfInterest
│   ├─ HandleBasedROI
│   │   ├─ RectangleROI
│   │   ├─ PolygonROI
│   │   ├─ ArcROI
│   │   ├─ CrossROI
│   │   ├─ LineROI
│   │   ├─ HorizontalLineROI
│   │   ├─ VerticalLineROI
│   │   ├─ CircleROI
│   │   ├─ EllipseROI
│   │   └─ BandROI
│   ├─ PointROI
│   ├─ HorizontalLineROI
│   ├─ VerticalLineROI
│   └─ HorizontalRangeROI
└─ ROI
```

I added unit tests which check stat results for all `RegionOfInterest`. These tests also pass for the main branch so there is not regression in terms of stat values.

See example below for increase in performance by this MR.

### DIFF
--- a/src/silx/gui/plot/CurvesROIWidget.py
+++ b/src/silx/gui/plot/CurvesROIWidget.py
@@ -1257,8 +1257,8 @@ class ROI(_RegionOfInterestBase):
         return rawArea, netArea
 
     @docstring(_RegionOfInterestBase)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        positions, is_single = self._normalize_positions_shape(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        positions, is_single = self._normalize_positions_shape(position)
         is_inside = self._fromdata <= positions[:, 0] <= self._todata
         return is_inside[0] if is_single else is_inside
 
--- a/src/silx/gui/plot/items/_arc_roi.py
+++ b/src/silx/gui/plot/items/_arc_roi.py
@@ -913,8 +913,8 @@ class ArcROI(HandleBasedROI, items.LineMixIn, InteractionModeMixIn):
         self._updateHandles()
 
     @docstring(HandleBasedROI)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        positions, is_single = self._normalize_positions_shape(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        positions, is_single = self._normalize_positions_shape(position)
 
         # geometry parameters (center is (x, y))
         center = numpy.array(self.getCenter())
--- a/src/silx/gui/plot/items/_band_roi.py
+++ b/src/silx/gui/plot/items/_band_roi.py
@@ -125,13 +125,15 @@ class BandGeometry(NamedTuple):
             self.begin.y + offset[1] - self.slope * (self.begin.x + offset[0]),
         )
 
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
         """Check which positions are inside the ROI.
 
-        :param positions: array-like of shape (N, 2) or (2,), each row is (x, y)
-        :return: bookean or boolean array of shape (N,), True if the point is inside the ROI
+        :param position: array-like of positions, where each position is given as ``(x, y)``.
+        If multiple positions are provided, the shape should be ``(N, 2)``.
+        For a single position, the shape should be ``(2,)``.
+        :return: boolean or boolean array of shape ``(N,)``, True if the point is inside the ROI.
         """
-        positions, is_single = self._normalize_positions_shape(positions)
+        positions, is_single = self._normalize_positions_shape(position)
         it = (Polygon(self.corners).is_inside(x, y) for x, y in positions)
         is_inside = numpy.fromiter(it, dtype=bool)
         return is_inside[0] if is_single else is_inside
@@ -394,8 +396,8 @@ class BandROI(HandleBasedROI, items.LineMixIn, InteractionModeMixIn):
         return tuple(geometry.center - offset * numpy.array(geometry.normal))
 
     @docstring(HandleBasedROI)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        return self.getGeometry().contains(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        return self.getGeometry().contains(position)
 
     def __str__(self):
         begin, end, width = self.getGeometry()
--- a/src/silx/gui/plot/items/_roi_base.py
+++ b/src/silx/gui/plot/items/_roi_base.py
@@ -99,14 +99,16 @@ class _RegionOfInterestBase(qt.QObject):
         """
         self.sigItemChanged.emit(event)
 
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
         """Check which positions are inside the ROI.
 
-        :param positions: array-like of shape (N, 2) or (2,), each row is (x, y)
-        :return: boolean or boolean array of shape (N,), True if the point is inside the ROI
+        :param position: array-like of positions, where each position is given as ``(x, y)``.
+        If multiple positions are provided, the shape should be ``(N, 2)``.
+        For a single position, the shape should be ``(2,)``.
+        :return: boolean or boolean array of shape ``(N,)``, True if the point is inside the ROI.
         """
         # Overwrite in subclass
-        positions, is_single = self._normalize_positions_shape(positions)
+        positions, is_single = self._normalize_positions_shape(position)
         if is_single:
             return False
         return numpy.zeros(len(positions), dtype=bool)
--- a/src/silx/gui/plot/items/roi.py
+++ b/src/silx/gui/plot/items/roi.py
@@ -115,8 +115,8 @@ class PointROI(RegionOfInterest, items.SymbolMixIn):
         self._marker.setPosition(*pos)
 
     @docstring(RegionOfInterest)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        positions, is_single = self._normalize_positions_shape(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        positions, is_single = self._normalize_positions_shape(position)
         roiPos = self.getPosition()
         is_inside = (positions[:, 0] == roiPos[0]) & (positions[:, 1] == roiPos[1])
         return is_inside[0] if is_single else is_inside
@@ -199,8 +199,8 @@ class CrossROI(HandleBasedROI, items.LineMixIn):
             self.sigRegionChanged.emit()
 
     @docstring(HandleBasedROI)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        positions, is_single = self._normalize_positions_shape(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        positions, is_single = self._normalize_positions_shape(position)
         roiPos = self.getPosition()
         is_inside = (positions[:, 0] == roiPos[0]) | (positions[:, 1] == roiPos[1])
         return is_inside[0] if is_single else is_inside
@@ -312,8 +312,8 @@ class LineROI(HandleBasedROI, items.LineMixIn):
             self.setEndPoints(start, end)
 
     @docstring(HandleBasedROI)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        positions, is_single = self._normalize_positions_shape(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        positions, is_single = self._normalize_positions_shape(position)
 
         points = self.__shape.getPoints()
         line_pt1, line_pt2 = points[0], points[1]
@@ -419,8 +419,8 @@ class HorizontalLineROI(RegionOfInterest, items.LineMixIn):
         self._marker.setPosition(0, pos)
 
     @docstring(RegionOfInterest)
-    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
-        positions, is_single = self._normalize_positions_shape(positions)
+    def contains(self, position: ArrayLike) -> Union[bool, numpy.ndarray]:
+        positions, is_single = self._normalize_positions_shape(position)
         roi_x = self.getPosition()
         is_inside = positions[:, 1] == roi_x
         return is_inside[0] if is_single else is_inside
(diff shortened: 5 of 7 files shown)


src/silx/gui/plot/items/_arc_roi.py
```diff
@@ -910,32 +911,57 @@ def setGeometry(self, center, innerRadius, outerRadius, startAngle, endAngle):
...
-            (position[1] - center[1]) ** 2 + (position[0] - center[0]) ** 2
-        )
-        is_in_distance = self.getInnerRadius() <= distance <= self.getOuterRadius()
-        if not is_in_distance:
-            return False
-        rel_pos = position[1] - center[1], position[0] - center[0]
-        angle = numpy.arctan2(*rel_pos)
-        # angle is inside [-pi, pi]
+    def contains(self, position: tuple[float, float]) -> bool:
+        return self.contains_multi(position)[0]
```
payno:
```suggestion
        return self.contains_multi((position,))[0]
```

src/silx/gui/plot/items/_band_roi.py
```diff
@@ -124,9 +125,21 @@ def edgesIntercept(self) -> tuple[float, float]:
...
+    def contains_multi(self, positions: ArrayLike) -> numpy.ndarray:
+        """Vectorized check for multiple positions.
+
+        :param positions: array-like of shape (N, 2), each row is (x, y) (col, row)
+        :return: boolean array of shape (N,), True if the point is inside the band
+        """
+        return numpy.fromiter(
+            (Polygon(self.corners).is_inside(x, y) for x, y in positions),
+            dtype=bool,
+            count=len(positions),
```
payno: Is providing `count` useful in this case ? We will read all the positions in this case

src/silx/gui/plot/CurvesROIWidget.py
```diff
@@ -1255,9 +1256,14 @@ def computeRawAndNetArea(self, curve):
         return rawArea, netArea
 
     @docstring(_RegionOfInterestBase)
-    def contains(self, position):
+    def contains(self, position: tuple[float, float]) -> bool:
         return self._fromdata <= position[0] <= self._todata
 
+    @docstring(_RegionOfInterestBase)
+    def contains_multi(self, positions: ArrayLike) -> numpy.ndarray:
+        positions = numpy.asarray(positions)
```
payno: Reusing `_RegionOfInterestBase._normalize_positions_shape` would also check the shapes
```suggestion
        positions = self._normalize_positions_shape(positions)
```

src/silx/gui/plot/items/roi.py
```diff
@@ -843,8 +887,18 @@ def handleDragUpdated(self, handle, origin, previous, current):
             self.setRadius(numpy.linalg.norm(center - current))
 
     @docstring(HandleBasedROI)
-    def contains(self, position):
-        return numpy.linalg.norm(self.getCenter() - position) <= self.getRadius()
+    def contains(self, position: tuple[float, float]) -> bool:
+        position = numpy.asarray(position)
+        center = numpy.array(self.getCenter())
+        distances = numpy.linalg.norm(center - position)
+        return distances <= self.getRadius()
```
payno:
```suggestion
        return self.contains_multi((position,))[0]
```

src/silx/gui/plot/items/roi.py
```diff
@@ -1123,14 +1177,37 @@ def _centerPositionChanged(self, event):
             self._updateGeometry()
 
     @docstring(HandleBasedROI)
-    def contains(self, position):
-        major, minor = self.getMajorRadius(), self.getMinorRadius()
+    def contains(self, position: tuple[float, float]) -> bool:
+        return self.contains_multi(position)[0]
```
payno:
```suggestion
        return self.contains_multi((position,))[0]
```

src/silx/gui/plot/items/roi.py
```diff
@@ -1327,9 +1404,10 @@ def __str__(self):
         return f"{self.__class__.__name__}({params})"
 
     @docstring(HandleBasedROI)
-    def contains(self, position):
+    def contains(self, position: tuple[float, float]) -> bool:
+        # First, filter positions outside the bounding box
         bb1 = _BoundingBox.from_points(self.getPoints())
-        if bb1.contains(position) is False:
+        if not bb1.contains(position):
```
payno: :+1:

src/silx/gui/plot/stats/stats.py
```diff
@@ -311,39 +310,40 @@ def clipData(self, item, plot, onlimits, roi):
...
-            if self.is_mask_valid(onlimits=onlimits, from_=minX, to_=maxX):
-                mask = self.mask
-            else:
-                mask = (minX <= xData) & (xData <= maxX)
-                mask = mask == 0
-                self._set_mask_validity(onlimits=onlimits, from_=minX, to_=maxX)
         elif roi:
             minX, maxX = roi.getFrom(), roi.getTo()
+        else:
+            minX, maxX = None, None
```
payno: Strange that we didn't saw it first. But from what I see nothing prevent to have at least `onlimits` or `roi` True :(

src/silx/gui/plot/stats/stats.py
```diff
@@ -311,39 +310,40 @@ def clipData(self, item, plot, onlimits, roi):
...
-        if len(unmasked_data) > 0:
-            self.min, self.max = min_max(unmasked_data)
-        else:
-            self.min, self.max = None, None
         self.data = (xData, yData)
         self.axes = (xData,)
 
+        if self.values.count() > 0:
+            self.min = self.values.min()
+            self.max = self.values.max()
```
payno: If I remember well the silx `math.combo.min_max` was to be faster than the `numpy` min / max one. Within the mask layer I would expect the same but not 100% sure.

src/silx/gui/plot/stats/stats.py
```diff
@@ -373,41 +373,40 @@ def __init__(self, item, plot, onlimits, roi):
...
-        unmasked_data = self.values.compressed()
-        if len(unmasked_data) > 0:
-            self.min, self.max = min_max(unmasked_data)
+        self.values = numpy.ma.array(yData, mask=mask)
+        self.data = (xData, yData)
+        self.axes = (xData,)
+
+        if self.values.count() > 0:
+            self.min = self.values.min()
+            self.max = self.values.max()
```
payno: Same interrogation. Using silx.math.combo.min_max is probably faster on large dataset / image.

src/silx/gui/plot/stats/stats.py
```diff
@@ -444,19 +443,21 @@ def clipData(self, item, plot, onlimits, roi):
...
 
-            # filter on X axis
-            valueData = valueData[(minX <= xData) & (xData <= maxX)]
-            yData = yData[(minX <= xData) & (xData <= maxX)]
-            xData = xData[(minX <= xData) & (xData <= maxX)]
-            # filter on Y axis
-            valueData = valueData[(minY <= yData) & (yData <= maxY)]
-            xData = xData[(minY <= yData) & (yData <= maxY)]
-            yData = yData[(minY <= yData) & (yData <= maxY)]
+            # Build a single boolean mask instead of repeated filtering
```
payno: :+1:

src/silx/gui/plot/stats/stats.py
```diff
@@ -465,15 +466,16 @@ def clipData(self, item, plot, onlimits, roi):
...
         self.data = (xData, yData, valueData)
         self.values = numpy.ma.array(valueData, mask=mask)
         self.axes = (xData, yData)
 
-        unmasked_values = self.values.compressed()
-        if len(unmasked_values) > 0:
-            self.min, self.max = min_max(unmasked_values)
+        if self.values.count() > 0:
+            self.min = self.values.min()
+            self.max = self.values.max()
```
payno: silx.math.combo.min_max ?

src/silx/gui/plot/stats/stats.py
```diff
@@ -544,54 +546,55 @@ def clipData(self, item, plot, onlimits, roi):
...
 
         if onlimits:
+            # Plot limits cropping
             minX, maxX = plot.getXAxis().getLimits()
             minY, maxY = plot.getYAxis().getLimits()
 
-            XMinBound = int((minX - self.origin[0]) / self.scale[0])
-            YMinBound = int((minY - self.origin[1]) / self.scale[1])
+            XMinBound = max(int((minX - self.origin[0]) / self.scale[0]), 0)
+            YMinBound = max(int((minY - self.origin[1]) / self.scale[1]), 0)
```
payno: To be honest I thought it was clearer before. Not a big fan of having many operations in the same line.
But I agree that this is more inline with the variable name...

src/silx/gui/plot/stats/stats.py
```diff
@@ -633,23 +636,32 @@ def __init__(self, item, plot, onlimits, roi):
...
             if self.values.ndim == 3:
                 axes.append(item.getZData(copy=False))
             self.axes = tuple(axes)
-            self.min, self.max = min_max(self.values)
-            self.values = numpy.ma.array(self.values, mask=mask)
+
+            # Use masked array methods instead of min_max
+            if self.values.count() > 0:
+                self.min = self.values.min()
+                self.max = self.values.max()
```
payno: silx.math.combo.min_max ?

src/silx/gui/plot/test/test_stats.py
```diff
@@ -1141,3 +1153,209 @@ def test(self):
...
+
+    def testRectangleROIStats(self):
+        roi = RectangleROI()
+        roi.setGeometry(origin=(2, 1), size=(1, 2))
+
+        # [[ 0,  1,  2,  3,  4],
+        #  [ 5,  6,  7*,  8*, 9],
+        #  [10, 11, 12*, 13*, 14],
+        #  [15, 16, 17*, 18*, 19],
+        #  [20, 21, 22, 23, 24]]
```
payno, line 1188: Not sure this is useful to have those outside the `setUp` function.

woutdenolf: The stars * are different for every test and it shows visually what we expect to be inside the ROI.

payno: ah my bad I missed it.

src/silx/gui/plot/stats/stats.py
woutdenolf: done

src/silx/gui/plot/stats/stats.py
woutdenolf: improved with numpy.clip

src/silx/gui/plot/stats/stats.py
woutdenolf: done

src/silx/gui/plot/stats/stats.py
woutdenolf: done

src/silx/gui/plot/stats/stats.py
woutdenolf: done

src/silx/gui/plot/stats/stats.py
woutdenolf: We now do this all all contexts

```python
if onlimits:
    ...
elif roi:
    ...
else:
    ...
```

src/silx/gui/plot/items/roi.py
woutdenolf: Refactored to one single `contains` that accepts a single point or a list of points.

src/silx/gui/plot/items/roi.py
woutdenolf: Refactored to one single `contains` that accepts a single point or a list of points.

src/silx/gui/plot/CurvesROIWidget.py
woutdenolf: Refactored to one single `contains` that accepts a single point or a list of points.

src/silx/gui/plot/items/_band_roi.py
woutdenolf: done

src/silx/gui/plot/items/_arc_roi.py
woutdenolf: Refactored to one single `contains` that accepts a single point or a list of points.

src/silx/gui/plot/items/_roi_base.py
```diff
@@ -96,15 +99,33 @@ def _updated(self, event=None, checkVisibility=True):
...
-        :param position: position to check
-        :return: True if the value / point is consider to be in the region of
-                 interest.
-        :rtype: bool
-        """
-        return False  # Override in subclass to perform actual test
+    def contains(self, positions: ArrayLike) -> Union[bool, numpy.ndarray]:
+        """Check which positions are inside the ROI.
+
+        :param positions: array-like of shape (N, 2) or (2,), each row is (x, y)
```
payno: Changing the function signature 'position' vs 'positions' could break some code...
After discussion between us I think we are fine with it. Nevertheless we need to make sure this is part of the CHANGELOG in the next release. 

I might state more clearly that users can provide a single position or a list

```suggestion
    :param positions: array-like of positions, where each position is given as `(x, y)`.
    If multiple positions are provided, the shape should be `(N, 2)`.
    For a single position, the shape should be `(2,)`. shape (N, 2).
```

src/silx/gui/plot/stats/stats.py
```diff
@@ -540,66 +532,63 @@ def is_mask_valid(self, xmin, xmax, ymin, ymax):
...
+            # Indices
+            xmin = int((minX - self.origin[0]) / self.scale[0])
+            ymin = int((minY - self.origin[1]) / self.scale[1])
+            xmax = int((maxX - self.origin[0]) / self.scale[0])
+            ymax = int((maxY - self.origin[1]) / self.scale[1])
 
-            XMinBound = max(XMinBound, 0)
-            YMinBound = max(YMinBound, 0)
+            xmin, xmax = numpy.clip([xmin, xmax], 0, xsize - 1)
+            ymin, ymax = numpy.clip([ymin, ymax], 0, ysize - 1)
```
payno, line 560: I was not sure about the -1 but indeed it make sense. Not sure it was properly done before.

src/silx/image/_boundingbox.py
```diff
@@ -47,21 +50,57 @@ def __init__(self, bottom_left, top_right):
         self.max_x = top_right[1]
         self.max_y = top_right[0]
 
-    def contains(self, item):
-        """
+    def contains(
```
payno, line 53: Same, we must make sure the update of the function signature is on the CHANGELOG

src/silx/image/_boundingbox.py
```diff
@@ -47,21 +50,57 @@ def __init__(self, bottom_left, top_right):
...
+                if ndim_org == 1:
+                    if items.shape[0] != 2:
+                        raise ValueError(err_msg)
+                    items = items.reshape(1, 2)
+                elif ndim_org != 2 or items.shape[1] != 2:
+                    raise ValueError(err_msg)
         else:
-            return (self.min_x <= item[1] <= self.max_x) and (
-                self.min_y <= item[0] <= self.max_y
+            raise ValueError(err_msg)
```
payno: I might expect more a TypeError

src/silx/gui/plot/items/_roi_base.py
woutdenolf: I changed the argument name back to `position` and improved the doctring (occurs in 3 places).

src/silx/image/_boundingbox.py
woutdenolf: I changed the argument name back to `item`.

src/silx/image/_boundingbox.py
woutdenolf: done

PR #4403, opened 2025-09-11
src.silx.io.url.DataUrl: Add `invalid_reason` property that tells why a DataUrl is not valid

### LINKED ISSUE #4266: DataUrl: provide more details when an URL is invalid
Today when we are creating an invalid url we get the following message:

``` bash
ValueError: URL 'xxx' is not valid
```

It would be useful to explain which part is invalid

loichuder:
Checklist:

<hr>

Fix #4266 

In essence, replaces the `__is_valid` inner member by a `__invalid_reason` member. `__invalid_reason` is `None` if the URL is valid or a `str` giving the reason why it is invalid if the URL is not valid.

The `__invalid_reason` is exposed through a `invalid_reason` property. The `is_valid` method is still there but now relies on `__invalid_reason` since `__is_valid` was removed.

### DIFF
--- a/src/silx/io/test/test_url.py
+++ b/src/silx/io/test/test_url.py
@@ -21,18 +21,32 @@
 #
 # ############################################################################*/
 """Tests for url module"""
+from __future__ import annotations
 
 __authors__ = ["V. Valls"]
 __license__ = "MIT"
 __date__ = "29/01/2018"
 
 
+from types import EllipsisType
 import pytest
 from ..url import DataUrl
 
 
-def assert_url(url, expected):
+def assert_url(
+    url: DataUrl,
+    expected: tuple[
+        bool,
+        bool,
+        str | None,
+        str,
+        str | None,
+        tuple[slice | int | EllipsisType, ...] | None,
+    ],
+):
     assert url.is_valid() == expected[0]
+    if url.is_valid():
+        assert url.invalid_reason is None
     assert url.is_absolute() == expected[1]
     assert url.scheme() == expected[2]
     assert url.file_path() == expected[3]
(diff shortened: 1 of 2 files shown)


src/silx/io/test/test_url.py
```diff
@@ -298,23 +324,32 @@ def test_scheme_none():
...
 
 def test_http_with_path():
     url = DataUrl("http://hsds-server.tld/home/file?/foo/bar")
-    expected = [True, False, "http", "hsds-server.tld/home/file", "/foo/bar", None]
+    expected = (True, False, "http", "hsds-server.tld/home/file", "/foo/bar", None)
     assert_url(url, expected)
+
+
+def test_invalid_slice():
+    url = DataUrl("silx://my_file.hdf5?path=/path/to/data&slice=not_a_slice")
```
loichuder, line 348: Note that slice validity is only checked when parsing from a path. We should properly check when the slice comes from `data_slice` as well ?

src/silx/io/url.py
```diff
@@ -188,38 +188,45 @@ def quote_string(string):
...
+                self.is_valid(),
                 quote_string(self.__scheme),
                 quote_string(self.__file_path),
                 quote_string(self.__data_path),
                 self.__data_slice,
             )
         else:
             template = "DataUrl(valid=%s, string=%s)"
-            return template % (self.__is_valid, self.__path)
+            return template % (self.is_valid(), self.__path)
```
loichuder: The invalid reason should perhaps be part of the string format of `DataUrl` now ?

src/silx/io/url.py
```diff
@@ -188,38 +188,45 @@ def quote_string(string):
...
+                "fabio URLs cannot have a data path"
+                if self.__data_path is not None
+                else None
+            )
         elif self.__scheme == "silx":
             # If there is a slice you must have a data path
-            # But you can have a data path without slice
-            slice_implies_data = (
-                self.__data_path is None and self.__data_slice is None
-            ) or self.__data_path is not None
```
loichuder, line 216: Simplified this slightly since we don't have to check if `__data_path` can be `None` (the URL can be valid in this case if there is no slice)

| `data_slice` | `data_path` | `valid` |
| ---------------- | ----------------- | -------- |
| None | None | Yes |
| None | str | Yes |
| str | str | Yes |
| str | None | **No** |

src/silx/io/test/test_url.py
payno: Yes I agree. Maybe just double check with @vallsv to make sure this is not used as 'a feature' in one of our project.
If this is the case we can still allow it to fail in this particular use case.

src/silx/io/url.py
payno: Seems like a good idea

src/silx/io/url.py
payno: Maybe this table could be added in the code if you find a good spot.

src/silx/io/url.py
```diff
@@ -188,38 +188,45 @@ def quote_string(string):
...
+            else:
+                self.__invalid_reason = None
+        elif self.__scheme in ("http", "https"):  # is an HSDS url for h5pyd
             # is an HSDS url for h5pyd
-            self.__is_valid = True
+            self.__invalid_reason = None
         else:
-            self.__is_valid = False
+            self.__invalid_reason = (
+                "Invalid scheme. It can only be 'fabio'. 'silx', 'http' or 'https'."
```
payno: Maybe it could be worth to have a (protected) variable with the 'valid_scheme' and use it here.

src/silx/io/test/test_url.py
loichuder: If we are unsure, I would suggest to keep the current behaviour for now

src/silx/io/url.py
loichuder: The table is there to explain why I simplified the previous condition. The only point of interest is that `data_slice` cannot be present if `data_path` is `None`.

Therefore, I don't think it is needed to have it in the current version of the code

src/silx/io/url.py
loichuder: `DataUrl(valid=True, string='foo:/foo/foobar.h5?path=/foo/bar&slice=5,1', invalid_reason='Invalid scheme. It can only be fabio, silx, http, https.')`

PR #4404, opened 2025-09-11
created LegendIconWidget  a similar  LegendIcon

abmajith:
In this pull request, created a unique LegendIcon, and LegendItem widget that is listed in vertical in qt.MainWindow,

the respective example are provided, but so far no unit test,

also I still not sure, whether I added some what generic use case to the LegendItem widget,

### DIFF
--- a/src/silx/gui/plot/test/test_legendswidget.py
+++ b/src/silx/gui/plot/test/test_legendswidget.py
@@ -27,6 +27,7 @@ def test_add_remove_items(qapp, plot_with_legend):
     qapp.processEvents()
     assert len(legend_widget._itemWidgets) == 0
 
+
 def test_visibility_toggle(qapp, qapp_utils, plot_with_legend):
     plot, legend_widget = plot_with_legend
     plot.addCurve([0, 1], [0, 1], legend="test_item")


src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+                self._item.setVisible(not self._item.isVisible())
+
+
+class LegendItemList(qt.QWidget):
+    def __init__(self, parent: PlotWidget):
+        super().__init__(parent, qt.Qt.Window)
+        self.setFixedWidth(200)
+        self._layout = qt.QVBoxLayout(self)
+        self._plot: PlotWidget | None = None
+        self._binding(parent)
```
t20100: The PlotWidget that is "observed" by the list of legends is not the "parent": "parent" is Qt widget tree structure, the plot can be anywhere in it.
Best to provide `get|setPlotWidget` methods to set it separately, and eventually pass it in `__init__`:
```
def __init__(
    self,
    parent: qt.QWidget | None = None,
    plotWidget: PlotWidget | None = None,
):
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        if hasattr(parent, "sigVisibilityChanged"):
+            parent.sigVisibilityChanged.connect(self._setVisibility)
+
+    def _setVisibility(self, status: bool) -> None:
+        if status:
+            self.show()
+        else:
+            self.hide()
```
t20100: This widget shouldn't need to care about this, Qt does it already:
```suggestion
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _binding(self, plotWidget: PlotWidget):
+        """Binds this widget to the signals of a parent PlotWidget."""
+        self._plot = plotWidget
+        self._plot.sigActiveCurveChanged.connect(self._onActiveItemChanged)
+        self._plot.sigActiveImageChanged.connect(self._onActiveItemChanged)
+        self._plot.sigActiveScatterChanged.connect(self._onActiveItemChanged)
```
t20100: The "active" state of the items is not used in `_onActiveItemChanged`, so no need to handle those signals
```suggestion
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self._plot.sigItemAdded.connect(self._onContentChanged)
+        self._plot.sigItemRemoved.connect(self._onContentChanged)
```
t20100: No need for wrapper method:
```suggestion
        self._plot.sigItemAdded.connect(self._updateAllItemsList)
        self._plot.sigItemRemoved.connect(self._updateAllItemsList)
```

But here, we don't want to redo the whole list each time an item is added/removed, just add/remove the corresponding `LegendItemWidget`

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self._updateHeigth()
+
+    def _updateHeigth(self):
+        _totalHeight = (self._layout.count() - 1) * self._layout.spacing()
+        _totalHeight += self._layout.contentsMargins().top()
+        _totalHeight += self._layout.contentsMargins().bottom()
+        if self._layout.count() > 0:
+            item = self._layout.itemAt(0)
+            _totalHeight += item.widget().sizeHint().height() * self._layout.count()
+        self.setFixedHeight(_totalHeight)
```
t20100: If there is issue with layout, it should be dealt otherwise.
It's always risky to play with this, for instance, the text of the legend is clipped on my laptop.

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _onActiveItemChanged(self, previous, current):
+        self._updateAllItemsList()
+
+    def _onContentChanged(self, item):
+        self._updateAllItemsList()
```
t20100:
```suggestion
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _clearLayout(self, layout):
+        """Helper to clear all widgets from a layout."""
+        if layout is not None:
+            while layout.count():
+                child = layout.takeAt(0)
+                if child.widget():
+                    child.widget().deleteLater()
+
+    def _updateAllItemsList(self):
```
t20100: Only do this for initializing the list, not for each update

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        if event in (
+            items.ItemChangedType.VISIBLE,
+            items.ItemChangedType.SYMBOL,
+            items.ItemChangedType.SYMBOL_SIZE,
+            items.ItemChangedType.LINE_WIDTH,
+            items.ItemChangedType.LINE_STYLE,
+            items.ItemChangedType.COLOR,
+            items.ItemChangedType.ALPHA,
+            items.ItemChangedType.HIGHLIGHTED,
+            items.ItemChangedType.HIGHLIGHTED_STYLE,
```
t20100: No need to handle all those "events", only the one that are actually handle here (name, visibility and highlight)

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+    def __init__(self, parent=None, item=None):
+        super().__init__(parent)
+        self._itemRef = None
+        self.setItem(item)
+
+    def getItem(self):
+        """Returns item associated to this widget
+
+        :rtype: Union[~silx.gui.plot.items,...]
+        """
```
t20100: Let's use typing:
```suggestion
    def getItem(self) -> items.Item:
        """Returns item associated to this widget"""
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        _item = self._item
+        if _item is None:
+            _logger.debug("Item no longer exists, disabling legend widget.")
+            self.setEnabled(False)
+            return
+
+        self.label.setText(_item.getName())
+
+        style = ""
+        if isinstance(_item, items.Curve) and _item.isHighlighted():
```
t20100: replace `items.Curve` with `HighlightedMixIn`

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+            style += "border: 1px solid black; "
+
+        if not _item.isVisible():
+            style += "color: gray;"
+        else:
+            style += "color: black;"
```
t20100: This does not work with dark mode.
Look at [`QPalette`](https://doc.qt.io/qt-6/qpalette.html#ColorGroup-enum) and maybe QStyle to change the color in a way that supports dark mode.

examples/plotLegendLists.py
```diff
@@ -0,0 +1,50 @@
...
+plotWindow = qt.QMainWindow()
+plotCentralWidget = qt.QWidget()
+plotLayout = qt.QVBoxLayout(plotCentralWidget)
+plotWindow.setCentralWidget(plotCentralWidget)
+plotWindow.setWindowTitle("Plot Window")
+plotWindow.resize(800, 600)
+
+plot = Plot2D()
+plotLayout.addWidget(plot)
+infoWidget = LegendItemList(parent=plot)
```
t20100: This is not shown

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        """Set the item with which to synchronize this widget.
+
+        :param item: Union[~silx.gui.plot.items,...]
+        """
+        assert (
+            item is None
+            or isinstance(item, items.ColormapMixIn)
+            or isinstance(item, items.SymbolMixIn)
+            or isinstance(item, items.ColorMixIn)
+            or isinstance(item, items.LineMixIn)
```
t20100: Not sure it's the right place to filter out, if it's not supported it should just display nothing

```suggestion
            or isinstance(item, items.Item)
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        if item is not None:
+            item.sigItemChanged.connect(self._itemChanged)
+
+        self._update()
+
+    def _update(self):
+        item = self.getItem()
+        if item is None:
+            _logger.debug("Item no longer exists, disabling legend widget.")
+            self.setEnabled(False)
```
t20100: Need to also remove symbol, line, cmap, ...

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+            items.ItemChangedType.HIGHLIGHTED_STYLE,
+        ):
+            self._update()
+
+
+class LegendItemWidget(qt.QWidget):
+    def __init__(self, parent, item: items):
+        super().__init__(parent)
+        self._item = item
+        self._itemRef = weakref.ref(item)
```
t20100: Refer twice to the same thing

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self.setLayout(qt.QHBoxLayout())
+        self.setFixedWidth(200)
+        self.layout().setSpacing(20)
+        self.layout().setContentsMargins(10, 0, 10, 0)
+        item.sigItemChanged.connect(self._itemChanged)
+
+        self.icon = LegendItemIcon(parent=self, item=item)
```
t20100: Let's keep it private:
```suggestion
        self._icon = LegendItemIcon(parent=self, item=item)
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self.layout().addWidget(self.icon)
+
+        self.label = qt.QLabel(self._item.getName())
```
t20100:
```suggestion
        self._label = qt.QLabel(self._item.getName())
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+            items.ItemChangedType.LINE_WIDTH,
+            items.ItemChangedType.LINE_STYLE,
+            items.ItemChangedType.COLOR,
+            items.ItemChangedType.ALPHA,
+            items.ItemChangedType.HIGHLIGHTED,
+            items.ItemChangedType.HIGHLIGHTED_STYLE,
+        ):
+            self._update()
+
+    def item(self) -> items.Item | None:
```
t20100: Project naming convention:
```suggestion
    def getItem(self) -> items.Item | None:
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        return self._itemRef
+
+    def _update(self):
+        _item = self._item
```
t20100: No need to make local variable _private
```suggestion
        item = self._item
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        if event.button() == qt.Qt.LeftButton:
+            if self._item:
+                self._item.setVisible(not self._item.isVisible())
+
+
+class LegendItemList(qt.QWidget):
+    def __init__(self, parent: PlotWidget):
+        super().__init__(parent, qt.Qt.Window)
+        self.setFixedWidth(200)
+        self._layout = qt.QVBoxLayout(self)
```
t20100: No need to keep a reference on the layout, it can be retrieved with `self.layout()`

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _onLabelClicked(self, event):
+        if event.button() == qt.Qt.LeftButton:
+            if self._item:
+                self._item.setVisible(not self._item.isVisible())
+
+
+class LegendItemList(qt.QWidget):
+    def __init__(self, parent: PlotWidget):
+        super().__init__(parent, qt.Qt.Window)
```
t20100: Either pass flags from `__init__` or use the default:
```suggestion
        super().__init__(parent)
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self.setFixedWidth(200)
```
t20100: Do use fixed size, best to let Qt handles it: rendering changes depending on OS & screen...

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self._layout = qt.QVBoxLayout(self)
+        self._plot: PlotWidget | None = None
```
t20100: Best to keep a weakreference

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,226 @@
...
+        self.icon = LegendItemIcon(parent=self, item=item)
+        self.layout().addWidget(self.icon)
+
+        self.label = qt.QLabel(self._item.getName())
+        self.label.setAlignment(qt.Qt.AlignLeft | qt.Qt.AlignVCenter)
+        self.layout().addWidget(self.label)
+        self.layout().addStretch()
+
+        self.label.setToolTip("Click to toggle visibility")
+        self.label.mousePressEvent = self._onLabelClicked
```
t20100: Not a good idea to "monkey patch" a Qt method.
Since it would make sense to also accept the press on the icon, you can implement `mousePressEvent` in this class.

src/silx/gui/plot/LegendItem.py
abmajith: the weekref not pointing to the item attribute class, it throws error, for example when i call ite,getName(), throws no methods called getName() something

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+from ..widgets.LegendIconWidget import LegendIconWidget
+from . import items
+from .items.core import HighlightedMixIn
+from .PlotWidget import PlotWidget
+
+_logger = logging.getLogger(__name__)
+
+
+class LegendItemWidget(qt.QWidget):
+    def __init__(self, parent, item: items):
```
t20100: `items` is a module, not  a class
```suggestion
    def __init__(self, parent, item: items.Item):
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+        """Binds this widget to the signals of a parent PlotWidget."""
+        self._plot = plotWidget
+        self._plot.sigItemAdded.connect(self._updateAllItemsList)
+        self._plot.sigItemRemoved.connect(self._onItemRemoved)
+        self._updateAllItemsList()
+        self.show()
+
+    def _clearLayout(self, layout):
+        """Helper to clear all widgets from a layout."""
+        if layout is not None:
```
t20100: No need to pass the layout as an argument:
```suggestion
    def _clear(self):
        """Helper to clear all widgets from a layout."""
        layout = self.layout()
        if layout is not None:
```
+ this does not clear `self._itemWidgets`

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+                child = layout.takeAt(0)
+                if child.widget():
+                    child.widget().deleteLater()
+
+    def _updateAllItemsList(self):
+
+        if self._plot is None:
+            return
+        self._clearLayout(self.layout())
+        _activeItem: tuple = self._plot.getItems()
```
t20100: No `_` prefix for local variables (there's others in this PR).
Already mentionned in https://github.com/silx-kit/silx/pull/4404#discussion_r2344427473

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+
+        for item in _activeItem:
+            self._onItemAdded(item)
+
+    def _onItemRemoved(self, item: items.Item):
+        if item in self._itemWidgets:
+            self.layout().removeWidget(self._itemWidgets[item])
```
t20100: Is this enough to fully remove the widget? What  is it's `parent` after the removal?
If it's the `LegendItemList` then it's not deleted.

Either this is enough and there is no need for `deletedLater` in `_clearLayout` or it's fully destroying the widget and something like in `_clearLayout` should be enough.

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+        self._icon = LegendIconWidget(parent=self)
+        self.layout().addWidget(self._icon)
+
+        self._label = qt.QLabel(self._item.getName())
+        self._label.setAlignment(qt.Qt.AlignLeft | qt.Qt.AlignVCenter)
+        self.layout().addWidget(self._label)
+        self.layout().addStretch()
+
+        self._label.setToolTip("Click to toggle visibility")
+        self._label.mousePressEvent = self._onLabelClicked
```
t20100: Already mentioned in https://github.com/silx-kit/silx/pull/4404#discussion_r2344449940
This is not a good pattern. you can implement `mousePressEvent` in this class instead

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+        self._label.setCursor(qt.Qt.PointingHandCursor)
```
t20100: You can set it for the whole widget
```suggestion
        self.setCursor(qt.Qt.PointingHandCursor)
```

src/silx/gui/plot/LegendItem.py
```diff
@@ -0,0 +1,156 @@
...
+            color = qt.QColor.fromRgbF(
+                color_rgba[0],
+                color_rgba[1],
+                color_rgba[2],
+                color_rgba[3],
+            )
+            self._icon.setLineColor(color)
+            self._icon.setSymbolColor(color)
+        if isinstance(item, items.ColormapMixIn):
+            self._icon.setColormap(item.getColormap())
```
t20100: :+1:

src/silx/gui/plot/LegendsWidget.py
```diff
@@ -0,0 +1,156 @@
...
+        self._update()
+
+    def _itemChanged(self, event: items.ItemChangedType):
+        """Handle update of curve/scatter item
+        :param event: Kind of change
+        """
+        if event in (
+            items.ItemChangedType.VISIBLE,
+            items.ItemChangedType.HIGHLIGHTED,
+            items.ItemChangedType.NAME,
```
t20100, line 39: Quite a few change events seems to be missing here: line, symbol, color...

t20100 (commented):
Second round of comments:
Much simpler with a single LegendItemWidget, ch and update only on add/remove!

There's something to check about removing/clearing widget, this needs to be robust: it should not let widgets alive that are no longer displayed.

There's a few comments from last review that where not taken into account nor commented. Before requesting a new review, can you address all comments, either by applying them if you agree or by commenting if you don't agree or need clarifications.

src/silx/gui/plot/LegendItem.py
t20100: Can you be more explicit?
Without the code and the complete error message, it is impossible to help.

src/silx/gui/plot/LegendItem.py
abmajith: okay for the consistency, I will follow the same mechanism as in _clearLayout

src/silx/gui/plot/LegendItem.py
abmajith:
```
python3 examples/plotLegendLists.py 
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 162, in _updateAllItemsList
    self._onItemAdded(item)
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 171, in _onItemAdded
    legendIcon = LegendItemWidget(None, item)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 29, in __init__
    self._label = qt.QLabel(self._item.getName())
                            ^^^^^^^^^^^^^^^^^^
AttributeError: 'weakref.ReferenceType' object has no attribute 'getName'
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 162, in _updateAllItemsList
    self._onItemAdded(item)
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 171, in _onItemAdded
    legendIcon = LegendItemWidget(None, item)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 29, in __init__
    self._label = qt.QLabel(self._item.getName())
                            ^^^^^^^^^^^^^^^^^^
AttributeError: 'weakref.ReferenceType' object has no attribute 'getName'
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 162, in _updateAllItemsList
    self._onItemAdded(item)
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 171, in _onItemAdded
    legendIcon = LegendItemWidget(None, item)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/noordhee/DAU_Projects/silx/src/silx/gui/plot/LegendItem.py", line 29, in __init__
    self._label = qt.QLabel(self._item.getName())
                            ^^^^^^^^^^^^^^^^^^
AttributeError: 'weakref.ReferenceType' object has no attribute 'getName'
```

abmajith: when i store the item with weakref, this is the error i got

t20100: `weakref.ref` cannot be accessed as the item it references.
You have to "call" it to get back the item:

```python
myweakref = weakref.ref(myitem)
...
item = myweakref()
if item is not None:  # It returns None if the item has been deleted
    item.getName()
```
An alternative is to use `weakref.proxy`, but best IMO to know if the object still exists

Look at: https://docs.python.org/3/library/weakref.html#weak-reference-objects

abmajith:
Hi @t20100,

I tested the LegendItem gui, 


there is one discomfort right now, When I am closing plot window that linked to LegendItem (as in this example/plotLegendItem), the legendItem has to be closed separately, Is this a problem, or we have to set up the parent widget to plot window so that it will automatically closed?

examples/plotLegendLists.py
```diff
@@ -0,0 +1,70 @@
+import numpy
+
+from silx.gui import qt
+from silx.gui.plot import Plot2D
+from silx.gui.plot.LegendItem import LegendItemList
+
+
+class FloatingLegend(LegendItemList):
```
t20100: This adds quite some code to the demo code that is out of the scope of showing the legend list widget.

And it has display problem:

Image

Best to use a QDockWidget to keep it simple

examples/plotLegendLists.py
```diff
@@ -0,0 +1,70 @@
...
+
+plot = Plot2D()
+plotLayout.addWidget(plot)
+
+
+legend = FloatingLegend(plot)
+plot.resizeEvent = lambda event: (
+    Plot2D.resizeEvent(plot, event),
+    legend.updatePosition(),
+)
```
t20100: Monkey-patching is a bad pattern, best to inherit and override.

Anyway here, best to use a QDockWidget so this is not needed.

src/silx/gui/plot/test/test_legendswidget.py
```diff
@@ -0,0 +1,45 @@
...
+    item_widget = legend_widget._itemWidgets[item]
+    assert item.isVisible() is True
+    from silx.gui.utils.testutils import QTest
+
+    QTest.mouseClick(item_widget, qt.Qt.LeftButton)
+    qapp.processEvents()
+    assert item.isVisible() is False
+    QTest.mouseClick(item_widget, qt.Qt.LeftButton)
+    qapp.processEvents()
+    assert item.isVisible() is True
```
t20100:
```suggestion
def test_visibility_toggle(qapp, qapp_utils, plot_with_legend):
    plot, legend_widget = plot_with_legend
    plot.addCurve([0, 1], [0, 1], legend="test_item")
    qapp.processEvents()
    item = plot.getCurve("test_item")
    item_widget = legend_widget._itemWidgets[item]
    assert item.isVisible() is True

    qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
    assert item.isVisible() is False
    qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
    assert item.isVisible() is True
```

There's a `qapp_utils` fixture which gives access to `mouseClick` (which includes an equivalent to `processEvents`).
BTW, best to avoid imports inside functions unless strictly necessary

PR #4313, opened 2025-04-17
dictdump: support external data and VDS in dictdump link target values

woutdenolf:
Checklist:

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

**Before this MR**

In `nexus_to_h5_dict` and `h5_to_nexus_dict`, a link is defined as `">link_name": <target>`.

- HDF5 dict: target of type `h5py.SoftLink` or `h5py.ExternalLink`
- Nexus dict: target of type `str`

- The target can be a **single string**: URL to an HDF5 dataset which leads to a `h5py.SoftLink` or `h5py.ExternalLink`.

**After this MR**

In `nexus_to_h5_dict` and `h5_to_nexus_dict`, a link is defined as `">link_name": <target>`.

- HDF5 dict: target of type  `h5py.SoftLink`, `h5py.ExternalLink`, `h5py.VirtualLayout`, `silx.io.dictdumplink.ExternalBinaryLink`
- Nexus dict: target of type `str`, `DataUrl`, `Sequence[str|DataUrl]`, `Mapping`

- The target can be a **single string or DataUrl**: URL to an HDF5 dataset which leads to a `h5py.SoftLink` or `h5py.ExternalLink`. More generic URL's are supported for EDF and TIFF which lead to a `ExternalBinaryLink` and HDF5 dataset slicing which leads to a `h5py.VirtualLayout`.
- The target can be a **list of strings or DataUrls**. Several URL's merged in a single virtual or external binary dataset (concatenate along the first or a new dimension).
- The target can be a **dictionary (type `VdsModelV1` or `ExternalLinkModelV1`)**. Full control on how multiple sources are merged (VDS or external binary).

Added documentation for nxtodict and h5todict:

image

PR #4405, opened 2025-09-11
dictdump updates

t20100:
Couple of suggestions splited by commit on #4313

### DIFF
--- a/src/silx/io/dictdumplinks/_from_serialized.py
+++ b/src/silx/io/dictdumplinks/_from_serialized.py
@@ -131,9 +131,9 @@ def _url_to_hdf5_link(source: DataUrl, target: str | DataUrl) -> Hdf5LinkType | 
         else:
             return _url_to_external_link(target)
     elif file_type == "tiff":
-        return _tiff_url_to_external_data(source, target)
+        return _tiff_urls_to_external_data(source, [target])
     elif file_type == "edf":
-        return _fabio_url_to_external_data(source, target)
+        return _fabio_urls_to_external_data(source, [target])
     else:
         return None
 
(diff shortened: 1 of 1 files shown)


src/silx/io/dictdumplinks/_external_binary.py
```diff
@@ -19,6 +19,14 @@ class ExternalLinkModelV1(BaseModel):
     dtype: Any  # DTypeLike gives pydantic.errors.PydanticUserError on Python < 3.12.
     sources: list[tuple[str, int, int]]  # file name, byte offset, byte size
 
+    def tolink(self, source: DataUrl) -> "ExternalBinaryLink":
```
t20100, line 22: Could also be 
`def deserialize_external_binary(source: DataUrl, target: Mapping)` if you prefer to keep helper functions.
But simpler to understand if called directly rather than going through `deserialize_mapping` and then casting the result.

src/silx/io/dictdumplinks/_from_serialized.py
```diff
@@ -88,6 +90,9 @@ def link_from_serialized(
         # Already a link instance.
         return target
 
+    if not isinstance(source, DataUrl):
+        source = DataUrl(source)
```
t20100, line 94: Do it once early

src/silx/io/dictdumplinks/_from_serialized.py
```diff
@@ -129,20 +129,18 @@ def _url_to_hdf5_link(
...
             return _url_to_vds(source, target)
         else:
-            return _url_to_external_link(source, target)
+            return _url_to_external_link(target)
     elif file_type == "tiff":
-        return _tiff_url_to_external_data(source, target)
+        return _tiff_urls_to_external_data(source, [target])
     elif file_type == "edf":
-        return _fabio_url_to_external_data(source, target)
+        return _fabio_urls_to_external_data(source, [target])
```
t20100, line 136: Does the same

woutdenolf: Good catch. In the beginning there used to be a difference between `target` and `[target]`. But this is no longer the case, they are equivalent now.

In the tests in `src/silx/io/test/test_dictdumplinks.py` we should probably expand parameters like this

```
@pytest.mark.parametrize(
    "nfiles", [1, 3], ids=lambda val: f"{val}_file" if val == 1 else f"{val}_files"
)
```

but there should be two types of 1: `target` and `[target]`.

src/silx/io/dictdumplinks/_external_binary.py
woutdenolf: I like the `def tolink`

We should make an abstract base model for this.

PR #4407, opened 2025-09-12
pyproject.toml: Add uv tool field to tell that the project is not managed by uv

loichuder:
Checklist:

<hr>

This is to be able to use `uv` seamlessly (i.e. so thatr `uv` does not try to do too much by thinking that it needs to handle the project its way)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,9 +114,12 @@ skip = ["*-win32", "*-manylinux_i686", "pp*", "*musllinux*"]
 
 
 [tool.black]
-required-version = 25
+required-version = "25"
 target-version = ["py310", "py311", "py312", "py313"]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
 testpaths = [ "src/silx" ]
+
+[tool.uv]
+managed = false


pyproject.toml
```diff
@@ -114,9 +114,12 @@ skip = ["*-win32", "*-manylinux_i686", "pp*", "*musllinux*"]
...
-required-version = 25
+required-version = "25"
 target-version = ["py310", "py311", "py312", "py313"]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
 testpaths = [ "src/silx" ]
+
+[tool.uv]
+managed = false
```
loichuder, line 125: https://docs.astral.sh/uv/reference/settings/#managed

pyproject.toml
```diff
@@ -114,9 +114,12 @@ skip = ["*-win32", "*-manylinux_i686", "pp*", "*musllinux*"]
 
 
 [tool.black]
-required-version = 25
+required-version = "25"
```
loichuder, line 117: According to https://www.schemastore.org/partial-black.json, this is supposed to be a string

PR #4408, opened 2025-09-12
silx view: Remove Image Stack visualization for 3D+ datasets

loichuder:
Checklist:

<hr>

Since the `Image` visualization is doing the same job and it can have performance issues since it wants to load the whole stack.

The `StackView` widget is unchanged.


t20100 (commented):
Technically `STACK_MODE` is part of the API and shows-up in the [documentation](https://www.silx.org/doc/silx/2.2.2/modules/gui/data/dataviewer.html).

So would be best to go through deprecation:
Printing a deprecation warning when receiving `STACK_MODE` in `DataViewer` method and convert it to `IMAGE_MODE` should be too complex.

src/silx/gui/data/DataViewer.py
```diff
@@ -349,27 +349,45 @@ def getViewFromModeId(self, modeId):
...
         :rtype: silx.gui.data.DataViews.DataView
         """
+        if modeId == STACK_MODE:
+            deprecated_warning(
+                "Argument",
+                "DataViews.STACK_MODE",
+                replacement="DataViews.IMAGE_MODE",
+                since_version="3.0.0",
+            )
+            return None
```
loichuder, line 359: I prefered to follow the docstring instead of setting `modeId` to `IMAGE_MODE`: `Return None if modeId does not correspond to an existing view`

src/silx/gui/data/DataViewer.py
```diff
@@ -607,6 +624,15 @@ def replaceView(self, modeId, newView):
...
         :return: True if replacement was successful, else False
         """
+        if modeId == STACK_MODE:
+            deprecated_warning(
+                "Argument",
+                "DataViews.STACK_MODE",
+                replacement="DataViews.IMAGE_MODE",
+                since_version="3.0.0",
+            )
+            return False
```
loichuder, line 634: Same. I don't think we want to replace `IMAGE_MODE` if `STACK_MODE` is asked.

src/silx/gui/data/DataViewer.py
t20100: Makes sense!

src/silx/gui/data/DataViewer.py
t20100: Indeed :sweat_smile:

PR #4414, opened 2025-09-12
silx.io.dictdumplinks: add VdsUrlsModelV1 schema

woutdenolf:
Checklist:

<hr>

In addition to

```python
">images_roi": [
  "data0.h5?path=/group/dataset&slice=:,20:30,40:50",
  "data1.h5?path=/group/dataset&slice=:,20:30,40:50",
  "data2.h5?path=/group/dataset&slice=:,20:30,40:50"
]
```

this MR adds the "vds_urls_v1" schema in case all sources are the same and you want to avoid opening files

```python
">images_roi": {
  "dictdump_schema": "vds_urls_v1",
  "source_shape": (5, 50, 60),
  "source_dtype": "uint16",
  "sources": [
      "data0.h5?path=/group/dataset&slice=:,20:30,40:50",
      "data1.h5?path=/group/dataset&slice=:,20:30,40:50",
      "data2.h5?path=/group/dataset&slice=:,20:30,40:50"
    ],
}
```

Renamed "virtual_dataset_v1" to "vds_v1" in analogy.

### DIFF
--- a/src/silx/io/test/test_dictdumplinks.py
+++ b/src/silx/io/test/test_dictdumplinks.py
@@ -404,3 +404,52 @@ def _assert_single_hdf5_source(tmp_path, target_type, source_index, target_index
             assert isinstance(link, h5py.HardLink)
         else:
             assert isinstance(link, h5py.ExternalLink)
+
+
+@pytest.mark.parametrize(
+    "target_type",
+    ["external_binary_link_v1", "external_binary_link_v1_extra", "string"],
+)
+def test_single_edf_source(tmp_path, target_type):
+    data = numpy.ones((3, 2), dtype=numpy.uint16)
+    data_file = str(tmp_path / "data.edf")
+    with EdfImage(data=data) as edfimage:
+        edfimage.write(data_file)
+
+    if target_type == "external_binary_link_v1_extra":
+        extra_dim = True
+        target = {
+            "dictdump_schema": "external_binary_link_v1",
+            "dtype": "uint16",
+            "shape": (1, 3, 2),
+            "sources": [{"file_path": "data.edf", "offset": 512, "size": 12}],
+        }
+    elif target_type == "external_binary_link_v1":
+        extra_dim = False
+        target = {
+            "dictdump_schema": "external_binary_link_v1",
+            "dtype": "uint16",
+            "shape": (3, 2),
+            "sources": [{"file_path": "data.edf", "offset": 512, "size": 12}],
+        }
+    elif target_type == "string":
+        extra_dim = False
+        target = "data.edf"
+    else:
+        raise ValueError(target_type)
+
+    master_file = tmp_path / "master.h5"
+    link = link_from_serialized(f"{master_file}::/data", target)
+    with h5py.File(master_file, "w") as h5f:
+        link.create(h5f, "data")
+
+    with h5py.File(master_file, "r") as h5f:
+        link = h5f.get("data", getlink=True)
+        dset = h5f["data"]
+        is_virtual = dset.is_virtual
+        data_final = dset[()]
+        if extra_dim:
+            data = data[numpy.newaxis, ...]
+        numpy.testing.assert_equal(data, data_final)
+        assert isinstance(link, h5py.HardLink)
+        assert is_virtual is False


src/silx/io/dictdumplinks/_vds.py
```diff
@@ -59,6 +61,37 @@ def tolink(self, source: DataUrl) -> h5py.VirtualLayout:
         return vds_layout
 
 
+class VdsUrlsModelV1(Hdf5LinkModel, arbitrary_types_allowed=True):
+    dictdump_schema: Literal["vds_urls_v1"]
+    source_shape: tuple[int, ...]
+    source_dtype: (
+        Any  # DTypeLike gives pydantic.errors.PydanticUserError on Python < 3.12.
```
t20100, line 82: 👌

src/silx/io/dictdumplinks/_vds.py
```diff
@@ -59,6 +61,37 @@ def tolink(self, source: DataUrl) -> h5py.VirtualLayout:
...
+    @field_validator("sources", mode="before")
+    @classmethod
+    def as_dataurl(
+        cls, value: DataUrl | list[DataUrl] | str | Sequence[str]
+    ) -> RawDsetIndex:
+        if isinstance(value, str):
+            return DataUrl(value)
+        if isinstance(value, (list, tuple)) and isinstance(value[0], str):
+            return [DataUrl(s) for s in value]
+        return value
```
t20100:
```suggestion
    def as_dataurl(
        cls, value: DataUrl | Sequence[DataUrl] | str | Sequence[str]
    ) -> RawDsetIndex:
        if isinstance(value, str):
            return DataUrl(value)
        if isinstance(value, Sequence) and isinstance(value[0], str):
            return [DataUrl(s) for s in value]
        return value
```
🤷

src/silx/io/dictdumplinks/_vds.py
```diff
@@ -59,6 +61,37 @@ def tolink(self, source: DataUrl) -> h5py.VirtualLayout:
...
+
+    def tolink(self, source: DataUrl) -> h5py.VirtualLayout:
+        if isinstance(self.sources, DataUrl):
+            target_schema = hdf5_url_to_vds_schema(
+                source, self.sources, self.source_shape, self.source_dtype
+            )
+        else:
+            target_schema = hdf5_urls_to_vds_schema(
+                source, self.sources, self.source_shape, self.source_dtype
+            )
```
t20100: Passing `self.source_*` as `target_*` args is misleading IMO, best to make it explicit:

```suggestion
            target_schema = hdf5_urls_to_vds_schema(
                source,
                targets=self.sources,
                target_shape=self.source_shape,
                target_dtype=self.source_dtype,
            )
```

or to use `source` only for "target" (since it is used this way for VDS in h5py) and use something else for `source` when it refer to the URL of the link (e.g,`link` or `link_url`)

src/silx/io/dictdumplinks/_parse_hdf5_targets.py
```diff
@@ -1,24 +1,37 @@
...
-    """Single HDF5 dataset: keep original shape (no new axis)."""
-    datasets = [_read_target_info(source, target, allow_new_axis=False)]
-    target_desc = _build_vds_schema(datasets)
+def hdf5_url_to_vds(
+    source: DataUrl,
+    target: DataUrl,
+    target_shape: tuple[int, ...] | None = None,
+    target_dtype: DTypeLike | None = None,
+) -> h5py.VirtualLayout:
+    """Single HDF5 dataset: keep original shape (no new axis).
```
t20100, line 18: Not for this PR, but it's not the case for external binary data, it always add a dimension, right?
Maybe good to have a consistent behavior between VDS and external data.

woutdenolf: It should keep the original shape. If not then that is a bug. I thought the tests covered this.

woutdenolf: I'm confused as to what you are saying. The comment is about external binary data but the function is related to virtual datasets. Can you describe the issue and perhaps code to illustrate it?

woutdenolf:
```python
import h5py
import numpy
from silx.io.dictdump import dicttonx

with h5py.File("data.h5", "w") as h5f:
    h5f["data"] = numpy.zeros((10,20), dtype=numpy.uint16)


data = {
    ">data": {
        "dictdump_schema": "vds_urls_v1",
        "source_shape": (10,20),
        "source_dtype": "uint16",
        "sources": [
            "data.h5?path=/data"
        ],
    }
}

with h5py.File("master.h5", "w") as h5f:
    dicttonx(
        treedict=data,
        h5file=h5f,
        h5path="/",
        update_mode="replace",
        add_nx_class=True,
    )
```

```
(Pdb) vsource
VdsSourceV1(file_path='data.h5', data_path='/data', shape=(10, 20), dtype='uint16', source_index=None, target_index=slice(0, 1, None))
```

source_index can be `None` because it is part of `RawDsetIndexItem`. This is fine for numpy but not for h5py VDS it seems.

woutdenolf: I made a fix + test for `None` in virtual dataset indices.

woutdenolf: We now have specific tests for single-source HDF5 and EDF.

Apart from the `None` index handling in VDS that I fixed it seems consistent to me.

When you have a single HDF5 source, you can choose to keep the original shape (no new axis) or not. When the link target is a single string the shape is preserved:

- VDS: `".../data.h5?path=/data&slice=:"`
- External binary: `".../data.edf"`
- External link: `".../data.h5?path=/data"`

PR #4419, opened 2025-09-14
build: Switch to building a single executable

mretegan:
Checklist:

<hr>

### DIFF
--- a/package/pyinstaller/create-installer.iss.template
+++ b/package/pyinstaller/create-installer.iss.template
@@ -26,7 +26,8 @@ Source: "dist\silx\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs
 Source: "silx.ico"; DestDir: "{app}"
 
 [Icons]
-Name: "{group}\silx"; Filename: "{app}\silx-view.exe"; IconFilename: "{app}\silx.ico"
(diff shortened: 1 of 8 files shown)


package/pyinstaller/create-installer.iss.template
```diff
@@ -26,7 +26,8 @@ Source: "dist\silx\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs
 Source: "silx.ico"; DestDir: "{app}"
 
 [Icons]
-Name: "{group}\silx"; Filename: "{app}\silx-view.exe"; IconFilename: "{app}\silx.ico"
+Name: "{group}\silx view"; Filename: "{app}\silx.exe"; Parameters: "view"; IconFilename: "{app}\silx.ico"
+Name: "{group}\silx compare"; Filename: "{app}\silx.exe"; Parameters: "compare"; IconFilename: "{app}\silx.ico"
```
t20100: Just checked, the images to compare needs to be passed to the command line, there is no menu to open them from the GUI, so no interest for now to expose `silx compare` as an icon.
I would remove for now it but keep the `silx` group.

```suggestion

```

t20100 (approved):
Thanks for looking into this and simplify the pyinstaller config.

package/pyinstaller/create-installer.iss.template
mretegan: Thank you for checking. I have not gone that far; I just openned the app :smile:

PR #4425, opened 2025-09-23
silx.gui.plot: add X inversion action to PlotWindow and ImageToolbar

### LINKED ISSUE #4253: [GUI] Offer the opportunity to flip the image horizontally in addition to vertically using the pull-down menu
Requested by BM29 (among other)

loichuder:
Checklist:

<hr>

Fix #4253 

[Screencast from 2025-09-23 10-43-25.webm](https://github.com/user-attachments/assets/308ad11c-c928-4a52-a3ef-df4d04bea618)

### DIFF
--- a/src/silx/gui/plot/PlotToolButtons.py
+++ b/src/silx/gui/plot/PlotToolButtons.py
@@ -42,7 +42,6 @@ __date__ = "27/06/2017"
 import functools
 import logging
 from typing import TypedDict
-from weakref import WeakMethod
 
 from .. import icons
 from .. import qt
@@ -142,15 +141,12 @@ class _AxisOriginToolButton(PlotToolButton):
     ):
         super().__init__(parent=parent, plot=plot)
 
-        # https://github.com/silx-kit/silx/pull/4425#discussion_r2378901544
-        inversionTriggered = WeakMethod(self.setAxisInverted)
-
         disableInversionAction = self._createAction(False)
-        disableInversionAction.triggered.connect(lambda: inversionTriggered(False))
+        disableInversionAction.triggered.connect(self._disableInversion)
         disableInversionAction.setIconVisibleInMenu(True)
 
         enableInversionAction = self._createAction(True)
-        enableInversionAction.triggered.connect(lambda: inversionTriggered(True))
+        enableInversionAction.triggered.connect(self._enableInversion)
         enableInversionAction.setIconVisibleInMenu(True)
 
         menu = qt.QMenu(self)
@@ -186,6 +182,16 @@ class _AxisOriginToolButton(PlotToolButton):
         if axis is not None:
             # This will trigger _axisInvertedChanged
             axis.setInverted(inverted)
+
+    def _enableInversion(self):
+        # Wrapper function to avoid mixed python/qt circular reference
+        # https://github.com/silx-kit/silx/pull/4425#discussion_r2378901544
+        self.setAxisInverted(True)
+
+    def _disableInversion(self):
+        # Wrapper function to avoid mixed python/qt circular reference
+        # https://github.com/silx-kit/silx/pull/4425#discussion_r2378901544
+        self.setAxisInverted(False)
 
     def _axisInvertedChanged(self, inverted: bool):
         state = self._getState(inverted)


src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -123,73 +124,99 @@ def _keepDataAspectRatioChanged(self, aspectRatio):
...
-            self.STATE = {}
-            # is down
-            self.STATE[False, "icon"] = icons.getQIcon("plot-ydown")
-            self.STATE[False, "state"] = "Y-axis is oriented downward"
-            self.STATE[False, "action"] = "Orient Y-axis downward"
-            # keep ration
-            self.STATE[True, "icon"] = icons.getQIcon("plot-yup")
-            self.STATE[True, "state"] = "Y-axis is oriented upward"
-            self.STATE[True, "action"] = "Orient Y-axis upward"
+class _AxisOriginToolButton(PlotToolButton):
```
loichuder, line 136: I completely refactored `YAxisOriginToolButton` to have a base class to abstract the common behaviour.

The only change between the `XAxisOriginToolButton` and the `YAxisOriginToolButton` is the axis to act on (abstracted by `getAxis`) and the icons/labels (abstracted by `getState`).

Note that `self.STATE` was converted to TypedDict, which should be fine given these were introducted in Python 3.8.

Note also that `STATE` is no longer lazy-loaded. Is this a problem ?

src/silx/gui/plot/PlotWindow.py
```diff
@@ -449,25 +452,19 @@ def _createToolBar(self, title, parent):
...
-                toolbar.addAction(obj)
-            else:
-                # Add action for toolbutton in order to allow changing
-                # visibility (see doc QToolBar.addWidget doc)
-                if obj is self.keepDataAspectRatioButton:
-                    self.keepDataAspectRatioAction = toolbar.addWidget(obj)
-                elif obj is self.yAxisInvertedButton:
-                    self.yAxisInvertedAction = toolbar.addWidget(obj)
-                else:
-                    raise RuntimeError("unknow action to be defined")
```
loichuder, line 470: Rewritten this part that was a bit hard to process. For this, instead of inserting manually after `self.colormapAction`, I used `insertWidget` to insert before `self.colorbarAction` (`insertWidget` needs the action that comes after not before).

src/silx/gui/plot/tools/toolbars.py
```diff
@@ -162,11 +165,12 @@ def getKeepDataAspectRatioButton(self):
...
-    def getYAxisInvertedButton(self):
-        """Returns the QToolButton controlling Y axis orientation.
+    def getXAxisInvertedButton(self) -> PlotToolButtons.XAxisOriginToolButton:
+        """Returns the QToolButton controlling X axis orientation."""
+        return self._xAxisInvertedButton
 
-        :rtype: QToolButton
-        """
+    def getYAxisInvertedButton(self) -> PlotToolButtons.YAxisOriginToolButton:
+        """Returns the QToolButton controlling Y axis orientation."""
```
loichuder, line 173: I mirrored getters when present but I don't know if that is really useful :shrug:

src/silx/resources/gui/icons/plot-xleft.svg
loichuder, line 1: You can review the SVG raw text but I think it is in line with the other icons: I saved them as "Optimized SVG" with Inkscape and exported them as 32x32 PNG.

src/silx/gui/plot/tools/PlotToolButton.py
```diff
@@ -32,31 +32,30 @@
 __date__ = "20/12/2023"
 
 
-import logging
 import weakref
 
 from silx.gui import qt
-
-
-_logger = logging.getLogger(__name__)
```
loichuder, line 41: Was not used

t20100: I would have expected flake8 to spot it 🤷

src/silx/gui/plot/tools/toolbars.py
t20100: Here it sounds good, the buttons need to be accessible from the API

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -123,73 +127,114 @@ def _keepDataAspectRatioChanged(self, aspectRatio):
...
+    """Tool button to switch the axis orientation of a plot."""
 
+    def __init__(self, parent=None, plot=None):
```
t20100:
```suggestion
    def __init__(
        self, parent: qt.QWidget | None = None, plot: PlotWidget | None = None
    ):
```

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -123,73 +127,114 @@ def _keepDataAspectRatioChanged(self, aspectRatio):
...
         super().__init__(parent=parent, plot=plot)
 
-        upwardAction = self._createAction(True)
-        upwardAction.triggered.connect(self.setYAxisUpward)
-        upwardAction.setIconVisibleInMenu(True)
+        disableInversionAction = self._createAction(False)
+        disableInversionAction.triggered.connect(lambda: self.setAxisInverted(False))
```
t20100: This creates a mixed python/qt circular reference (lambda -> self -> menu -> disableInversionAction -> signal -> lambda) which may delay proper deletion, so best to avoid.

[WeakMethod](https://docs.python.org/3/library/weakref.html#weakref.WeakMethod) can be used here:

```suggestion
        inversionTriggered = WeakMethod(self.setAxisInverted)
        disableInversionAction.triggered.connect(lambda: inversionTriggered(False))
```

or wrapper methods since `signal.connect(self.slot)` is handled correctly.

Illustration of the problem:
```python
>>> import weakref
>>> from silx import sx
>>> class A(sx.qt.QObject):
...     def __init__(self):
...         super().__init__()
...         self.objectNameChanged.connect(self.changed)
...     def changed(self, name):
...         print(name)
... 
>>> a = A()
>>> r = weakref.ref(a, lambda ref: print('dead', ref))
>>> del a
dead <weakref at 0x7f48eff53ef0; dead>

>>> class A(sx.qt.QObject):
...     def __init__(self):
...         super().__init__()
...         self.objectNameChanged.connect(lambda n: self.changed(n))
...     def changed(self, name):
...         print(name)
... 
>>> a = A()
>>> r = weakref.ref(a, lambda ref: print('dead', ref))
>>> del a
>>> 
>>> r()
<__main__.A object at 0x7f48fe72eb80>
```

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -123,73 +127,114 @@ def _keepDataAspectRatioChanged(self, aspectRatio):
...
+        else:
+            return {
+                "icon": icons.getQIcon("plot-yup"),
+                "state": "Y-axis is oriented upward",
+                "action": "Orient Y-axis upward",
+            }
+
+    def setYAxisUpward(self):
+        deprecated_warning(
+            "Method", name="setYAxisUpward", replacement="setAxisInverted(False)"
```
t20100: `since_version` helps finding deprecated code to remove:
```suggestion
            "Method",
            name="setYAxisUpward",
            since_version="3.0.0",
            replacement="setAxisInverted(False)",
```

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -123,73 +127,114 @@ def _keepDataAspectRatioChanged(self, aspectRatio):
...
+        )
+        self.setAxisInverted(False)
+
+    def setYAxisDownward(self):
+        deprecated_warning(
+            "Method", name="setYAxisDownward", replacement="setAxisInverted(True)"
```
t20100:
```suggestion
            "Method",
            name="setYAxisDownward",
            since_version="3.0.0",
            replacement="setAxisInverted(True)",
```

src/silx/resources/gui/icons/plot-xleft.svg
t20100: Looks good!

src/silx/gui/plot/PlotWindow.py
```diff
@@ -213,6 +213,9 @@ def __init__(
         )
         self.keepDataAspectRatioButton.setVisible(aspectRatio)
 
+        self.xAxisInvertedButton = PlotToolButtons.XAxisOriginToolButton(
```
t20100: Since it is new, I would just keep it private (as all the other should be):
```suggestion
        self._xAxisInvertedButton = PlotToolButtons.XAxisOriginToolButton(
```

src/silx/gui/plot/PlotWindow.py
```diff
@@ -449,25 +452,19 @@ def _createToolBar(self, title, parent):
...
+        toolbar.addActions(self.group.actions())
 
         # Add push buttons to list
-        index = objects.index(self.colormapAction)
-        objects.insert(index + 1, self.keepDataAspectRatioButton)
-        objects.insert(index + 2, self.yAxisInvertedButton)
+        self.keepDataAspectRatioAction = toolbar.insertWidget(
+            self.colorbarAction, self.keepDataAspectRatioButton
+        )
+        self.xAxisInvertedAction = toolbar.insertWidget(
```
t20100:
```suggestion
        self._xAxisInvertedAction = toolbar.insertWidget(
```

src/silx/gui/plot/PlotWindow.py
```diff
@@ -787,11 +784,21 @@ def getKeepDataAspectRatioAction(self):
...
+
+    def getXAxisInvertedAction(self):
+        """Action associated to xAxisInvertedButton.
+        Use this to change the visibility xAxisInvertedButton in the toolbar.
+        (See :meth:`QToolBar.addWidget` documentation).
 
-        :rtype: PlotToolButtons.YAxisOriginToolButton
+        :rtype: actions.PlotAction
         """
+        return self.xAxisInvertedAction
```
t20100:
```suggestion
        return self._xAxisInvertedAction
```

src/silx/gui/plot/PlotWindow.py
```diff
@@ -787,11 +784,21 @@ def getKeepDataAspectRatioAction(self):
         """
         return self.keepDataAspectRatioAction
 
-    def getYAxisInvertedButton(self):
-        """Button to switch the Y axis orientation
+    def getXAxisInvertedButton(self) -> PlotToolButtons.XAxisOriginToolButton:
+        """Button to switch the X axis orientation"""
+        return self.xAxisInvertedButton
```
t20100:
```suggestion
        return self._xAxisInvertedButton
```

src/silx/gui/plot/PlotWindow.py
```diff
@@ -787,11 +784,21 @@ def getKeepDataAspectRatioAction(self):
...
+
+    def getXAxisInvertedAction(self):
+        """Action associated to xAxisInvertedButton.
+        Use this to change the visibility xAxisInvertedButton in the toolbar.
+        (See :meth:`QToolBar.addWidget` documentation).
 
-        :rtype: PlotToolButtons.YAxisOriginToolButton
+        :rtype: actions.PlotAction
```
t20100:
```suggestion
    def getXAxisInvertedAction(self) -> qt.QAction:
        """Action associated to xAxisInvertedButton.

        Use this to change the visibility xAxisInvertedButton in the toolbar.
        (See :meth:`QToolBar.addWidget` documentation).
```

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -123,73 +127,114 @@ def _keepDataAspectRatioChanged(self, aspectRatio):
...
+
+    def _getState(self, inverted: bool) -> _AxisState:
+        if inverted:
+            return {
+                "icon": icons.getQIcon("plot-xleft"),
+                "state": "X-axis goes from right to left",
+                "action": "Orient X-axis from right to left",
+            }
+        else:
+            return {
```
t20100, line 214: Could be an early-return 😁

src/silx/gui/plot/PlotToolButtons.py
t20100: Refactoring sounds good.
The state info is still lazy-loaded, however it's no longer cached.
It shouldn't be an issue: icons are already cached.

src/silx/gui/plot/PlotToolButtons.py
loichuder: Eeeeeh, I don't know. I prefer `else` in this case and I'll now start a pedantic discussion you did not ask for:

We have the following pattern:

```
if cond:
  do A
else:
  do B
```

In case, `A` and `B` are conceptually equivalent (no error treatment, I do this OR I do that), I prefer to keep the equivalence in the code via the `else`.

If `A` is treating an error, or a `None`, i.e. this is doing some check before doing `B`, in this case, I think it makes sense to have

```
if cond:
  do A
  return;

do B
```

`A` and `B` are not equivalent in this case and the asymmetry is reflected in the code

:nerd_face:

t20100: :+1:

t20100 (approved):

loichuder:
Maaan, there is a segfault in the CI that I cannot reproduce locally :face_exhaling:

src/silx/gui/plot/PlotToolButtons.py
loichuder: In the end, I used wrapper functions (108c45f57e6c198284fc691be39a9ba299a0550f) instead of weakref.

It seems to have fixed the elusive segfault in the CI !

t20100: Good!

IMO we should prevent using `lambda` functions, `functools.partial` or anything keeping references to `self` in signals even if that requires some boilerplate methods.

loichuder: Yep. It looks to be recipes for disaster :disappointed:

PR #4427, opened 2025-09-23
silx.gui.data.NXdataWidget: Use axis names as slider labels (when present)

### LINKED ISSUE #4213: Metadata handling in DataViewer
A question similar to #4199 for the DataViewer: Are there any plans to include support for axis metadata (units, labels, scaling) for the DataViewer? This would be a nice addition to its functionality.

loichuder:
Checklist:

<hr>

Fix #4213

When the NXdata has named axes, `silx view` will now use these as labels for the slicing sliders.

image

I only added it for images for now. If we agree that the implementation is sensible, I could add it to other visualizations as well.

### DIFF
--- a/src/silx/gui/data/NumpyAxesSelector.py
+++ b/src/silx/gui/data/NumpyAxesSelector.py
@@ -30,6 +30,7 @@ __license__ = "MIT"
 __date__ = "29/01/2018"
 
 import logging
+from collections.abc import Sequence
 import numpy
 import functools
 from silx.gui.widgets.FrameBrowser import HorizontalSliderWithBrowser
@@ -54,19 +55,31 @@ class _Axis(qt.QWidget):
     axisNameChanged = qt.Signal(object)
     """Emitted when the user change the name of the axis."""
 
-    def __init__(self, parent=None):
+    def __init__(self, number: int, size: int, label: str | None = None, parent=None):
         """Constructor
 
+        :param number: The number of the axis (from the original numpy array)
+        :param size: The size of this axis (0..n)
         :param parent: Parent of the widget
         """
         super().__init__(parent)
-        self.__axisNumber = None
+        self.__axisNumber = number
         self.__customAxisNames = set()
-        self.__label = qt.QLabel(self)
+
+        if label is None:
+            axisLabel = f"Dimension {number}"
+        else:
+            axisLabel = f"{label} (Dim{number})"
+
+        self.__label = qt.QLabel(axisLabel, parent=self)
+
         self.__axes = qt.QComboBox(self)
         self.__axes.currentIndexChanged[int].connect(self.__axisMappingChanged)
+
         self.__slider = HorizontalSliderWithBrowser(self)
         self.__slider.valueChanged[int].connect(self.__sliderValueChanged)
+        self.__slider.setMaximum(size - 1)
+
         layout = qt.QHBoxLayout(self)
         layout.setContentsMargins(0, 0, 0, 0)
         layout.addWidget(self.__label)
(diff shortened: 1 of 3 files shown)


src/silx/gui/data/NumpyAxesSelector.py
loichuder, line 1: Diff is hard to parse here as I added typing everywhere. Sorry about that.

I'll underline changes when relevant.

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -75,39 +89,21 @@ def __init__(self, parent=None):
...
-    def slider(self):
-        """Returns the slider used to display axes location.
-
-        :rtype: HorizontalSliderWithBrowser
-        """
+    def slider(self) -> HorizontalSliderWithBrowser:
+        """Returns the slider used to display axes location."""
         return self.__slider
 
-    def setAxis(self, number, position, size):
```
loichuder, line 85: `setAxis` was always called just after `_Axis` was constructed so I merged it in the constructor.

Note that the `position` arg was removed since it was not used.

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -54,19 +55,32 @@ class _Axis(qt.QWidget):
     axisNameChanged = qt.Signal(object)
     """Emitted when the user change the name of the axis."""
 
-    def __init__(self, parent=None):
+    def __init__(self, number: int, size: int, label: str | None = None, parent=None):
```
loichuder, line 58: The new argument is `label`. `number` and `size` come from the merging of `setAxis` in the constructor.

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -244,9 +233,10 @@ def __init__(self, parent=None):
 
         self.__data = None
         self.__selectedData = None
-        self.__axis = []
-        self.__axisNames = []
-        self.__customAxisNames = set()
+        self.__axis: list[_Axis] = []
+        self.__axisNames: list[str] = []
+        self.__customAxisNames: set[str] = set()
+        self.__labels: list[str] = []
```
loichuder, line 238: New member `__labels` to store the axes names.

You may ask: is it not duplicating the other members `_axisNames` and `_customAxisNames` ?

And then I'll answer that I have no clue. 

`_axisNames` seem to the options of the axis combo box (`Y` for curve, `X,Y` for image). So it appears to be the names of the axes **of the plot** rather than of the data. Any way, these only appear when the axis combo boxes are displayed which is not the case for NXdata widget.

`_customAxisNames` are, according to the docstring, "names of axes which can be custom by the user and provided to the view." which I don't understand. It also seem to be used to check the visibility of the sliders?

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -30,6 +30,7 @@
 __date__ = "29/01/2018"
 
 import logging
+from typing import Sequence
```
t20100:
```suggestion
from collections.abc import Sequence
```

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -54,19 +55,32 @@ class _Axis(qt.QWidget):
     axisNameChanged = qt.Signal(object)
     """Emitted when the user change the name of the axis."""
 
-    def __init__(self, parent=None):
+    def __init__(self, number: int, size: int, label: str | None = None, parent=None):
         """Constructor
 
+        :param int number: The number of the axis (from the original numpy
+            array)
+        :param int size: The size of this axis (0..n)
```
t20100:
```suggestion
        :param number: The number of the axis (from the original numpy array)
        :param size: The size of this axis (0..n)
```

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -123,22 +119,20 @@ def setAxisName(self, axisName):
...
-
-        :rtype: str
         """
         index = self.__axes.currentIndex()
         if index == -1:
             return ""
         return self.__axes.itemData(index)
 
-    def setAxisNames(self, axesNames):
+    def setAxisNames(self, axesNames: list[str]):
```
t20100, line 131:
```suggestion
    def setAxisNames(self, axesNames: Sequence[str]):
```

src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -148,18 +142,18 @@ def setAxisNames(self, axesNames):
...
 
-    def setCustomAxis(self, axesNames):
+    def setCustomAxis(self, axesNames: set[str]):
         """Set the available list of named axis which can be set to a value.
 
-        :param List[str] axesNames: List of customable axis names
+        :param axesNames: List of customable axis names
         """
-        self.__customAxisNames = set(axesNames)
+        self.__customAxisNames = axesNames
```
t20100, line 149: We should really start to enable type checking!

src/silx/gui/data/NumpyAxesSelector.py
t20100: 😅

src/silx/gui/data/NumpyAxesSelector.py
loichuder: Oh yeah, I have to update my import habits to Python 3.10 :smile:

PR #4437, opened 2025-10-04
test: run tests in a directory which is actually writable

kif:

Fix the issue discussed about CI/CD not able to run on edna-site.

### DIFF
--- a/src/silx/test/__init__.py
+++ b/src/silx/test/__init__.py
@@ -29,6 +29,7 @@ import importlib
 import os.path
 import subprocess
 import sys
+import tempfile
 from collections.abc import Sequence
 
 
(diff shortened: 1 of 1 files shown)


src/silx/test/__init__.py
```diff
@@ -43,6 +44,7 @@ def run_tests(
     :param verbosity: Requested level of verbosity
     :param args: List of extra arguments to pass to `pytest`
     """
+    workdir = tempfile.mkdtemp(prefix="silx-")
```
t20100: Doc says:

> The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

IMO best to use `tempfile.TemporaryDirectory` context manager to take care of it

t20100 (commented):
Sounds good if it no longer writes in the installed folder!

vallsv:
Could we link the issue related to this change, because it don't really see why the tests would not be read only.

vallsv:
> It's `pytest` writing some cache folder where the Python package is installed.... It really shouldn't.

Thanks. Yes, i think it would be good to feedback the problem upstream if it's not already the case. So what we could add a related comment in code of this PR, so what it could be drop in the future (in case the pytest people think that's a problem on there side).

My 2 cents.

PR #4444, opened 2025-11-10
silx.io.open, silx view: Added blissdata's redis_hdf5 support

t20100:

This is an early prototype of wrapping blissdata's redis_hdf5 in silx commonh5.
It currently require a patched version of blissdata to "inject" the blissdata to hdf5 mapping.

This is similar to zarr #4350 and tiled #4121 wrappers.

### DIFF
--- a/src/silx/io/blissdatah5.py
+++ b/src/silx/io/blissdatah5.py
@@ -1,0 +1,91 @@
+import logging
+from typing import Generator, Literal
+
+import numpy
+
+from . import commonh5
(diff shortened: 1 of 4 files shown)


vallsv:
I think it would be good to drop `"bliss"`. bliss is not blissdata.

t20100:
Tested with latest blissdata: attributes issue fixed.

Still need to investigate issue with access to group containing a stack of image dataset

PR #4448, opened 2025-12-02
test: clean up remaining files after test

kif:

Remove files created during the test to avoid the accumulation of them in tmp-folder
Also: clean-up a config of pytest which is incompatible with former version of `pyopencl`.

### DIFF
--- a/src/silx/conftest.py
+++ b/src/silx/conftest.py
@@ -87,7 +87,8 @@ try:
 except Exception:
     pass
 else:
-    if "RepeatedKernelRetrieval" in dir(pyopencl):
+    pyopen_version = tuple(int(i) for i in pyopencl.__version__.split(".")[:2])
+    if pyopen_version >= (2025, 2):
         _FILTERWARNINGS = _FILTERWARNINGS + (
             "ignore::pyopencl.RepeatedKernelRetrieval",
         )


src/silx/conftest.py
```diff
@@ -77,12 +77,18 @@ def pytest_configure(config):
...
     "ignore:__array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments.:DeprecationWarning",
-    "ignore::pyopencl.RepeatedKernelRetrieval",
     # Deprecated pyparsing usage in matplotlib: https://github.com/matplotlib/matplotlib/issues/30617
     "ignore::DeprecationWarning:matplotlib._fontconfig_pattern",
     "ignore::DeprecationWarning:matplotlib._mathtext",
     "ignore::DeprecationWarning:pyparsing.util",
 )
+try:
+    import pyopencl
+except:
```
t20100: Best to avoid catching bare exceptions:
```suggestion
except Exception:
```
or 
```suggestion
except ImportError:
```

src/silx/conftest.py
```diff
@@ -77,12 +77,18 @@ def pytest_configure(config):
...
+    pass
+else:
+    if "RepeatedKernelRetrieval" in dir(pyopencl):
```
t20100: Why is this needed?

When does this occurs? Is it limited to a few specific tests? if so, best to filter this warning only for those tests with [pytest.mark.filterwarnings](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=filterwarnings#pytest.mark.filterwarnings)

Which version of pyopencl have it/does not have it?
If it's is old enough, IMO it reasonable to enforce a minimum version in the dependencies and spare the if statement
If not, it's best to test the version of pyopencl rather than relying on a `hasattr`-like test so it documents why this check is needed for (and when we can get rid of it).

kif: Elder version of pyopencl (2025.1) do not have this attribute. Silencing a warning that does not exist simply prevents pytest from launching.

t20100: Could you either check the version number instead or add a comment with the required pyopencl version?
So we have a chance to remove this at some point

src/silx/conftest.py
```diff
@@ -77,12 +77,21 @@ def pytest_configure(config):
...
+try:
+    import pyopencl
+except Exception:
+    pass
+else:
+    pyopen_version = tuple(int(i) for i in pyopencl.__version__.split(".")[:2])
+    if pyopen_version >= (2025, 2):
+        _FILTERWARNINGS = _FILTERWARNINGS + (
+            "ignore::pyopencl.RepeatedKernelRetrieval",
+        )
```
t20100, line 94: What is this for? is this still relevant or was it for a specific or old pyopencl version?

t20100: @kif, this question is still relevant: What is this for? is this still relevant or was it for a specific or old pyopencl version?

kif: Removed in 
https://github.com/silx-kit/silx/pull/4527/changes#diff-cbe18314f5f85f974d031fec51982de03a7708695fb6259c240e22ca045ecb48

PR #4449, opened 2025-12-08
OpenCL: Explicit kernel cache

kif:

Built on top of #4448 to be able to run tests :)


pierrepaleo:
Looks good. I agree with the approach, we should try to cache the kernels compilation and pyopencl extra memory allocations. 

It was a good improvement of `KernelContainer`: we go from `self.__setattr__(kernel.function_name, kernel)` and querying `self.__dict__` to something more explicit.

PR #4453, opened 2026-01-06
silx.gui.data.DataViewer: split composite ImageView into Plot2dView and ComplexImageView

### LINKED ISSUE #4423: silx view: max value of autoscale colormap set to 1, regardless of values in the image
Silx 2.2.2

When viewing an image with values ranging up to - say - 25000, autoscale correctly displays the image but seems to set the max value of the colormap to one (see attachment1). It then completely messes small manual adjustment of the colormap max value (a very small displacement of the upper bound makes the whole image completely saturated (see attachment2).

Image

Image

loichuder:

Fix #4423 

The root cause of #4423 is the fact that the two views (`Plot2dView` and `ComplexImageView`) of the composite `ImageView` are updating simultaneously the colorbar dialog (since it is shared for the whole `DataViewer`).

In specific circumstances, the limits can be set by the view that is **not** displayed on screen, mismatching the limits of the actual displayed image.

After a long ponder on how to proceed, we decided to split the composite `ImageView` into two simple views `Plot2dView` and `ComplexImageView` so that only one is instantiated at a time. This matches the previous behaviour where only one view over the two is visible at all time but required some minor changes:
- `_ImageView` is now deprecated
- `ComplexImageView` label was set to `Image` for consistency with the previous behaviour
- `Plot2dView` does not support complex anywore
- The mode (i.e. `id`) of the available visualization changes (from `20`, it becomes `21` for a regular image and `22` for a complex image)

### DIFF
--- a/src/silx/gui/data/DataViewer.py
+++ b/src/silx/gui/data/DataViewer.py
@@ -32,6 +32,7 @@ from silx.gui import qt
 from silx.gui.data import DataViews
 from silx.gui.data.DataViews import (
     NXDATA_STACK_MODE,
+    PLOT2D_MODE,
     STACK_MODE,
     IMAGE_MODE,
     _normalizeData,
@@ -166,7 +167,8 @@ class DataViewer(qt.QFrame):
             DataViews._Hdf5View,
             DataViews._NXdataView,
             DataViews._Plot1dView,
-            DataViews._ImageView,
+            DataViews._Plot2dView,
+            DataViews._ComplexImageView,
             DataViews._Plot3dView,
             DataViews._RawView,
             DataViews._Plot2dRecordView,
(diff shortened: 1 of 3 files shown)


src/silx/gui/data/DataViewer.py
```diff
@@ -377,21 +388,31 @@ def setDisplayMode(self, modeId: int):
 
             - `DataViews.EMPTY_MODE`: display nothing
             - `DataViews.PLOT1D_MODE`: display the data as a curve
-            - `DataViews.IMAGE_MODE`: display the data as an image
+            - `DataViews.PLOT2D_MODE`: display real data as an image
+            - `DataViews.COMPLEX_IMAGE_MODE`: display complex data as an image
```
t20100: `COMPLEX_PLOT2D_MODE` for consistency with `PLOT2D_MODE` maybe?

t20100 (approved):
Thanks for delving into this!

Good to remove some intermediate layers.

@vallsv do you recall why there was this `_ImageView` using a `QStackedWidget`?

src/silx/gui/data/DataViewer.py
loichuder: Sure!

PR #4454, opened 2026-01-07
silx.gui.data.test.test_dataviewer: migrate to pytest

loichuder:

### DIFF
--- a/src/silx/gui/data/test/test_dataviewer.py
+++ b/src/silx/gui/data/test/test_dataviewer.py
@@ -25,10 +25,7 @@ __authors__ = ["V. Valls"]
 __license__ = "MIT"
 __date__ = "19/02/2019"
 
-import os
-import tempfile
 import pytest
-from contextlib import contextmanager
 
 import numpy
 from ..DataViewer import DataViewer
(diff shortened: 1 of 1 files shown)


src/silx/gui/data/test/test_dataviewer.py
```diff
@@ -60,136 +60,185 @@ def getDataPriority(self, data, info):
...
+    os.close(fd)
+    data = numpy.arange(3 * 3 * 3)
+    data.shape = 3, 3, 3
+    # create h5 data
+    h5file = h5py.File(tmp_name, "w")
+    h5file["data"] = data
+    yield h5file
+    # clean up
+    h5file.close()
+    os.unlink(tmp_name)
```
t20100: This is now only used in `test_3d_h5_dataset`, so it would be simpler to read if this test was creating the file either in memory or by using the `tmp_path` fixture.

src/silx/gui/data/test/test_dataviewer.py
```diff
@@ -60,136 +60,185 @@ def getDataPriority(self, data, info):
         return 0
 
 
-class _TestAbstractDataViewer(TestCaseQt):
-    __test__ = False  # ignore abstract class
+# https://docs.pytest.org/en/stable/how-to/fixtures.html#parametrizing-fixtures
+@pytest.fixture(params=[DataViewer, DataViewerFrame])
+def widget(request):
```
t20100: IMO, a more specific name would be good here, e.g., `dataViewerWidget`

PR #4475, opened 2026-02-03
silx.app.view: Fix Ubuntu taskbar icon by setting app.setDesktopFileName

### LINKED ISSUE #4474: Silx icon is not showing up in Ubuntu24 task bar
Basically we currently have cog showing up instead of the silx logo in Ubuntu24:

Image

poautran:

This helps desktop environments (notably Ubuntu/GNOME) match the running app window with the installed desktop entry, so the correct taskbar icon is used.

Should not impact any other distros.


src/silx/app/view/main.py
```diff
@@ -143,6 +143,8 @@ def mainQt(options):
     import silx.gui.utils.matplotlib  # noqa
 
     app = qt.QApplication([])
+    if hasattr(app, "setDesktopFileName"):
```
t20100: No need for this check, `setDesktopFileName` is available since qt5.7 (https://doc.qt.io/archives/qt-5.15/qguiapplication.html#desktopFileName-prop) and silx works with qt>=5.9 (https://www.silx.org/doc/silx/latest/install.html#dependencies)

src/silx/app/view/main.py
```diff
@@ -143,6 +143,8 @@ def mainQt(options):
...
+        app.setDesktopFileName("org.silx.SilxView")
```
t20100: Unless I miss something, this relies on what's in [`package/desktop` ](https://github.com/silx-kit/silx/tree/main/package/desktop) to be installed on the host. So one must have installed silx as an Ubuntu package to have the icon... Did you tested on a computer without silx installed on the system?

poautran: Indeed, it does not work if we just pip install...

t20100: From what I tested/searched, there seems to be no neat solution to display the icon in the menu bar that only relies on the application code.

However this fix is needed even if it is not the only required step: It makes it work when the icon and .desktop file are already available on the host. So it is required to install the .desktop and icon another way.

PR #4459, opened 2026-01-09
silx.gui.data.NumpyAxesSelector: Fix alignment of axis labels

loichuder:
- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

For #4428 

| **Before** | **After** |
| ----------------- | -------------  |
|image | image |

I didn't see a way of doing this without a `GridLayout`. Unfortunately, this means breaking the encapsulation of `_Axis` since the elements of `_Axis` (slider, label, and combobox) need to be added individually to the layout.

Not very happy about this, tell me what you think.

Issue #4428, opened 2025-09-25
NXdataWidget axis selector improvements


loichuder:
- [ ] Fix alignment of axis labels (implemented in #4427 )
- [ ] Remove/deprecate `customAxes` (https://github.com/silx-kit/silx/pull/4427#discussion_r2372674307)
- [ ] Implement axis labels for other `Array...Plot` widgets

loichuder:
With https://github.com/silx-kit/silx/pull/4705, axes can be changed when viewing NXdata as curves, images and volumes.

I don't know if it is worth doing it for the remaining vis (scatter, RGB, complex) since they are quite specific.

loichuder:
After discussion with @t20100, it may be useful for scatters. Leaving this open then.

PR #4476, opened 2026-02-06
silx.gui.data.NumpyAxesSelector: Fix label alignment by normalizing widths

loichuder:
For https://github.com/silx-kit/silx/issues/4428

While reworking https://github.com/silx-kit/silx/pull/4459, I noticed the presence of a hook `__normalizeAxisGeometry` to have consistent slider widths.

In this PR, I extend this hook to also have consistent label widths, making it an alternative of #4459. Even though we are sort of wrestling against Qt layouts, the amount of changes is way less important.

What do you think @t20100 ?

### DIFF
--- a/src/silx/gui/data/NumpyAxesSelector.py
+++ b/src/silx/gui/data/NumpyAxesSelector.py
@@ -352,7 +352,7 @@ class NumpyAxesSelector(qt.QWidget):
             a.slider().lineEdit().setFixedWidth(lineEditWidth)
             a.slider().limitWidget().setFixedWidth(limitWidth)
 
-        labelWidth = max([a.label().width() for a in self.__axis])
+        labelWidth = max([a.label().minimumSizeHint().width() for a in self.__axis])
         for a in self.__axis:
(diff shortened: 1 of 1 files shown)


src/silx/gui/data/NumpyAxesSelector.py
```diff
@@ -349,6 +352,10 @@ def __normalizeAxisGeometry(self):
             a.slider().lineEdit().setFixedWidth(lineEditWidth)
             a.slider().limitWidget().setFixedWidth(limitWidth)
 
+        labelWidth = max([a.label().width() for a in self.__axis])
```
t20100:
```suggestion
        labelWidth = max([a.label().minimumSizeHint().width() for a in self.__axis])
```
as for `limitWidget()`  which is also a QLabel.

`minimumSizeHint` changes according to the label text, `width` and `minimumWidth` are set by `setFixedWidth`

loichuder: Good to know, thanks !

PR #4479, opened 2026-02-10
common api to add/remove items

abmajith:
In the LegendWidgets, we are moving from default add/remove Items (based on the internal Items trigger action from the plotwidget), to application explicit add/remove Items calls

### DIFF
--- a/src/silx/gui/plot/test/test_legendswidget.py
+++ b/src/silx/gui/plot/test/test_legendswidget.py
@@ -35,11 +35,11 @@ def test_visibility_toggle(qapp, qapp_utils, plot_with_legend):
     qapp.processEvents()
     item = plot.getCurve("test_item")
     item_widget = legend_widget._itemWidgets[item]
-    assert item.isVisible() is True
+    assert item.isVisible()
     qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
-    assert item.isVisible() is False
+    assert not item.isVisible()
     qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
-    assert item.isVisible() is True
+    assert item.isVisible()
 
 
 def test_manual_add_same_item(qapp, plot_with_legend):
@@ -90,12 +90,12 @@ def test_orphan_item_interaction(qapp, qapp_utils, qWidgetFactory):
     legend_widget.addItem(orphan_item)
     qapp.processEvents()
     item_widget = legend_widget._itemWidgets[orphan_item]
-    assert orphan_item.isVisible() is True
-    assert item_widget._label.isEnabled() is True
+    assert orphan_item.isVisible()
+    assert item_widget._label.isEnabled()
     qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
     qapp.processEvents()
-    assert orphan_item.isVisible() is False
-    assert item_widget._label.isEnabled() is False
+    assert not orphan_item.isVisible()
+    assert not item_widget._label.isEnabled()
     qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
-    assert orphan_item.isVisible() is True
-    assert item_widget._label.isEnabled() is True
+    assert orphan_item.isVisible()
+    assert item_widget._label.isEnabled()


src/silx/gui/plot/LegendsWidget.py
```diff
@@ -150,15 +147,18 @@ def _clear(self):
...
 
-    def _onItemRemoved(self, item: items.Item):
+    def removeItem(self, item: items.Item):
         widget: qt.QWidget | None = self._itemWidgets.pop(item, None)
         if widget is not None:
             self.layout().removeWidget(widget)
             widget.deleteLater()
+
+    def clearItems(self):
+        self._clear()
```
t20100: Just make the `_clear` method public.
I propose to name it `clear` rather than `clearItems` since Qt seems to use this naming: https://doc.qt.io/qt-6/qlistwidget.html

src/silx/gui/plot/LegendsWidget.py
```diff
@@ -121,26 +121,23 @@ def __init__(
...
         self.setPlotWidget(plotWidget)
 
     def setPlotWidget(self, plot: PlotWidget | None):
         previousPlot = None if self._plotRef is None else self._plotRef()
         if previousPlot is not None:
-            previousPlot.sigItemAdded.disconnect(self._onItemAdded)
-            previousPlot.sigItemRemoved.disconnect(self._onItemRemoved)
-
-        self._clear()
+            previousPlot.sigItemRemoved.disconnect(self._onItemRemovedFromPlot)
```
t20100: There should be no need for `_onItemRemovedFromPlot`:
```suggestion
            previousPlot.sigItemRemoved.disconnect(self.removeItem)
```

src/silx/gui/plot/LegendsWidget.py
```diff
@@ -121,26 +121,23 @@ def __init__(
...
-
         self.setPlotWidget(plotWidget)
 
     def setPlotWidget(self, plot: PlotWidget | None):
         previousPlot = None if self._plotRef is None else self._plotRef()
         if previousPlot is not None:
-            previousPlot.sigItemAdded.disconnect(self._onItemAdded)
-            previousPlot.sigItemRemoved.disconnect(self._onItemRemoved)
-
-        self._clear()
```
t20100, line 133: clear should not be removed, instead it should remove the items that belonged to `previousPlot` if any.

src/silx/gui/plot/LegendsWidget.py
```diff
@@ -121,26 +121,23 @@ def __init__(
...
         if plot is None:
             self._plotRef = None
         else:
-            plot.sigItemAdded.connect(self._onItemAdded)
-            plot.sigItemRemoved.connect(self._onItemRemoved)
+            plot.sigItemRemoved.connect(self._onItemRemovedFromPlot)
             self._plotRef = weakref.ref(plot)
 
-            for item in plot.getItems():
-                self._onItemAdded(item)
```
t20100, line 143: The plot items still need to be added.

src/silx/gui/plot/LegendsWidget.py
```diff
@@ -121,26 +121,23 @@ def __init__(
...
-        self._clear()
+            previousPlot.sigItemRemoved.disconnect(self._onItemRemovedFromPlot)
+            self.clearItems()
 
         if plot is None:
             self._plotRef = None
         else:
-            plot.sigItemAdded.connect(self._onItemAdded)
-            plot.sigItemRemoved.connect(self._onItemRemoved)
+            plot.sigItemRemoved.connect(self._onItemRemovedFromPlot)
```
t20100: Why removing `sigItemAdded` connection?

The idea is to have 2 ways to manage the items:
- Automatically from a `PlotWidget` through `sigItemAdded`, `sigItemRemoved`
- Manually with explicit `addItem`, `removeItem`
- 
Both mode should co-exist and work in corner cases e.g.:
- remove an item added automatically and then remove it from the plot
- add an item manually that was already added automatically (in this case it should not be added)
- `setPlotWidget(None)` should not remove item added manually that do not belong to the previous plot

Note: We could remove `setPlotWidget` and have users do the connections  `sigItemAdded` ->  `addItem`, `sigItemRemoved` -> `removeItem`, but quite a few widgets provides an automatic update through a `setPlotWidget` (or `setPlot`) method, so it is worth keeping it IMO.

examples/plotLegendsWidget.py
```diff
@@ -18,10 +18,14 @@
...
+c1 = plot.addCurve(x, numpy.sin(x), legend="Sin Curve", color="blue")
+legends.addItem(plot.getCurve("Sin Curve"))
+c2 = plot.addCurve(x, numpy.cos(x), legend="Cosine Curve", color="red")
+legends.addItem(plot.getCurve("Cosine Curve"))
 xScatter = numpy.random.rand(50) * 10
 yScatter = numpy.random.rand(50) * 2 - 1
-plot.addScatter(xScatter, yScatter, value=yScatter * 10, legend="Points")
+s1 = plot.addScatter(xScatter, yScatter, value=yScatter * 10, legend="Points")
+legends.addItem(plot.getScatter("Points"))
+s2 = plot.addScatter(xScatter - 1, yScatter - 1, value=yScatter * 10, legend="Points_2")
```
t20100: I would keep the sample code as it was: it was not showing the manual way, but it's a simple on

src/silx/gui/plot/test/test_legendswidget.py
```diff
@@ -14,23 +14,29 @@ def plot_with_legend(qWidgetFactory):
...
     plot.addCurve(numpy.arange(10), numpy.arange(10), legend="curve1")
     qapp.processEvents()
+    assert len(legend_widget._itemWidgets) == 0
+    item = plot.getCurve("curve1")
+    legend_widget.addItem(item)
+    qapp.processEvents()
     assert len(legend_widget._itemWidgets) == 1
-    plot.remove(legend="curve1")
+    legend_widget.removeItem(item)
     qapp.processEvents()
```
t20100: Just keep the tests as they were and add new ones, it should still work as it was

src/silx/gui/plot/test/test_legendswidget.py
```diff
@@ -14,23 +14,29 @@ def plot_with_legend(qWidgetFactory):
...
     assert len(legend_widget._itemWidgets) == 0
 
 
 def test_visibility_toggle(qapp, qapp_utils, plot_with_legend):
     plot, legend_widget = plot_with_legend
     plot.addCurve([0, 1], [0, 1], legend="test_item")
+    item = plot.getCurve("test_item")
+    legend_widget.addItem(item)
```
t20100: Just keep the tests as they were and add new ones, it should still work as it was

src/silx/gui/plot/test/test_legendswidget.py
```diff
@@ -14,23 +14,29 @@ def plot_with_legend(qWidgetFactory):
 
 
 def test_initial_state(plot_with_legend):
-    plot, legend_widget = plot_with_legend
+    _, legend_widget = plot_with_legend
```
t20100:
```
def test_initial_state(qWidgetFactory):
    legend_widget = qWidgetFactory(LegendsWidget)
```

src/silx/gui/plot/test/test_legendswidget.py
```diff
@@ -39,3 +45,27 @@ def test_visibility_toggle(qapp, qapp_utils, plot_with_legend):
...
+
+def test_sync_on_plot_remove(qapp, plot_with_legend):
+    plot, legend_widget = plot_with_legend
+    plot.addCurve([0, 1], [0, 1], legend="c1")
+    legend_widget.addItem(plot.getCurve("c1"))
+    qapp.processEvents()
+    assert len(legend_widget._itemWidgets) == 1
+    plot.remove("c1")
+    qapp.processEvents()
+    assert len(legend_widget._itemWidgets) == 0
```
t20100: Those tests needs rework to test the manual add/remove and calling `setPlotWidget` to tests corner cases (see other comment)

t20100 (commented):
The previous behavior was fine and should not change IMO.
The manual add/remove can co-exist with the "automatic" sync with the plot: We just need to make sure in work in all conditions

src/silx/gui/plot/LegendsWidget.py
abmajith: Okay, Now I understand features, by default it should be like previous way, 

additionally, if the user wants to remove an item, then we could remove from legend and from plot widget,

if user wants to addItem to legend, it should also add it into plot.

src/silx/gui/plot/test/test_legendswidget.py
abmajith: yep,

src/silx/gui/plot/LegendsWidget.py
abmajith: Actually, if we implement this feature, we dont have to go for the sorting of items based on the Items.type,

user can manually arange Items using multiple LegendWidgets, and order them as they want,

because, if we try to do manually, then it complicates the code, may be we need to find some clever sorting insertion, deletion, data structure, ?

src/silx/gui/plot/test/test_legendswidget.py
```diff
@@ -39,3 +40,62 @@ def test_visibility_toggle(qapp, qapp_utils, plot_with_legend):
...
+    item_widget = legend_widget._itemWidgets[orphan_item]
+    assert orphan_item.isVisible() is True
+    assert item_widget._label.isEnabled() is True
+    qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
+    qapp.processEvents()
+    assert orphan_item.isVisible() is False
+    assert item_widget._label.isEnabled() is False
+    qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
+    assert orphan_item.isVisible() is True
+    assert item_widget._label.isEnabled() is True
```
t20100: No  need for `is True` and `is False` is `not`:

```suggestion
    assert orphan_item.isVisible()
    assert item_widget._label.isEnabled()
    qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
    qapp.processEvents()
    assert not orphan_item.isVisible()
    assert not item_widget._label.isEnabled()
    qapp_utils.mouseClick(item_widget, qt.Qt.LeftButton)
    assert orphan_item.isVisible()
    assert item_widget._label.isEnabled()
```

Issue #4351, opened 2025-07-15
Uniform lighting effects on Sphere class


abmajith:
When we change the angle of view, the lighting effects changes the color,

is there a way to switch off the lighting effects on the 3d plots?

PR #4491, opened 2026-03-03
silx.gui.plot3d: Fixed lighting of Spheres

t20100:

This PR fixes the lighting of sphere impostors. As opposed to other 3D primitives, the lighting is not performed in the object frame of reference but in camera space.
It looks to work fine with directional light. There is probably some issues if we add a spot light one day.

related to issue #4351 and PR #4483

attn @abmajith, review welcomed!

### DIFF
--- a/src/silx/gui/plot3d/scene/function.py
+++ b/src/silx/gui/plot3d/scene/function.py
@@ -370,33 +370,35 @@ class DirectionalLight(event.Notifier, ProgramFunction):
                                   It MUST be in use and using this function.
         :param frame: The frame of reference in which to apply the lighting
         """
-        if self.isOn and self._direction is not None:
-            if frame == "object":
-                # Transform light direction from camera space to object coords
-                lightdir = context.objectToCamera.transformDir(
-                    self._direction, direct=False
-                )
-                lightdir /= numpy.linalg.norm(lightdir)
+        if not self.isOn or self._direction is None:
+            return
 
-                # Convert view position to object coords
-                viewpos = context.objectToCamera.transformPoint(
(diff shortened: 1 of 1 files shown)


src/silx/gui/plot3d/scene/function.py
```diff
@@ -356,28 +357,40 @@ def fragmentCall(self):
...
+        frame: Literal["camera", "object"] = "object",
+    ):
         """Sets-up uniforms of a program using this shader function.
 
         :param RenderContext context: The current rendering context
         :param GLProgram program: The program to set-up.
                                   It MUST be in use and using this function.
+        :param frame: The frame of reference in which to apply the lighting
         """
         if self.isOn and self._direction is not None:
```
abmajith: We could use early return to avoid nested if logic

t20100: yes! done

src/silx/gui/plot3d/scene/function.py
```diff
@@ -356,34 +357,48 @@ def fragmentCall(self):
         """Function name to call in fragment shader"""
         return "lighting"
 
-    def setupProgram(self, context, program):
+    def setupProgram(
+        self,
+        context,
+        program,
```
loichuder, line 363: Is it possible to add typings for the two other parameters?

t20100: I'll do another PR for it: typing the context requires splitting a module to avoid circular references

loichuder: Sure, no worries. I figured it might be difficult which is why I asked if it was possible :smile:

PR #4494, opened 2026-03-06
silx.gui.plot3d: Added support for array of symbol size to `Scatter2D` and `Scatter3D`

t20100:

This PR adds support for arrays of symbol values for `plot3d` scatter items.

To do so, it reworks the `plot` item `SymbolMixIn` which is also used by `plot3d` to add supports for arrays.
A `SymbolSingleSizeMixIn` class is added to restrict symbol size to be only a single float. For now this is used throughout items in `silx.gui.plot` but it paves the way for 2d scatters with an array of sizes.

For 3D scatter items, the size can be set through `item.setSymbolSize` and tools like the param tree view are updated to support array of sizes.

attn @abmajith @loichuder 

- [x] Test for regression of curves and scatters in plot widget.

### DIFF
--- a/src/silx/gui/plot3d/items/mixins.py
+++ b/src/silx/gui/plot3d/items/mixins.py
@@ -200,6 +200,14 @@ class SymbolMixIn(_SymbolMixIn):
             self.__primitive.marker = symbol
             self.__primitive.setAttribute("size", size, copy=False)
 
+    def _getPickingDistances(self) -> float | numpy.ndarray:
+        """Returns distances below which to consider a point as picked
+
+        Distances are in screen pixels
+        """
+        _, size = self._getSceneSymbol()
+        return numpy.maximum(size, 3.0)
+
     def _getSceneSymbol(self) -> tuple[str, float | ArrayLike]:
         """Returns a symbol name and size suitable for scene primitives.
 
--- a/src/silx/gui/plot3d/items/scatter.py
+++ b/src/silx/gui/plot3d/items/scatter.py
@@ -178,9 +178,7 @@ class Scatter3D(DataItem3D, ColormapMixIn, SymbolMixIn):
         # Perform picking
         distancesNdc = numpy.abs(pointsNdc[:, :2] - rayNdc[0, :2])
         # TODO issue with symbol size: using pixel instead of points
-        threshold = numpy.maximum(0.5 * self.getSymbolSize(copy=False), 3.0).reshape(
-            -1, 1
-        )
+        threshold = self._getPickingDistances().reshape(-1, 1)
         thresholdNdc = 2.0 * threshold / primitive.viewport.size
         picked = numpy.where(
             numpy.logical_and(
@@ -504,10 +502,8 @@ class Scatter2D(DataItem3D, ColormapMixIn, SymbolMixIn, ScatterVisualizationMixI
         mode = self.getVisualization()
         if mode is self.Visualization.POINTS:
             # TODO issue with symbol size: using pixel instead of points
-            # Get "corrected" symbol size
-            _, size = self._getSceneSymbol()
             return self._pickPoints(
-                context, points, threshold=numpy.maximum(0.5 * size, 3.0)
+                context, points, threshold=self._getPickingDistances()
             )
 
         elif mode is self.Visualization.LINES:


src/silx/gui/plot/items/_roi_base.py
```diff
@@ -644,7 +644,7 @@ def getCurrentStyle(self) -> items.CurveStyle:
             baseLinewidth = self._DEFAULT_LINEWIDTH
         if isinstance(self, core.SymbolMixIn):
             baseSymbol = self.getSymbol()
-            baseSymbolsize = self.getSymbolSize()
+            baseSymbolsize = self.getSymbolSize() if self.isSingleSymbolSize() else 1
```
t20100, line 647: Fallback value for array of sizes

src/silx/gui/plot/items/core.py
```diff
@@ -1440,7 +1469,7 @@ def getCurrentVisualizationParameter(self, parameter):
         return self.getVisualizationParameter(parameter)
 
 
-class PointsBase(DataItem, SymbolMixIn, AlphaMixIn):
+class PointsBase(DataItem, SymbolSingleSizeMixIn, AlphaMixIn):
```
t20100, line 1472: So for now Curve and Scatter items do not support array of sizes

src/silx/gui/plot3d/_model/items.py
```diff
@@ -1059,15 +1059,44 @@ class SymbolSizeRow(ItemProxyRow):
...
     """
 
-    def __init__(self, item):
+    def __init__(self, item, setflags: bool = True):
         super().__init__(
             item=item,
             name="Marker size",
             fget=item.getSymbolSize,
             fset=item.setSymbolSize,
+            toModelData=lambda data: None if isinstance(data, numpy.ndarray) else data,
```
t20100: Nothing displayed for size array

src/silx/gui/plot3d/_model/items.py
```diff
@@ -1059,15 +1059,44 @@ class SymbolSizeRow(ItemProxyRow):
...
+            self.setFlags(qt.Qt.ItemIsEnabled, 0)
+            self.setFlags(qt.Qt.ItemIsEnabled | qt.Qt.ItemIsEditable, 1)
+        else:
+            self.setFlags(qt.Qt.NoItemFlags)
+
+    def _itemChanged(self, event):
+        """Set flags to enable/disable the row"""
+        if event == items.ItemChangedType.SYMBOL_SIZE:
+            item = self.sender()
+            self.__isEnabled = item.isSingleSymbolSize()
```
t20100, line 1094: disable tree row for size array

src/silx/gui/plot3d/_model/items.py
```diff
@@ -1679,8 +1711,18 @@ class Scatter2DSymbolSizeRow(Scatter2DPropertyMixInRow, SymbolSizeRow):
     """
 
     def __init__(self, item):
-        SymbolSizeRow.__init__(self, item)
-        Scatter2DPropertyMixInRow.__init__(self, item, "symbolSize")
+        SymbolSizeRow.__init__(self, item, setflags=False)
+        Scatter2DPropertyMixInRow.__init__(self, item, "symbolSize", setflags=False)
+        self._setFlags()
```
t20100, line 1719: I faced issue with parent classes calling `_setFlags` in their `__init__` to make a dual inheritance that works.
This is not nice but it makes it work without duplicating or reworking the whole thing. I'd take any better option

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -167,13 +168,37 @@ class SymbolMixIn(_SymbolMixIn):
...
+    def setSymbolSize(self, size: float | ArrayLike | None, copy: bool = True):
+        super().setSymbolSize(size, copy)
+        self._syncPointsPrimitive()
+
+    def _syncPointsPrimitive(self):
+        """Synchronize scene object's symbol and size"""
+        if self.__primitive is not None:
+            symbol, size = self._getSceneSymbol()
+            self.__primitive.marker = symbol
+            self.__primitive.setAttribute("size", size, copy=False)
```
t20100, line 201: There is 2 layers: "high level" items (Scatter3D and Scatter2D) as for the 2D PlotWidget and "low level" scene graph rendering primitives (Points). This updates the low level one when the high level gets updated

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -51,33 +51,19 @@ class Scatter3D(DataItem3D, ColormapMixIn, SymbolMixIn):
...
 
-    def _updated(self, event=None):
-        """Handle mix-in class updates"""
-        if event in (ItemChangedType.SYMBOL, ItemChangedType.SYMBOL_SIZE):
-            symbol, size = self._getSceneSymbol()
-            self._scatter.marker = symbol
-            self._scatter.setAttribute("size", size, copy=True)
-
-        super()._updated(event)
-
```
t20100, line 80: now in the `plot3d` `SymbolMixIn` class

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -194,8 +178,10 @@ def _pickFull(self, context, threshold=0.0, sort="depth"):
         # Perform picking
         distancesNdc = numpy.abs(pointsNdc[:, :2] - rayNdc[0, :2])
         # TODO issue with symbol size: using pixel instead of points
-        threshold += self.getSymbolSize()
-        thresholdNdc = 2.0 * threshold / numpy.array(primitive.viewport.size)
+        threshold = numpy.maximum(0.5 * self.getSymbolSize(copy=False), 3.0).reshape(
+            -1, 1
+        )
+        thresholdNdc = 2.0 * threshold / primitive.viewport.size
```
t20100, line 182: Supports array and fix issue with threshold that was twice too big: The distance threshold is between the point center and the mouse and it should be half the size of the symbol.

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -194,8 +178,10 @@ def _pickFull(self, context, threshold=0.0, sort="depth"):
         # Perform picking
         distancesNdc = numpy.abs(pointsNdc[:, :2] - rayNdc[0, :2])
         # TODO issue with symbol size: using pixel instead of points
-        threshold += self.getSymbolSize()
-        thresholdNdc = 2.0 * threshold / numpy.array(primitive.viewport.size)
+        threshold = numpy.maximum(0.5 * self.getSymbolSize(copy=False), 3.0).reshape(
```
t20100: Fix issue with using symbol size for picking: Use 0.5 * size since it compares a distance, it was twice too big before

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -257,14 +243,7 @@ def __init__(self, parent=None):
 
     def _updated(self, event=None):
         """Handle mix-in class updates"""
-        if event in (ItemChangedType.SYMBOL, ItemChangedType.SYMBOL_SIZE):
-            symbol, size = self._getSceneSymbol()
-            for child in self._getScenePrimitive().children:
-                if isinstance(child, primitives.Points):
-                    child.marker = symbol
-                    child.setAttribute("size", size, copy=True)
-
```
t20100, line 266: This logic is in SymbolMixIn now

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -445,7 +424,7 @@ def _pickPoints(self, context, points, threshold=1.0, sort="depth"):
 
         # Perform picking
         distancesNdc = numpy.abs(pointsNdc[:, :2] - rayNdc[0, :2])
-        thresholdNdc = threshold / numpy.array(primitive.viewport.size)
+        thresholdNdc = numpy.array(threshold).reshape(-1, 1) / primitive.viewport.size
```
t20100, line 425: This supports both array and scalar

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -526,8 +505,10 @@ def _pickFull(self, context):
         if mode is self.Visualization.POINTS:
             # TODO issue with symbol size: using pixel instead of points
             # Get "corrected" symbol size
-            _, threshold = self._getSceneSymbol()
-            return self._pickPoints(context, points, threshold=max(3.0, threshold))
+            _, size = self._getSceneSymbol()
+            return self._pickPoints(
+                context, points, threshold=numpy.maximum(0.5 * size, 3.0)
```
t20100: 0.5 * size to fix issue with picking distance twice too big

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -167,13 +168,45 @@ class SymbolMixIn(_SymbolMixIn):
...
+    def setSymbol(self, symbol):
+        super().setSymbol(symbol)
+        self._syncPointsPrimitive()
+
+    def getSymbolSize(self, copy: bool = True) -> float | numpy.ndarray:
+        """Return the marker size in pixels."""
+        return super().getSymbolSize(copy)
+
+    def setSymbolSize(self, size: float | ArrayLike | None, copy: bool = True):
+        """Set the marker size in pixels.
```
t20100, line 189: As opposed to 2D plots, for now the symbol size is in screen pixel.
It would be nice to be consistent, but for another PR, so advertising it in the doc for now.

src/silx/gui/plot/items/roi.py
```diff
@@ -72,12 +72,12 @@ class PointROI(RegionOfInterest, items.SymbolMixIn):
     _DEFAULT_SYMBOL = "+"
     """Default symbol of the PointROI
 
-    It overwrite the `SymbolMixIn` class attribte.
+    It overwrite the `SymbolSingleSizeMixIn` class attribute.
```
loichuder:
```suggestion
    It overwrites the `SymbolSingleSizeMixIn` class attribute.
```

src/silx/gui/plot3d/_model/items.py
loichuder: Could we not print `array <size:?>` or something?

src/silx/gui/plot3d/_model/items.py
```diff
@@ -1059,15 +1059,44 @@ class SymbolSizeRow(ItemProxyRow):
...
         super().__init__(
             item=item,
             name="Marker size",
             fget=item.getSymbolSize,
             fset=item.setSymbolSize,
+            toModelData=lambda data: None if isinstance(data, numpy.ndarray) else data,
             events=items.ItemChangedType.SYMBOL_SIZE,
             editorHint=(1, 20),
         )  # TODO link with OpenGL max point size
+        self.__isEnabled = item.isSingleSymbolSize()
```
loichuder, line 1075:
```suggestion
        # Disable tree row if the item size is an array
        self.__isEnabled = item.isSingleSymbolSize()
```

src/silx/gui/plot3d/_model/items.py
loichuder: Yeah not much choice.

The only alternative I can think of is to create some kind of `__post_init__` method that could be overwritten by child classes to call the appropriate `setFlags`.

I did it in [ewoks3dxrd](https://gitlab.esrf.fr/workflow/ewoksapps/ewoks3dxrd/-/merge_requests/208) but it does not feel super clean either.

src/silx/gui/plot3d/items/scatter.py
```diff
@@ -421,7 +400,7 @@ def _pickPoints(self, context, points, threshold=1.0, sort="depth"):
         """Perform picking while in 'points' visualization mode
 
         :param PickContext context: Current picking context
-        :param float threshold: Picking threshold in pixel.
+        :param Union[float,numpy.ndarray threshold: Picking threshold in pixel.
```
loichuder:
```suggestion
        :param Union[float,numpy.ndarray] threshold: Picking threshold in pixel.
```

loichuder: It would be cooler to have it as type annotation 😉

src/silx/gui/plot3d/items/scatter.py
loichuder: Feels like this formula should perhaps be in an util `pickingThreshold` or similar

src/silx/gui/plot3d/_model/items.py
t20100: I'll keep it as it is for now, it's only one model item.
If it is needed for another, I'll consider your approach.

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -167,13 +168,53 @@ class SymbolMixIn(_SymbolMixIn):
...
+            self.__primitive.marker = symbol
+            self.__primitive.setAttribute("size", size, copy=False)
+
+    def _getPickingDistances(self) -> float | numpy.ndarray:
+        """Returns distances below which to consider a point as picked
+
+        Distances are in screen pixels
+        """
+        _, size = self._getSceneSymbol()
+        return numpy.maximum(size, 3.0)
```
t20100, line 209: Added a method to get the picking thresholds.

Also I removed the 0.5 factor... It actually seems to be more an issue of screen scaling than size being to big. I opened an issue for it: #4511

src/silx/gui/plot3d/_model/items.py
```diff
@@ -1059,15 +1059,47 @@ class SymbolSizeRow(ItemProxyRow):
...
 
-    def __init__(self, item):
+    def __init__(self, item, setflags: bool = True):
         super().__init__(
             item=item,
             name="Marker size",
             fget=item.getSymbolSize,
             fset=item.setSymbolSize,
+            toModelData=lambda data: (
+                "From data" if isinstance(data, numpy.ndarray) else data
```
t20100, line 1069: Image

Issue #4446, opened 2025-11-17
ImageStack: avoid keep all images in memory


payno:
Today the [ImageStack](https://github.com/silx-kit/silx/blob/main/src/silx/gui/plot/ImageStack.py#L148) is keeping all the loaded image in memory.

But those images can be pretty large. It would be nice to have an option (or open the API) to have a [LRUCache](https://github.com/silx-kit/silx/blob/main/src/silx/gui/plot/items/_cache.py#L7) like.

PR #4498, opened 2026-03-09
silx.gui.plot.Imagestack: use a LRUCache for data displayed

payno:
close https://github.com/silx-kit/silx/issues/4446

### DIFF
--- a/src/silx/gui/plot/ImageStack.py
+++ b/src/silx/gui/plot/ImageStack.py
@@ -164,7 +164,6 @@ class ImageStack(qt.QMainWindow):
         super().__init__(parent)
         self.__n_prefetch = ImageStack.N_PRELOAD
         self._urlData = LRUCache(maxsize=None)
-        """How many image we prefetch"""
         self._loadingThreads = []
         self.setWindowFlags(qt.Qt.Widget)
         self._current_url = None
@@ -305,7 +304,7 @@ class ImageStack(qt.QMainWindow):
             raise ValueError("'n' should be positive")
         if self._urlData.maxsize < self.__n_prefetch:
             _logger.warning(
-                "Incoherent value set for n-prefetch. Will overwrite NKeepInMemory"
+                "Number of prefetchs lower that data cache size: This is not optimal"
             )
             self._urlData.maxsize = None
 
@@ -324,18 +323,18 @@ class ImageStack(qt.QMainWindow):
 
     def getDataCacheSize(self) -> int | None:
         """
-        Return the number of image we want to keep in cache at most.
+        Return the maximum number of images kept in cache.
         """
         return self._urlData.maxsize
 
     def setDataCacheSize(self, n: int | None) -> None:
         """
-        Define the number of image we want to keep in cache at most.
+        Define the maximum number of images to cache.
 
-        :param n: number of images to keeps in cache at most. If -1: all images will be kept.
+        :param n: Maximum number of images to keep in cache. If None: All images are cached.
         """
         if n is not None and n < self.__n_prefetch:
-            raise ValueError("'n' should be higher than 'n' prefetch")
+            raise ValueError("'Data cache size should be higher than 'n' prefetch")
         self._urlData.maxsize = n
 
     def setUrlsEditable(self, editable: bool):
--- a/src/silx/gui/plot/items/_cache.py
+++ b/src/silx/gui/plot/items/_cache.py
@@ -47,7 +47,7 @@ class LRUCache:
 
         If None, the cache has no size limit.
 
-        .. warning: modifying the maximal size might affect the cache.
+        .. warning: decreasing the maximal size affects the cache.
         """
         if maxsize is not None and maxsize < 0:
             raise ValueError("cache max size should be None or higher than 0")


src/silx/gui/plot/ImageStack.py
```diff
@@ -311,6 +320,22 @@ def getNPrefetch(self) -> int:
         """
         return self.__n_prefetch
 
+    def getDataCacheSize(self) -> int:
+        """
+        Return the number of image we want to keep in cache at most.
+        """
+        return self._urlData.maxsize
+
+    def setDataCacheSize(self, n: int) -> None:
```
payno: Expose the `maxsize` of the LRUCache

src/silx/gui/plot/ImageStack.py
```diff
@@ -298,6 +301,12 @@ def setNPrefetch(self, n: int) -> None:
         :param int n: number of url to prefetch on left and right sides.
                       In total n*2 DataUrl will be prefetch
         """
+        if n < 0:
+            raise ValueError("'n' should be positive")
+        if self._urlData.maxsize < self.__n_prefetch:
+            _logger.warning("Incoherent value set for n-prefetch. Will overwrite NKeepInMemory")
+            self._urlData.maxsize = -1
```
payno: A lot of things could be done here. I toke the simplest and most robust option: ignore the maxsize if this is not coherent. But maybe better / most constraining approach could be implemented.

src/silx/gui/plot/ImageStack.py
```diff
@@ -159,9 +160,11 @@ class ImageStack(qt.QMainWindow):
     sigCurrentUrlChanged = qt.Signal(str)
     """Signal emitted when the current url change"""
 
-    def __init__(self, parent=None) -> None:
+    def __init__(self, parent=None, cacheMaxSize=-1) -> None:
         super().__init__(parent)
         self.__n_prefetch = ImageStack.N_PRELOAD
+        self._urlData = LRUCache(maxsize=cacheMaxSize)
```
payno: Creating it in the constructor instead of during reset allow us to keep trace of the `maxsize` value if this evolve over time.

Setting it to `-1` make sure we keep compatibility with existing.

src/silx/gui/plot/items/_cache.py
```diff
@@ -33,6 +33,32 @@ def __getitem__(self, key):
...
+
+        .. warning: modifying the maximal size might affect the cache.
+        """
+        if maxsize < 1 and maxsize != -1:
+            raise ValueError("cache max size should be -1 or higher than 0")
+        new_cache = OrderedDict()
+        max_elmts_kepts = min(len(self._cache), maxsize if maxsize is not None else len(self._cache))
+        # Preserve the most recently added element as the final entry in the cache to maintain state continuity between operations.
+        items_to_copy = tuple(reversed(self._cache.items()))[:max_elmts_kepts]
+        for key, value in reversed(items_to_copy):
```
payno, line 60: IMO having twice `reversed` is easier that computing the indexes to kept.
The 'last' element read / set is at the end of the `OrderedDict`.

src/silx/gui/plot/items/_cache.py
```diff
@@ -13,16 +13,16 @@ class LRUCache:
...
+            raise ValueError("cache max size should be -1 or higher than 0")
         self._maxsize = maxsize
         self._cache = OrderedDict()
 
     def __setitem__(self, key, value):
         self._cache[key] = value
         self._cache.move_to_end(key)
 
-        if len(self) > self._maxsize:
+        if (self._maxsize != -1) and len(self) > self._maxsize:
```
t20100:
```suggestion
        if self._maxsize != -1 and len(self) > self._maxsize:
```

src/silx/gui/plot/items/_cache.py
```diff
@@ -13,16 +13,16 @@ class LRUCache:
     """
 
     def __init__(self, maxsize: int = 128):
-        if maxsize < 1:
-            raise ValueError("cache max size should be higher than 0")
+        if maxsize < 1 and maxsize != -1:
```
t20100: What about using `None` instead of `-1` as for `functools.lru_cache`?

src/silx/gui/plot/items/_cache.py
t20100: Maybe simpler to loop and use [`popitem`](https://docs.python.org/3/library/collections.html#collections.OrderedDict.popitem) 🤷

src/silx/gui/plot/ImageStack.py
t20100: Why not using `max(self._urlData.maxsize, self.__n_prefetch)`?

src/silx/gui/plot/items/_cache.py
payno: Fine with me 👍

src/silx/gui/plot/ImageStack.py
payno: No strong Opinion. But IMO it would be clearer to remove the limit instead of setting it artificially...

src/silx/gui/plot/items/_cache.py
payno: No strong opinion

src/silx/gui/plot/items/_cache.py
```diff
@@ -33,6 +33,34 @@ def __getitem__(self, key):
...
+        """
+        Return cache maximal number of element kept. If None the cache has no size limit.
+        """
+        return self._maxsize
+
+    @maxsize.setter
+    def maxsize(self, maxsize: int | None) -> None:
+        """
+        Modify the number of elements kept in the cache.
+        If -1 the cache has no size limit.
```
t20100:
```suggestion

        If None, the cache has no size limit.
```

src/silx/gui/plot/items/_cache.py
```diff
@@ -33,6 +33,34 @@ def __getitem__(self, key):
...
+
+        .. warning: modifying the maximal size might affect the cache.
```
t20100:
```suggestion
        .. warning: decreasing the maximal size affects the cache.
```

src/silx/gui/plot/ImageStack.py
```diff
@@ -162,6 +163,8 @@ class ImageStack(qt.QMainWindow):
     def __init__(self, parent=None) -> None:
         super().__init__(parent)
         self.__n_prefetch = ImageStack.N_PRELOAD
+        self._urlData = LRUCache(maxsize=None)
+        """How many image we prefetch"""
```
t20100: Why using a docstring here? It doesn't seems related to `_urlData` but more to `__n_prefetch`


```suggestion
        self._urlData = LRUCache(maxsize=None)
```

src/silx/gui/plot/ImageStack.py
```diff
@@ -311,6 +322,22 @@ def getNPrefetch(self) -> int:
         """
         return self.__n_prefetch
 
+    def getDataCacheSize(self) -> int | None:
+        """
+        Return the number of image we want to keep in cache at most.
```
t20100:
```suggestion
        Return the maximum number of images kept in cache.
```

src/silx/gui/plot/ImageStack.py
```diff
@@ -311,6 +322,22 @@ def getNPrefetch(self) -> int:
...
+        """
+        return self._urlData.maxsize
+
+    def setDataCacheSize(self, n: int | None) -> None:
+        """
+        Define the number of image we want to keep in cache at most.
+
+        :param n: number of images to keeps in cache at most. If -1: all images will be kept.
+        """
```
t20100:
```suggestion
        Define the maximum number of images to cache.

        :param n: Maximum number of images to keep in cache. If None: All images are cached.
        """
```

src/silx/gui/plot/ImageStack.py
```diff
@@ -298,6 +301,14 @@ def setNPrefetch(self, n: int) -> None:
         :param int n: number of url to prefetch on left and right sides.
                       In total n*2 DataUrl will be prefetch
         """
+        if n < 0:
+            raise ValueError("'n' should be positive")
+        if self._urlData.maxsize < self.__n_prefetch:
+            _logger.warning(
+                "Incoherent value set for n-prefetch. Will overwrite NKeepInMemory"
```
t20100: This message is not clear to me.

Maybe something like:
```suggestion
                "Number of prefetchs lower that data cache size: This is not optimal"
```

src/silx/gui/plot/ImageStack.py
```diff
@@ -311,6 +322,22 @@ def getNPrefetch(self) -> int:
...
+        return self._urlData.maxsize
+
+    def setDataCacheSize(self, n: int | None) -> None:
+        """
+        Define the number of image we want to keep in cache at most.
+
+        :param n: number of images to keeps in cache at most. If -1: all images will be kept.
+        """
+        if n is not None and n < self.__n_prefetch:
+            raise ValueError("'n' should be higher than 'n' prefetch")
```
t20100:
```suggestion
            raise ValueError("'Data cache size should be higher than 'n' prefetch")
```

t20100: BTW why raising here and not logging a warning as fror n prefetch?

PR #4501, opened 2026-03-09
silx.gui.data.DataViews: Break it down in several modules

loichuder:

- `DataInfo` goes into `_DataInfo.py`
- `DataView` and `DataViewHooks` go into `_DataView.py`
- Util functions go into `_utils.py`
- `CompositeDataView` and views inheriting from it go into `composite.py` 
- Modes (view ids) go into `modes.py`
- NXdata views go into `nexus.py`
- Views inherithing from `DataView`  go into `views.py`


loichuder:
Having slept on it, perhaps we don't have to be so divisive: we can keep `DataViews` for `composite`, `nexus` and `views` views. Perhaps also keeping `modes` inside.

PR #4502, opened 2026-03-10
silx.gui.widgets.IntEdit: Add IntEdit widget (rather make it public)

loichuder:

As discussed previsouly @t20100 , `IntEdit` could be useful in other projects as `FloatEdit`. 

This PR makes it public and adds tests and references to it in the doc.

### DIFF
--- a/doc/source/modules/gui/widgets/editwidgets.rst
+++ b/doc/source/modules/gui/widgets/editwidgets.rst
@@ -1,0 +1,24 @@
+
+.. currentmodule:: silx.gui.widgets
+
+Edition widgets
+===============
+
+:mod:`FloatEdit`
+--------------------
+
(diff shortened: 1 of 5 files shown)


src/silx/gui/widgets/IntEdit.py
loichuder, line 1: Moved without change

doc/source/modules/gui/widgets/editwidgets.rst
```diff
@@ -0,0 +1,24 @@
+
+.. currentmodule:: silx.gui.widgets
+
+Edition widgets
+===============
+
+:mod:`FloatEdit`
```
loichuder, line 7: `FloatEdit` was mentioned in `sample_code/index` but there was no "real" reference in the doc

PR #4508, opened 2026-03-10
silx.io.nxdata, silx.gui.data: Changed NeXus `@interpretation=scalar` is now ignored

t20100:

This PR makes plot selection to behave the same when `@interpretation=scalar|scaler` as if no `@interpretation` is set.
Beforehand, signal with "scalar" interpretation was displayed with a matrix vis for all but 3D datasets.
With this PR, it is displayed with:
- 0D: matrix vis
- 1D: curve
- 2D: image
- 3D: stack (special case to be removed in another PR IMO)
- 4D: image with axes selection

### DIFF
--- a/src/silx/gui/data/DataViews.py
+++ b/src/silx/gui/data/DataViews.py
@@ -1108,7 +1108,7 @@ class _NXdataScalarView(_NXdataBaseDataView):
         from silx.gui.data.ArrayTableWidget import ArrayTableWidget
 
         widget = ArrayTableWidget(parent)
-        # widget.displayAxesSelector(False)
+        widget.displayAxesSelector(False)
         return widget
 
     def axesNames(self, data, info):


src/silx/gui/data/DataViews.py
```diff
@@ -1113,7 +1108,7 @@ def createWidget(self, parent):
         from silx.gui.data.ArrayTableWidget import ArrayTableWidget
 
         widget = ArrayTableWidget(parent)
-        # widget.displayAxesSelector(False)
+        widget.displayAxesSelector(False)
```
t20100, line 1111: This widget is now meant to only display scalars, no longer arrays, so hiding the axes selection (BTW it was at the top of the widget which was inconsistent with the other data views).

For another PR: change the matrix vis to a better view widget

src/silx/io/nxdata/parse.py
```diff
@@ -846,7 +851,7 @@ def is_curve(self) -> bool:
             2,
         ]:
             return False
-        if self.interpretation is None:
+        if self.interpretation in (None, "scalar", "scaler"):
```
t20100, line 854: Needed to be added here to provide the same behavior

loichuder: I think the conditions could be rewritten so that `scalar`/`scaler` should not need to be mentioned explicitly

loichuder: Eh, it's fine.

Issue #4495, opened 2026-03-08
[Bug] silx convert crashes with AttributeError ('list' object has no attribute 'split') when parsing SSRF TIFF metadata


TianyiMa96:
**Description:**

When attempting to convert TIFF files generated by the SSRF (Shanghai Synchrotron Radiation Facility) using `silx convert`, the process crashes with an **AttributeError.**

The TIFF file in question is an SSRF (Shanghai Synchrotron Radiation Facility) dataset that we pre-processed (background subtraction) and saved using fabio. During our saving process, some custom header data was inadvertently saved as Python lists (e.g., [825], [9906100]) and dicts (e.g., {}).

While the metadata origin is custom, it exposes a lack of robustness in silx.io.fabioh5. The _convert_metadata_vector method assumes all non-scalar values are space-separated strings and blindly calls .split(" ") on them, causing a hard crash when encountering lists or dicts.

**Steps to Reproduce:**

Run the following command on the attached SSRF TIFF file:
`silx convert processed_HDPE-STANDARD-sd1-5s_00015.tiff -o output.h5`

**Error Traceback:**

Traceback (most recent call last):

  ...
  File "...\site-packages\silx\io\fabioh5.py", line 612, in _convert_metadata_vector
    array.append(v.split(" "))
AttributeError: 'list' object has no attribute 'split'

**Root Cause Analysis:**

At line 612 in silx/io/fabioh5.py:

```
if c.shape != tuple():
    array.append(v.split(" ")) # <--- Crashes here when v is [825] or {}
    is_array = True
```

When v is a list (e.g., [825]) or an empty dict ({}) parsed directly from the TIFF header by fabio, calling .split() raises the AttributeError.

Furthermore, we noticed that if we bypass this split error, the conversion eventually fails at the HDF5 writing stage with:
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
This happens because h5py cannot natively serialize a Python dict ({}) into HDF5 format.

**Suggested Fix:**

In silx/io/fabioh5.py, _convert_metadata_vector needs type checking before splitting(in line 612). For dict types, serializing them to JSON strings ensures HDF5 compatibility.

```
if c.shape != tuple():
    if isinstance(v, str):
        array.append(v.split(" "))
    elif isinstance(v, (list, tuple, numpy.ndarray)):
        array.append(list(v))
    elif isinstance(v, dict):
        import json
        array.append([json.dumps(v)])
    else:
        array.append([v])
    is_array = True
[...]

PR #4515, opened 2026-03-11
silx.io.fabioh5.FabioReader: Fix Attribute error when reading TIFF files

loichuder:

For #4495 

The `_convert_metadata_vector` seemed to hold some weird parts. One of those was the `.split` call on a value that was not type-checked.

Rather than trying to typecheck `v`, I analyzed the code further to remove some logic (including the `split`). The tests are still passing so I guess it is fine :shrug: 

I also added a new test to prevent regression and some typings to the best of my inferred knowledge.

### DIFF
--- a/src/silx/io/fabioh5.py
+++ b/src/silx/io/fabioh5.py
@@ -575,7 +575,6 @@ class FabioReader:
         return previous
 
     def _normalize_vector_type(self, dtype):
-        """Normalize the"""
         if self.__at_least_32bits:
             if numpy.issubdtype(dtype, numpy.signedinteger):
                 dtype = numpy.result_type(dtype, numpy.uint32)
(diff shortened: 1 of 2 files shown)


src/silx/io/fabioh5.py
```diff
@@ -591,46 +591,27 @@ def _normalize_vector_type(self, dtype):
...
-        result_type = numpy.result_type(*types)
-
-        if issubclass(result_type.type, numpy.bytes_):
-            # use the raw data to create the array
-            result = values
-        elif issubclass(result_type.type, numpy.str_):
-            # use the raw data to create the array
-            result = values
-        else:
-            result = converted
```
loichuder, line 631: Letting `numpy` handle this for us. `converted` is now used as the `result` except in the particular case where precision is needed (see below)

src/silx/io/test/test_fabioh5.py
```diff
@@ -422,18 +423,13 @@ def test_dirty_header(self):
...
-            _logger.error(e.args[0])
-            _logger.debug("Backtrace", exc_info=True)
-            self.skipTest("fabio do not allow to create the resource")
+        header = {}
+        header["foo"] = b"abc"
+        data = numpy.array([[0, 0], [0, 0]], dtype=numpy.int8)
+        fabio_image = fabio.edfimage.edfimage(data=data, header=header)
+        header = {}
+        header["foo"] = "abc\u2764"
+        fabio_image.append_frame(data=data, header=header)
```
loichuder, line 432: I think we want the test to fail in case of exceptions, not skip it.

src/silx/io/fabioh5.py
```diff
@@ -591,46 +591,27 @@ def _normalize_vector_type(self, dtype):
...
                 converted.append(None)
                 has_none = True
-                array.append(None)
             else:
                 c = self._convert_value(v)
-                if c.shape != tuple():
-                    array.append(v.split(" "))
-                    is_array = True
-                else:
-                    array.append(v)
```
t20100, line 614: From what I understand, this should already be handled by `_convert_value`

src/silx/io/fabioh5.py
```diff
@@ -591,46 +591,27 @@ def _normalize_vector_type(self, dtype):
...
-            # use the raw data to create the array
-            result = values
-        elif issubclass(result_type.type, numpy.str_):
-            # use the raw data to create the array
-            result = values
-        else:
-            result = converted
-
-        result_type = self._normalize_vector_type(result_type)
+        result_type = self._normalize_vector_type(numpy.result_type(*types))
```
t20100, line 617: BTW, passing arrays, scalar to result_type also works.
This should also work:
```suggestion
        result_type = self._normalize_vector_type(numpy.result_type(*[c for c in converted if c is not None]))
```
and spare the `types` variable

src/silx/io/fabioh5.py
t20100: No sure about this. This sound safer to avoid compression to number and back to str/bytes in this case.

t20100: I propose to keep this for now, sound unsafe to change

src/silx/io/fabioh5.py
loichuder: You mean passing the values themselves rather than the types?

t20100: Yes, this works.
Up to you

loichuder: This breaks the tests for some reason.

I refactored the `_convert_metadata_vector` function but kept the `types` variable.

src/silx/io/fabioh5.py
```diff
@@ -575,7 +575,6 @@ def _set_vector_normalization(self, at_least_32bits, signed_type):
         return previous
 
     def _normalize_vector_type(self, dtype):
-        """Normalize the"""
```
t20100, line 578: 😅

src/silx/io/fabioh5.py
```diff
@@ -591,85 +590,62 @@ def _normalize_vector_type(self, dtype):
...
+    def _get_none_value(self, dtype: numpy.dtype):
+        """Finds the fill value for missing data according to the array type"""
+        if dtype.kind == "S":
+            return b""
+        if dtype.kind == "U":
+            return ""
+        if dtype.kind == "f":
+            return dtype.type("NaN")
+        if dtype.kind in ("i", "u", "b"):
+            return dtype.type(0)
```
t20100, line 602: 👍

PR #4516, opened 2026-03-11
silx.io.fabioh5.FabioReader: Convert to scalar strings to np.str_ rather than np.bytes_

### LINKED ISSUE #4495: [Bug] silx convert crashes with AttributeError ('list' object has no attribute 'split') when parsing SSRF TIFF metadata

loichuder:

Fix #4495

Since h5py v3, `np.str` is handled just fine so we don't have to manually convert to `np.bytes`. This simplifies the code and solves the bug reported in https://github.com/silx-kit/silx/issues/4495#issuecomment-4035846647.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = ["Development Status :: 5 - Production/Stable",
 dependencies = [
    'numpy',
    'packaging',
-   'h5py',
+   'h5py >= 3',
    'fabio',
    'pydantic >= 2',
 ]
--- a/src/silx/io/commonh5.py
+++ b/src/silx/io/commonh5.py
@@ -168,36 +168,7 @@ class Dataset(Node):
 
     def __init__(self, name, data, parent=None, attrs=None):
         Node.__init__(self, name, parent, attrs=attrs)
-        if data is not None:
-            self._check_data(data)
         self.__data = data
-
-    def _check_data(self, data):
-        """Check that the data provided by the dataset is valid.
-
-        It is valid when it can be stored in a HDF5 using h5py.
-
-        :param numpy.ndarray data: Data associated to the dataset
-        :raises TypeError: In the case the data is not valid.
-        """
-        if isinstance(data, (str, bytes)):
-            return
-
-        chartype = data.dtype.char
-        if chartype == "U":
-            pass
-        elif chartype == "O":
-            d = h5py.special_dtype(vlen=data.dtype)
-            if d is not None:
-                return
-            d = h5py.special_dtype(ref=data.dtype)
-            if d is not None:
-                return
-        else:
-            return
-
-        msg = "Type of the dataset '%s' is not supported. Found '%s'."
-        raise TypeError(msg % (self.name, data.dtype))
 
     def _set_data(self, data):
         """Set the data exposed by the dataset.
@@ -207,7 +178,6 @@ class Dataset(Node):
 
         :param numpy.ndarray data: Data associated to the dataset
         """
-        self._check_data(data)
         self.__data = data
 
     def _get_data(self):
--- a/src/silx/io/fabioh5.py
+++ b/src/silx/io/fabioh5.py
@@ -630,13 +630,7 @@ class FabioReader:
     def _convert_value(
         self, value: list | dict | str | bytes | None
     ) -> (
-        numpy.ndarray
-        | numpy.void
-        | numpy.floating
-        | numpy.integer
-        | numpy.bytes_
-        | numpy.str_
-        | None
+        numpy.ndarray | numpy.void | numpy.floating | numpy.integer | numpy.str_ | None
     ):
         """Convert an object into a numpy object (scalar or array).
 
@@ -674,7 +668,7 @@ class FabioReader:
 
     def _convert_scalar_value(
         self, value: str
-    ) -> numpy.bytes_ | numpy.integer | numpy.floating:
+    ) -> numpy.str_ | numpy.integer | numpy.floating:
         """Convert a string into a numpy int or float.
 
         If it is not possible it returns a numpy string.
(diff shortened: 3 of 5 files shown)


src/silx/io/test/test_commonh5.py
```diff
@@ -268,11 +265,8 @@ def test_readonly_create_group(self):
 
     def test_create_unicode_dataset(self):
         f = commonh5.File(name="Foo", mode="w")
-        try:
-            f.create_dataset("foo", data=numpy.array("aaaa"))
-            self.fail()
-        except TypeError:
-            pass
+        f.create_dataset("foo", data=numpy.array("aaaa"))
+        self.assertEqual(f["foo"][()], "aaaa")
```
loichuder, line 269: This is now supported  🎉

src/silx/io/commonh5.py
```diff
@@ -168,37 +168,8 @@ class Dataset(Node):
 
     def __init__(self, name, data, parent=None, attrs=None):
         Node.__init__(self, name, parent, attrs=attrs)
-        if data is not None:
-            self._check_data(data)
         self.__data = data
 
-    def _check_data(self, data):
```
loichuder, line 175: This function was raising an error if `data` was of type `U` since it could not be handled by h5py.

This is no longer the case so it can be removed.

src/silx/io/fabioh5.py
```diff
@@ -675,47 +669,37 @@ def _convert_value(
...
             numpy_type = silx.utils.number.min_numerical_convertible_type(value)
-            converted = numpy_type(value)
+            return numpy_type(value)
         except ValueError:
-            converted = numpy.bytes_(value)
-        return converted
+            return numpy.str_(value)
 
-    def _convert_list(self, value: str) -> numpy.ndarray | numpy.bytes_ | numpy.str_:
+    def _convert_list(self, value: str) -> numpy.ndarray | numpy.str_:
```
loichuder, line 689: Refactored this to reduce the extent of the `try`/`except` and use more functional code.

There are two main changes in addition of this: 
- the `if` branch `issubclass(result_type.type, (numpy.bytes_, bytes))` was removed since `_convert_scalar_value` does not return `bytes` anymore
- the fallback `np.bytes(value)` was replaced by `np.str(value)`

src/silx/io/fabioh5.py
```diff
@@ -675,47 +669,37 @@ def _convert_value(
...
+
+        converted_values = [self._convert_scalar_value(v) for v in raw_values]
+        types = set(v.dtype.type for v in converted_values)
+        result_type = numpy.result_type(*types)
+
+        if issubclass(result_type.type, (numpy.str_, str)):
+            return numpy.str_(value)
+
+        # Keep raw values if there are multiple types to avoid loss of precision
+        result = raw_values if len(types) > 1 else converted_values
```
t20100: With PR #4519 this will not be needed anymore

pyproject.toml
```diff
@@ -33,7 +33,7 @@ classifiers = ["Development Status :: 5 - Production/Stable",
 dependencies = [
    'numpy',
    'packaging',
-   'h5py < 3.16',
+   'h5py >2, < 3.16',
```
t20100:
```suggestion
   'h5py >= 3, < 3.16',
```

2.1 should be >2

src/silx/io/fabioh5.py
loichuder: I'll let you merge first then :wink:

t20100: PR #4519 merged

PR #4518, opened 2026-03-11
silx.gui.data.Hdf5TableView: Add a 'copyable' button

payno:
- [ ] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

Screenshot from 2026-03-11 17-11-42

# Note on qtawesome

## distribution
qtawesome is distributed with conda and debian

Screenshot from 2026-03-10 10-53-00

Screenshot from 2026-03-10 10-52-31

## license

The qtawesome on itself is MIT but [each part has a different version](https://github.com/spyder-ide/qtawesome/tree/master?tab=readme-ov-file#license)
The one used here ~(fontAwesome & elusive Icon) is under the [SIL Open Font License](http://scripts.sil.org/OFL).~
Material Design Icons is under Apache License version 2.0.

### DIFF
--- a/src/silx/gui/data/Hdf5TableView.py
+++ b/src/silx/gui/data/Hdf5TableView.py
@@ -601,7 +601,7 @@ class _CopyableQLineEdit(qt.QWidget):
 
         self.setLayout(qt.QHBoxLayout())
         self.layout().addWidget(self._qLineEdit)
-        self._button = qt.QPushButton(icon=qtawesome.icon("fa6.clipboard"))
+        self._button = qt.QPushButton(icon=qtawesome.icon("mdi6.content-copy"))
         self._button.setFlat(True)
         self.layout().addWidget(self._button)
 


src/silx/gui/data/Hdf5TableView.py
```diff
@@ -35,6 +35,8 @@
 import logging
 import h5py
 import numpy
+import qtawesome
```
t20100, line 38: This needs to be added to pyproject.toml and documentation

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -71,6 +74,10 @@ def isHeader(self):
         """
         return self.__isHeader
 
+    def isCopyable(self):
```
t20100:
```suggestion
    def isCopyable(self) -> bool:
```

src/silx/gui/widgets/HierarchicalTableView.py
```diff
@@ -66,6 +66,9 @@ class HierarchicalTableModel(qt.QAbstractTableModel):
     UserRole = qt.Qt.UserRole + 2
     """First index of user defined roles"""
 
+    IsCopyableRole = qt.Qt.UserRole + 3
+    """Role returning True if the content of the cell can be copyable by the user"""
```
t20100:
```suggestion
    """Role returning True if the content of the cell can be copied to clipboard by the user"""
```

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+        self.layout().setContentsMargins(0, 0, 0, 0)
+        self.layout().setSpacing(0)
+
+        # set up
+        self._qLineEdit.setReadOnly(True)
+        self._qLineEdit.deselect()
+        self._isCopyable = True
+
+        # connect signal / slot
+        self._button.released.connect(self.copyToClipBoard)
```
t20100:
```suggestion
        self._button.clicked.connect(self.copyToClipBoard)
```

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+        self._button = qt.QPushButton(icon=_CopyableQLineEdit._getIcon())
+        self._button.setFlat(True)
+        self.layout().addWidget(self._button)
+
+        self.layout().setContentsMargins(0, 0, 0, 0)
+        self.layout().setSpacing(0)
+
+        # set up
+        self._qLineEdit.setReadOnly(True)
+        self._qLineEdit.deselect()
```
t20100: is it needed here?

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+
+    def setCopyable(self, copyable: bool):
+        self._isCopyable = copyable
+
+    def copyToClipBoard(self) -> None:
+        """Copy data to the clipboard"""
+        qt.QApplication.clipboard().setText(self._qLineEdit.text())
+
+    @staticmethod
+    @lru_cache(maxsize=1)
```
t20100: Isn't `qtawesome` caching icons?

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+    def _getIcon() -> qt.QIcon | qt.QColor | qt.QPixmap:
```
t20100: Can't the type be restricted?

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+    textChanged = qt.Signal(str)
+
+    def __init__(self, qlineEdit: qt.QLineEdit):
+        super().__init__(parent=qlineEdit.parent())
+        self._qLineEdit = qt.QLineEdit(text=qlineEdit.text())
+        self._qLineEdit.setFrame(False)
+
+        self.setLayout(qt.QHBoxLayout())
+        self.layout().addWidget(self._qLineEdit)
+        self._button = qt.QPushButton(icon=_CopyableQLineEdit._getIcon())
```
t20100: No need to keep a reference on the button as an attribute

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+    """A non editable QLineEdit embedding an optional button to copy the text contained"""
+    textChanged = qt.Signal(str)
+
+    def __init__(self, qlineEdit: qt.QLineEdit):
+        super().__init__(parent=qlineEdit.parent())
+        self._qLineEdit = qt.QLineEdit(text=qlineEdit.text())
+        self._qLineEdit.setFrame(False)
+
+        self.setLayout(qt.QHBoxLayout())
+        self.layout().addWidget(self._qLineEdit)
```
t20100, line 603: no need to set a stretch to keep the icon on the right side?

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+        self._isCopyable = copyable
+
+    def copyToClipBoard(self) -> None:
+        """Copy data to the clipboard"""
+        qt.QApplication.clipboard().setText(self._qLineEdit.text())
+
+    @staticmethod
+    @lru_cache(maxsize=1)
+    def _getIcon() -> qt.QIcon | qt.QColor | qt.QPixmap:
+        return qtawesome.icon("fa6.clipboard")
```
t20100: BTW, how does this behaves in dark mode? and when switching?

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
+    def deselect(self):
+        self._qLineEdit.deselect()
+
+
 class Hdf5TableItemDelegate(HierarchicalTableView.HierarchicalItemDelegate):
     """Item delegate the :class:`Hdf5TableView` with read-only text editor"""
 
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
```
t20100: not needed

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
     def createEditor(self, parent, option, index):
         """See :meth:`QStyledItemDelegate.createEditor`"""
         editor = super().createEditor(parent, option, index)
         if isinstance(editor, qt.QLineEdit):
-            editor.setReadOnly(True)
-            editor.deselect()
+            editor = _CopyableQLineEdit(editor)
```
t20100: IMO best to not call `createEditor` if not using the widget it returns, this is looking for trouble.
Something like:
```
if copyable_editor_needed:
    editor = _CopyableQLineEdit(parent)
    ...
    return editor
else:
    return super().createEditor(parent, option, index)
```
would be safer

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
...
             self.installEventFilter(editor)
         return editor
 
+    def setEditorData(self, editor, index):
+        if isinstance(editor, _CopyableQLineEdit):
+            is_copyable = index.model().data(index, HierarchicalTableView.HierarchicalTableModel.IsCopyableRole)
+            editor.setCopyable(is_copyable)
+            editor.setText(index.model().data(index, qt.Qt.DisplayRole))
+
+        super().setEditorData(editor, index)
```
t20100:
```suggestion
        else:
            super().setEditorData(editor, index)
```

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,77 @@ def __formatChanged(self):
         self.reset()
 
 
+class _CopyableQLineEdit(qt.QWidget):
+    """A non editable QLineEdit embedding an optional button to copy the text contained"""
+    textChanged = qt.Signal(str)
+
+    def __init__(self, qlineEdit: qt.QLineEdit):
+        super().__init__(parent=qlineEdit.parent())
+        self._qLineEdit = qt.QLineEdit(text=qlineEdit.text())
```
t20100: Sounds risky, best to pass parent

src/silx/gui/data/Hdf5TableView.py
payno: Doesn't seems like. The QLineEdit already has an 'expand' behavior. But if you prefer I can add one.

Image

src/silx/gui/data/Hdf5TableView.py
payno: Not sure I get it. You mean `self._qLineEdit=qlineEdit` ?
If I do so I get some segfault. I guess this is deleted somehow by the Qt underlying mechanism

src/silx/gui/data/Hdf5TableView.py
payno: This is useful downstream because making the button visible is optional.

src/silx/gui/data/Hdf5TableView.py
payno: Indeed not.

src/silx/gui/data/Hdf5TableView.py
payno: Indeed nice catch 👍

src/silx/gui/data/Hdf5TableView.py
payno: call to the super is done upstream. This is what returns the original `QLineEdit`.
Or I might not understand

src/silx/gui/data/Hdf5TableView.py
payno: Indeed the switching seems to fails:

Image

Image

Image

I guess the icon should be regenerate when we switch to dark mode. Is there some existing mechanism... ?

payno: @loichuder was not very happy with the icon.
Here are other possible options:

Image

Image

Image

Image

Image

Image

Or if you have better option I would be happy to consider it. Else I will be fine with it.

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -552,19 +568,68 @@ def __formatChanged(self):
         self.reset()
 
 
+class _CopyableQLineEdit(qt.QWidget):
+    """A non editable QLineEdit embedding an optional button to copy the text contained"""
+    textChanged = qt.Signal(str)
+
+    def __init__(self, qlineEdit: qt.QLineEdit):
+        super().__init__(parent=qlineEdit.parent())
+        self._qLineEdit = qt.QLineEdit()
```
payno: Taking directly the `qlineEdit` makes segfault. I guess this is removed by the Qt editor mechanism behind the scene.

src/silx/gui/data/Hdf5TableView.py
```diff
@@ -412,7 +448,7 @@ def link(x):
                     def link(x):
                         return x.path
 
-                self.__data.addHeaderValueRow("Link", link)
+                self.__data.addHeaderValueRow("Link", link, isCopyable=True)
```
payno, line 450: What is copyable or not is very debatable. We can modify it if needed.

src/silx/gui/data/Hdf5TableView.py
loichuder: `ri.file-copy-line` looks good.

Similar to what you have on GitHub:

```
This is copyable
```

t20100: Great if icons uses dark/light mode.
Let's keep the handling of dark/light mode updates for later.

payno: Small updated: the previous icon (`fa6.clipboard`) use to handle the light / dark mode.
The new one (`mdi6.content-copy`) seems not to handle it...

PR #4521, opened 2026-03-12
silx.gui.plot3d.scene.functions: Added typing

t20100:

This PR adds type hints and simplifies docstrings accordingly.

Follow-up of comment in #4491: https://github.com/silx-kit/silx/pull/4491/changes#r2923570505

### DIFF
--- a/src/silx/gui/plot3d/scene/function.py
+++ b/src/silx/gui/plot3d/scene/function.py
@@ -51,6 +51,9 @@ if TYPE_CHECKING:
 _logger = logging.getLogger(__name__)
 
 
+Vec3 = tuple[float, float, float]
+
+
 class ProgramFunction:
     """Class providing a function to add to a GLProgram shaders."""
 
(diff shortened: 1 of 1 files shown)


src/silx/gui/plot3d/scene/function.py
```diff
@@ -32,26 +32,34 @@
...
+
 
 from ... import _glutils
 from ..._glutils import gl
 
 from . import event
 from . import utils
 
+if TYPE_CHECKING:
+    from .viewport import RenderContext, Viewport
```
t20100, line 48: Avoids circular references

loichuder: Neat trick! Thanks for the idea!

src/silx/gui/plot3d/scene/function.py
```diff
@@ -186,30 +187,34 @@ class ClippingPlane(ProgramFunction):
     void clipping(vec4 position) {}
     """
 
-    def __init__(self, point=(0.0, 0.0, 0.0), normal=(0.0, 0.0, 0.0)):
+    def __init__(
+        self,
+        point: Sequence[float] = (0.0, 0.0, 0.0),
+        normal: Sequence[float] = (0.0, 0.0, 0.0),
```
loichuder:
```suggestion
        point: tuple[float, float, float] = (0.0, 0.0, 0.0),
        normal: tuple[float, float, float] = (0.0, 0.0, 0.0),
```

Note that we could introduce a custom type `Vec3 = tuple[float, float, float]`

src/silx/gui/plot3d/scene/function.py
t20100: First time I use it ;)

src/silx/gui/plot3d/scene/function.py
t20100: Done

PR #4522, opened 2026-03-12
silx.app.view.DataPanel: allow copy header text

payno:
close #4518 

### DIFF
--- a/src/silx/app/view/DataPanel.py
+++ b/src/silx/app/view/DataPanel.py
@@ -43,18 +43,22 @@ class _HeaderFrame(qt.QFrame):
         self.setFrameShape(qt.QFrame.StyledPanel)
         self._qLineEdit = qt.QLabel(parent)
 
-        self.setLayout(qt.QHBoxLayout())
-        self.layout().addWidget(self._qLineEdit)
-        self._button = qt.QPushButton(icon=qtawesome.icon("mdi6.content-copy"))
+        layout = qt.QHBoxLayout()
+        layout.addWidget(self._qLineEdit)
+        self._button = qt.QPushButton(
+            icon=qtawesome.icon("mdi6.content-copy", active="mdi6.check")
+        )
         self._button.setFlat(True)
 
-        self.layout().addWidget(self._button)
+        layout.addWidget(self._button)
+        self.setLayout(layout)
 
         self._qLineEdit.setSizePolicy(
             qt.QSizePolicy.Expanding, qt.QSizePolicy.Preferred
         )
 
-        self.layout().setContentsMargins(3, 0, 3, 0)
+        layout.setContentsMargins(3, 0, 3, 0)
+        self.setLayout(layout)
 
         # connect signal / slot
         self._button.clicked.connect(self.copyToClipBoard)


src/silx/app/view/DataPanel.py
```diff
@@ -28,17 +28,18 @@
...
+import qtawesome
 
 from silx.gui import qt
 from silx.gui.data.DataViewerFrame import DataViewerFrame
 
 _logger = logging.getLogger(__name__)
 
 
-class _HeaderLabel(qt.QLabel):
+class _HeaderQLineEdit(qt.QLineEdit):
```
payno: To have a coherent display I needed to move this from a `Frame` to a `QLineEdit`.
The frame style has been moved to `_HeaderFrame`.

src/silx/app/view/DataPanel.py
```diff
@@ -86,13 +87,44 @@ def paintEvent(self, event):
         super().paintEvent(event)
 
 
+class _HeaderFrame(qt.QFrame):
```
payno, line 39: This is very close to `_CopyableQLineEdit`. Could have a common mix-in class but to me it would add complexity for not much.

src/silx/app/view/DataPanel.py
```diff
@@ -86,13 +87,44 @@ def paintEvent(self, event):
...
+        self.setLayout(qt.QHBoxLayout())
+        self.layout().addWidget(self._qLineEdit)
+        self._button = qt.QPushButton(icon=qtawesome.icon("mdi6.content-copy"))
+        self._button.setFlat(True)
+
+        self.layout().addWidget(self._button)
+
+        self._qLineEdit.setSizePolicy(
+            qt.QSizePolicy.Expanding, qt.QSizePolicy.Preferred
+        )
```
payno, line 58: Make sure the button keeps a minimal space.

src/silx/app/view/DataPanel.py
```diff
@@ -86,13 +87,44 @@ def paintEvent(self, event):
...
+        # connect signal / slot
+        self._button.clicked.connect(self.copyToClipBoard)
+
+    def copyToClipBoard(self) -> None:
+        """Copy data to the clipboard"""
+        qt.QApplication.clipboard().setText(self._qLineEdit.text())
+
+    # expose API
+    def setData(self, filename, path):
+        self._qLineEdit.setData(filename=filename, path=path)
```
payno: `setData` is the already existing API of `_HeaderQLineEdit`. No much interest to refactor or rename here.

src/silx/app/view/DataPanel.py
payno, line 39: The bigger work is to keep something coherent with existing and the previous version

Today it looks like:
Image
And it used to looks like (silx 2.2.2):
Image

t20100: Isn't it possible to make the height smaller? setContentsMargins maybe?

payno: > Isn't it possible to make the height smaller? setContentsMargins maybe?

I removed margins:

image

Is it better ?

src/silx/app/view/DataPanel.py
t20100: I would merge _HeaderQLineEdit and _HeaderFrame in a single class rather than having proxy methods for setData.

Also why using a LineEdit instead of a QLabel as before?

src/silx/app/view/DataPanel.py
t20100: Yes, thanks!

src/silx/app/view/DataPanel.py
t20100: I would just merge both classes

src/silx/app/view/DataPanel.py
```diff
@@ -28,17 +28,40 @@
...
 from silx.gui import qt
 from silx.gui.data.DataViewerFrame import DataViewerFrame
 
 _logger = logging.getLogger(__name__)
 
 
-class _HeaderLabel(qt.QLabel):
-    def __init__(self, parent=None):
-        qt.QLabel.__init__(self, parent=parent)
+class _HeaderFrame(qt.QFrame):
```
payno, line 39: Maybe `_Header` would be a better name...

src/silx/app/view/DataPanel.py
```diff
@@ -28,17 +28,40 @@
...
+
+        self.layout().addWidget(self._button)
+
+        self._qLineEdit.setSizePolicy(
+            qt.QSizePolicy.Expanding, qt.QSizePolicy.Preferred
+        )
+
+        self.layout().setContentsMargins(3, 0, 3, 0)
+
+        # connect signal / slot
```
loichuder, line 63: Comment is not needed IMO

src/silx/app/view/DataPanel.py
```diff
@@ -28,17 +28,40 @@
...
+        self._button.clicked.connect(self.copyToClipBoard)
+
+    def copyToClipBoard(self) -> None:
+        """Copy data to the clipboard"""
```
loichuder: Same IMO, this docstring not needed 🤡

src/silx/app/view/DataPanel.py
```diff
@@ -28,17 +28,40 @@
...
-    def __init__(self, parent=None):
-        qt.QLabel.__init__(self, parent=parent)
+class _HeaderFrame(qt.QFrame):
+
+    def __init__(self, parent):
+        super().__init__(parent=parent)
         self.setFrameShape(qt.QFrame.StyledPanel)
+        self._qLineEdit = qt.QLabel(parent)
+
+        self.setLayout(qt.QHBoxLayout())
```
loichuder: I am nit-picking here but for type-inference, it is better to store the layout in a variable and to use it to add widgets since `layout()` can return `None`.

```python
self.setLayout(qt.QHBoxLayout())
self.layout().addWidget(self._qLineEdit) # <--- Type-inference will tell you that `addWidget` cannot be called on None
```
```python
layout = qt.QHBoxLayout()
layout.addWidget(self._qLineEdit) # <--- OK
...
self.setLayout(layout)
```

or, to avoid calling `setLayout`
```python
layout = qt.QHBoxLayout(self)
layout.addWidget(self._qLineEdit)
```

src/silx/app/view/DataPanel.py
payno: Well to me having the signal / slot connection on a specific section helps finding out where things are when you read back.
Here the piece of code is very small but still help IMO

src/silx/app/view/DataPanel.py
payno: e35fcbc33a3926249ed1d4ae3146e8e7d59b1c9d

src/silx/app/view/DataPanel.py
loichuder: Yeah ok fine :+1:

src/silx/app/view/DataPanel.py
loichuder: `_TitleFrame` ?

src/silx/app/view/DataPanel.py
loichuder: img

Issue #4434, opened 2025-10-01
[silx view] Some toolbar actions are not relevant for RGB images


loichuder:
- Change colormap
- Show/hide Colorbar
- Aggregation

PR #4525, opened 2026-03-12
silx.gui.data.RgbaImagePlot: Added new widget to plot NXdata RGBA images

loichuder:

For #4434 

Last refactoring step before moving to the actual bug fix.

In this PR, I breakdown the `ArrayImagePlot` widget that handles RGBA images in two widgets: 
- `ArrayImagePlot` that only handle images (regular or irregular, but no RGBA)
- `RgbaImagePlot` that only handle regular RGBA images

To avoid copy-pasting too much code, I created a new base widget `BaseImagePlot` that abstract a lot of common behaviour of the two aforementionned widgets.

I also introduced some utility functions that can be used elsewhere (e.g. `ArrayComplexImagePlot`). There is still some duplication but I think it was reduced and I had to stop refactoring at some point :clown_face: 

The changelist may be a bit too big but I'll try to underline the changes in comments. If needed, I can break down in several commits.

### DIFF
--- a/src/silx/gui/data/DataViews.py
+++ b/src/silx/gui/data/DataViews.py
@@ -40,6 +40,7 @@ from silx.gui.plot.items.image import ImageDataAggregated
 from silx.gui.plot.actions.image import AggregationModeAction
 from ._DataInfo import DataInfo
 from ._DataView import DataView
+from ._RgbaImagePlot import RgbaImagePlot
 
 # DataViewHooks is part of the public API of this module
 from ._DataView import DataViewHooks  # noqa: F401
@@ -1303,7 +1304,6 @@ class _NXdataImageView(_NXdataBaseDataView):
             axes_names=nxd.axes_names,
             axes_scales=nxd.plot_style.axes_scale_types,
             title=nxd.title,
-            isRgba=False,
         )
 
     def getDataPriority(self, data, info: DataInfo):
@@ -1327,10 +1327,7 @@ class _NXDataRgbaImageView(_NXdataBaseDataView):
         _NXdataBaseDataView.__init__(self, parent, modeId=NXDATA_RGBA_IMAGE_MODE)
 
     def createWidget(self, parent):
-        from silx.gui.data.NXdataWidgets import ArrayImagePlot
-
-        widget = ArrayImagePlot(parent)
-        return widget
+        return RgbaImagePlot(parent)
 
     def axesNames(self, data, info):
         # disabled (used by default axis selector widget in Hdf5Viewer)
@@ -1345,17 +1342,13 @@ class _NXDataRgbaImageView(_NXdataBaseDataView):
         if nxd is None:
             return
 
-        self._updateColormap(nxd)
-
-        widget: ArrayImagePlot = self.getWidget()
+        widget: RgbaImagePlot = self.getWidget()
         widget.setImageData(
             [nxd.signal] + nxd.auxiliary_signals,
             axes=nxd.axes,
-            signals_names=[nxd.signal_name] + nxd.auxiliary_signals_names,
-            axes_names=nxd.axes_names,
-            axes_scales=nxd.plot_style.axes_scale_types,
+            signalsNames=[nxd.signal_name] + nxd.auxiliary_signals_names,
+            axesNames=nxd.axes_names,
             title=nxd.title,
-            isRgba=True,
         )
 
     def getDataPriority(self, data, info: DataInfo):
--- a/src/silx/gui/data/NXdataWidgets.py
+++ b/src/silx/gui/data/NXdataWidgets.py
@@ -29,21 +29,24 @@ __date__ = "12/11/2018"
 
 import logging
 from typing import Literal
-import numpy
+
 import h5py
+import numpy
 
 from silx.gui import qt
+from silx.gui.colors import Colormap
+from silx.gui.data._RgbaImagePlot import BaseImagePlot
+from silx.gui.data._SignalSelector import SignalSelector
 from silx.gui.data.NumpyAxesSelector import NumpyAxesSelector
-from silx.gui.plot import Plot1D, Plot2D, StackView, ScatterView, items
+from silx.gui.plot import Plot1D, ScatterView, StackView, items
+from silx.gui.plot.actions.image import AggregationModeAction
 from silx.gui.plot.ComplexImageView import ComplexImageView
 from silx.gui.plot.items.image_aggregated import ImageDataAggregated
-from silx.gui.plot.actions.image import AggregationModeAction
-from silx.gui.colors import Colormap
-from silx.gui.data._SignalSelector import SignalSelector
-
 from silx.io.commonh5 import Dataset
-from silx.io.nxdata._utils import get_attr_as_unicode
-from silx.math.calibration import ArrayCalibration, NoCalibration, LinearCalibration
+from silx.math.calibration import ArrayCalibration, LinearCalibration, NoCalibration
+
+from ._utils import getAxesCalib, setImageCoords
+from ..utils import blockSignals
 
 _logger = logging.getLogger(__name__)
 
@@ -366,7 +369,7 @@ class XYVScatterPlot(qt.QWidget):
         self._plot.getPlotWidget().clear()
 
 
-class ArrayImagePlot(qt.QWidget):
+class ArrayImagePlot(BaseImagePlot):
     """
     Widget for plotting an image from a multi-dimensional signal array
     and two 1D axes array.
@@ -387,11 +390,8 @@ class ArrayImagePlot(qt.QWidget):
         :param parent: Parent QWidget
         """
         super().__init__(parent)
+        self._axesScales = None
 
-        self.__signals = None
-        self.__signals_names = None
-
-        self._plot = Plot2D(self)
         self._plot.setDefaultColormap(
             Colormap(
                 name="viridis", vmin=None, vmax=None, normalization=Colormap.LINEAR
@@ -399,23 +399,6 @@ class ArrayImagePlot(qt.QWidget):
         )
         self._plot.getIntensityHistogramAction().setVisible(True)
         self._plot.setKeepDataAspectRatio(True)
-        maskToolWidget = self._plot.getMaskToolsDockWidget().widget()
-        maskToolWidget.setItemMaskUpdated(True)
-
-        self._axesSelector = NumpyAxesSelector(self)
-        self._axesSelector.selectionChanged.connect(self._updateImage)
-        self._axesSelector.selectedAxisChanged.connect(self._updateImageAxes)
-
-        self._signalSelector = SignalSelector(parent=self)
-        self._signalSelector.selectionChanged.connect(self._signalChanges)
-        self._signalSelector.setToolTip("Select signal")
-
-        layout = qt.QVBoxLayout()
-        layout.addWidget(self._plot)
-        layout.addWidget(self._axesSelector)
-        layout.addWidget(self._signalSelector)
-
-        self.setLayout(layout)
 
         self.__aggregationModeAction = AggregationModeAction(parent=self)
         self.getPlot().toolBar().addAction(self.__aggregationModeAction)
@@ -438,16 +421,6 @@ class ArrayImagePlot(qt.QWidget):
                 self.getAggregationModeAction().getAggregationMode()
             )
 
-    def _signalChanges(self, value):
-        self._updateImageAxes()
-
-    def getPlot(self):
-        """Returns the plot used for the display
-
-        :rtype: Plot2D
-        """
-        return self._plot
-
     def setImageData(
         self,
         signals: list[h5py.Dataset | Dataset],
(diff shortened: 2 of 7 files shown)


src/silx/gui/data/DataViews.py
```diff
@@ -1345,17 +1343,13 @@ def setData(self, data):
...
-        widget: ArrayImagePlot = self.getWidget()
+        widget: RgbaImagePlot = self.getWidget()
         widget.setImageData(
             [nxd.signal] + nxd.auxiliary_signals,
             axes=nxd.axes,
             signals_names=[nxd.signal_name] + nxd.auxiliary_signals_names,
             axes_names=nxd.axes_names,
-            axes_scales=nxd.plot_style.axes_scale_types,
             title=nxd.title,
-            isRgba=True,
```
loichuder, line 1358: Note that `isRgba` was removed since it is no longer relevant.

Also, the plot of RGBA image does not need scales since it can only plot linear axes.

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -469,153 +441,75 @@ def setImageData(
         :param title: Graph title
         :param isRgba: True if data is a 3D RGBA image
         """
-        self._axesSelector.selectionChanged.disconnect(self._updateImage)
-        self._axesSelector.selectedAxisChanged.disconnect(self._updateImageAxes)
-        self._signalSelector.selectionChanged.disconnect(self._signalChanges)
```
loichuder, line 474: I think the aim of the "disconnect at the beginning, reconnect at the end" pattern was to block unwanted updates due to signal firings.

Therefore, I changed it to use the `blockSignals` context manager since it looked more robust and explicit.

src/silx/gui/data/_BaseImagePlot.py
```diff
@@ -0,0 +1,115 @@
...
+        except KeyError:
+            raise KeyError("No image found. Was an image loaded?")
+        return signal[self._axesSelector.selection()]
+
+    def _getImageName(self):
+        if len(self._signalSelector.getSignalNames()) > 0:
+            return self._signalSelector.getCurrentSignalName()
+        return ""
+
+    def _updateImageAxes(self):
```
loichuder, line 66: This is basically the same function as `_updateImageAxes` from `ArrayImagePlot`, except that it calls an abstract `_addItemToPlot` method that must be implemented by subclasses.

src/silx/gui/data/_BaseImagePlot.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            signal = self._signals[signal_index]
+        except KeyError:
+            raise KeyError("No image found. Was an image loaded?")
+        return signal[self._axesSelector.selection()]
+
+    def _getImageName(self):
+        if len(self._signalSelector.getSignalNames()) > 0:
+            return self._signalSelector.getCurrentSignalName()
+        return ""
```
loichuder, line 64: Instead of keeping a list of signal names in `_signal_names`, I am now relying on `_signalSelector` to hold the signal names. 

To easily retrieve the current signal name, I added a new method `getCurrentSignalName` to `SignalSelector`.

src/silx/gui/data/_utils.py
```diff
@@ -39,3 +45,43 @@ def isRgba(nxd: NXdata) -> bool:
...
+        ycalib = NoCalibration()
+
+    return xcalib, ycalib
+
+
+def setImageCoords(
+    item: ImageBase, xcalib: AbstractCalibration, ycalib: AbstractCalibration
+):
+    item.setOrigin((xcalib(0), ycalib(0)))
+    item.setScale((xcalib.get_slope(), ycalib.get_slope()))
```
loichuder, line 81: New util functions that are used in the subclasses of `BaseImagePlot`

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -469,153 +441,75 @@ def setImageData(
...
-                y_axis = y_axis[0] * numpy.arange(image.shape[0]) + y_axis[1]
-
-            try:
-                xcalib = ArrayCalibration(x_axis)
-            except ValueError:
-                xcalib = NoCalibration()
-            try:
-                ycalib = ArrayCalibration(y_axis)
-            except ValueError:
-                ycalib = NoCalibration()
```
loichuder, line 579: Abstracted in `getAxesCalib`

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -469,153 +441,75 @@ def setImageData(
...
         )
-
+        image = self._getImageToDisplay()
+        xcalib, ycalib = getAxesCalib(image, x_axis, y_axis)
         if xcalib.is_affine() and ycalib.is_affine():
-            # regular image
-            xorigin, xscale = xcalib(0), xcalib.get_slope()
-            yorigin, yscale = ycalib(0), ycalib.get_slope()
-            origin = (xorigin, yorigin)
-            scale = (xscale, yscale)
```
loichuder, line 593: Abstracted in `setImageCoords`

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -469,153 +441,75 @@ def setImageData(
...
-                yaxisscale = self.__axes_scales[y_axis_index]
+            if self._axes_scales:
+                axis_indices = self._axesSelector.getIndicesOfNamedAxes()
+                try:
+                    x_axis_index = axis_indices["X"]
+                    y_axis_index = axis_indices["Y"]
+                except KeyError:
+                    raise KeyError("Axes X and Y not found. Was an image loaded?")
+                xaxisscale = self._axes_scales[x_axis_index]
+                yaxisscale = self._axes_scales[y_axis_index]
```
loichuder: Rather than passing the `x_axis_index` and `y_axis_index` variables as parameters of `addItemToPlot`, I am retrieving them here.

src/silx/gui/plot/PlotWidget.py
```diff
@@ -2317,10 +2317,10 @@ def getActiveImage(self, just_legend: bool = False) -> items.ImageBase | str | N
         """
         return self._getActiveItem(kind="image", just_legend=just_legend)
 
-    def setActiveImage(self, legend: str) -> str | None:
+    def setActiveImage(self, legend: str | items.ImageBase) -> str | None:
```
loichuder, line 2320: Little typing fix since `legend` can actually be the item itself (as it was used in `ArrayImagePlot`)

src/silx/gui/data/_BaseImagePlot.py
```diff
@@ -0,0 +1,115 @@
...
+        layout.addWidget(self._axesSelector)
+        layout.addWidget(self._signalSelector)
+
+        self.setLayout(layout)
+
+    def getPlot(self) -> Plot2D:
+        return self._plot
+
+    def _getImageToDisplay(self):
+        signal_index = self._signalSelector.getSignalIndex()
```
payno:
```suggestion
        signalIndex = self._signalSelector.getSignalIndex()
```

src/silx/gui/data/_BaseImagePlot.py
```diff
@@ -0,0 +1,115 @@
...
+            return self._signalSelector.getCurrentSignalName()
+        return ""
+
+    def _updateImageAxes(self):
+        """Updates the image axes. Called when the user selects a different axis than the displayed one."""
+
+        axis_indices = self._axesSelector.getIndicesOfNamedAxes()
+        try:
+            x_axis_index = axis_indices["X"]
+            y_axis_index = axis_indices["Y"]
```
payno: Looks like I am not the only one having troubles to switch from camelCase and snake_case ;)
```suggestion
        axisIndices = self._axesSelector.getIndicesOfNamedAxes()
        try:
            xAxisIndex = axisIndices["X"]
            yAxisIndex = axisIndices["Y"]
```

src/silx/gui/data/_RgbaImagePlot.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        with blockSignals(self._axesSelector, self._signalSelector):
+            if axes is None:
+                axes = []
+            if signals_names is None:
+                signals_names = []
+
+            self._signals = signals
+            self._axes = axes
+            self._title = title
```
payno: I would say at least this processing can be done outside the context manager.

src/silx/gui/data/_RgbaImagePlot.py
```diff
@@ -0,0 +1,88 @@
...
+            self._axesSelector.setAxisNames(["Y", "X", "RGB(A) channel"])
+            self._axesSelector.setNamedAxesSelectorVisibility(False)
+
+            # Labels need to be set before the data
+            if axes_names:
+                self._axes_names = axes_names
+                self._axesSelector.setLabels(axes_names)
+            self._axesSelector.setData(signals[0])
+
+            if len(signals[0].shape) <= 3:
```
payno, line 56: Are you sure signals always has at least one element ?

src/silx/gui/data/_BaseImagePlot.py
```diff
@@ -0,0 +1,115 @@
...
+    Widget for plotting images with a multi-dimensional signal array
+    and two 1D axes array.
+    """
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+        self._signals: list[h5py.Dataset | Dataset] = []
+        self._axes: list[h5py.Dataset | Dataset] = []
+        self._axes_names: list[str] = []
```
payno: Should we benefit from the refactoring to have snakeCase in the Qt widget ?
I am always very bad at switching from one to the other but if this is the rule :(

```suggestion
        self._axesNames: list[str] = []
```

src/silx/gui/data/_utils.py
```diff
@@ -39,3 +45,43 @@ def isRgba(nxd: NXdata) -> bool:
...
+    if x_axis is None:
+        # no calibration
+        return ArrayCalibration(numpy.arange(axis_length))
+    if numpy.isscalar(x_axis) or len(x_axis) == 1:
+        # constant axis
+        return ArrayCalibration(x_axis * numpy.ones((axis_length,)))
+    if len(x_axis) == 2:
+        # linear calibration
+        return ArrayCalibration(x_axis[0] * numpy.arange(axis_length) + x_axis[1])
+    return NoCalibration()
```
payno: Reading this sounds a bit weird.
Line 54 you have a comment 'no calibration' and then you return an 'ArrayCalibration'

But for others if you cannot fulfilled condition you will return a `NoCalibration` object.

src/silx/gui/data/_BaseImagePlot.py
loichuder: Absolutely! Good catch

src/silx/gui/data/_utils.py
loichuder: Agreed! 

I reproduced the algorithm already present in `ArrayImagePlot`

```python
            if x_axis is None:
                # no calibration
                x_axis = numpy.arange(image.shape[1])
            elif numpy.isscalar(x_axis) or len(x_axis) == 1:
                # constant axis
                x_axis = x_axis * numpy.ones((image.shape[1],))
            elif len(x_axis) == 2:
                # linear calibration
                x_axis = x_axis[0] * numpy.arange(image.shape[1]) + x_axis[1]
```

and added the `NoCalibration` as a fallback (there was no fallback in the original algorithm which could lead to unexpected things).

I can raise an error instead if that sounds better?

src/silx/gui/data/_RgbaImagePlot.py
loichuder: Good point, I'll add a check

src/silx/gui/data/_BaseImagePlot.py
```diff
@@ -0,0 +1,116 @@
...
+        self._plot.setGraphYLabel(self._axesNames[yAxisIndex])
+        self._plot.resetZoom()
+
+    def _addItemToPlot(self, xAxis: ImageAxis, yAxis: ImageAxis):
+        raise NotImplementedError()
+
+    def clear(self):
+        old = self._axesSelector.blockSignals(True)
+        self._axesSelector.clear()
+        self._axesSelector.blockSignals(old)
```
payno: Maybe we could take advantage of the refactoring to use `with blockSignals` (from `silx.gui.utils`)

```suggestion
        with blockSignals(self._axesSelector):
            self._axesSelector.clear()
```

src/silx/gui/data/_utils.py
payno: +1 for raising an exception. This is better to break clearly ASAP when we are on an unexpected case.

src/silx/gui/data/_RgbaImagePlot.py
```diff
@@ -0,0 +1,89 @@
...
+    Widget for plotting a RGB(A) image with a multi-dimensional signal array
+    and two 1D axes array.
+    """
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+        self._plot.getXAxis().setScale("linear")
+        self._plot.getYAxis().setScale("linear")
+
```
payno: Even if this comes from the original implementation, after a discussion with @t20100 axis scale are always linear by default. So we can avoid defining them.


```suggestion
```

src/silx/gui/data/_utils.py
loichuder: Then, I should also remove the `try`/`except` from `getAxesCalib` right?

PR #4529, opened 2026-03-13
silx.gui.plot.PlotWidget: Added `asinh` scale for axes

EdgarGF93:
- [ ] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

Continue from #4306


src/silx/gui/plot/backends/BackendBase.py
```diff
@@ -493,13 +493,27 @@ def setXAxisLogarithmic(self, flag):
...
+
+        : param str scale: linear, log or asinh
+        """
+        pass
+
+    def setYAxisScale(self, scale):
+        """Set the scale of Y axis between (linear / log / asinh)
+
+        : param str scale: linear, log or asinh
+        """
```
t20100: Let's add typing for those 2 functions and reduce the docstring:
```
from typing import Literal
...
def setXAxisScale(self, scale: Literal["linear", "log", "asinh"]):
    ...
```

src/silx/gui/plot/backends/BackendBase.py
```diff
@@ -493,13 +493,27 @@ def setXAxisLogarithmic(self, flag):
...
         """
-        pass
+        self.setXAxisScale(scale="log" if flag is True else "linear")
 
     def setYAxisLogarithmic(self, flag):
         """Set the Y axis scale between linear and log.
 
         :param bool flag: If True, the left axis will use a log scale
         """
+        self.setYAxisScale(scale="log" if flag is True else "linear")
```
t20100: Are `setXAxisLogarithmic` and `setYAxisLogarithmic` still needed?
The backends are not really supposed to be called directly so best to remove them.
If there is a good reason to keep those, best to deprecate them with the `@deprecated` decorator from `silx.utils.deprecation` module

src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -1581,20 +1581,35 @@ def isXAxisTimeSeries(self):
     def setXAxisTimeSeries(self, isTimeSeries):
         self._plotFrame.xAxis.isTimeSeries = isTimeSeries
 
-    def setXAxisLogarithmic(self, flag):
-        if flag != self._plotFrame.xAxis.isLog:
-            if flag and self._keepDataAspectRatio:
+    def setXAxisScale(self, scale):
+        if scale in ("arcsinh", "asinh"):
```
t20100: No need to support more than what is advertised in the BackendBase class:
```suggestion
        if scale == "asinh":
```

src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -1581,20 +1581,35 @@ def isXAxisTimeSeries(self):
...
+        is_log = True if scale == "log" else False
+        if is_log != self._plotFrame.xAxis.isLog:
+            if is_log and self._keepDataAspectRatio:
                 _logger.warning("KeepDataAspectRatio is ignored with log axes")
 
-            self._plotFrame.xAxis.isLog = flag
+            self._plotFrame.xAxis.isLog = is_log
+
+    def setYAxisScale(self, scale):
+        if scale in ("arcsinh", "asinh"):
```
t20100:
```suggestion
        if scale == "asinh":
```

src/silx/gui/plot/items/axis.py
```diff
@@ -224,40 +227,22 @@ def getScale(self) -> AxisScaleType:
     def setScale(self, scale: AxisScaleType):
         """Set the scale to be used by this axis.
 
-        :param scale: Name of the scale ("log", or "linear")
+        :param scale: Name of the scale ("log", "linear", "arcsinh")
```
t20100: `AxisScaleType` should be updated as well:
```
AxisScaleType = typing.Literal["linear", "log", "asinh"]
```

src/silx/gui/plot/items/axis.py
```diff
@@ -67,7 +67,10 @@ class Axis(qt.QObject):
     LOGARITHMIC = "log"
     """Constant defining a logarithmic scale"""
 
-    _SCALES = {LINEAR, LOGARITHMIC}
+    ARCSINH = "arcsinh"
```
t20100: Simpler to use the same string as the backend here
```suggestion
    ARCSINH = "asinh"
```

src/silx/gui/plot/items/axis.py
```diff
@@ -424,6 +406,27 @@ def _setRangeConstraints(self, minRange=None, maxRange=None):
...
+                else:
+                    if vmax > 0 and dataRange[0] < vmax:
+                        self.setLimits(dataRange[0], vmax)
+                    else:
+                        self.setLimits(*dataRange)
+        elif scale == self.ARCSINH:
+            self._getBackend().setXAxisScale(scale="asinh")
+        elif scale == self.LINEAR:
+            self._getBackend().setXAxisScale(scale="linear")
+        else:
```
t20100, line 422:
```suggestion
        elif self._scale in (self.ARCSINH, self.LINEAR):
            self._getBackend().setXAxisScale(scale=self._scale)
```

src/silx/gui/plot/items/axis.py
```diff
@@ -464,8 +467,26 @@ def _internalGetLimits(self):
...
+                    self.setLimits(1.0, 100.0)
+                else:
+                    if vmax > 0 and dataRange[0] < vmax:
+                        self.setLimits(dataRange[0], vmax)
+                    else:
+                        self.setLimits(*dataRange)
+        elif scale == self.ARCSINH:
+            self._getBackend().setYAxisScale(scale="asinh")
+        elif scale == self.LINEAR:
+            self._getBackend().setYAxisScale(scale="linear")
```
t20100:
```suggestion
        if self._scale == self.LOGARITHMIC:
            self._getBackend().setYAxisScale(scale="log")
            vmin, vmax = self.getLimits()
            if vmin <= 0:
                dataRange = self._getDataRange()
                if dataRange is None:
                    self.setLimits(1.0, 100.0)
                else:
                    if vmax > 0 and dataRange[0] < vmax:
                        self.setLimits(dataRange[0], vmax)
                    else:
                        self.setLimits(*dataRange)
        elif self._scale == (self.ARCSINH, self._LINEAR):
            self._getBackend().setYAxisScale(scale=self._scale)
```

src/silx/gui/plot/items/axis.py
```diff
@@ -424,6 +406,27 @@ def _setRangeConstraints(self, minRange=None, maxRange=None):
         updated = constrains.update(minXRange=minRange, maxXRange=maxRange)
         return updated
 
+    def _internalSetScale(self):
+        scale = self._scale
+        vmin, vmax = self.getLimits()
+        if scale == self.LOGARITHMIC:
+            self._getBackend().setXAxisScale(scale="log")
```
t20100, line 410: I propose to:
- use `self._scale` directly, when searching through code for an attribute usage, having to go through local variable allocation makes it harder.
- move `vmin, vmax` inside the `if` since they are only used there

```suggestion
        if self._scale == self.LOGARITHMIC:
            self._getBackend().setXAxisScale(scale="log")
            vmin, vmax = self.getLimits()
```

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -50,9 +50,23 @@
...
 
 _logger = logging.getLogger(__name__)
 
+ICON_YSCALE = "mdi6.arrow-up-down"
+ICON_XSCALE = "mdi.arrow-left-right"
+ICON_LINEAR = "ph.line-segment-bold"
+ICON_LOG = "mdi.math-log"
+ICON_ASINH = "mdi6.math-sin"
+ICONS_OPTIONS = {
+    "color": "green",
```
t20100: Best to leave the default color so it is fine with dark mode:
```suggestion
```

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -50,9 +50,23 @@
...
+    "options": [
+        {"offset": (-0.25, 0.0), "scale_factor": 0.7},
+        {"offset": (0.25, 0.0), "scale_factor": 0.7},
```
t20100: Why is there 2 offset & scale_factor?

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -50,9 +50,23 @@
...
 from .tools.PlotToolButton import PlotToolButton
+import qtawesome
 
 _logger = logging.getLogger(__name__)
 
+ICON_YSCALE = "mdi6.arrow-up-down"
+ICON_XSCALE = "mdi.arrow-left-right"
+ICON_LINEAR = "ph.line-segment-bold"
+ICON_LOG = "mdi.math-log"
+ICON_ASINH = "mdi6.math-sin"
```
t20100: arcsinh is not sin.
There's some `colormap-norm-[arcsinh|linear|log]]` is silx icons maybe they could be reused here

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -224,6 +238,174 @@ def setYAxisDownward(self):
...
+        """Set Y-scale to Linear"""
+        self.plot().getYAxis().setScale(scale="linear")
+
+    def setYAxisScaleLog(self):
+        """Set Y-scale to Log"""
+        self.plot().getYAxis().setScale(scale="log")
+
+    def setYAxisScaleAsinh(self):
+        """Set Y-scale to Arcsinh"""
+        self.plot().getYAxis().setScale(scale="arcsinh")
```
t20100: Let's keep those private with a `_` prefix

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -50,9 +50,23 @@
 from .items.axis import XAxis, YAxis, YRightAxis
 from .PlotWidget import PlotWidget
 from .tools.PlotToolButton import PlotToolButton
+import qtawesome
 
 _logger = logging.getLogger(__name__)
 
+ICON_YSCALE = "mdi6.arrow-up-down"
+ICON_XSCALE = "mdi.arrow-left-right"
```
t20100: Let's keep those `ICON_*` private with a `_`.

src/silx/gui/plot/PlotToolButtons.py
```diff
@@ -224,6 +238,174 @@ def setYAxisDownward(self):
...
+    def setXAxisScaleLinear(self):
+        """Set X-scale to Linear"""
+        self.plot().getXAxis().setScale(scale="linear")
+
+    def setXAxisScaleLog(self):
+        """Set X-scale to Log"""
+        self.plot().getXAxis().setScale(scale="log")
+
+    def setXAxisScaleAsinh(self):
+        """Set X-scale to Arcsinh"""
```
t20100: Let's keep it private

src/silx/gui/plot/PlotWindow.py
```diff
@@ -106,6 +106,7 @@ def __init__(
         resetzoom=True,
         autoScale=True,
         logScale=True,
+        chooseScale=True,
```
t20100: Can you remove the change in `PlotWindow` from this PR?
I'm not sure we'll have the OpenGL implementation before next release and we should not add those buttons by default until then.
Then there's the question of the ever growing `PlotWindow`. So probably best to leave the PlotWindow as it is and add the new axes scale in the `Plot1D` and hide the log scale buttons there.

t20100 (commented):
Thanks for working back on this feature!

src/silx/gui/plot/backends/BackendBase.py
EdgarGF93: it's only called from https://github.com/silx-kit/silx/blob/0583c7bdbc5123e16a9906415b3e75c54058c2fb/src/silx/gui/plot/PlotWidget.py#L631
so I can change that

t20100: Yes, you can remove those

src/silx/gui/plot/PlotToolButtons.py
EdgarGF93: im combining two icons in one, without the offset they stack one of top of the other, scale factors two make them smaller

src/silx/gui/plot/PlotToolButtons.py
EdgarGF93: can i combine qtawesome icons with silx icons?

t20100: I would not expect so, but I don't know

EdgarGF93: what about this, all silx icons again

src/silx/gui/plot/items/axis.py
```diff
@@ -47,7 +47,7 @@ class TickMode(enum.Enum):
     TIME_SERIES = 1  # Ticks are datetime objects
 
 
-AxisScaleType = typing.Literal["linear", "log"]
+AxisScaleType = typing.Literal["linear", "log", "asinh"]
```
t20100, line 50: matplotlib uses `asinh` but we use `arcsinh` for the colormap scaling... Would be best to use only one.
I propose to be consistent with `matplotlib` and use `asinh` and add support for this name in colormap scales (in another PR)

t20100 (approved):
LGTM, I may rework this a bit while adding the OpenGL support.

PR #4532, opened 2026-03-13
silx.gui.plot.PlotWindow: move the 'Options' menu to a toolbar.

payno:

Plot options will now be presented as a 'burger' menu on the top right corner.

Screenshot from 2026-03-16 10-02-28

### DIFF
--- a/src/silx/gui/plot/tools/_PlotOptionButton.py
+++ b/src/silx/gui/plot/tools/_PlotOptionButton.py
@@ -44,7 +44,7 @@ class PlotOptionButton(PlotToolButton):
         self._menu.exec(self.mapToGlobal(self.rect().bottomLeft()))
 
     def setPlot(self, plot: PlotWindow):
-        from ..PlotWindow import PlotWindow  # noqq: F811 avoid cyclic import
+        from ..PlotWindow import PlotWindow  # noqa: F811 avoid cyclic import
 
         if not isinstance(plot, PlotWindow):
             raise TypeError(


src/silx/gui/plot/PlotWindow.py
```diff
@@ -256,14 +258,13 @@ def __init__(
...
-            self.controlButton = qt.QToolButton()
-            self.controlButton.setText("Options")
-            self.controlButton.setToolButtonStyle(qt.Qt.ToolButtonTextBesideIcon)
-            self.controlButton.setAutoRaise(True)
-            self.controlButton.setPopupMode(qt.QToolButton.InstantPopup)
-            menu = qt.QMenu(self)
-            menu.aboutToShow.connect(self._customControlButtonMenu)
-            self.controlButton.setMenu(menu)
+            self._plotOptionButton = qt.QPushButton()
+            self._plotOptionButton.setFlat(True)
```
payno: I think this is more 'main stream' to have a `QPushButton` as a `burger menu` than a `QToolButton` with a menu attach to it.

But can be discussed.

src/silx/gui/plot/PlotWindow.py
```diff
@@ -256,14 +258,13 @@ def __init__(
         self._updateColorBarBackground()
 
         if control:  # Create control button only if requested
-            self.controlButton = qt.QToolButton()
```
payno, line 259: `controlButton` is a public API. But existing only if `control` is True.

So as next release is a major version I didn't handled any deprecation.

But we could add if needed.

src/silx/gui/plot/PlotWindow.py
```diff
@@ -256,14 +257,13 @@ def __init__(
...
-            self.controlButton.setToolButtonStyle(qt.Qt.ToolButtonTextBesideIcon)
-            self.controlButton.setAutoRaise(True)
-            self.controlButton.setPopupMode(qt.QToolButton.InstantPopup)
-            menu = qt.QMenu(self)
-            menu.aboutToShow.connect(self._customControlButtonMenu)
-            self.controlButton.setMenu(menu)
+            self._plotOptionButton = qt.QPushButton()
+            self._plotOptionButton.setFlat(True)
+            # should be presented on the right of the toolbar and with the burger menu
+            self._plotOptionButton.setIcon(qtawesome.icon("ei.lines"))
```
loichuder: `'fa6.bars` might look less blocky if you wish

src/silx/gui/plot/tools/_PlotOptionButton.py
```diff
@@ -0,0 +1,59 @@
...
+if TYPE_CHECKING:
+    from ..PlotWindow import PlotWindow
+
+
+class PlotOptionButton(qt.QPushButton):
+    """Button presented as a 'burger' menu to present user contextual actions such as 'Legend' or 'Region Of Interest'."""
+
+    def __init__(self, parent: PlotWindow | None = None):
+        super().__init__(parent)
+        self._plot = None
```
payno: When created we expect the parent to be the `_plot`. But as this widget will in the current code change parent (for the toolbar) and to have a more generic implementation I defined the `getPlot` and `setPlot` functions.

src/silx/gui/plot/tools/_PlotOptionButton.py
```diff
@@ -0,0 +1,59 @@
...
+        self._menu = qt.QMenu(self)
+
+        self.clicked.connect(self._showPlotActionMenu)
+        self._menu.aboutToShow.connect(self._customControlButtonMenu)
+
+    def setPlot(self, plot: PlotWindow):
+        from ..PlotWindow import PlotWindow  # Avoid cyclic import
+        if not isinstance(plot, PlotWindow):
+            raise TypeError(f"plot should be an instanec of {PlotWindow}. Got {type(plot)}")
+        self._plot = ref(plot)
```
payno: As this is now in a dedicated class we need to explicit relation with the plot and clarify life-cycle management (weakref on the plot). 

This is more decouple from the `PlotWindow` but this complexify a bit...

src/silx/gui/plot/tools/_PlotOptionButton.py
loichuder: As far as I can see, this button is not useful if there is no plot attached to it.

Therefore, I would suggest to pass the `plot` in the constructor to avoid the need to call `get/setPlot` later and to deal with `None`.

Personally, I would use `parent` to set the `_plot` but if you really wish to decouple `parent` and `plot`, we could have a new parameter `plot` in the constructor.

src/silx/gui/plot/tools/_PlotOptionButton.py
loichuder: I still think I prefer the dedicated class. Setting `_plot` in the constructor should also simplify things.

src/silx/gui/plot/tools/_PlotOptionButton.py
payno: I agree. But I think now the policy is to reduce as much as possible class constructor signature to the benefit of dedicated setter / getter.
But I don't have a strong opinion. Any preferences @t20100 ?

t20100: It's best to decouple `parent` and attached plot, they are not always the same.

For me it is OK to pass both `parent` and `plot` even if Qt-ish tends to limit the amount of arguments.
BTW, there is a base class that implements it: [silx.gui.plot.tools.PlotToolButton](https://github.com/silx-kit/silx/blob/main/src/silx/gui/plot/tools/PlotToolButton.py)
Why not using it here?

src/silx/gui/plot/tools/_PlotOptionButton.py
```diff
@@ -0,0 +1,61 @@
...
+from weakref import ref
+from typing import TYPE_CHECKING
+from silx.gui import qt
+import qtawesome
+
+if TYPE_CHECKING:
+    from ..PlotWindow import PlotWindow
+
+
+class PlotOptionButton(qt.QPushButton):
```
t20100: Best to use a tool button for toolbar buttons (see other comment)

src/silx/gui/plot/tools/_PlotOptionButton.py
```diff
@@ -0,0 +1,61 @@
...
+    """Button presented as a 'burger' menu to present user contextual actions such as 'Legend' or 'Region Of Interest'."""
+
+    def __init__(self, parent: PlotWindow | None = None):
```
t20100:
```suggestion
    def __init__(self, parent: PlotWidget | None = None):
```

Why restricting to `PlotWindow`?

src/silx/gui/plot/PlotWindow.py
t20100: I think it is needed, PlotWindow is fairly used so if we can go through deprecation it is best. When using a library it is always annoying when things are removed without notice.
A `controlButton` property returning `self._plotOptionButton` would do the job.

src/silx/gui/plot/PlotWindow.py
t20100, line 342: Is this `bottomBar` widget and `hbox` layout still needed for containing a single widget?

src/silx/gui/plot/PlotWindow.py
```diff
@@ -304,6 +299,16 @@ def __init__(
             for action in toolbar.actions():
                 self.addAction(action)
 
+        if control:
+            self._plotOptionToolBar = qt.QToolBar()
```
t20100: No need to keep a reference:
```suggestion
            plotOptionToolBar = qt.QToolBar()
```

src/silx/gui/plot/PlotWindow.py
```diff
@@ -256,14 +257,8 @@ def __init__(
...
         if control:  # Create control button only if requested
-            self.controlButton = qt.QToolButton()
-            self.controlButton.setText("Options")
-            self.controlButton.setToolButtonStyle(qt.Qt.ToolButtonTextBesideIcon)
-            self.controlButton.setAutoRaise(True)
-            self.controlButton.setPopupMode(qt.QToolButton.InstantPopup)
-            menu = qt.QMenu(self)
-            menu.aboutToShow.connect(self._customControlButtonMenu)
-            self.controlButton.setMenu(menu)
+            self._plotOptionButton = PlotOptionButton(self)
```
t20100, line 261: To be consistent with the other actions and tool buttons, it's worth adding a `getOptionButton()`

src/silx/gui/plot/tools/_PlotOptionButton.py
t20100: Decoupling the button is inline with other tool buttons and actions: This allows to reuse things.

src/silx/gui/plot/PlotWindow.py
payno: My thought was that this variable is existing only if 'control' is True.

So if used it should be with 'hasattr' and if we deprecate through a property then 'hasattr' condition would always be True... And might not advance much things.

But thinking again I guess it would cover at least cases where users deduce the existence of this `controlButton` from the 'control' parameter of the constructor. Which should be most of the cases.
Even if it modify somehow the original API.

src/silx/gui/plot/tools/_PlotOptionButton.py
payno: Nice catch indeed :+1: . I forget about the `PlotToolButton`. Will remove some code.

src/silx/gui/plot/tools/_PlotOptionButton.py
payno: Because of `_customControlButtonMenu`. We need to link it with the different actions (`getStatsAction`...) that are defined at the `PlotWindow` level.

src/silx/gui/plot/tools/_PlotOptionButton.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self._menu.addSeparator()
+        self._menu.addAction(plot.getCrosshairAction())
+        self._menu.addAction(plot.getPanWithArrowKeysAction())
+
+    def _showPlotActionMenu(self):
+        self._menu.exec(self.mapToGlobal(self.rect().bottomLeft()))
+
+    def setPlot(self, plot: PlotWindow):
+        from ..PlotWindow import PlotWindow  # noqq: F811 avoid cyclic import
```
loichuder:
```suggestion
        from ..PlotWindow import PlotWindow  # noqa: F811 avoid cyclic import
```

PR #4534, opened 2026-03-16
silx view: Updated nxdata display

t20100:

This PR reworks the display of NXdata in silx view to:
- display multiple views instead of the one chosen by the `@interpretation`. The `@interpretation` is now used to select the "best" view.
- remove the use of `StackView`

It also deprecates enum and widget related to StackView that is no longer used.
`silx.io.nxdata.NXdata.is_*` (`is_curve`, `is_image`, ...) are no longer used (not deprecated in this PR, see #4535).

Example for h5web_mock nexus_entry/image NXdata group:

image

### DIFF
--- a/src/silx/gui/data/test/test_nexus_dataviewer.py
+++ b/src/silx/gui/data/test/test_nexus_dataviewer.py
@@ -16,11 +16,14 @@ def test_image_no_interpretation(qWidgetFactory, tmp_path):
 
         widget.setData(h5file["/"])
 
-        currentCompositeView = widget.currentAvailableViews()[0]
-        assert isinstance(currentCompositeView, DataViews._NXdataView)
-        currentView = currentCompositeView.getCurrentView()
-        assert isinstance(currentView, DataViews._NXdataImageView)
-        plot = currentView.getWidget().getPlot()
+        viewClasses = tuple(view.__class__ for view in widget.currentAvailableViews())
+        assert viewClasses == (
+            DataViews._NXdataImageView,
+            DataViews._NXdataCurveView,
+            DataViews._Hdf5View,
+        )
+        imageView = widget.currentAvailableViews()[0]
+        plot = imageView.getWidget().getPlot()
         isinstance(plot.getImage("signal"), ImageDataAggregated)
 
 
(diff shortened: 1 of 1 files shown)


src/silx/gui/data/DataViews.py
```diff
@@ -1103,12 +1100,32 @@ def _updateColormap(self, nxdata):
...
+    def getDataPriority(self, data, info) -> int:
+        if not info.hasNXdata:
+            return DataView.UNSUPPORTED
+
+        data = self.normalizeData(data)
+        nxd = nxdata.get_default(data, validate=False)
+        if nxd is None:
+            return DataView.UNSUPPORTED
+
+        return self._getNXDataPriority(nxd)
```
t20100, line 1114: This was duplicated in all `_NXdata*` classes

src/silx/gui/data/DataViews.py
```diff
@@ -1103,12 +1100,32 @@ def _updateColormap(self, nxdata):
...
     """DataView using a table view for displaying NXdata 0-D signal"""
 
     def __init__(self, parent):
-        _NXdataBaseDataView.__init__(self, parent, modeId=NXDATA_SCALAR_MODE)
+        _NXdataBaseDataView.__init__(
+            self,
+            parent,
+            modeId=NXDATA_SCALAR_MODE,
+            label="Raw",
+            icon=icons.getQIcon("view-raw"),
```
t20100, line 1129: Choose to use the same icons as the one used for datasets

src/silx/gui/data/DataViews.py
```diff
@@ -1185,20 +1202,25 @@ def setData(self, data):
...
 
-    def getDataPriority(self, data, info):
-        data = self.normalizeData(data)
-        if info.hasNXdata and not info.isInvalidNXdata:
-            if nxdata.get_default(data, validate=False).is_curve:
-                return 100
-        return DataView.UNSUPPORTED
+    def _getNXDataPriority(self, nxd: NXdata) -> int:
+        if nxd.signal.ndim >= 1:
+            return 200 if nxd.interpretation == "spectrum" else 110
```
t20100, line 1209: Default priority is 110 for curves, 120 for images so if no `@interpretation` is defined the view showing most dimensions is used.

src/silx/gui/data/DataViews.py
```diff
@@ -1667,21 +1551,19 @@ def __init__(self, parent):
...
+
+        view = SelectManyDataView(parent)
+        self.addView(view)
+        view.addView(_NXdataScalarView(parent))
+        view.addView(_NXdataCurveView(parent))
+        view.addView(_NXdataXYVScatterView(parent))
+        view.addView(_NXdataComplexImageView(parent))
+        view.addView(_NXdataImageView(parent))
+        view.addView(_NXDataRgbaImageView(parent))
+        view.addView(_NxDataScatter3D(parent))
```
t20100, line 1566: Rework view selection to allow multiple.
Keep invalid view outside of it.

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -877,6 +878,7 @@ def clear(self):
         self._plot.setData(None)
 
 
+@deprecated(since_version="3.0.0")
```
t20100, line 704: This works for class as well

src/silx/gui/data/DataViews.py
loichuder: Should we make a dictionary out of them?

src/silx/gui/data/DataViews.py
loichuder: How do you feel about making an enum of priorities? It is something I had in the back of my mind now.

src/silx/gui/data/DataViews.py
t20100: Yes, that would avoid duplicating this for now.
As discussed, let's wait to see if we can avoid having duplicated for NXdata/dataset first

src/silx/gui/data/DataViews.py
t20100: Yes! I felt something was missing when I added this comment.
Create issue #4536 for this

PR #4535, opened 2026-03-16
silx.io.ndata.parse.NXdata: Deprecated attributes related to kind of plot and `signal_is_Nd`

t20100:

Following PR #4534, `NXdata.signal_is_Xd|signal_ndim` and `NXdata.is_[scatter|curve|...]` are no longer used in silx code base, so this PR proposes to deprecate them:
- `signal_is_Xd` can be replaced with `signal.ndim == X`
- `is_[scatter|curve|...]` are opinionated interpretation of NXdata information, the aim here would be to have the `NXdata` class more focus and handling: the axes to signal mapping, the axes/signal names and not the interpretation of the NXdata for plots.

### DIFF
--- a/src/silx/io/nxdata/parse.py
+++ b/src/silx/io/nxdata/parse.py
@@ -47,6 +47,7 @@ import h5py
 import numpy
 
 from silx.io.utils import is_group, is_file, is_dataset, h5py_read_dataset
+from silx.utils.deprecation import deprecated
 
 from ._utils import (
     Interpretation,
@@ -192,11 +193,11 @@ class NXdata:
         """Signal long name, as specified in the @long_name attribute of the
         signal dataset. If not specified, the dataset name is used."""
 
-        self.signal_ndim = None
-        self.signal_is_0d = None
-        self.signal_is_1d = None
-        self.signal_is_2d = None
-        self.signal_is_3d = None
+        self._signal_ndim = None
+        self._signal_is_0d = None
+        self._signal_is_1d = None
+        self._signal_is_2d = None
+        self._signal_is_3d = None
 
         self.axes_names = None
         """List of axes names in a NXdata group.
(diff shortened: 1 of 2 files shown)


src/silx/io/nxdata/parse.py
```diff
@@ -212,26 +213,50 @@ def __init__(self, group: h5py.Group, validate: bool = True):
...
-            self.signal_is_1d = (
-                self.signal_is_1d and len(self.axes) <= 1
+            self._signal_is_1d = (
+                self._signal_is_1d and len(self.axes) <= 1
             )  # excludes n-D scatters
 
             self._plot_style = _SilxStyle(self)
 
+    @property
+    @deprecated(since_version="3.0.0")
```
loichuder, line 236:
```suggestion
    @deprecated(since_version="3.0.0", replacement="signal.ndim")
```

src/silx/io/nxdata/parse.py
```diff
@@ -788,16 +813,13 @@ def plot_style(self) -> _SilxStyle | None:
...
 
-    @property
-    def is_scatter(self) -> bool:
-        """True if the signal is 1D and all the axes have the
-        same size as the signal."""
+    def __is_scatter(self) -> bool:
         if not self.is_valid:
             raise InvalidNXdataError("Unable to parse invalid NXdata")
 
         if self._is_scatter is not None:
```
loichuder, line 820: `_is_scatter` is caching the value of the `__is_scatter` function.

Since this is now deprecated, I would not bother to keep `_is_scatter` but up to you.

t20100: `is_scatter` was used l853 and l862. I did it this way to avoid double deprecation warning and to do a minimal change.

I can remove the caching and only keep `__is_scatter()` to make it simpler.

loichuder: Yeah as I said: not essential

Eventually everything will be removed anyway

PR #4538, opened 2026-03-16
silx.gui.data.DataViews._NxDataScatter3D: Use first auxiliary signal as scatter sizes

loichuder:

Follow-up of  https://github.com/silx-kit/silx/pull/4507

### DIFF
--- a/src/silx/gui/data/DataViews.py
+++ b/src/silx/gui/data/DataViews.py
@@ -38,6 +38,7 @@ from silx.io.nxdata import get_attr_as_unicode
 from silx.io.nxdata.parse import NXdata
 from silx.gui.plot.items.image import ImageDataAggregated
 from silx.gui.plot.actions.image import AggregationModeAction
+from silx.math.combo import min_max
 from ._DataInfo import DataInfo
 from ._DataView import DataView
 from ._RgbaImagePlot import RgbaImagePlot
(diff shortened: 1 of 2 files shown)


src/silx/gui/data/DataViews.py
```diff
@@ -1631,6 +1632,15 @@ def setData(self, data: Any):
             nxd.axes[0], nxd.axes[1], nxd.axes[2], nxd.signal, copy=False
         )
 
+        if len(nxd.auxiliary_signals) >= 1:
```
loichuder, line 1635: I prefer to be a bit lenient and not check that it is **exactly** equal to 1.

src/silx/gui/data/DataViews.py
```diff
@@ -1631,6 +1632,15 @@ def setData(self, data: Any):
...
+            sig_min, sig_max = min_max(nxd.auxiliary_signals[0])
+
+            # Scale between 1 and 5
+            sizes = 1 + 4 * (nxd.auxiliary_signals[0][()] - sig_min) / (
+                sig_max - sig_min
+            )
```
loichuder, line 1641: This is up for discussion.

We can't really keep raw values since they could be anything and there is only a limited range of sizes that will look good on the 3d plot.

I therefore made the arbitrary choice to scale the sizes between 1 and 5. 

So that the user can still play a bit with this, I already prepared a PR where the global marker size is used to scale the symbols with array sizes.

t20100: Symbol sizes are in pixels (should be in points=1/72 inch) maybe [1,5] is a bit small, but the data definitely needs to be normalized before display.

> So that the user can still play a bit with this, I already prepared a PR where the global marker size is used to scale the symbols with array sizes.

Sounds good!

loichuder: > maybe [1,5] is a bit small

Yeah I leaned towards a small difference since the global marker size can be turned up but not down. Let's keep this and see how it feels with the scaling in the next PR.

PR #4539, opened 2026-03-16
silx.gui.plot3d: Global size parameters now scales symbols with array sizes

loichuder:

https://github.com/silx-kit/silx/pull/4538#discussion_r2941041391

### DIFF
--- a/src/silx/gui/plot3d/items/mixins.py
+++ b/src/silx/gui/plot3d/items/mixins.py
@@ -171,6 +171,7 @@ class SymbolMixIn(_SymbolMixIn):
     def __init__(self):
         super().__init__()
         self.__primitive = None
+        self.__sizeScaleFactor = 1.0
 
     def _setPrimitive(self, primitive: primitives.Points):
         """Set the scene primitive on which to set the symbol and size"""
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot3d/items/mixins.py
loichuder, line 153: Is it fine to implement it here in `SymbolMixIn` ? Or must I drill deeper in the hierarchy of classes?

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -208,7 +213,7 @@ def _getPickingDistances(self) -> float | numpy.ndarray:
         _, size = self._getSceneSymbol()
         return numpy.maximum(size, 3.0)
 
-    def _getSceneSymbol(self) -> tuple[str, float | ArrayLike]:
+    def _getSceneSymbol(self) -> tuple[str, float | numpy.ndarray]:
```
loichuder, line 221: Based on the implementation, this can only return a `float` or `numpy.ndarray`, which makes the multiplication l.206 safe.

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -171,6 +171,7 @@ class SymbolMixIn(_SymbolMixIn):
     def __init__(self):
         super().__init__()
         self.__primitive = None
+        self.__scaleFactor = 1
```
t20100:
```suggestion
        self.__scaleFactor = 1.0
```

src/silx/gui/plot3d/items/mixins.py
```diff
@@ -193,12 +194,16 @@ def setSymbolSize(self, size: float | ArrayLike | None, copy: bool = True):
         super().setSymbolSize(size, copy)
         self._syncPointsPrimitive()
 
+    def setSizeScaleFactor(self, scaleFactor: float):
```
t20100: I propose to name it `getSymbolSizeScaleFactor` or `getSymbolSizeScale`. It's a bit long but makes it unambiguous which size it is.

Can you add a getter for it as well?

src/silx/gui/plot3d/items/mixins.py
t20100: This is OK for me. I already hesitated to add the symbol size here as well instead of the base class in `plot`.
If we add multiple sizes support to 2d scatters, this will have to move to `plot` but that shouldn't break anything

t20100 (changes requested):
It is OK for me, but this adds a second way to set the symbol size. I'm wondering if it would not be best to have a different approach:
- Revert code to keep `get|setSymbolSize` with a unique value
- Add `get|setSymbolSizeData` to pass an array of values or None to disable which gets normalized to range [1, getSymbolSize()] for display

What do you think?

t20100:
Let's add the scale factor and make it public

t20100 (approved):
LGTM

PR #4544, opened 2026-03-17
silx.gui.data.NXdataWidgets.ArrayImagePlot: fix error when an image has axis

loichuder:

Henri and I pondered about the calibration to put if the axis was not a scalar or a 2-value array: https://github.com/silx-kit/silx/pull/4525#discussion_r2929586117

We decided to throw an error but in fact, we should return an `ArrayCalibration` since if it is not a scalar nor a 2-value array, it is a regular array :slightly_smiling_face: 

I also updated the test to add axes to prevent regression

### DIFF
--- a/src/silx/gui/data/NXdataWidgets.py
+++ b/src/silx/gui/data/NXdataWidgets.py
@@ -485,7 +485,7 @@ class ArrayImagePlot(BaseImagePlot):
             )
         )
         image = self._getImageToDisplay()
-        xCalib, yCalib = getAxesCalib(image, xAxis, yAxis)
+        xCalib, yCalib = getAxesCalib(image.shape[:2], xAxis, yAxis)
         if xCalib.is_affine() and yCalib.is_affine():
             if image.ndim != 2:
                 raise ValueError(f"image dims should be 2. Got {image.ndim}")
--- a/src/silx/gui/data/_utils.py
+++ b/src/silx/gui/data/_utils.py
@@ -62,17 +62,17 @@ def isScatter(nxd: NXdata, naxes: int) -> bool:
     return len(nxd.axes) == naxes
 
 
-def _getAxisCalib(x_axis: ImageAxis, axis_length: int) -> AbstractCalibration:
-    if x_axis is None:
-        # no calibration
+def _getAxisCalib(axis: ImageAxis, axis_length: int) -> AbstractCalibration:
+    if axis is None:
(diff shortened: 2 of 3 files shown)


src/silx/gui/data/test/test_nexus_dataviewer.py
```diff
@@ -3,28 +3,36 @@
...
         assert viewClasses == (
             DataViews._NXdataImageView,
             DataViews._NXdataCurveView,
             DataViews._Hdf5View,
         )
         imageView = widget.currentAvailableViews()[0]
-        plot = imageView.getWidget().getPlot()
-        isinstance(plot.getImage("signal"), ImageDataAggregated)
+        plot: Plot2D = imageView.getWidget().getPlot()
+        assert isinstance(plot.getImage("signal"), ImageDataAggregated)
```
loichuder, line 32: Note that this line was not testing anything since the `assert` was missing...

I had to introduce a call to `processEvents` so the assert went through

src/silx/gui/data/_utils.py
```diff
@@ -62,17 +62,17 @@ def isScatter(nxd: NXdata, naxes: int) -> bool:
...
-        return ArrayCalibration(x_axis * numpy.ones((axis_length,)))
-    if len(x_axis) == 2:
+        return ArrayCalibration(axis * numpy.ones((axis_length,)))
+    if len(axis) == 2:
         # linear calibration
-        return LinearCalibration(slope=x_axis[0], y_intercept=x_axis[1])
-    raise ValueError("Expected a scalar or two values. Got {x_axis}.")
+        return LinearCalibration(slope=axis[0], y_intercept=axis[1])
+    # array
+    return ArrayCalibration(axis)
```
loichuder, line 75: The fix. I also rename the parameter `axis` since it can be any axis, not only a x-axis

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -485,7 +485,7 @@ def _addItemToPlot(self, xAxis, yAxis):
             )
         )
         image = self._getImageToDisplay()
-        xCalib, yCalib = getAxesCalib(image, xAxis, yAxis)
+        xCalib, yCalib = getAxesCalib(image.shape[:2], xAxis, yAxis)
```
loichuder, line 488: Another bug that slip by because the test was not testing anything 😅 

Everything should be good now 🤞

PR #4578, opened 2026-04-02
Tools: updated tool to prepare release notes

t20100:

### PR summary

Update github release note conversion to sort and group PRs

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/tools/format_GH_release_notes.py
+++ b/tools/format_GH_release_notes.py
@@ -16,39 +16,35 @@ options = parser.parse_args()
 
 content = Path(options.file[0]).read_text().split("\n")
 
-pr_descriptions = sorted(
+pr_descriptions = tuple(
     GH_PR_REGEXP.sub(r"\1 (PR #\2)", line)
     for line in content
     if GH_PR_REGEXP.match(line)
 )
 
-current_topic = None
-current_descriptions = []
-
-for index in range(len(pr_descriptions)):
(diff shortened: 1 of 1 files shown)


tools/format_GH_release_notes.py
```diff
@@ -1,18 +1,54 @@
...
+    topic_and_description = tuple(
+        text.strip() for text in pr_descriptions[index].split(":", 1)
+    )
+    if len(topic_and_description) == 2:
+        new_topic, new_description = topic_and_description
+    else:
+        new_topic = pr_descriptions[index]
+        new_description = ""
+
+    if new_topic == current_topic:
```
loichuder: It may be less dependent in ordering if the loop populate first a dict `{topic: descriptions}` before flushing with prints

But I guess it also works like that since descriptions are sorted first 😉

loichuder (approved):
So the changelog is no longer updated and we have to copy paste the print results?

tools/format_GH_release_notes.py
t20100: Yes, it's best to first generate a `dict` !
I've updated this PR to do it

t20100 (commented):
> So the changelog is no longer updated and we have to copy paste the print results?

Yes, that was the idea. It was simpler while making the script 😅

Issue #4590, opened 2026-04-29
`FitManager` with uncertainties should not be "pre-declared"


kif:
While trying to follow this tutorial:
https://www.silx.org/doc/silx/latest/Tutorials/fit.html#fitting-gaussians

I stumbled upon a bug, or at least a very **unexpected** behaviour:

```
from silx.math.fit.functions import sum_gauss
from silx.math.fit import fittheories
from silx.math.fit.fitmanager import FitManager
result = results["CdTe:Si"]
print(len(result.radial), len(result.intensity), len(result.sigma))
>>> 4000 4000 4000
fit = FitManager()
# fit.fitconfig["WeightFlag"] = True
fit.setdata(result.radial, result.intensity, sigmay=result.sigma, xmin=8)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[56], line 1
----> 1 fit.setdata(result.radial, result.intensity, sigmay=result.sigma, xmin=8)

File ~/.venv/py313/lib/python3.13/site-packages/silx/math/fit/fitmanager.py:792, in FitManager.setdata(self, x, y, sigmay, xmin, xmax)
    790         self.xdata = self.xdata[bool_array]
    791         self.ydata = self.ydata[bool_array]
--> 792         self.sigmay = self.sigmay[bool_array] if sigmay is not None else None
    794 self._finite_mask = numpy.logical_and(
    795     numpy.all(
    796         numpy.isfinite(self.xdata), axis=tuple(range(1, self.xdata.ndim))
    797     ),
    798     numpy.isfinite(self.ydata),
    799 )

TypeError: 'NoneType' object is not subscriptable
```

To me, `setdata` with some uncertainties should automatically set the flag "WeightFlag"

kif:
Another bug a couple of lines later:
```
fit.runfit()

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[63], line 1
----> 1 fit.runfit()

File ~/.venv/py313/lib/python3.13/site-packages/silx/math/fit/fitmanager.py:879, in FitManager.runfit(self, callback)
    876 xwork = self.xdata[self._finite_mask]
    878 try:
--> 879     params, covariance_matrix, infodict = leastsq(
    880         self.fitfunction,  # bg + actual model function
    881         xwork,
    882         ywork,
    883         param_val,
    884         sigma=self.sigmay,
    885         constraints=param_constraints,
[...]

PR #4591, opened 2026-04-29
FitManager: make uncertainty trigger the use of weighted fit

kif:
related to #4590

### PR summary

More intuitive handling of uncertainties

#### AI Disclosure
- [X] No AI used


t20100:
From what I understand, your fix doe not correspond to the`FitManager` docstring:
```
    :param weight_flag: If this parameter is ``True`` and ``sigmay``
        uncertainties are not specified, the square root of ``y`` is used
        as weights in the least-squares problem. If ``False``, the
        uncertainties are set to 1.
    :type weight_flag: boolean
```
https://github.com/kif/silx/blob/afdef395d150e0052ec6c484793ea472ef9b8f66/src/silx/math/fit/fitmanager.py#L78-L82

-> If weight flag is False, uncertainties should be ignored...

kif:
`False` is the default value, which is a mistake to me: it should be `None` and switch to `False` if no `sigmay` is provided. 
When a user provides uncertainties, it should not crash, especially that I am following the example of the documentation ...

src/silx/math/fit/fitmanager.py
```diff
@@ -777,10 +779,8 @@ def setdata(self, x, y, sigmay=None, xmin=None, xmax=None):
                     numpy.sqrt(self.ydata) if self.fitconfig["WeightFlag"] else None
                 )
             else:
-                self.sigmay0 = numpy.array(sigmay)
-                self.sigmay = (
-                    numpy.array(sigmay) if self.fitconfig["WeightFlag"] else None
-                )
+                self.fitconfig["WeightFlag"] = True
```
t20100, line 784: Then maybe good to add to the `setdata` docstring that setting `sigmay` enables `WeightFlag`. I agree it is expected but since there looks to be a few ways to set this best to make it explicit.

Did you check how this works from the fit widgets?

kif: No, I did not check it ... BTW, I doubt this code is used anywhere since the "split Pseudo-Voigt 2" mode could not work at all, since array are not initialized at the proper size.

The code-style is very PyMca-like, fit results would deserve to be dataclass or namedtuple instances, but this is not the use-case for this PR.

kif:
I just checked, the GUI still works, and the bug about "Split pseudo-Voigt 2" is gone with this PR.

src/silx/_version.py
```diff
@@ -71,18 +71,22 @@
...
+class VersionInfo(NamedTuple):
+    major: int
+    minor: int
+    micro: int
+    releaselevel: str
+    serial: int
 
-version_info = _version_info(MAJOR, MINOR, MICRO, RELEV, SERIAL)
+
+version_info = VersionInfo(MAJOR, MINOR, MICRO, RELEV, SERIAL)
```
t20100: Let's keep it private:

```suggestion
class _VersionInfo(NamedTuple):
    major: int
    minor: int
    micro: int
    releaselevel: str
    serial: int


version_info = _VersionInfo(MAJOR, MINOR, MICRO, RELEV, SERIAL)
```

src/silx/math/fit/fitmanager.py
```diff
@@ -747,7 +749,8 @@ def setdata(self, x, y, sigmay=None, xmin=None, xmax=None):
         :type y: Sequence or numpy array or None
         :param sigmay: The uncertainties in the ``ydata`` array. These are
             used as weights in the least-squares problem.
-            If ``None``, the uncertainties are assumed to be 1.
+            If provided, enables weighted fit, unless (i.e. ``None``),
+            the uncertainties are assumed to be 1.
```
t20100: The `None` case is not correct, suggestion:
```suggestion
            If provided, enables weighted fit.
            If not provided (i.e. ``None``): use the square root of ``y`` if ``weight_flag`` is ``True``
            else uncertainties are assumed to be 1.
```

t20100 (approved):
Thanks for checking the GUI and for looking into this!

Two minor comments, otherwise LGTM

> BTW, I doubt this code is used anywhere

Provided the issues you fixed, probably...

src/silx/math/fit/fitmanager.py
vasole: > The code-style is very PyMca-like, fit results would deserve to be dataclass or namedtuple instances, but this is not the use-case for this PR.

I would prefer the expression "pre python2.6" (released in October 2008). PyMca could not use something that was not existing when it was written.

Issue #4485, opened 2026-02-22
Hdf5TreeView (silx view) unusably slow on Win11 Qt6


tifuchs:
Hello silx team,

There is a bug related to upgrading to Qt6 on Windows 11. Linux is not affected. I haven't tested MacOS.

1. Hdf5TreeView (and the dependent silx view) encounter UI locks if Qt6 is used. Just hovering the mouse cursor above an opened hdf5 file locks the UI for minutes.

2. This is just aesthetics and can be fixed by the changing the visual style to windowsvista or fusion: 

<img width="1775" height="865" alt="Image" src="https://github.com/user-attachments/assets/5c08508e-a2dc-48e8-b1df-0fd435556011" />

bug confirmed on:
OS: Win11, silx 2.2.2., PySide6, Qt 6.9.2

current main branch (a87091be4463f5da4ab413a77d2d016c82120558):
Win11/python 3.13.12, Qt 6.10.2 (PyQt6), Qt installed through anaconda
Win11/python 3.13.12, Qt 6.10.2 (PyQt6), Qt installed through pypi

Linux seems not to be affected.

This is related to #4450.

tifuchs:
This issue occurs on Win 11, but not on GNU/Linux.
Possibly a Windows specific issue.

Linux confirmed bug-free:
miniconda3/python 3.13.11
qt, pyqt = 6.9.2

Affected:
win11/anaconda/python 3.12.3

Maybe it is just my install or a specific package version.
I will try to reinstall and upgrade to python 3.13.

t20100:
We are making the release without fixing the slowness issue on Windows. We'll investigate this for the next one.

Is your file a Nexus-compliant file? with some NXdata? In this case, there is more access to the file to extract nexus information, that may be one cause of the slowness.

tifuchs:
Hdf5TreeView slowness on Win 11 is caused by 

```python
qt.QApplication.style().standardIcon(icon)
```
lookups, which are called for each entry in the Hdf5TreeView (i.e. hundreds of times once the file becomes large!) and are not cached. 
The slowness of the calls is probably caused by the new scaling introduced by Qt6 on Win11.
Depending on `QT_SCALE_FACTOR`, or `screen.devicePixelRatio()`, the function `qt.QApplication.style().standardIcon()` returns differently sized icons.

```python
from silx.gui import qt
app = qt.QApplication([])
%timeit qt.QApplication.style().standardIcon(qt.QStyle.SP_FileIcon)
[...]

Issue #1771, opened 2018-04-09
[gui] Exception when leaving silx view (with the close button, not the menu) Py35, debian9


kif:
```Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
  File "/usr/lib/python3.5/weakref.py", line 117, in remove
TypeError: 'NoneType' object is not callable
Exception ignored in: <function WeakValueDictionary.__init__.<locals>.remove at 0x7f9d2b4db488>
Traceback (most recent call last):
[...]

PR #4592, opened 2026-04-30
Win11 Hdf5TreeView: gui.hdf5.Hdf5Item: cache expensive qt.QApplication.style().standardIcon() lookups

tifuchs:
Addresses #4485

### PR summary

Fixes Hdf5TreeView slowness on Win 11 caused by slow `qt.QApplication.style().standardIcon(icon)` calls in `gui.hdf5.Hdf5Item`.

This PR just adds simple caching of the icons in a `dict`. 

The scaling factor of the primary screen is added as key to the icon cache. This should account for multi-monitor setups with different DPI scaling. - Works on my two screens.

There is a related issue addressing silx icon caching, which causes Exceptions when leaving silx
https://github.com/silx-kit/silx/issues/1771

But I haven't encountered this issue with the simple dict solution.

pytests fail on my machine with a probably unrelated issue:

```python

tmp_path = WindowsPath('C:/Users/timof/AppData/Local/Temp/pytest-of-timof/pytest-0/testErrorInInsertFilenameAsync0')
caplog = <_pytest.logging.LogCaptureFixture object at 0x00000246CAEE6900>

    def testErrorInInsertFilenameAsync(tmp_path, caplog, qapp_utils):
        filename = str(tmp_path / "corrupt.h5")
        with open(filename, "wb") as h5file:
            h5file.write("8736553".encode())

        model = hdf5.Hdf5TreeModel()
        assert model.rowCount(qt.QModelIndex()) == 0
        model.insertFileAsync(filename)

        assert qt.silxGlobalThreadPool().waitForDone(2000)  # Needed for PySide6
>       waitForPendingOperations(qapp_utils, model)

test\test_hdf5.py:85:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

model = <silx.gui.hdf5.Hdf5TreeModel.Hdf5TreeModel(0x246c59908c0) at 0x00000246CB5F9E00>

    def waitForPendingOperations(qapp_utils: TestCaseQt, model):
        for _ in range(20):
            if not model.hasPendingOperations():
                break
            qapp_utils.qWait(200)
        else:
>           raise RuntimeError("Still waiting for a pending operation")
E           RuntimeError: Still waiting for a pending operation
```

#### AI Disclosure
AI tool Codex used for automated bug finding, benchmarks and code generation.

### DIFF
--- a/src/silx/gui/hdf5/Hdf5Item.py
+++ b/src/silx/gui/hdf5/Hdf5Item.py
@@ -45,19 +45,12 @@ _hdf5Formatter = Hdf5Formatter(textFormatter=_formatter)
 _standardIconCache = {}
 
 
-def _getStandardIcon(icon):
-    style = qt.QApplication.style()
-    devicePixelRatio = None
-    app = qt.QApplication.instance()
-    if app is not None:
-        screen = app.primaryScreen()
-        if screen is not None:
-            devicePixelRatio = screen.devicePixelRatio()
-    key = style.objectName(), devicePixelRatio, icon
-    qicon = _standardIconCache.get(key)
+def _getStandardIcon(icon_type):
+    qicon = _standardIconCache.get(icon_type)
     if qicon is None:
-        qicon = style.standardIcon(icon)
-        _standardIconCache[key] = qicon
+        style = qt.QApplication.style()
+        qicon = style.standardIcon(icon_type)
+        _standardIconCache[icon_type] = qicon
     return qicon
 
 
--- a/src/silx/gui/hdf5/Hdf5Item.py
+++ b/src/silx/gui/hdf5/Hdf5Item.py
@@ -42,6 +42,17 @@ _logger = logging.getLogger(__name__)
 _formatter = TextFormatter()
 _hdf5Formatter = Hdf5Formatter(textFormatter=_formatter)
 # FIXME: The formatter should be an attribute of the Hdf5Model
+_standardIconCache = {}
+
+
+def _getStandardIcon(icon):
+    style = qt.QApplication.style()
+    key = style.objectName(), icon
+    qicon = _standardIconCache.get(key)
+    if qicon is None:
+        qicon = style.standardIcon(icon)
+        _standardIconCache[key] = qicon
+    return qicon
 
 
 class DescriptionType(enum.Enum):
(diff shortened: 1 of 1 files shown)


src/silx/gui/hdf5/Hdf5Item.py
```diff
@@ -42,6 +42,23 @@
...
+
+def _getStandardIcon(icon):
+    style = qt.QApplication.style()
+    devicePixelRatio = None
+    app = qt.QApplication.instance()
+    if app is not None:
+        screen = app.primaryScreen()
+        if screen is not None:
+            devicePixelRatio = screen.devicePixelRatio()
+    key = style.objectName(), devicePixelRatio, icon
```
t20100: `devicePixelRatio` shouldn't be needed in the cache key: the `QIcon` contains multiple-size pixmaps so there is no need to change it with `devicePixelRatio`.

```suggestion
    key = style.objectName(), icon
```

BTW, if `devicePixelRatio` is needed, best to use `QWindow.devicePixelRatio` (see https://doc.qt.io/qt-6/qscreen.html#devicePixelRatio-prop)

t20100 (commented):
Thanks for looking into this and for proposing a fix!
This is unfortunate qt does not cache standard icons.

Just one comment: the cache key could be simplified.

src/silx/gui/hdf5/Hdf5Item.py
tifuchs: You are correct, I have confirmed that QIcon returns differently sized icons. However, depending on `QT_SCALE_FACTOR`, the size is scaled for some icons, (used QIcon.availableSizes() to check it):

`QT_SCALE_FACTOR = 1` : SP_DirIcon: QSize(100, 100) ... QSize(1600, 1600)

`QT_SCALE_FACTOR = 2` : SP_DirIcon : QSize(200, 200) ... QSize(3200, 3200)

But the behavior is not really consistent. For example SP_DialogOkButton
does not get scaled (16, 16) -> (128, 128) independent of `QT_SCALE_FACTOR`.

Thus, devicePixelRatio might still be required. Though, QIcon has a built-in scaling feature. So it might not make a big difference.
I think we could remove it. Let me know.

`style.objectName()` is something like `windowsvista`, `win11`, ...
Do we expect this to change during runtime? - Could also be removed.

tifuchs:
I reverted it back to a minimal working version. 

This version does not account for style changes. And also not for full scaling (High DPI) changes. 

I think this would still be fine, if we accept that:

1. We don't account for the user to change the visual style at runtime. I think that this is a good assumption.
2. Even though the size of stored QIcon changes depending on `QT_SCALE_FACTOR`, QIcon *does* store multiple versions and will display an icon in any case (presumably with lower quality). This issue only exists on multi-screen setups.  

https://doc.qt.io/qt-6/qicon.html#pixmap
`
The pixmap might be smaller than requested, but never larger, unless the device-pixel ratio of the returned pixmap is larger than 1.
`
So the returned pixmaps might be smaller if the QIcon was created on a low DPI screen and then the screen is changed to a high DPI screen.

t20100 (approved):
LGTM, thanks for the fix!

From what I tested on Linux with PySide6, icons behaves correctly when moving the window on different screens with different resolutions and changing screen display ratio.

Issue #4486, opened 2026-02-23
Icons adaptation for dark mode


sergey-yaroslavtsev:
Several icons are barely visible if used in dark mode:

<img width="510" height="70" alt="Image" src="https://github.com/user-attachments/assets/43e5f828-b800-4633-99c7-e012a96201c0" />

Since `visa` by default use the dark mode and a lot of people use dark mode as well would be nice to adapt the icons.
I mean it is possible to parse icons afterwards and to reverse not saturated colors. But if it could be handled at creation level that would be better.

t20100:
On going work on this issue:

- #4593
- https://github.com/spyder-ide/qtawesome/issues/306

PR #4593, opened 2026-05-06
silx.gui.icons: Improved silx icons to support dark color theme

t20100:

### PR summary

This PR adds a [QIconEngine](https://doc.qt.io/qt-6/qiconengine.html) to hook into QIcon rendering and "patch" the svg when the application is in dark mode.
This adds a bit of complexity to the icon module but it is fully transparent from a user point of view: Icons automatically adapts to color scheme changes.

There is still some thing to be done:

- [x] how to handle icons that explicitly want to keep a black color
- [x] remove white background in some icons
~~- [ ] use slightly brighter green&blue, but it already gives this:~~

lightimage

related to #4486

#### AI Disclosure
- [x] AI tool(s) copilot used to bootstrap SvgIconEngine + tabby/qwen & copilot for autocompletion


src/silx/gui/icons.py
```diff
@@ -333,8 +414,12 @@ def getQIcon(name):
...
     cached_icons = getIconCache()
     if name not in cached_icons:
-        qfile = getQFile(name)
-        icon = qt.QIcon(qfile.fileName())
+        try:
+            icon = qt.QIcon(_SvgIconEngine(name))
+        except ValueError:
+            _logger.warning("Failed to load SVG icon")
+            qfile = getQFile(name)
+            icon = qt.QIcon(qfile.fileName())
```
t20100, line 432: Kept previous implementation for other application using silx icon mechanism but in silx we can consider completely remove .png icons.

src/silx/gui/icons.py
```diff
@@ -70,6 +71,86 @@ def cleanIconCache():
...
+        palette = qt.QApplication.palette()
+        foregroundQColor = palette.color(qt.QPalette.Active, qt.QPalette.Text)
+        foreground = foregroundQColor.name()
+
+        svgpath = self._lightSvgPath.replace(
+            b"<svg", f'<svg fill="{foreground}"'.encode()
+        )
+        svgpath = re.sub(
+            b'fill="(#000000|#000)"', f'fill="{foreground}"'.encode(), svgpath
+        )
```
t20100: svg default `fill` is black and this can be overridden in with `fill` declared in the `<svg>` tag.
To complete this PR, the idea is to patch existing icons to remove unnecessary `fill="#000"` so this substitution is not needed. This would allow to explicitly set a fill color to black even in dark mode.

src/silx/gui/icons.py
```diff
@@ -70,6 +71,86 @@ def cleanIconCache():
...
+        svgpath = re.sub(
+            b'stroke="(#000000|#000)"', f'stroke="{foreground}"'.encode(), svgpath
+        )
```
t20100: svg default `stroke` is `"none"` so overriding the default in `<svg>` does not work as for `fill`....

src/silx/gui/icons.py
```diff
@@ -70,6 +71,86 @@ def cleanIconCache():
...
+            isDark = (
+                qt.QApplication.styleHints().colorScheme() == qt.Qt.ColorScheme.Dark
+            )
+        else:
+            isDark = False
+
+        if isDark != self._isDarkColorScheme:
+            self._renderer.load(self._darkSvgPath if isDark else self._lightSvgPath)
+            self._isDarkColorScheme = isDark
+
```
t20100: Reload icon only when needed

t20100:
Ready for review.

I left green&blue colors unchanged, except for the dark blue "triangle" icons that were too dark.



Light/Dark scheme:



Before/After dark scheme support:

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+            raise ValueError(f"Cannot open SVG icon resource: {name}")
+
+        svgpath = bytes(qfile.readAll())
+        qfile.close()
+        return svgpath
+
+    @property
+    def _darkSvgPath(self) -> bytes:
+        svgpath = self._lightSvgPath.replace(b"<svg", b'<svg fill="#fff"')
+        svgpath = re.sub(b'stroke="(#000000|#000)"', b'stroke="#fff"', svgpath)
```
t20100: hard coded `#fff` color: From what I tested, getting the foreground color from the application's palette was not working correctly when changing the color scheme at the system level (still get the previous foreground color)

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+        pixmap = qt.QPixmap(size)
+        pixmap.fill(qt.Qt.transparent)
+
+        painter = qt.QPainter(pixmap)
+        self.paint(painter, qt.QRect(0, 0, size.width(), size.height()), mode, state)
+        painter.end()
+        return pixmap
+
+    def clone(self):
+        return _SvgIconEngine(self._lightSvgPath)
```
woutdenolf: Constructor accepts name, not path.

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+
+        if not qfile.open(qt.QIODevice.ReadOnly | qt.QIODevice.Text):
+            raise ValueError(f"Cannot open SVG icon resource: {name}")
+
+        svgpath = bytes(qfile.readAll())
+        qfile.close()
+        return svgpath
+
+    @property
+    def _darkSvgPath(self) -> bytes:
```
woutdenolf: Can we cache it? Now we recompute on every access.

src/silx/gui/icons.py
woutdenolf: I don't know enough about SVG but there might be others?

stroke="#000000ff"
stroke="black"
stroke="rgb(0,0,0)"
stroke='#000'
style="stroke:#000"

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+        svgpath = self._lightSvgPath.replace(b"<svg", b'<svg fill="#fff"')
```
woutdenolf: I don't know enough about SVG but this assumes:

- there's exactly one <svg
- no existing fill attribute
- that changing the root fill is sufficient

Is all that true?

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+        )
+        qfile = qt.QFile(filename)
+        if not qfile.exists():
+            raise ValueError(f"SVG icon resource not found: {name}")
+
+        if not qfile.open(qt.QIODevice.ReadOnly | qt.QIODevice.Text):
+            raise ValueError(f"Cannot open SVG icon resource: {name}")
+
+        svgpath = bytes(qfile.readAll())
+        qfile.close()
```
woutdenolf: A try...finally... would be appropriate here.

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+        painter: qt.QPainter,
+        rect: qt.QRect,
+        mode: qt.QIcon.Mode,
+        state: qt.QIcon.State,
+    ):
+        renderer = self._currentRenderer()
+        if renderer.isValid():
+            renderer.render(painter, rect)
+
+    def pixmap(self, size: qt.QSize, mode: qt.QIcon.Mode, state: qt.QIcon.State):
```
woutdenolf:
```suggestion
    def pixmap(self, size: qt.QSize, mode: qt.QIcon.Mode, state: qt.QIcon.State) -> qt.QPixmap:
```

src/silx/gui/icons.py
```diff
@@ -70,6 +71,75 @@ def cleanIconCache():
...
+        pixmap = qt.QPixmap(size)
+        pixmap.fill(qt.Qt.transparent)
+
+        painter = qt.QPainter(pixmap)
+        self.paint(painter, qt.QRect(0, 0, size.width(), size.height()), mode, state)
+        painter.end()
+        return pixmap
+
+    def clone(self):
```
woutdenolf:
```suggestion
    def clone(self) -> "_SvgIconEngine":
```

Note that I often use

```python
if sys.version_info < (3, 11):
    from typing_extensions import Self
else:
    from typing import Self
```

src/silx/gui/icons.py
woutdenolf: There are other places where we use `getQFile`.

Perhaps make the original `getQFile` private and use this block in the new `getQFile`?

woutdenolf (changes requested):

loichuder:
@payno found a way to handle dark mode for SVG automatically by editing the `svg` to add CSS under the media query `@media (prefers-color-scheme: dark)`.

It seems the media query does not work in Qt. But instead of replacing `stroke` and `fill` (which may be brittle as @woutdenolf pointed out), we could use the same mechanism to add an attribute in the `svg` (simulating the media query) and handle the dark mode in the CSS of each icon specifically.

t20100:
> we could use the same mechanism to add an attribute in the svg (simulating the media query) and handle the dark mode in the CSS of each icon specifically

:+1: this sounds much better than the current patching

loichuder:
As an illustration of my previous comment, we could dynamically add a `dark` class to the enclosing `svg` to change  the stroke color of `add.svg` by using CSS:

```svg
<?xml version="1.0" encoding="UTF-8"?>
<svg class="dark" id="svg2" enable-background="new 0 0 32 32" version="1.1" viewBox="0 0 32 32" xml:space="preserve" xmlns="http://www.w3.org/2000/svg" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" dark='true'>
<metadata id="metadata24">
<rdf:RDF>
    <cc:Work rdf:about="">
    <dc:format>image/svg+xml</dc:format>
    <dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/><dc:title/>
    </cc:Work>
</rdf:RDF>
</metadata>
<style>
.dark * {
    stroke: #ffffff;
}
</style>
<rect id="rect3059" x="4.2759" y="4.2462" width="23.381" fill="none" height="23.236" stroke="#000" stroke-miterlimit="10" stroke-width="1.912"/>
<path id="path3877" d="m7.9205 16.397h15.555" fill="none" stroke="#000" stroke-miterlimit="10" stroke-width="3.9434"/>
<path id="path3877-5" d="m15.697 8.6682v15.459" fill="none" stroke="#000" stroke-miterlimit="10" stroke-width="3.9434"/>
</svg>
```

In light mode, the `dark` class would not be added and the stroke color would stay `#000`.

This would also to handle dark mode robustly across icons but giving at the same time, the flexibility for each icon to handle the dark mode differently depending on the styles.

Media queries and [data-attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes) are out of the picture since they are not supported by Qt handling of CSS so it has to be a CSS class.

src/silx/gui/icons.py
t20100: Not really: `readAll` returns an empty bytearray in case of errors rather than raising.

But this was complex for reading a file, replaced qtish code with a `pathlib.Path` single liner and let it raise if it fails.

src/silx/gui/icons.py
t20100: > - there's exactly one <svg

Yes, or at least we can assume it

> - no existing fill attribute
> - that changing the root fill is sufficient

This was a convention, changing to use `class="dark"` and style: more robust and more versatile

src/silx/gui/icons.py
t20100: I keep the "_SvgIconEngine" for this PR, but we can consider adding `typing_extensions` as a dependency.

src/silx/gui/icons.py
t20100: `getQFile` returns a `QFile`, this block instantiates a `QIcon` and so can't replace `getQFile`

t20100 (commented):
Addressed review comments and changed patching to add `class="dark"`.

I started to convert a few icons (`3d*.svg` and `add.svg`), but happy to get feedback before going through all of them.

I propose to do:

- global `fill`/`stroke` patching when appropriate (e.g. `add.svg`):
  ```
  <style>
  .dark * {
      stroke: #ffffff;
  }
  </style>
  ```

  - use `currentColor` when only few places needs to be custom. It defaults to `black` and can be defined from CSS through `color`:
  ```
  <style>
  .dark * {
      color: #ffffff;
  }
  </style>
  ```
  and then use `fill="currentColor"` or `stroke="currentColor"`. For me it makes it explicit what will change. The down side is that with inkscape there is currently no way to set `currentColor`.

attn @loichuder @woutdenolf feedbacks welcome on the overall approach before doing for all icons

src/silx/gui/icons.py
```diff
@@ -70,6 +72,68 @@ def cleanIconCache():
...
+        self._renderer = qt.QSvgRenderer(self._lightSVG)
+
+    @staticmethod
+    def _readSVG(name: str) -> bytes:
+        filename = silx.resources._resource_filename(
+            f"{name}.svg", default_directory="gui/icons"
+        )
+        return Path(filename).read_bytes()
+
+    @cached_property
```
loichuder: Is it really that expensive to compute `_darkSVG` ? If not, I would simply add it a property in the constructor.

src/silx/gui/icons.py
```diff
@@ -70,6 +72,68 @@ def cleanIconCache():
...
+
+    def _currentRenderer(self) -> qt.QSvgRenderer:
+        if qt.BINDING != "PyQt5":
+            isDark = (
+                qt.QApplication.styleHints().colorScheme() == qt.Qt.ColorScheme.Dark
+            )
+        else:
+            isDark = False
+
+        if isDark != self._isDarkColorScheme:
```
loichuder:
```suggestion
        if isDark != self._hasDarkColorScheme:
```

🤷

src/silx/gui/icons.py
```diff
@@ -70,6 +72,68 @@ def cleanIconCache():
...
+            self._isDarkColorScheme = isDark
+
+        return self._renderer
+
+    def paint(
+        self,
+        painter: qt.QPainter,
+        rect: qt.QRect,
+        mode: qt.QIcon.Mode,
+        state: qt.QIcon.State,
```
loichuder, line 95: `mode` and `state` are ignored. Is that ok?

loichuder (commented):
> feedbacks welcome on the overall approach before doing for all icons

Yeah it is fine.

I mean, the advantage of the `.dark` class is that each icon can implement its own way of dealing with the dark mode but best be consistent.

src/silx/gui/icons.py
woutdenolf: I will stop interfering @loichuder and leave the rest up to you. We are saying opposite things. This was added on my request.

loichuder: Nah, it is not opposite. If this is a `property`, it is indeed computed at each access which is a problem. In this case, `cached_property` makes sense.

My idea was to solve this differently by not making it a `property` but a proper member computed it once in the constructor.

loichuder: Sorry if that wasn't clear, I didn't mean to cross you

t20100: Both are fine for me.
Let's store it in an attribute, it's a single liner and a bit more memory used

src/silx/gui/icons.py
t20100: oups!

That's a critical issue: `state` is apparently not taken into account, but `mode` is definitely needed to support greyed icons for disabled widgets!
I'm preparing the change, it's going to add more complexity

t20100: This is fixed at the expense of rewriting in Python some private parts of Ot

src/silx/gui/icons.py
```diff
@@ -70,6 +71,96 @@ def cleanIconCache():
...
+        return Path(filename).read_bytes()
+
+    def paint(
+        self,
+        painter: qt.QPainter,
+        rect: qt.QRect,
+        mode: qt.QIcon.Mode,
+        state: qt.QIcon.State,
+    ):
+        # Adapated from QSvgIconEngine::paint
```
t20100, line 97: `QSvgIconEngine` is the private engine used by default by Qt for SVG icons.
This is a rewrite of its `paint` method.

src/silx/gui/icons.py
```diff
@@ -70,6 +71,96 @@ def cleanIconCache():
...
+        elif qt.BINDING != "PyQt5":
+            isDark = (
+                qt.QApplication.styleHints().colorScheme() == qt.Qt.ColorScheme.Dark
+            )
+        else:
+            isDark = False
+
+        # Adapted from QSvgIconEngine::scaledPixmap
+        key = f"silx_svgicon_{self._name}_{'dark' if isDark else 'light'}_{size.width()}_{size.height()}_{mode}"
+        pixmap = qt.QPixmapCache.find(key)
```
t20100, line 132: `QPixmapCache` is the global cache Qt uses for icons.
Qt uses `qt_svgicon_` prefix for the icons it stores, so no risk of key conflict.

src/silx/gui/icons.py
```diff
@@ -70,6 +71,96 @@ def cleanIconCache():
...
+        if pixmap is not None:
+            return pixmap
+
+        renderer = qt.QSvgRenderer(self._darkSVG if isDark else self._lightSVG)
```
t20100, line 136: This is how Qt is doing it in `QSvgIconEngine`: Cache the pixmaps and use them in `paint`, do not keep the rendered alive.

t20100 (commented):
- Reworked of the SVG icon engine to support icon mode and updated all SVG icon to use `class="dark"`:



- Added documentation about `class="dark"` in contributing guidelines
Image

doc/source/contribute/icons.rst
```diff
@@ -12,13 +12,44 @@ There are three steps to add an icon:
...
+
+       <?xml version="1.0" encoding="UTF-8"?>
+       <svg version="1.1" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg">
+       <style>
+       .dark * {
+           color: #ffffff;
+       }
+       </style>
+       <rect x="3" y="3" width="26" height="26" rx="2" fill="none" stroke="currentColor" stroke-width="2px"/>
+       </svg>
```
loichuder, line 52: Add that you can look at the existing icons for examples of this?

src/silx/resources/gui/icons/next.svg
loichuder, line 1: Is not this icon duplicated?

src/silx/resources/gui/icons/ruler.svg
loichuder, line 1: Size is not consistent with the others

loichuder (commented):
👏

loichuder (approved):

src/silx/resources/gui/icons/next.svg
t20100: No, it's similar to `last.svg` but it's not the same.

doc/source/contribute/icons.rst
t20100: Done

src/silx/resources/gui/icons/ruler.svg
t20100: Fixed, the `viewBox` was scaled compared to the others

Issue #4594, opened 2026-05-07
silx view: No longer disable hdf5 file locking


t20100:
Introduced in #4376  by moving imports at the top

t20100:
Add a non regression test for this

PR #4376, opened 2025-09-08
silx.app: Fixed flake8 warnings and rework imports

t20100:
Checklist:

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

This PR fixes flake8 warnings in src/silx/app.
It also removes some lazy/optional imports:
- in `convert.py`: Optional imports where from a while ago when `h5py` was an optional dependency...
- Moving hdf5plugin import to the top as well: it's still optional but it no longer need to be imported before h5py

PR #4560, opened 2026-03-27
silx.gui.plot3d.scene: Added typing

t20100:
- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

### PR summary

This PR adds typing to 2 modules of `plot3d.scene`.
This was initiated by sphinx-build warnings (see #4559) and aimed at fixing it while typing the project.

Not need for v3.0.0 release.

#### AI Disclosure
- [x] No AI used
- [ ] AI tool(s) ... used for ...

PR #4595, opened 2026-05-07
silx view: Fixed ensure hdf5 file locking is disabled

t20100:

### PR summary

Set `HDF5_USE_FILE_LOCKING` **before** importing `hdf5plugin`.

related to #4594
partly revert #4376

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,22 @@
 Release Notes
 =============
 
-3.0.0: 2026/04/24
+This is a bug fix release fixing `silx view` v3.0.0 not disabling hdf5 file locking.
+
+3.0.1: 2026/05/07
 -----------------
+
+* `silx view`:
+
+  * Fixed to ensure hdf5 file locking is disabled (PR #4595)
+  * Fixed Ubuntu taskbar icon by setting deskop file name (PR #4475)
+
+* `silx.gui.hdf5.Hdf5Item`: Fixed performance issue by caching icons (PR #4592)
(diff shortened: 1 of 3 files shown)


t20100:
Add preparation of a release.

I propose to make a minor release since there is no new features. The only questionable PR for me is #4560 which adds some typing.

loichuder (approved):

loichuder:
Is it possible to add a test to avoid this in the future?

t20100:
> Is it possible to add a test to avoid this in the future?

Yes! I leave issue #4594 opened to add a non regression test, I prefer not to delay the bug fix release.

woutdenolf:
@sergey-yaroslavtsev To be checked that pymca has the same issue.

sergey-yaroslavtsev:
> @sergey-yaroslavtsev To be checked that pymca has the same issue.

Locking should be set before importing `hdf5plugin` and `h5py`. No sure why but it is so.
I think it is OK in PyMca now:
1st was introduced https://github.com/silx-kit/pymca/pull/1198 (`_init_hdf5()` in PyMcaAppInit.py)
2nd was fixed in https://github.com/silx-kit/pymca/pull/1209

t20100:
> Locking should be set before importing hdf5plugin and h5py. No sure why but it is so.

Now, `libhdf5` look at the locking environment variable once when it is loaded.

This issue was only affecting `silx view`.

payno:
Maybe we could add some comment on the code linking to this issue in order to avoid reproducing it ? (because personally in 2 months I will have forgot :( )

PR #4601, opened 2026-06-01
lz4: unbloc compressed stream using cython code when input data are on the cpu

### LINKED ISSUE #4598: Rewrite the `lz4_unblock` kernel into cython for performances
Single threaded OpenCL kernel should be replaced with host-side compiled code for optimal performance.

kif:

#### AI Disclosure
- [X] No AI used

Close #4598

The performance variation: was 1.2ms, down to 650µs for a 4M Eiger image.

### DIFF
--- a/src/silx/opencl/codec/ext/unblock_lz4.pyx
+++ b/src/silx/opencl/codec/ext/unblock_lz4.pyx
@@ -11,10 +11,9 @@ __authors__ = ["Jérôme Kieffer"]
 __license__ = "MIT"
 __date__ = "29/05/2026"
 
-
 import os
 cimport cython
-from libc.stdint cimport int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t, uint64_t
+from libc.stdint cimport uint8_t, uint32_t, uint64_t
 import logging
 import numpy
 
@@ -22,8 +21,7 @@ __all__ = ['_unblock_lz4']
 
 _logger = logging.getLogger(__name__)
 
-cdef:
-    bint SWAP_BE, SWAP_LE
+cdef bint SWAP_BE, SWAP_LE
 if numpy.little_endian:
     SWAP_BE = True
     SWAP_LE = False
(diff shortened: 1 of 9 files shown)


kif:
With this patch, azimuthal integration as described in https://pyfai.readthedocs.io/en/stable/usage/tutorial/Parallelization/GPU-decompression.html goes from 1.76ms for processing an Eiger4M images to 1.09ms, on the same computer.

kif:
There is likely still a bug somewhere ... OpenCL get randomly killed.

src/silx/opencl/codec/ext/unblock_lz4.pyx
```diff
@@ -0,0 +1,86 @@
+#cython: embedsignature=True, language_level=3
+## This is for optimization
+# cython: boundscheck=False, wraparound=False, cdivision=True, initializedcheck=False,
+## This is for developing:
+##cython: profile=True, warn.undeclared=True, warn.unused=True, warn.unused_result=False, warn.unused_arg=True
+
+"""This module provides :func:`_unblock_lz4` which scans a dataset for different blocks.
```
t20100: Why not putting  `unblock_lz4.pyx` directly in `silx.opencl.codec`?

Regarding naming, it should be a private module `_unblock_l4.pyx` providing a public function `unblock_lz4`.

src/silx/opencl/codec/ext/unblock_lz4.pyx
```diff
@@ -0,0 +1,86 @@
...
+    """
+    cdef uint32_t result
+    result = (<uint32_t>(src[pos + 0]) << 24) | \
+                (<uint32_t>(src[pos + 1]) << 16) | \
+                (<uint32_t>(src[pos + 2]) <<  8) | \
+                (<uint32_t>(src[pos + 3]))
+    return result
+
+
+def _unblock_lz4(bytes src):
```
t20100: this is more `unlock_bitshuffle_lz4` rather than bare LZ4 code block

src/silx/opencl/codec/ext/unblock_lz4.pyx
kif: I tend to dislike mixing python and cython code into the same directory, same for headers and c-code ... especially in the development phase where `.c` and `.html` files are likely to be generated in place and could end up polluting the repo.

kif: For the naming, it is correct that the module should be private.

src/silx/opencl/codec/ext/unblock_lz4.pyx
kif: Indeed, it looks to be specific to bitshuffle-LZ4

t20100 (approved):
LGTM

PR #4604, opened 2026-06-06
numpy: setting an ndarray shape is deprecated since 2.5.0

### LINKED ISSUE #4602: numpy 2.4.6 -> 2.5.0rc1: setting array shape is deprecated
https://gitlab.esrf.fr/workflow/ewoksapps/ewoksndreg/-/work_items/30

https://numpy.org/doc/stable/release/2.4.0-notes.html#in-place-modification-of-ndarray-shape-is-pending-deprecation

woutdenolf:

#### AI Disclosure
- [x] No AI used

For reference: https://github.com/silx-kit/silx/issues/4602#issuecomment-4637994839

### DIFF
--- a/src/silx/gui/data/ArrayTableModel.py
+++ b/src/silx/gui/data/ArrayTableModel.py
@@ -351,10 +351,10 @@ class ArrayTableModel(qt.QAbstractTableModel):
             self._array = numpy.array([])
         elif copy:
             # copy requested (default)
-            self._array = numpy.array(data, copy=True)
-            if hasattr(data, "dtype"):
-                # Avoid to lose the monkey-patched h5py dtype
-                self._array.dtype = data.dtype
+            # Avoid to lose the monkey-patched h5py dtype
+            self._array = numpy.array(
+                data, copy=True, dtype=getattr(data, "dtype", None)
+            )
         elif not _is_array(data):
             raise TypeError(
                 "data is not a proper array. Try setting"
--- a/src/silx/gui/data/ArrayTableModel.py
+++ b/src/silx/gui/data/ArrayTableModel.py
@@ -351,10 +351,7 @@ class ArrayTableModel(qt.QAbstractTableModel):
             self._array = numpy.array([])
         elif copy:
             # copy requested (default)
-            # Avoid to lose the monkey-patched h5py dtype
-            self._array = numpy.array(
-                data, copy=True, dtype=getattr(data, "dtype", None)
-            )
+            self._array = numpy.array(data, copy=True)
         elif not _is_array(data):
             raise TypeError(
                 "data is not a proper array. Try setting"
--- a/src/silx/gui/plot/tools/compare/profile.py
+++ b/src/silx/gui/plot/tools/compare/profile.py
@@ -74,8 +74,8 @@ class ProfileImageLineROI(rois.ProfileImageLineROI):
         title = title + "; width = %d" % lineWidth
         _coords, profile2, _title, _xLabel = createProfile2(currentData2)
 
-        profile1.shape = -1
-        profile2.shape = -1
+        profile1 = profile1.reshape(-1)
+        profile2 = profile2.reshape(-1)
 
         title = title.format(xlabel="width", ylabel="height")
         xLabel = xLabel.format(xlabel="width", ylabel="height")
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,4 +1,4 @@
 CONTRIBUTING
 ============
 
-See https://www.silx.org/doc/silx/dev/contributing.html#how-to-contribute
+See https://www.silx.org/doc/silx/latest/contribute
--- a/doc/source/Tutorials/array_widget.rst
+++ b/doc/source/Tutorials/array_widget.rst
@@ -18,7 +18,7 @@ Let's look at a simple usage example:
     from silx.gui.data.ArrayTableWidget import ArrayTableWidget
     import numpy
     array = numpy.arange(1000)
-    array.shape = (5, 10, 20)
+    array = array.reshape(5, 10, 20)
     app = qt.QApplication([])
     w = ArrayTableWidget()
     w.setArrayData(array, labels=True)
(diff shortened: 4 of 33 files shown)


src/silx/gui/data/ArrayTableModel.py
```diff
@@ -351,10 +351,10 @@ def setArrayData(self, data, copy=True, perspective=None, editable=False):
             self._array = numpy.array([])
         elif copy:
             # copy requested (default)
-            self._array = numpy.array(data, copy=True)
-            if hasattr(data, "dtype"):
-                # Avoid to lose the monkey-patched h5py dtype
-                self._array.dtype = data.dtype
+            # Avoid to lose the monkey-patched h5py dtype
```
woutdenolf: This was introduced in https://github.com/silx-kit/silx/pull/1164/changes/bcf65d8bd59ad95a07f829079a64ef180ef1bf8b

@vallsv @t20100 I would assume `numpy.array(data, copy=True)` would preserve the `dtype`. Can you provide an example that demonstrates that this is not the case?

woutdenolf: I tried to reproduce a situation where the dtype is not copied but failed at doing so

```python
import h5py
import numpy

path = "test.h5"

with h5py.File(path, "w") as f:
    vlen_dt = h5py.string_dtype(encoding="utf-8")
    vlen_data = numpy.array([f"str_{i}" for i in range(10)], dtype=object)
    f.create_dataset("vlen_data", data=vlen_data, dtype=vlen_dt)

    comp_dt = numpy.dtype(
        [
            ("id", numpy.int32),
            ("value", numpy.float64),
        ]
    )

    comp_data = numpy.zeros(10, dtype=comp_dt)
    comp_data["id"] = numpy.arange(10)
    comp_data["value"] = numpy.random.rand(10)

    f.create_dataset("comp_data", data=comp_data)


def run_tests(name, dset):
    copy1 = numpy.array(dset, copy=True)
    copy2 = numpy.array(dset, copy=True, dtype=dset.dtype)

    assert copy1.dtype == dset.dtype, name
    assert copy2.dtype == dset.dtype, name

    assert copy1.dtype is dset.dtype, name
    assert copy2.dtype is dset.dtype, name


with h5py.File(path, "r") as f:
    run_tests("vlen_data (string_dtype)", f["vlen_data"])
    run_tests("comp_data (structured dtype)", f["comp_data"])
```

woutdenolf:
Needs the equivalent fix in fabio: https://github.com/silx-kit/fabio/pull/641. Or we could ignore the fabio warnings.

loichuder (approved):
Looks okay to me.

Do you think we should put `copy=False` everywhere to be sure that no copy is done?

woutdenolf:
> Do you think we should put copy=False everywhere to be sure that no copy is done?

I was thinking about that but decided not to in the end. 

I doubt the `array.shape = ...` pattern was actually used to ensure data did not get copied. We want to reshape the array and if the array happens to be non-contiguous or has an incompatible stride pattern/view that leads to copying, we still want to reshape. There are many other numpy operations that could lead to copying or not depending on the array/view. We rely on numpy to not copy when not needed.

I always hated the `array.shape = ...` pattern, it smells and I never used it. Seems like the numpy developers agree.

loichuder:
>  We want to reshape the array and if the array happens to be non-contiguous or has an incompatible stride pattern/view that leads to copying, we still want to reshape. 

Yeah, ok, I do agree.

src/silx/gui/data/ArrayTableModel.py
woutdenolf: The numpy docs do seem to suggest that providing the `dtype` explicitly might be different than providing `None`

```python
numpy.array(object, dtype=object.dtype)
```

> The desired data-type for the array. If not given, NumPy will try to use a default dtype that can represent the values (by applying promotion rules when necessary.)

The alternative would be

```python
numpy.array(object).astype(dtype=object.dtype)
```

Without an actual use case I have no way of testing.

In any case, setting the dtype is deprecated.

t20100: Provided the date of the commit, this is likely something for `h5py < v3` and/or python 2.7 support, so most likely no longer needed.
I'm having a look, but I would be in favor of removing the special handling of dtype if we don't find a clear need for it.

t20100: > The numpy docs do seem to suggest that providing the dtype explicitly might be different than providing None

From what I tested, if the passed `object` is a `numpy.ndarray` or a `h5py.Dataset`, it uses it's dtype.

I'm 👍 for `self._array = numpy.array(data, copy=True)`

woutdenolf: Done.

src/silx/gui/plot/tools/compare/profile.py
```diff
@@ -74,8 +74,8 @@ def createProfile2(currentData):
         title = title + "; width = %d" % lineWidth
         _coords, profile2, _title, _xLabel = createProfile2(currentData2)
```
kif, line 76: I would prefer to `ravel` or `flatten`  here ...

t20100: `flatten` always makes a copy, so not the best option IMO.

t20100 (approved):

src/silx/gui/plot/tools/compare/profile.py
woutdenolf: Sure. I would do this:

```python
profile1 = profile1.ravel()
profile2 = profile2.ravel()
```

There are several other instances in this PR and all equivalent PRs in other projects that remove setting the shape. However I refrained from doing this. These PRs are to replace `.shape=` with `.reshape(`, nothing else.

t20100: :+1: for merging this PR as is, it removes numpy warnings with minimal changes.

Using `ravel` can be done in another step.

loichuder:
`src/silx/math/fit/test/test_fit.py::Test_leastsq::testDataWithNaN` seem to get an extra `Need to reshape input xdata` warning, making the test fail

loichuder:
It was due to the fact that the test was mutating the input array. The extra warning should indeed be raised now that we use `reshape`.

Merging this and waiting for Fabio's release to fix the issue for good.

src/silx/gui/plot/tools/compare/profile.py
woutdenolf: Follow-up in https://github.com/silx-kit/silx/pull/4620

Issue #4531, opened 2026-03-13
Several issues spotted by ruff


kif:
```
E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
   --> src/silx/gui/data/TextFormatter.py:259:16
    |
257 |         vlen = h5py.check_dtype(vlen=dtype)
258 |         if vlen is not None:
259 |             if vlen == str:
    |                ^^^^^^^^^^^
260 |                 # HDF5 UTF8
261 |                 # With h5py>=3 reading dataset returns bytes
    |

E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
   --> src/silx/gui/data/TextFormatter.py:268:18
    |
266 |                         self.__formatSafeAscii(data)
267 |                 return self.__formatText(data)
268 |             elif vlen == bytes:
    |                  ^^^^^^^^^^^^^
269 |                 # HDF5 ASCII
270 |                 return self.__formatCharString(data)
    |

E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
   --> src/silx/gui/hdf5/Hdf5Formatter.py:125:12
    |
124 |     def humanReadableDType(self, dtype, full=False):
125 |         if dtype == bytes or numpy.issubdtype(dtype, numpy.bytes_):
    |            ^^^^^^^^^^^^^^
126 |             text = "string"
127 |             if full:
    |

E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
   --> src/silx/gui/hdf5/Hdf5Formatter.py:130:14
    |
128 |                 text = "ASCII " + text
129 |             return text
130 |         elif dtype == str or numpy.issubdtype(dtype, numpy.str_):
    |              ^^^^^^^^^^^^
131 |             text = "string"
132 |             if full:
    |

E402 Module level import not at top of file
  --> src/silx/gui/plot/_utils/dtime_ticklayout.py:33:1
   |
33 | from collections.abc import Sequence
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
34 | import datetime as dt
35 | import enum
   |

E402 Module level import not at top of file
  --> src/silx/gui/plot/_utils/dtime_ticklayout.py:34:1
   |
33 | from collections.abc import Sequence
34 | import datetime as dt
   | ^^^^^^^^^^^^^^^^^^^^^
35 | import enum
36 | import logging
   |

E402 Module level import not at top of file
[...]

PR #4611, opened 2026-06-15
Lint/Format: Use ruff instead of black/flake8

loichuder:

### PR summary

For #4531 

#### AI Disclosure
- [x] No AI used


src/silx/gui/data/TextFormatter.py
```diff
@@ -256,7 +256,7 @@ def __formatH5pyObject(self, data, dtype):
                 return "NULL_REF"
         vlen = h5py.check_dtype(vlen=dtype)
         if vlen is not None:
-            if vlen == str:
+            if vlen is str:
```
loichuder: https://docs.astral.sh/ruff/rules/type-comparison/

src/silx/gui/plot/test/test_legendselector.py
```diff
@@ -83,13 +83,12 @@ def signalReceived(self, **kw):
         # layout.setContentsMargins(0,0,0,0)
         llist = []
 
-        for _idx, (l, c, s) in enumerate(zip(legends, colors, symbols)):
+        for _idx, (legend, c, s) in enumerate(zip(legends, colors, symbols)):
```
loichuder, line 86: https://docs.astral.sh/ruff/rules/ambiguous-variable-name/

This is the only place where there was a violation of the rule so I kept it enabled for now

pyproject.toml
```diff
@@ -131,7 +130,26 @@ target-version = ["py310", "py311", "py312", "py313", "py314"]
...
+    "build",
+    "dist",
+    "*.pyi",                                # Don't lint Cython code
+    "doc/source/Tutorials/Image.ipynb",     # Problem when linting magic commands
+    "doc/source/Tutorials/Sift/sift.ipynb", # Problem when linting magic commands
+]
+#
+ignore = [
+    "E402", # Import not at top of file not compatible with COPYRIGHT notice
+    "E501", # Line too long
```
loichuder, line 147: There were some additional rules ignored in `.flake8` but they do not seem to trigger any violation so I did not port them to ruff

pyproject.toml
```diff
@@ -131,7 +130,26 @@ target-version = ["py310", "py311", "py312", "py313", "py314"]
...
+
+[tool.ruff.lint]
+select = ["E", "F"]
+exclude = [
+    ".eggs",
+    "build",
+    "dist",
+    "*.pyi",                                # Don't lint Cython code
+    "doc/source/Tutorials/Image.ipynb",     # Problem when linting magic commands
+    "doc/source/Tutorials/Sift/sift.ipynb", # Problem when linting magic commands
```
loichuder, line 142: When defining variables like

```
&time var = 1
```

`ruff` does not detect that `var` is defined.

Also, the notebooks use `%pylab nbagg` to populate the env with numpy/matplotlib commands and ruff does therefore not detect that these are defined

payno (approved):
Looks good thanks 👍 
If you feel like adding it to the Changelog would be perfect (else we will add it later)

PR #4615, opened 2026-06-15
Revert "lz4: unbloc compressed stream using cython code when input data are on the cpu"

loichuder:
Reverts silx-kit/silx#4601


loichuder:
Closing this: it didn't fix anything

PR #4618, opened 2026-06-15
silx.math.fit: Updated `leastsq` test to use pytest

t20100:

### PR summary

Reworking/splitting test and use pytest

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/math/fit/test/test_fit.py
+++ b/src/silx/math/fit/test/test_fit.py
@@ -1,5 +1,5 @@
 # /*##########################################################################
-# Copyright (C) 2016-2021 European Synchrotron Radiation Facility
+# Copyright (C) 2016-2026 European Synchrotron Radiation Facility
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/silx/math/fit/test/test_fit.py
```diff
@@ -24,365 +24,282 @@
...
-    def testImport(self):
-        self.assertTrue(
-            self.instance is not None, "Cannot import leastsq from silx.math.fit"
+def _my_exp(x):
+    # put a (bad) filter to avoid over/underflows
+    # with no python looping
+    with numpy.errstate(invalid="ignore"):
+        return numpy.exp(x * numpy.less(abs(x), 250)) - 1.0 * numpy.greater_equal(
+            abs(x), 250
         )
```
woutdenolf, line 105: Very weird test.

src/silx/math/fit/test/test_fit.py
```diff
@@ -24,365 +24,282 @@
...
+    constraints_factor = constraints_delta_position * 1
+    constraints_factor[2] = [CFACTOR, 5, 2]
+    constraints_list = [
+        None,
+        constraints_all_free,
+        constraints_all_positive,
+        constraints_delta_position,
+        constraints_sum_position,
+        constraints_factor,
+    ]
```
woutdenolf: Added more to the list than originally. Ok.

src/silx/math/fit/test/test_fit.py
```diff
@@ -24,365 +24,282 @@
...
+
+@pytest.mark.parametrize("check_finite", [True, False])
+def test_dadly_shaped_data(check_finite):
+    parameters_actual = [10.5, 2, 1000.0, 20.0, 15]
+    x = numpy.arange(10000.0).reshape(1000, 10)
+    y = _gauss(x, *parameters_actual)
+    sigma = numpy.sqrt(y)
+    parameters_estimate = [0.0, 1.0, 900.0, 25.0, 10]
+
+    with testutils.LoggingValidator(fitlogger.name, warning=1):
```
woutdenolf: Used to be 2 warnings. I see this all over the test suite: we count logs but do not regex-match their content so hard to judge.

src/silx/math/fit/test/test_fit.py
```diff
@@ -24,365 +24,282 @@
...
-                msg += f"Expected {parameters_actual[i]:g} obtained {fittedpar[i]:g}\n"
-            self.assertTrue(test_condition, msg)
-
-        # testing now with ydata containing NaN
-        x = numpy.arange(10000.0).reshape(1000, 10)
-        y[500] = numpy.nan
-        fittedpar, cov = self.instance(
-            model_function, x, y, parameters_estimate, sigma=sigma, check_finite=False
-        )
+    assert numpy.allclose(parameters_actual, fittedpar), "Unsuccessfull fit"
```
woutdenolf:
```suggestion
    assert numpy.allclose(parameters_actual, fittedpar), "Unsuccessful fit"
```

The error message used to be more detailed. I don't mind.

src/silx/math/fit/test/test_fit.py
woutdenolf: I suggest `@pytest.mark.parametrize`.

src/silx/math/fit/test/test_fit.py
woutdenolf: On the other hand, it is reused often so perhaps add a helper:

```python
def assert_fit_success(
    expected,
    actual,
    *,
    constraints=None,
    model_deriv=None,
    with_sigma=None,
    full_output=None,
):
    msg_parts = []

    if constraints is not None:
        msg_parts.append(f"constraints={constraints}")

    if model_deriv is not None:
        msg_parts.append(f"model_deriv={model_deriv.__name__}")

    if with_sigma is not None:
        msg_parts.append(f"with_sigma={with_sigma}")

    if full_output is not None:
        msg_parts.append(f"full_output={full_output}")

    context = ", ".join(msg_parts)

    numpy.testing.assert_allclose(
        actual,
        expected,
        err_msg=context,
    )
```

woutdenolf: Or more lightweight

```python
def assert_fit_success(expected, actual, **context):
    assert numpy.allclose(expected, actual), (
        f"Fit failed: {context}\n"
        f"Expected: {expected}\n"
        f"Actual:   {actual}"
    )
```

src/silx/math/fit/test/test_fit.py
```diff
@@ -24,365 +24,282 @@
...
+
+    # test that uncertainties are not 0.
+    fittedpar, cov, infodict = leastsq(
+        _gauss, x, y, parameters_estimate, full_output=True
+    )
+    uncertainties = infodict["uncertainties"]
+    assert len(uncertainties) == len(parameters_actual)
+    assert len(uncertainties) == len(fittedpar)
+    for uncertainty in uncertainties:
+        assert round(uncertainty, 7) != 0
```
woutdenolf: Some suggestions but not sure it is better, I just never saw this pattern

```python
assert uncertainty != pytest.approx(0.0, abs=1e-7, rel=0)
```

```python
assert not numpy.isclose(uncertainty, 0.0, atol=1e-7, rtol=0)
```

woutdenolf (approved):
Some nits but look ok. Thanks!

src/silx/math/fit/test/test_fit.py
woutdenolf: Or in fact more clear to me

```python
assert abs(uncertainty) > 1e-7
```

src/silx/math/fit/test/test_fit.py
woutdenolf: Since we are refactoring to use pytest, why not make something like `pytest.warns` or `pytest.raises`?

```python
with logs(
    caplog,
    logging.INFO, # just count
    (logging.WARNING, re.escape("...")), # count an match
    allow_others=True,  # the default
    ordered=False,  # the default
):
    run()
```

woutdenolf: LLM generated code:

```python
import logging
import re
from typing import NamedTuple, Optional, Union

from _pytest.logging import LogCaptureFixture


def logs(
    caplog: LogCaptureFixture,
    *expected: "_ExpectedArg",
    ordered: bool = True,
    allow_others: bool = False,
):
    expected_norm = [_normalize(e) for e in expected]

    capture_level = min(
        (e.level for e in expected_norm),
        default=logging.WARNING,
    )

    start = len(caplog.records)

    with caplog.at_level(capture_level):
        yield

    records = caplog.records[start:]

    if ordered:
        if allow_others:
            pos = 0
            for exp in expected_norm:
                while pos < len(records) and not _matches(records[pos], exp):
                    pos += 1
                assert pos < len(records), f"Missing log: {exp}"
                pos += 1
        else:
            assert len(records) == len(expected_norm)
            for r, e in zip(records, expected_norm):
                assert _matches(r, e)
    else:
        unmatched = list(records)

        for exp in expected_norm:
            for i, r in enumerate(unmatched):
                if _matches(r, exp):
                    del unmatched[i]
                    break
            else:
                raise AssertionError(f"Missing log: {exp}")

        if not allow_others:
            assert not unmatched


class _ExpectedLog(NamedTuple):
    level: int
    match: Optional[str] = None


_ExpectedArg = Union[int, tuple[int, str]]


def _normalize(expected: _ExpectedArg) -> _ExpectedLog:
    if isinstance(expected, int):
        return _ExpectedLog(level=expected)

    level, match = expected
    return _ExpectedLog(level=level, match=match)


def _matches(record, expected: _ExpectedLog) -> bool:
    if record.levelno != expected.level:
        return False

    if expected.match is not None:
        return bool(re.search(expected.match, record.getMessage()))

    return True
```

woutdenolf: Note: don't forget this in conftest.py when doing asserts in helper modules

```python
pytest.register_assert_rewrite("silx.utils.assertlogs")
```

This is to have proper diffs in asserts etc.

t20100: Updated to use `caplog` but without a helper, `caplog.record_tuples` makes it simple enough to check the log messages

src/silx/math/fit/test/test_fit.py
t20100: I took the lightweight approach + now that all contitions are parameterized, there is not much interest for the context.

src/silx/math/fit/test/test_fit.py
t20100: I used the definition of `assertAlmostEqual` in `unittest` doc, but much simpler this way

woutdenolf (approved):
Thanks!

PR #4619, opened 2026-06-15
silx.gui.plot.actions.io: add an option to save a plot with the colorbar

### LINKED ISSUE #4231: Save Image Snapshot w/ Colorbar
Would be great if when saving a snapshot of an image with the colorbar displayed, the colorbar saved in the snapshot.

Today I get: 

Image

Would like to get:

Image

payno:
### PR summary

silx.gui.plot.actions.io: add an option to save a plot with the colorbar

!!! Warning: this option (`Plot Snapshot with ColorBar as PNG (*.png)`) is only available if the `plot` of the action is an instance of `PlotWindow`.

Closes #4231 

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/actions/io.py
+++ b/src/silx/gui/plot/actions/io.py
@@ -191,7 +191,7 @@ class SaveAction(PlotAction):
                 and nameFilter == self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG
             ):
                 _logger.info(
-                    f"Ignore {nameFilter} because requires to have acces to the ColorBarWidget"
+                    f"Ignore {nameFilter} because requires to have access to the ColorBarWidget"
                 )
                 continue
             self.setFileFilter(
--- a/src/silx/gui/plot/actions/io.py
+++ b/src/silx/gui/plot/actions/io.py
@@ -52,7 +52,6 @@ from silx.io.nxdata import save_NXdata
 
 from . import PlotAction
 from ...utils.image import convertArrayToQImage
-from contextlib import contextmanager
 
 _logger = logging.getLogger(__name__)
 
@@ -85,12 +84,10 @@ class SaveAction(PlotAction):
     """
 
     SNAPSHOT_FILTER_SVG = "Plot Snapshot as SVG (*.svg)"
-    SNAPSHOT_FILTER_PNG = "Plot Snapshot as PNG (*.png)"
-    SNAPSHOT_WITH_COLORBAR_FILTER_PNG = "Plot Snapshot with ColorBar as PNG (*.png)"
+    SNAPSHOT_FILTER_PNG = "View Snapshot as PNG (*.png)"
 
     DEFAULT_ALL_FILTERS = (
         SNAPSHOT_FILTER_PNG,
-        SNAPSHOT_WITH_COLORBAR_FILTER_PNG,
         SNAPSHOT_FILTER_SVG,
     )
 
@@ -183,17 +180,7 @@ class SaveAction(PlotAction):
         self._appendFilters = list(self.DEFAULT_APPEND_FILTERS)
 
         # Initialize filters
-        from ..PlotWindow import PlotWindow  # noqa: F811 avoid cyclic import
-
         for nameFilter in self.DEFAULT_ALL_FILTERS:
-            if (
-                not isinstance(plot, PlotWindow)
-                and nameFilter == self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG
-            ):
-                _logger.info(
-                    f"Ignore {nameFilter} because requires to have access to the ColorBarWidget"
-                )
-                continue
             self.setFileFilter(
                 dataKind="all", nameFilter=nameFilter, func=self._saveSnapshot
             )
@@ -248,31 +235,18 @@ class SaveAction(PlotAction):
         :return: False if format is not supported or save failed,
                  True otherwise.
         """
-        if nameFilter in (
-            self.SNAPSHOT_FILTER_PNG,
-            self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG,
-        ):
-            fileFormat = "png"
+        if nameFilter == self.SNAPSHOT_FILTER_PNG:
+            # Note: The contents of the view depend on the plot type:
+            # - When `plot` is a PlotWindow instance, the view includes the color bar (if visible).
+            # - When `plot` is a PlotWidget, the view contains only the plot.
+            centralWidget = self.plot.centralWidget()
+            pixmap = centralWidget.grab()
+            pixmap.save(filename)
         elif nameFilter == self.SNAPSHOT_FILTER_SVG:
-            fileFormat = "svg"
+            plot.saveGraph(filename, fileFormat="svg")
         else:  # Format not supported
             _logger.error("Saving plot snapshot failed: format not supported")
             return False
-        if nameFilter == self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG:
-            # Note: today taking a pixmap of the widget works because the
-            # Central widget is only composed of the plot and the ColorBarWidget.
-            from ..PlotWindow import PlotWindow  # noqa: F811 avoid cyclic import
-
-            if not isinstance(plot, PlotWindow):
-                raise RuntimeError(
-                    f"An instance of {PlotWindow} is expected. Got {type(self.plot)}"
-                )
-            with _colorbar_visible(self.plot):
-                centralWidget = self.plot.centralWidget()
-                pixmap = centralWidget.grab()
-                pixmap.save(filename)
-        else:
-            plot.saveGraph(filename, fileFormat=fileFormat)
         return True
 
     def _getAxesLabels(self, item):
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/actions/io.py
```diff
@@ -232,15 +244,29 @@ def _saveSnapshot(self, plot, filename, nameFilter):
...
             _logger.error("Saving plot snapshot failed: format not supported")
             return False
+        if nameFilter == self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG:
+            from ..PlotWindow import PlotWindow  # noqa: F811 avoid cyclic import
 
-        plot.saveGraph(filename, fileFormat=fileFormat)
+            if not isinstance(plot, PlotWindow):
+                raise RuntimeError(
+                    f"An instance of {PlotWindow} is expected. Got {type(self.plot)}"
+                )
```
payno: This should never happen. Maybe this is a bit over-killed ? (but usually I like to have a clear error when those things happen)

src/silx/gui/plot/actions/io.py
```diff
@@ -177,7 +183,17 @@ def __init__(self, plot, parent=None):
...
+
         for nameFilter in self.DEFAULT_ALL_FILTERS:
+            if (
+                not isinstance(plot, PlotWindow)
+                and nameFilter == self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG
+            ):
+                _logger.info(
+                    f"Ignore {nameFilter} because requires to have acces to the ColorBarWidget"
+                )
+                continue
```
payno: I don't know if having this action only for PlotWindow is a big issue or not and / or if it should be advise somewhere else ?

src/silx/gui/plot/actions/io.py
```diff
@@ -873,3 +905,23 @@ def copyPlot(self):
         pngData = _plotAsPNG(self.plot)
         image = qt.QImage.fromData(pngData, "png")
         qt.QApplication.clipboard().setImage(image)
+
+
+@contextmanager
+def _colorbar_visible(plot):
```
payno: Not sure if it worth moving to a dedicated file ?

src/silx/gui/plot/actions/io.py
loichuder: I think it makes sense to have the check.

Would it help to have it in a helper?

src/silx/gui/plot/actions/io.py
```diff
@@ -177,7 +183,17 @@ def __init__(self, plot, parent=None):
...
         # Initialize filters
+        from ..PlotWindow import PlotWindow  # noqa: F811 avoid cyclic import
+
         for nameFilter in self.DEFAULT_ALL_FILTERS:
+            if (
+                not isinstance(plot, PlotWindow)
+                and nameFilter == self.SNAPSHOT_WITH_COLORBAR_FILTER_PNG
+            ):
+                _logger.info(
+                    f"Ignore {nameFilter} because requires to have acces to the ColorBarWidget"
```
loichuder:
```suggestion
                    f"Ignore {nameFilter} because requires to have access to the ColorBarWidget"
```

loichuder (commented):
I don't know if adding another filter is the right way.

Could we not add the colorbar if it is visible when saving, whatever the format?

src/silx/gui/plot/actions/io.py
payno: Depends on the appreciation. To me this is very light but I can add one if someone has a strong opinion on it.

payno:
> n't know if adding another filter is the right way.
> 
> Could we not add the colorbar if it is visible when saving, whatever the format?

I guess it depends on the understanding of the issue. For me the goal is to have a snapshot that embed the colorbar. But it could be to have a snapshot of the widget. But for me the second one has less interest.

src/silx/gui/plot/actions/io.py
```diff
@@ -85,8 +86,13 @@ class SaveAction(PlotAction):
 
     SNAPSHOT_FILTER_SVG = "Plot Snapshot as SVG (*.svg)"
     SNAPSHOT_FILTER_PNG = "Plot Snapshot as PNG (*.png)"
+    SNAPSHOT_WITH_COLORBAR_FILTER_PNG = "Plot Snapshot with ColorBar as PNG (*.png)"
```
t20100: why not changing `SNAPSHOT_FILTER_PNG` rather than adding an option?

t20100 (commented):
I would make it much simpler: just `grab` the central widget and do not care to show the colorbar.

The code would be simpler + no need for lazy import/specific treatment of `PlotWindow`.
From a user point of view, this is what I would expect: take a screenshot of what is currently displayed.
What do you think?

loichuder:
> From a user point of view, this is what I would expect: take a screenshot of what is currently displayed.

Agreed

payno:
Well I have an interrogation on being consistent with what exists.

If I go for saving the the `centralWidget` then I would expect the current `SNAPSHOT_FILTER_SVG` and `SNAPSHOT_FILTER_PNG` to have the same behavior. But today the `QPixmap` is not able to save as `.svg`. This is also the reason of `PlotWidget.saveGraph` to exists.

So having to different behavior between `SNAPSHOT_FILTER_PNG` and `SNAPSHOT_FILTER_SVG` doesn't make much sense IMO.

We can rename `SNAPSHOT_FILTER_SVG` to make it clear the result will be different and move it from `"Plot Snapshot as PNG (*.png)"` to `"Widget Snapshot as PNG (*.png)"`. But in this case do users will expect to have buttons as well ?

t20100:
> But it could be to have a snapshot of the widget

By providing the snapshot of the widget, it provides the colorbar snapshot when it matters.

Initial issue #4231 statement was:

> when saving a snapshot of an image with the colorbar displayed, the colorbar saved in the snapshot.

We could implement it for SVG as well but that would be clunky so I would not do it (it's possible to embed a base64 encoded png image in a svg).

+1 to provide slightly different descriptions for png and svg.
I propose:
- "Plot Area Snapshot as SVG (*.svg)"
- "Plot Snapshot as PNG (*.png)"
but no strong opinion.

src/silx/gui/plot/actions/io.py
```diff
@@ -84,9 +85,12 @@ class SaveAction(PlotAction):
     """
 
     SNAPSHOT_FILTER_SVG = "Plot Snapshot as SVG (*.svg)"
-    SNAPSHOT_FILTER_PNG = "Plot Snapshot as PNG (*.png)"
+    SNAPSHOT_FILTER_PNG = "View Snapshot as PNG (*.png)"
```
payno, line 87: The class variable `SNAPSHOT_FILTER_PNG` could also be renamed. But not sure this is worth breaking the API

src/silx/gui/plot/actions/io.py
```diff
@@ -233,14 +237,17 @@ def _saveSnapshot(self, plot, filename, nameFilter):
                  True otherwise.
         """
         if nameFilter == self.SNAPSHOT_FILTER_PNG:
-            fileFormat = "png"
+            # Note: The contents of the view depend on the plot type:
+            # - When `plot` is a PlotWindow instance, the view includes the color bar (if visible).
+            # - When `plot` is a PlotWidget, the view contains only the plot.
+            centralWidget = self.plot.centralWidget()
```
payno: Doing this directly is simpler and better from the code point of view. My concern is more that depending on `plot` I don't know if users will know what they will generate as output.

src/silx/gui/plot/actions/io.py
t20100: +1 to keep it this way

src/silx/gui/plot/actions/io.py
t20100: Yes, it takes whatever is the central widget.

Looking at `PlotWindow`, it also takes the bottom `PositionInfo` as well (it was already the case with the first implementation of this PR).

If this is an issue, an alternative is, rather than changing the action, is to override `saveGraph()` in `PlotWindow` which knows it's content and is responsible for selecting what goes in the screenshot.

payno: Yes, but this is where you start to have specific treatment according to the type of `plot`. This might become more complex and harder to maintain.
So IMO this is either grabbing the view and get something simple or having something more advance and we would go more for the plot + the colorbar. I vote for 1 :innocent:
If this is reported as an issue for users then we might go for 2... or something else

src/silx/gui/plot/actions/io.py
loichuder: No longer relevant.

src/silx/gui/plot/actions/io.py
loichuder: 1. is good enough I'd say.

Not having the colormap is kind of a big deal so personally, I am willing to compromise by having the `PositionInfo` as well.

loichuder: Feels weird to use `plot` (the argument) for SVG and `self.plot` for PNG though.

Could we use the `plot` for both?

payno: Indeed nice catch thanks :)

f578a0196284d15ed02d9ac4cc7a785d563b9fc7 will fix it

t20100 (approved):
LGTM, thanks for the rework!

t20100 (approved):
LGTM, thanks!

PR #4624, opened 2026-06-16
Refactor profile test

payno:
Profile: preparatory PR to rework test before https://github.com/silx-kit/silx/pull/4625

### PR summary

This PR rework some test of the profile (move to pytest) and add test on the result of the profile calculated.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/tools/test/test_profile.py
+++ b/src/silx/gui/plot/tools/test/test_profile.py
@@ -29,7 +29,6 @@ __date__ = "28/06/2018"
 import contextlib
 import numpy
 import logging
-import pytest
 
 from silx.gui import qt
 from silx.gui.utils.testutils import qWaitForWindowExposedAndActivate, QTest
@@ -529,38 +528,11 @@ class TestProfile3DToolBar(TestCaseQt):
         numpy.testing.assert_almost_equal(data, expected)
 
 
-@pytest.fixture
-def plot2D_for_profile(qapp):
-    plot = Plot2D()
+def testProfile1D(qWidgetFactory):
+    plot = qWidgetFactory(Plot2D)
     plot.show()
     qWaitForWindowExposedAndActivate(plot)
 
-    yield plot
-
-    manager = plot.getProfileToolbar().getProfileManager()
-    manager.clearProfile()
-    manager = None
-    plot.setAttribute(qt.Qt.WA_DeleteOnClose)
-    plot.close()
-
-
-@pytest.fixture
-def plotStackView_for_profile(qapp):
-    plot = StackView()
-    plot.show()
-    qWaitForWindowExposedAndActivate(plot)
-
-    yield plot
-
-    manager = plot.getProfileToolbar().getProfileManager()
-    manager.clearProfile()
-    manager = None
-    plot.setAttribute(qt.Qt.WA_DeleteOnClose)
-    plot.close()
-
-
-def testProfile1D(plot2D_for_profile):
-    plot = plot2D_for_profile
     plot.addImage([[0, 1], [2, 3], [4, 5], [6, 7], [8, 9]])
 
     toolBar = plot.getProfileToolbar()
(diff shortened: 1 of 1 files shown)


src/silx/gui/plot/tools/test/test_profile.py
```diff
@@ -29,8 +29,10 @@
 import contextlib
 import numpy
 import logging
+import pytest
 
 from silx.gui import qt
+from silx.gui.utils.testutils import qWaitForWindowExposedAndActivate, QTest
```
payno, line 34: Used to move to pytest. Not the entire tests are moved to pytest but only the `TestGetProfilePlot` part I was interest about.

src/silx/gui/plot/tools/test/test_profile.py
```diff
@@ -526,83 +529,123 @@ def testMethodSumLine(self):
...
+    assert isinstance(plotWidget, Plot1D)
 
-        profileWindow = roi.getProfileWindow()
-        self.assertIsInstance(roi.getProfileWindow(), qt.QMainWindow)
-        self.assertIsInstance(profileWindow.getCurrentPlotWidget(), Plot2D)
+    # check result
+    curves = plotWidget.getAllCurves()
+    assert len(curves) == 1
+    profile = curves[0]
+    numpy.testing.assert_almost_equal(profile.getYData(), numpy.array([4.0, 5.0]))
```
payno, line 563: The test on the result has been added for the two. This is the only difference.

src/silx/gui/plot/tools/test/test_profile.py
```diff
@@ -526,83 +529,123 @@ def testMethodSumLine(self):
         numpy.testing.assert_almost_equal(data, expected)
 
 
-class TestGetProfilePlot(TestCaseQt):
-    def setUp(self):
-        self.plot = None
-        super().setUp()
+@pytest.fixture
+def plot2D_for_profile(qapp):
```
t20100: There is a [`qWidgetFactory` fixture](https://github.com/silx-kit/silx/blob/3c88cd028557f931b5691798852f80292810b19d/src/silx/conftest.py#L184) to handle widget life-cycle. Best to use IMO

src/silx/gui/plot/tools/test/test_profile.py
```diff
@@ -526,83 +529,123 @@ def testMethodSumLine(self):
...
-            manager = None
-            self.plot.setAttribute(qt.Qt.WA_DeleteOnClose)
-            self.plot.close()
-            self.plot = None
+    yield plot
 
-        super().tearDown()
+    manager = plot.getProfileToolbar().getProfileManager()
+    manager.clearProfile()
+    manager = None
```
t20100: do you need to do this clean up? It should be done when the widget is deleted, right?

src/silx/gui/plot/tools/test/test_profile.py
```diff
@@ -526,83 +529,123 @@ def testMethodSumLine(self):
...
 
-    def testProfile1D(self):
-        self.plot = Plot2D()
-        self.plot.show()
-        self.qWaitForWindowExposed(self.plot)
-        self.plot.addImage([[0, 1], [2, 3]])
 
-        toolBar = self.plot.getProfileToolbar()
+@pytest.fixture
+def plotStackView_for_profile(qapp):
```
t20100: same, best to use qWidgetFactory if possible

t20100 (approved):
I propose to use the `qWidgetFactory` fixture instead of `plot2D_for_profile` and `plotStackView_for_profile`.

Otherwise LGTM!

src/silx/gui/plot/tools/test/test_profile.py
payno: I was not even aware this was existing :rofl: thanks for the tip

src/silx/gui/plot/tools/test/test_profile.py
payno: I keep it as close as it was. But this is not done at least explicitly from the code I can see

src/silx/gui/plot/tools/test/test_profile.py
woutdenolf: Yes probably lots of utils like this that could be used in the `ewoksorange` test suite and the associate test suites in many ewoks apps.

t20100: I started to look at `pytest-qt`, IMO that is the best place to share such utils.
They have similar features: https://pytest-qt.readthedocs.io/en/latest/reference.html#pytestqt.qtbot.QtBot

src/silx/gui/plot/tools/test/test_profile.py
t20100: I would expect deleting the plot to do the clean up automatically, right?

PR #4625, opened 2026-06-16
silx.gui.plot.tools.profile: fix make profile take into account the mask properly

### LINKED ISSUE #3920: plot: Side histogram does not support mark
There is maybe a way to reuse the mark profile when the side histograms are displayed

There is probably a way to display something cleaner.

image

### LINKED ISSUE #4493: profile does not take into account the mask properly
Hello,

I am trying to integrate an image along a direction and taking into account a mask. Has you can see on the image, the integrated plot has an hole corresponding to the mask X coordinates.
But the width of the projection is largen than the mask, So it should be possible to compute a value in the blank area.

Is it a bug in silx ?

Cheers

Fred

Image

payno:
Closes #4493 
Closes https://github.com/silx-kit/silx/issues/3920
https://github.com/silx-kit/silx/pull/4624 to be merged first.

### PR summary

This PR improve handling of mask when computing profile. When a mask is applied values of the data will be `nan`.

### Demo

[Screencast from 2026-06-16 13-32-31.webm](https://github.com/user-attachments/assets/76218b10-db75-4937-949c-d44f4bb4ae1f)

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/tools/profile/core.py
+++ b/src/silx/gui/plot/tools/profile/core.py
@@ -219,6 +219,8 @@ def _alignedFullProfile(data, origin, scale, position, roiWidth, axis, method):
     assert axis in (0, 1)
     assert len(data.shape) == 3
     assert method in ("mean", "sum", "none")
+    # filter nan values
+    data = numpy.ma.masked_array(data=data, mask=numpy.isnan(data))
 
     # Convert from plot to image coords
     imgPos = int((position - origin[1 - axis]) / scale[1 - axis])
--- a/src/silx/gui/plot/tools/profile/core.py
+++ b/src/silx/gui/plot/tools/profile/core.py
@@ -219,8 +219,6 @@ def _alignedFullProfile(data, origin, scale, position, roiWidth, axis, method):
     assert axis in (0, 1)
     assert len(data.shape) == 3
     assert method in ("mean", "sum", "none")
-    # filter nan values
-    data = numpy.ma.masked_array(data=data, mask=numpy.isnan(data))
 
     # Convert from plot to image coords
     imgPos = int((position - origin[1 - axis]) / scale[1 - axis])
@@ -243,9 +241,9 @@ def _alignedFullProfile(data, origin, scale, position, roiWidth, axis, method):
     else:
         if start < height and end > 0:
             if method == "mean":
-                fct = numpy.mean
+                fct = numpy.nanmean
             elif method == "sum":
-                fct = numpy.sum
+                fct = numpy.nansum
             else:
                 raise ValueError("method not managed")
             profile = fct(data[:, max(0, start) : min(end, height), :], axis=1).astype(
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/tools/profile/core.py
```diff
@@ -219,6 +219,8 @@ def _alignedFullProfile(data, origin, scale, position, roiWidth, axis, method):
     assert axis in (0, 1)
     assert len(data.shape) == 3
     assert method in ("mean", "sum", "none")
+    # filter nan values
+    data = numpy.ma.masked_array(data=data, mask=numpy.isnan(data))
```
payno: Alternative exists like:
* using a masked_array only when the data contains `nan`
* using `nansum` and `nanmean` instead

If someone has a strong opinion on those alternative I might be receptive.

As those are numpy native I didn't looked for performances bottleneck.

src/silx/gui/plot/tools/test/test_profile.py
```diff
@@ -559,9 +559,19 @@ def plotStackView_for_profile(qapp):
     plot.close()
 
 
-def testProfile1D(plot2D_for_profile):
+@pytest.mark.parametrize("with_mask", (True, False))
```
payno, line 532: We could also have a dict that integrate the expected result for the different use cases or include several mask. Having a bool was the simplest IMO.

loichuder: `with_mask` didn't seem used in `testProfile1D` ?

Also, should't the expected profile change depending on `with_mask`?

src/silx/gui/plot/tools/profile/core.py
loichuder: So with the masked array, the `sum` and `mean` operation implicitly skip NaNs?

Personally, if it is enough the replace `mean` and `sum` by `nanmean` and `nansum` below, I would prefer this solution that is a bit more explicit.

src/silx/gui/plot/tools/test/test_profile.py
payno: This is what I added on a note of the test. On this case (with the current values) the expected output is the same. Which was not the case before.

src/silx/gui/plot/tools/profile/core.py
payno: Could, no strong opinion. Having masked array could ease other processing. But I doubt there will be others.

src/silx/gui/plot/tools/test/test_profile.py
loichuder: Ok!

But still no mask in created in `testProfile1D` as far as I can see.

payno: ah :sob: looks like a mess up a rebase thanks. d73e843f0d5851bd8063b07af5eb2b774f64e050 will fix it.

loichuder (approved):
Excellent stuff 👌

loichuder:
@payno Would be good to rename the PR to include the relevant module

Issue #4605, opened 2026-06-06
Qt bug: QWidget.destroyed callback referencing the QWidget itself can cause segfaults


woutdenolf:
More details on how to reproduce the issue and the gdb traceback https://gitlab.esrf.fr/workflow/ewoksapps/ewoksfluo/-/work_items/91#note_563654

This is the design problem in silx

```python
class Hdf5TreeModel(qt.QAbstractItemModel):

    def _init(self):
        ...

        onDestroy = functools.partial(self._closeFileList, self.__openedFiles)
        self.destroyed.connect(onDestroy)

    @staticmethod
    def _closeFileList(fileList):
        """Static method to close explicit references to internal objects."""
        _logger.debug("Clear AbstractDataFileDialog")
        for obj in fileList:
            _logger.debug("Close file %s", obj.filename)
            obj.close()
        fileList[:] = []
```

If self is part of a reference cycle, GC may be tearing it down at the same time Qt emits destroyed. This causes a SEGFAULT.

The top of the traceback is the partial call

```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
Downloading source file /tmp/python-build.20250823125345.35687/Python-3.12.11/./Modules/_functoolsmodule.c
0x00007ffff7b884d5 in partial_vectorcall (pto=0x7fffc618c360, args=0x7fffc43f70d8, nargsf=1, kwnames=0x0)                      
    at ./Modules/_functoolsmodule.c:215
warning: 215    ./Modules/_functoolsmodule.c: No such file or directory
(gdb) bt
#0  0x00007ffff7b884d5 in partial_vectorcall (pto=0x7fffc618c360, args=0x7fffc43f70d8, nargsf=1, kwnames=0x0)
    at ./Modules/_functoolsmodule.c:215
#1  0x00007fffe1405c25 in PyQtSlot::call(_object*, _object*) const ()
   from /home/denolf/.local/wdnbashutils/virtualenvs/pybox_4v53Ca/lib/python3.12/site-packages/PyQt5/QtCore.abi3.so
#2  0x00007fffe1406128 in PyQtSlot::invoke(void**, _object*, void*, bool) const ()
   from /home/denolf/.local/wdnbashutils/virtualenvs/pybox_4v53Ca/lib/python3.12/site-packages/PyQt5/QtCore.abi3.so
#3  0x00007fffe14063d0 in PyQtSlotProxy::unislot(void**) ()
[...]

PR #4627, opened 2026-06-16
4262 doc update silx compare doc

### LINKED ISSUE #4262: Doc: update silx compare doc
On [this page](https://www.silx.org/doc/silx/2.0.0/applications/compare.html)
It would be good to provide at least a screenshot to the silx compare doc and provide a link a tutorial to it (if any)

payno:
### PR summary

Update the silx-compare (and fix it)

#### AI Disclosure
- [ ] No AI used
- [x] lobchat used for generating doc of the different mode


doc/source/applications/img/silx-compare.png
payno, line 1: I changed the image to have something a bit more 'nicer' but this is totally debatable. I can roll back to the old one if needed.

doc/source/applications/compare.rst
```diff
@@ -8,13 +8,62 @@ Purpose
...
+- Displays the raw concatenated data without color encoding
+- Shows unprocessed dataset values for direct inspection
+
+
+More
+----
+
+You can add files by dragging and dropping them from ``silx view``.
+
+  .. figure:: http://www.silx.org/doc/silx/img/silx_compare_drag_and_drop.gif
```
payno, line 52: Two points:

* This is probably overkilled but this is adding some substance :(
* At the moment the two .gif are stored at silx.org. This might be something we want to change in the future ... But at least this doesn't pollute the git project.

src/silx/gui/widgets/UrlSelectionTable.py
```diff
@@ -45,7 +45,7 @@
 class _IntegratedRadioButton(qt.QWidget):
     """RadioButton integrated in the QTableWidget as a centered widget"""
 
-    toggled = qt.Signal()
+    toggled = qt.Signal(bool)
```
payno, line 48: There was a bug here so I fix it in this PR because `self._radio.toggled` propagate a boolean.

src/silx/gui/widgets/UrlSelectionTable.py
```diff
@@ -233,15 +233,15 @@ def setError(self, urlPath: str, message: str):
             item.setIcon(icon)
             item.setToolTip(f"Error: {message}")
 
-    def _activeImgAChanged(self, row):
+    def _activeImgAChanged(self, row: int, checked: bool):
```
payno, line 236: This is continuation of the fix as we also have a book now (which is the effect of the `toggled` signal)

doc/source/applications/compare.rst
```diff
@@ -8,13 +8,62 @@ Purpose
...
-This list can be updated with drag and drop from silx view.
+This tools allow you to list list and open multiple datasets.
 
+.. |imgViewImg| figure:: img/silx-compare.png
+  :height: 300px
+  :align: center
 
-.. |imgViewImg| image:: img/silx-compare.png
-   :height: 300px
-   :align: middle
```
payno, line 17: the `image`  routine cannot handle the `middle` alignment. So moved to `figure` -> `center`. More consistent with existing as well.

doc/source/applications/compare.rst
```diff
@@ -8,13 +8,62 @@ Purpose
 
 The *silx compare* command provides a graphical user interface to compare 2D data from files.
 
-It displays a list of datasets, and allows to open multiple datasets.
-This list can be updated with drag and drop from silx view.
+This tools allow you to list list and open multiple datasets.
```
t20100:
```suggestion
This tool allows you to list land open multiple datasets.
```

src/silx/gui/widgets/UrlSelectionTable.py
t20100: What was the bug? `self._radio.toggled.connect(self.toggled)` didn't work with `toggled = qt.Signal()`?

It would be interesting to know, since here some `functools.partial` are connected to a signal and this is the same pattern as in #4605 which leads to seg fault, so this may be a generic pattern

src/silx/gui/widgets/UrlSelectionTable.py
t20100: If `checked` is not there, it should not be a problem... but it may be thanks to `functools.partial` 🤷

doc/source/applications/compare.rst
```diff
@@ -8,13 +8,62 @@ Purpose
...
+----------------
+
+Two selected datasets (labeled ``A`` and ``B``) can be compared using different visualization modes:
+
+Vertical Compare Mode
+"""""""""""""""""""""
+
+- Splits the display horizontally into two panels
+- Dataset ``A`` is displayed on the **left** side
+- Dataset ``B`` is displayed on the **right** side
```
t20100:
```suggestion
Splits the display horizontally into two panels:

- Dataset ``A`` is displayed on the **left** side
- Dataset ``B`` is displayed on the **right** side
```

🤷

doc/source/applications/compare.rst
```diff
@@ -8,13 +8,62 @@ Purpose
...
+
+Horizontal Compare Mode
+"""""""""""""""""""""""
+
+- Splits the display vertically into two panels
+- Dataset ``A`` is displayed at the **top**
+- Dataset ``B`` is displayed at the **bottom**
```
t20100:
```suggestion
Splits the display vertically into two panels: 

- Dataset ``A`` is displayed at the **top**
- Dataset ``B`` is displayed at the **bottom**
```

doc/source/applications/compare.rst
```diff
@@ -8,13 +8,62 @@ Purpose
...
+''''''''''''''''''''''''
+
+- Visualizes differences using a yellow/cyan color scheme
+- Provides alternative color mapping for difference visualization
+
+Raw Compare Mode
+''''''''''''''''
+
+- Displays the raw concatenated data without color encoding
+- Shows unprocessed dataset values for direct inspection
```
t20100: What about making this a bit more compact?

t20100 (approved):
Would be good to know what the bug actually is, it's maybe a more general issue with `functools.partial` used with qt signals.

src/silx/gui/widgets/UrlSelectionTable.py
payno: `self._radio.toggled` emits a `boolean` value, but when connected directly to `_IntegratedRadioButton.toggled`, the latter's slot does not expect a `bool` parameter. This causes a type mismatch error at runtime.

Introduce a pass-through wrapper function to bridge the signal connection:
```python
   self._radio.toggled.connect(self._tmpFct)

def self._tmpFct(self, checked):
    self.toggled.emit()
```

src/silx/gui/widgets/UrlSelectionTable.py
payno: This comes from the fact that now we propagate `checked` from `_IntegratedRadioButton.toggled`

doc/source/applications/compare.rst
payno: https://github.com/silx-kit/silx/pull/4627/commits/58c5c5322e915e22486f14b80279e35d78224841 will fix it.

PR #4256, opened 2025-04-04
silx.gui.plot.PlotWidget: Fixed matplotlib backend issue with plot axes limits

t20100:
Checklist:

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

This PR fixes issues & warnings in the PlotWidget with matplotlib backend related to plot axes limits and matplotlib autoscaling.

reverts #4204
related to #4203
closes #4250

PR #4394, opened 2025-09-09
silx.gui.plot.PlotWidget: Allow to invert X axis

loichuder:
Checklist:

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/CONTRIBUTING.rst#pull-request-title-format))

<hr>

For #4253

This PR adds the internal machinery to allow to invert X axis programatically in `PlotWidget` for both backends (see example code below).

GUI control implementation is left to completely handle #4253

## Example code
```python
from silx.gui import qt
from silx.gui.plot.PlotWidget import PlotWidget

app = qt.QApplication()

X = [-4, -3, -2, -1, 0, 1, 2, 3, 4]
Y = [x**3 for x in X]

w = qt.QWidget()
layout = qt.QHBoxLayout(w)


w1 = PlotWidget(backend="mpl")
w1.addCurve(X, Y)
w1.setXAxisInverted(True)
w1.setYAxisInverted(True)

w2 = PlotWidget(backend="gl")
w2.addCurve(X, Y)
w2.setXAxisInverted(True)
w2.setYAxisInverted(True)

layout.addWidget(w1)
layout.addWidget(w2)

w.show()
app.exec()
```

image

Issue #4335, opened 2025-06-03
Double plot with sync axes and aspect ratio gives matplotlib warning


loichuder:
Using the following code stills shows the warning given in #4203

```python
import numpy
from silx.gui import qt
from silx.gui.plot import Plot2D
from silx.gui.plot.utils.axis import SyncAxes

class DoublePlot2D(qt.QWidget):
    def __init__(self, parent: qt.QWidget | None = None) -> None:
        super().__init__(parent=parent)

        self._left_plot = Plot2D(self)
        self._right_plot = Plot2D(self)
        self._left_plot.setKeepDataAspectRatio(True)
        self._right_plot.setKeepDataAspectRatio(True)

        layout = qt.QHBoxLayout()
        layout.addWidget(self._left_plot)
        layout.addWidget(self._right_plot)
        self.setLayout(layout)

        self._x_axis_sync = SyncAxes(
            (self._left_plot.getXAxis(), self._right_plot.getXAxis())
        )
        self.y_axis_sync = SyncAxes(
            (self._left_plot.getYAxis(), self._right_plot.getYAxis())
        )

app = qt.QApplication([])

widget = DoublePlot2D()
widget.show()
data = numpy.random.rand(100).reshape(10, 10)
widget._left_plot.setKeepDataAspectRatio(True)
widget._left_plot.addImage(data)

app.exec()
```

## Relevant versions
- silx 2.2.1
- matplotlib 3.10.3
- PyQt6 6.8.1

loichuder:
Closed by #4628

PR #4628, opened 2026-06-16
silx.gui.plot.backends.BackendMatplotlib: move aspect ratio handling from matplotlib to the backend

### LINKED ISSUE #4588: silx view: Y-axis orientation toolbutton state issue
There is an inconsistency in the state of y-axis orientation in `silx view` image view (at least).
It does not seems to be the case with the x-axis.

## To reproduce

- Download https://www.silx.org/pub/h5web/water_224.h5
- Run `silx view water_224.h5`
- Open `/entry_0000/0_measurement/images` and set "y-axis origin at the top" with the toolbutton

Image

- Double click on `/entry_0000`
- Double click again `/entry_0000/0_measurement/images`: The toolbutton icon is "origin at the top" but the plot is no more.

Image

## Expected behavior

y-axis orientation should be consistent between the toolbutton and the plot and persisted in silx view.

loichuder:

### PR summary

Fix #4588 

Linked to https://github.com/silx-kit/silx/pull/4256 and https://github.com/silx-kit/silx/pull/4204 among others

I tracked down the origin of #4588 to https://github.com/silx-kit/silx/pull/4204. While seemingly unrelated, setting the label width programatically triggers a resize of the plot window with a wrong value of `YAxisInverted`.

@t20100 suspected something fishy with the treatement of the aspect ratio in the matplotlib backend that tries to delegate a bit the aspect ratio but not totally.

In this PR, I move the handling of the aspect ratio **completely** in the backend to not rely at all on matplotlib. For this, I reused the handling of the aspect ratio used in OpenGL by putting it in helper functions.

This seems to have fixed the issue.

<s>Unfortunately, a little problem remains: the first render of the matplotlib figure does not respect the fixed aspect ratio. </s>

**Now fixed:** I forgot to disable autoscaling for X in the last commit

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/backends/BackendBase.py
+++ b/src/silx/gui/plot/backends/BackendBase.py
@@ -60,9 +60,9 @@ class BackendBase:
         """
         self.__xLimits = 1.0, 100.0
         self.__yLimits = {"left": (1.0, 100.0), "right": (1.0, 100.0)}
-        self.__xAxisInverted = True
+        self.__xAxisInverted = False
         self.__yAxisInverted = False
-        self.__keepDataAspectRatio = False
+        self._keepDataAspectRatio = False
         self.__xAxisTimeSeries = False
         self._xAxisTimeZone = None
         # Store a weakref to get access to the plot state.
@@ -523,7 +523,7 @@ class BackendBase:
 
     def isKeepDataAspectRatio(self):
         """Returns whether the plot is keeping data aspect ratio or not."""
-        return self.__keepDataAspectRatio
+        return self._keepDataAspectRatio
 
     def setKeepDataAspectRatio(self, flag):
         """Set whether to keep data aspect ratio or not.
@@ -531,7 +531,7 @@ class BackendBase:
         :param flag:  True to respect data aspect ratio
         :type flag: Boolean, default True
         """
-        self.__keepDataAspectRatio = bool(flag)
+        self._keepDataAspectRatio = bool(flag)
 
     def setGraphGrid(self, which):
         """Set grid.
--- a/src/silx/gui/plot/backends/BackendMatplotlib.py
+++ b/src/silx/gui/plot/backends/BackendMatplotlib.py
@@ -564,9 +564,8 @@ class BackendMatplotlib(BackendBase.BackendBase):
             axis.set_major_formatter(DefaultTickFormatter())
 
         # Autoscale is handled manually by the backend
-        self.ax.set_autoscalex_on(False)
-        self.ax.set_autoscaley_on(False)
-        self.ax2.set_autoscaley_on(False)
+        self.ax.autoscale(enable=False)
+        self.ax2.autoscale(enable=False)
 
         # this works but the figure color is left
         self.ax.set_facecolor("none")
@@ -1216,7 +1215,7 @@ class BackendMatplotlib(BackendBase.BackendBase):
         return self.ax.get_xbound()
 
     def setGraphXLimits(self, xmin: float, xmax: float):
-        self._setPlotBounds(xRange=(xmin, xmax))
+        self._setPlotBounds(xRange=(xmin, xmax), keepDim="x")
 
     def getGraphYLimits(self, axis: Literal["left", "right"]) -> Range | None:
         assert axis in ("left", "right")
@@ -1229,15 +1228,16 @@ class BackendMatplotlib(BackendBase.BackendBase):
 
     def setGraphYLimits(self, ymin: float, ymax: float, axis: Literal["left", "right"]):
         if axis == "right":
-            self._setPlotBounds(y2Range=(ymin, ymax))
+            self._setPlotBounds(y2Range=(ymin, ymax), keepDim="y")
         else:
-            self._setPlotBounds(yRange=(ymin, ymax))
+            self._setPlotBounds(yRange=(ymin, ymax), keepDim="y")
 
     def _setPlotBounds(
         self,
         xRange: Range | None = None,
         yRange: Range | None = None,
         y2Range: Range | None = None,
+        keepDim: Literal["x", "y"] | None = None,
     ):
         # Keep data aspect ratio
         if xRange is None:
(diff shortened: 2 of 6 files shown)


t20100:
Thanks for digging into this :muscle:

src/silx/gui/plot/backends/BackendBase.py
```diff
@@ -60,9 +60,9 @@ def __init__(self, plot, parent=None):
         """
         self.__xLimits = 1.0, 100.0
         self.__yLimits = {"left": (1.0, 100.0), "right": (1.0, 100.0)}
-        self.__xAxisInverted = True
```
loichuder, line 63: I don't see why it should be `True` by default. Oversight from me in https://github.com/silx-kit/silx/pull/4394/

src/silx/gui/plot/backends/BackendBase.py
```diff
@@ -60,9 +60,9 @@ def __init__(self, plot, parent=None):
...
+        self.__xAxisInverted = False
         self.__yAxisInverted = False
-        self.__keepDataAspectRatio = False
+        self._keepDataAspectRatio = False
```
loichuder, line 65: Making this slightly less private to reuse it in both backend subclasses

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -519,13 +522,6 @@ class BackendMatplotlib(BackendBase.BackendBase):
     def __init__(self, plot, parent=None):
         super().__init__(plot, parent)
 
-        # matplotlib is handling keep aspect ratio at draw time
-        # When keep aspect ratio is on, and one changes the limits and
-        # ask them *before* next draw has been performed he will get the
-        # limits without applying keep aspect ratio.
-        # This attribute is used to ensure consistent values returned
-        # when getting the limits at the expense of a replot
-        self._dirtyLimits = True
```
loichuder, line 528: As the comment indicates, `_dirtyLimits` was used to force a replot so that matplotlib could apply the aspect ratio for us.

Since we now handle the aspect ratio completely, the attribute and the related code can be removed.

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -1200,74 +1197,71 @@ def setGraphYLabel(self, label, axis):
 
     # Graph limits
 
-    def _setXLimits(self, xmin: float, xmax: float):
```
loichuder, line 1203: Replaced by `set_xbound` to be consistent with y axis

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -1200,74 +1197,71 @@ def setGraphYLabel(self, label, axis):
...
-            self._setXLimits(xcenter - 0.5 * newXRange, xcenter + 0.5 * newXRange)
-
-        ax.set_ybound(ymin, ymax)
+    def setGraphYLimits(self, ymin: float, ymax: float, axis: Literal["left", "right"]):
+        if axis == "right":
+            self._setPlotBounds(y2Range=(ymin, ymax))
+        else:
+            self._setPlotBounds(yRange=(ymin, ymax))
 
+    def _setPlotBounds(
```
loichuder, line 1235: New method that handles the aspect ratio.

All limit changes now go through this method.

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -1358,13 +1327,6 @@ def isXAxisInverted(self) -> bool:
     def isYRightAxisVisible(self):
         return self.ax2.yaxis.get_visible()
 
-    def isKeepDataAspectRatio(self):
-        return self.ax.get_aspect() in (1.0, "equal")
-
-    def setKeepDataAspectRatio(self, flag):
-        self.ax.set_aspect(1.0 if flag else "auto")
-        self.ax2.set_aspect(1.0 if flag else "auto")
```
loichuder, line 1366: Moved to `BackendBase`

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -1505,8 +1467,6 @@ def __init__(self, plot, parent=None):
         FigureCanvasQTAgg.__init__(self, self.fig)
         self.setParent(parent)
 
-        self._limitsBeforeResize = None
```
loichuder, line 1508: Same as `_dirtyLimits`. There was some conditional logic to force the aspect ratio when resizing.

No longer needed.

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -1626,19 +1586,10 @@ def pickItem(self, x, y, item):
...
-            self.ax.get_ybound(),
-            self.ax2.get_ybound(),
-        )
-        self.ax.set_autoscaley_on(True)
-        self.ax2.set_autoscaley_on(True)
-
         FigureCanvasQTAgg.resizeEvent(self, event)
-        if self.isKeepDataAspectRatio() or self._hasOverlays():
-            # This is needed with matplotlib 1.5.x and 2.0.x
-            self._plot._setDirtyPlot()
```
loichuder, line 1641: No longer needed

src/silx/gui/plot/backends/BackendMatplotlib.py
```diff
@@ -567,8 +563,9 @@ def __init__(self, plot, parent=None):
         for axis in (self.ax.yaxis, self.ax.xaxis, self.ax2.yaxis, self.ax2.xaxis):
             axis.set_major_formatter(DefaultTickFormatter())
 
-        self.ax.set_autoscaley_on(False)
-        self.ax2.set_autoscaley_on(False)
+        # Autoscale is handled manually by the backend
+        self.ax.autoscale(enable=False)
+        self.ax2.autoscale(enable=False)
```
loichuder, line 568: This disables autoscale and the automatic margins added by matplotlib when plotting a new element or changing scales: https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.autoscale.html#matplotlib.axes.Axes.autoscale

loichuder:
I added a new test for #4335 

Unfortunately, this broke the `ruff check`. I'll push a fix a the next rebase.

t20100 (approved):
LGTM, thanks a lot for digging into this, this is much simpler now!

PR #4629, opened 2026-06-17
silx.gui.plot.StackView: deprecated getStack and getCurrentView

### LINKED ISSUE #627: [plot] Make StackView getters return objects
StackWidget methods getStack and getCurrentView returns a tuple (data, dict of params).
To follow Plot API refactoring, those methods should return an object with getters... or return just the data and make new methods in StackWidget to access information stored in params (e.g., getStackColormap, getStackScale, ...).

loichuder:

### PR summary

Fix https://github.com/silx-kit/silx/issues/627

After thinking a bit on how to avoid breaking the API, I thought that the best way forward was to deprecate `getStack` and `getCurrentView` and provide replacements that don't return the parameter dict.

The unfortunate side-effect is that we can't reuse the same method names for the replacements. I therefore settled for `getStackData` and `getCurrentStackData`. 

So far, I did not add additional getters for other parameters since we only used `colormap` that is accessible via `getColormap`. We'll have to see if more are needed.

I updated the tests to use the new methods. I plan to refactor them and migrate them to pytest in a follow-up MR.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/StackView.py
+++ b/src/silx/gui/plot/StackView.py
@@ -70,11 +70,14 @@ __authors__ = ["P. Knobel", "H. Payno"]
 __license__ = "MIT"
 __date__ = "10/10/2018"
 
+from typing import Any
+
 import numpy
 import logging
 
 import silx
 from silx.gui import qt
+from silx.utils.deprecation import deprecated
 from .. import icons
 from . import items, PlotWindow, actions
 from .items.image import ImageStack
@@ -326,7 +329,7 @@ class StackView(qt.QMainWindow):
         return save_NXdata(
             filename,
             nxentry_name=entryPath,
-            signal=self.getStack(copy=False, returnNumpyArray=True)[0],
+            signal=self.getData(copy=False, returnNumpyArray=True),
             signal_name="image_stack",
         )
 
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/StackView.py
```diff
@@ -625,29 +628,14 @@ def setStack(self, stack, perspective=None, reset=True, calibrations=None):
...
-            returnNumpyArray is True, a copy will be made anyway.
-        :param bool returnNumpyArray: If True, the returned object is
-            guaranteed to be a numpy array.
-        :return: 3D stack and parameters.
-        :rtype: (numpy.ndarray, dict)
-        """
-        if self._stack is None:
-            return None
-
+    def _getParams(self) -> dict[str, Any]:
```
loichuder, line 631: Using this new helper in both `getStack` and `getCurrentView`

payno:
And what about `getData and getCurrentData` instead of `getStackData` and `getStack` ? This might be a bit more on the today `silx` style.

loichuder:
> And what about `getData and getCurrentData` instead of `getStackData` and `getStack` ? This might be a bit more on the today `silx` style.

Sounds good to me as well!

src/silx/gui/plot/StackView.py
maxenceprog: why this refacto if we are going to deprecated both functions ?

src/silx/gui/plot/test/test_stackview.py
```diff
@@ -246,14 +247,16 @@ def tearDown(self):
...
-        self.assertEqual(params["colormap"]["name"], "viridis")
+
+        colormap = self.stackview.getColormap()
+        self.assertEqual(colormap["name"], "viridis")
 
     def testSetStackPerspective(self):
         self.stackview.setStack(self.mystack, perspective=1)
-        my_trans_stack, params = self.stackview.getCurrentView()
+        my_trans_stack = self.stackview.getCurrentStackData()
         # get stack returns the transposed data, depending on the perspective
```
maxenceprog, line 260:
```suggestion
        # `getCurrentStackData` returns the transposed data, depending on the perspective
```

src/silx/gui/plot/test/test_stackview.py
maxenceprog, line 131:
```suggestion
        )  # returnNumpyArray=False by default in `getStackData`
```

maxenceprog (commented):
One last thing i have in mind (but maybe not relevant)

`getCurrentView` still exist in `DataView`

is it weird in user side to use getCurrentView in DataView

and getCurrentStackData in  stackview ?

src/silx/gui/plot/StackView.py
loichuder: Cause I have OCD

loichuder: That fixes also an inconsistency between `getStack` and `getCurrentView`: the case where that `image` is not a `items.ColormapMixIn` was handled in `getCurrentView` and not `getStack`.

With the refacto, both are now consistent.

loichuder:
> One last thing i have in mind (but maybe not relevant)
> 
> `getCurrentView` still exist in `DataView`
> 
> is it weird in user side to use getCurrentView in DataView
> 
> and getCurrentStackData in stackview ?

It is relevant. This is actually why I didn't keep the `View` word for the new method: these are two different concepts:
- `getCurrentView` of `DataView` is to get the current visualisation (Image, Curve, ...) in `silx view`
- `getCurrentView` of `StackView` was to get the data as it is displayed.

With the new name `getCurrentStackData` (or rather `getCurrentData` with Henri's suggestion), we remove the ambiguity.

src/silx/gui/plot/test/test_stackview.py
loichuder: Thanks for the comments!

I won't apply them right away as I plan to tackle the tests in a follow-up MR. I'll ask you for a review then :wink:

src/silx/gui/plot/StackView.py
payno, line 631: I have a doubt about the scope of this function: Should this be public or not ?

At the moment the `params` seems not used anywhere (if I am correct) **internally**. But it might be used elsewhere. How much are we sure this is 'useless'.
I guess this API was done on-purpose and with the new API there is no access to this information for users.
So I would have tendency to make it public. What do you think @t20100 do you know a place where those params could be used ?

payno (commented):
looks good, thanks for the modernization ;)
Except for the API scope of `_getParams` I am fine

src/silx/gui/plot/StackView.py
loichuder: The idea of the deprecation warning is to give time for consumers of `getStack` to report upstream if something else is needed.

I'd prefer that we use individual getters to expose what it needed based on the feedback we get rather than exposing `getParams`

loichuder: I can't find any use of `getStack` in GitHub 🤷

t20100: If there is no obvious need for it, I would keep it private.

BTW, `StackView.getActiveImage` returns the image item from which you can get the info

t20100 (approved):

PR #4630, opened 2026-06-17
Remove silx.gui.plot._utils.timestamp and future annotations

### LINKED ISSUE #4235: Clean-up code base from code supporting Python<3.10
PR #4234 drops supports of Python<3.10.
Let's check the code base and remove compatibility code for Python<3.10 support.

loichuder:

### PR summary

Fix #4235 

- [silx.gui.plot._utils: Remove timestamp function](https://github.com/silx-kit/silx/commit/be6eb315e10db12c80b93ca436390a2aa0ea7910) Could have been removed earlier
- [Remove from __future__ import annotations](https://github.com/silx-kit/silx/commit/2471a2f15c4573e3de38420dc7ec71541617ca04)

There is still some mention to Python 2.7 in very specific code (e.g. `EdfFile`). To be discussed if we want to remove it.

`from __future__ import annotations` is actually still needed if the class name is used as a type hint before the class is defined (forward references, present in e.g. `colors`, `PlotWidget`). This is because the lazy evaluation of annotations is only implemented in Python 3.14 (https://peps.python.org/pep-0749/), even though the `from __future__ import annotations` behaviour dates as far as Python 3.7 (https://peps.python.org/pep-0563/).

Instead of keeping `from __future__ import annotations`  in only a few files, I converted the forward references to string annotations (doing explicitely what `from __future__` is doing) where needed.

#### AI Disclosure
- [x] I chatted with Claude to build an understanding of the need of the  `from __future__ import annotations`

### DIFF
--- a/src/silx/gui/plot/_utils/dtime_ticklayout.py
+++ b/src/silx/gui/plot/_utils/dtime_ticklayout.py
@@ -35,9 +35,7 @@ import datetime as dt
 import enum
 import logging
 import math
-import time
 
-import dateutil.tz
 
 from dateutil.relativedelta import relativedelta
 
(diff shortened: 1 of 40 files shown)


src/silx/gui/plot/_utils/dtime_ticklayout.py
```diff
@@ -54,49 +52,6 @@
 SECONDS_PER_MONTH_AVERAGE = SECONDS_PER_YEAR / 12  # Seconds per average month
 
 
-# No dt.timezone in Python 2.7 so we use dateutil.tz.tzutc
-_EPOCH = dt.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc())
-
-
-def timestamp(dtObj):
```
loichuder, line 61: `timestamp` is available on datetime objects since Python 3.3: https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

src/silx/gui/widgets/StackedProgressBar.py
```diff
@@ -67,7 +66,7 @@ class StackedProgressBar(qt.QProgressBar):
     Multiple stacked progress bar in single component
     """
 
-    def __init__(self, parent: qt.Qwidget | None = None):
+    def __init__(self, parent: qt.QWidget | None = None):
```
loichuder, line 69: Removing `from __future__ import annotations` allowed to catch this typo in the type hint

t20100 (approved):
LGTM, thanks for the clean-up

PR #4631, opened 2026-06-17
silx.gui.plot.PlotWidget: Allow directive zoom with shortcut ALT and SHIFT like h5web

maxenceprog:

### PR summary

Add 2 short cut for directive zoom.

Alt for X and Shift for Y (like h5web)

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/backends/BackendOpenGL.py
+++ b/src/silx/gui/plot/backends/BackendOpenGL.py
@@ -321,7 +321,7 @@ class BackendOpenGL(BackendBase.BackendBase, glu.OpenGLWidget):
         event.accept()
 
     def wheelEvent(self, event):
-        delta = event.angleDelta().y()
+        delta = event.angleDelta().x() + event.angleDelta().y()
         angleInDegrees = delta / 8.0
         x, y = qt.getMouseEventPosition(event)
         self._plot.onMouseWheel(x, y, angleInDegrees)
--- a/src/silx/gui/plot/backends/BackendOpenGL.py
+++ b/src/silx/gui/plot/backends/BackendOpenGL.py
@@ -321,7 +321,14 @@ class BackendOpenGL(BackendBase.BackendBase, glu.OpenGLWidget):
         event.accept()
 
     def wheelEvent(self, event):
-        delta = event.angleDelta().x() + event.angleDelta().y()
+        delta = event.angleDelta().y()
+        #  https://github.com/qt/qtbase/blob/120883a028a59864dc7691dd1efa318bd602755d/src/plugins/platforms/xcb/qxcbwindow.cpp#L1955-L1956
+        # qt xcb plugin put angleDelta in x rather than in y when alt modifier is pressed.
+        # matplotlib/matplotlib#25671 marked as 'won't fix'
+        # Fix here as we want to be able to use alt modifier in silx plot.
+        # https://github.com/silx-kit/silx/pull/4631
+        if event.angleDelta().y() == 0:
+            delta = event.angleDelta().x()
         angleInDegrees = delta / 8.0
         x, y = qt.getMouseEventPosition(event)
         self._plot.onMouseWheel(x, y, angleInDegrees)
--- a/src/silx/gui/plot/tools/menus.py
+++ b/src/silx/gui/plot/tools/menus.py
@@ -52,9 +52,9 @@ class ZoomEnabledAxesMenu(qt.QMenu):
         self.__plotRef = weakref.ref(plot)
 
         self.addSection("Enabled axes")
-        self.__xAxisAction = qt.QAction("X axis", parent=self)
-        self.__yAxisAction = qt.QAction("Y left axis", parent=self)
-        self.__y2AxisAction = qt.QAction("Y right axis", parent=self)
+        self.__xAxisAction = qt.QAction("X axis (alt)", parent=self)
+        self.__yAxisAction = qt.QAction("Y left axis (shift)", parent=self)
+        self.__y2AxisAction = qt.QAction("Y right axis (shift)", parent=self)
 
         for action in (self.__xAxisAction, self.__yAxisAction, self.__y2AxisAction):
             action.setCheckable(True)
--- a/src/silx/gui/plot/tools/menus.py
+++ b/src/silx/gui/plot/tools/menus.py
@@ -52,9 +52,9 @@ class ZoomEnabledAxesMenu(qt.QMenu):
         self.__plotRef = weakref.ref(plot)
 
         self.addSection("Enabled axes")
-        self.__xAxisAction = qt.QAction("X axis (alt)", parent=self)
-        self.__yAxisAction = qt.QAction("Y left axis (shift)", parent=self)
-        self.__y2AxisAction = qt.QAction("Y right axis (shift)", parent=self)
+        self.__xAxisAction = qt.QAction("X axis (Alt+Wheel)", parent=self)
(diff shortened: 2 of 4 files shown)


src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -323,7 +323,7 @@ def mouseReleaseEvent(self, event):
         event.accept()
 
     def wheelEvent(self, event):
-        delta = event.angleDelta().y()
+        delta = event.angleDelta().x() + event.angleDelta().y()
```
maxenceprog: when alt is pressed in opengl, event.angleDelta().y() has a value and event.angleDelta().x() is 0

loichuder:
@maxenceprog Thanks for the PR! Could you change the name to add the full path of `PlotWidget` ? See other PR for example: https://github.com/silx-kit/silx/pull/4626

src/silx/gui/plot/tools/menus.py
```diff
@@ -52,9 +52,9 @@ def __init__(self, plot: PlotWidget, parent: qt.QWidget | None = None):
         self.__plotRef = weakref.ref(plot)
 
         self.addSection("Enabled axes")
-        self.__xAxisAction = qt.QAction("X axis", parent=self)
-        self.__yAxisAction = qt.QAction("Y left axis", parent=self)
-        self.__y2AxisAction = qt.QAction("Y right axis", parent=self)
+        self.__xAxisAction = qt.QAction("X axis (alt)", parent=self)
+        self.__yAxisAction = qt.QAction("Y left axis (shift)", parent=self)
+        self.__y2AxisAction = qt.QAction("Y right axis (shift)", parent=self)
```
payno: For me here there is clear need to clarify the policy we want to have.
I think this is not something we want to "impose" at the library side but to have at `silx view` side right @t20100 ?

So from my understanding either this is added at the silx view application level and/or this should be optional at the lib side.

src/silx/gui/plot/backends/BackendOpenGL.py
payno: I would expect x to always be `0` except for fancy mouses no ? According to [the doc](https://doc.qt.io/qt-6/qwheelevent.html#angleDelta-prop)

And even if x is not zero I am not sure we want to take it into account.

> Discussed with @maxenceprog having `event.angleDelta().x() != 0` comes from the OpenGL backend.

maxenceprog: sorry my comment was wrong when alt is pressend in open gl  x value is set else y


that's why. DO you want me to be more specific like really use x when alt is pressed ?

src/silx/gui/plot/tools/menus.py
maxenceprog: i think if we add this shortcut it can benefit to library too ?

loichuder: > I think this is not something we want to "impose" at the library side but to have at silx view side right @t20100 ?

Agreed. It should not enabled by default in the lib but be in opt-in. And `silx view` would opt-in.

> i think if we add this shortcut it can benefit to library too ?

Yes, it would be good to still have in the library so that other consumer apps can enable it. But not enabled by default.

src/silx/gui/plot/backends/BackendOpenGL.py
t20100: This behavior is hard-coded in qt xcb plugin: https://github.com/qt/qtbase/blob/120883a028a59864dc7691dd1efa318bd602755d/src/plugins/platforms/xcb/qxcbwindow.cpp#L1955-L1956

Rather than summing, it's also possible to use  `x()` only when `y() == 0`

Since this is not documented in qt doc, it's worth a comment to explain!

src/silx/gui/plot/tools/menus.py
t20100: So far, plot interactions are mostly wired in `PlotWidget` in `Plotinteraction` and associated tools through `PlotWidget.setInteractiveMode`.
It's possible to add an API to toggle use of modifer keys, but I'm not sure it is worth it here.

I agree silx should leave it up to the application to choose it's keyboard short-cuts (unless they are standard ones). However here it is about modifier keys for the mouse wheel which is a behavior I don't think one cannot infer with from the API.

t20100: `(alt)` -> `(Alt+Wheel)`
`(shift)` -> `(Shift+Wheel)`

maxenceprog: i m not sure i understand the final decision @t20100 . So i can keep thit like that and let it in silx core ?

t20100: Unless someone disagree, I would leave it as it is in silx core.

src/silx/gui/plot/backends/BackendOpenGL.py
maxenceprog: hey i'm facing a new issue actually i have the same problem with matplotlib backend but i have no clue how i solve it as when alt pressed,  _onMouseWheel is just not send at all...

(you can try with current version of silx by the way and xcb plugin)

should i open an issue an matplotlib ?

or give up alt and use an other modifier ?

maxenceprog: I look at the problem with Loic

a solution may be to override silx.gui.plot.backends.BackendMatplotlibQt.wheelEvent

and do the same as we do for opengl

Do you want to implement that @t20100  or do you prefer to use another modifier (like ctrl) ?

t20100: IMO best to be consistent with h5web modifier keys if possible.

Provided https://github.com/matplotlib/matplotlib/issues/25671 is marked as 'won't fix', it indeed requires to override matplotlib `wheelEvent`.
BTW going this way, we prevent silx to support horizontal scrolling in the future without changing this (one of the point that made matplotlib not patching this Qt behavior).

What do you think?

loichuder: > and do the same as we do for opengl

I thought it was a behaviour specific to matplotlib so why should something be done for opengl?

>  we prevent silx to support horizontal scrolling in the future 

Only in the visualization area. And resizing this one does not make a scroll bar appear since the figure has no minimum size (as far as I can see).

t20100: The behavior of using `Alt` modifier to change x/y of [`QWheelEvent.angleDelta`](https://doc.qt.io/qt-6/qwheelevent.html#angleDelta-prop) is in [Qt xcb plugin](https://github.com/qt/qtbase/blob/120883a028a59864dc7691dd1efa318bd602755d/src/plugins/platforms/xcb/qxcbwindow.cpp#L1955-L1956).

The [OpenGL backend was doing the same](https://github.com/silx-kit/silx/blob/66212df92fe8eee39bbd7cbe6c90399bd1f3bf56/src/silx/gui/plot/backends/BackendOpenGL.py#L323) as [what matplotlib does](https://github.com/matplotlib/matplotlib/blob/1cba20781634880c7b1327fb54033d9d71c6aeb2/lib/matplotlib/backends/backend_qt.py#L364).
So change to support `Alt+Wheel` is required in both or we have to use another key.

> >    we prevent silx to support horizontal scrolling in the future
>
> Only in the visualization area.

Yes.

t20100 (approved):
LGTM, thanks for the feature!

Once #4633 is merged too, we may need to have a second look over it all: alt/shift wheel modifier, axis autoscale and zoom options since it all takes care of similar features.

PR #4632, opened 2026-06-17
silx.gui.plot.MaskTools: add signals when the file is loaded and saved.

### LINKED ISSUE #649: [mask] Expected events
To be able to have a nice integration of the mask cool in pyFAI calibration i need to have few signals.

- `sigMaskSaved(filename)`
- `sigMaskLoaded(filename)`
- `sigChanged`

Right now i only use change event from a protected API: `masktoolswidget._mask.sigChanged`.

payno:
### PR summary

Closes #649 

I felt ambitious for the end of the day. It seems to match what is done at pyFAI side.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/ScatterMaskToolsWidget.py
+++ b/src/silx/gui/plot/ScatterMaskToolsWidget.py
@@ -425,7 +425,9 @@ class ScatterMaskToolsWidget(BaseMaskToolsWidget):
             msg = "Extension '%s' is not supported."
             raise RuntimeError(msg % extension)
 
-        self.setSelectionMask(mask, copy=False)
+        effectiveMaskShape = self.setSelectionMask(mask, copy=False)
+        if effectiveMaskShape is not None:
+            self.sigMaskLoaded.emit(filename)
 
     def _loadMask(self):
         """Open load mask dialog"""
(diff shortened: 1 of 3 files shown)


src/silx/gui/plot/ScatterMaskToolsWidget.py
```diff
@@ -463,6 +463,8 @@ def _loadMask(self):
             msg.setIcon(qt.QMessageBox.Critical)
             msg.setText("Cannot load mask from file. " + message)
             msg.exec()
+        else:
+            self.sigMaskLoaded.emit(filename)
```
t20100: why not doing this in the `load` method?

payno: You are absolutely right. I guess I need 👓 my bad 😢

loichuder: That's why you should not feel too ambitious at the end of the day :rofl:

payno: Exactly I will remember, next it will be :arrow_forward:  :house:

t20100 (approved):
LGTM

PR #4633, opened 2026-06-17
silx.gui.plot.PlotWidget: fix resetting zoom for curves/scatters to take the visible data in the X-range into acount

### LINKED ISSUE #179: [Plot] When autoscale is disabled on an axis, reset zoom should only consider
When autoscale is disabled on an axis, reset zoom should only consider the visible part of the data on this axis for resizing the other axis.

woutdenolf:

### PR summary

- `Item.getBounds()` now returns a named tuple `Bounds` instead of `tuple`
- `Item.getResetBounds()` introduced for the bounds used when resetting the zoom: fixing one axis has an effect on autoscaling of another (in case of dependent vs. independent variable, so not images).
- Refactor `PlotWidget`:
  -  `_forceResetZoom` calls `resetZoom`
  - `resetZoom` gets optional arguments `force` and `notify`
  - `resetZoom` no longer fixed axis limits because `Item.getResetBounds()` respects the limits (as opposed to `Item.resetBounds()` which was used perviously)
  - `_PlotDataRange` -> `PlotDataRange` because this type is exposed through public functions.
- Refactor `Item`:
  - Introduce helper `_getYAxisInstance` (for left vs. right)
  - Introduce helper `_getXAxisInstance` (for analogy)
  - Introduce helper `_getAxisInstances` (both X and Y)
 
#### AI Disclosure
- [x] AI tool(s) chatgpt used for adding the fix to PlotWidget based on https://silx.readthedocs.io/en/latest/_modules/silx/gui/plot/PlotWidget.html#PlotWidget

#### Before this PR: reset zoom takes entire X-range to determine Y-range

before

#### After this PR: reset zoom takes current X-range to determine Y-range

after

#### Code for the example

```python
import sys
import numpy

from silx.gui import qt
from silx.gui.plot import PlotWindow

app = qt.QApplication(sys.argv)

plot = PlotWindow()

x = numpy.linspace(0, 100, 5000)

y_left_main = numpy.sin(x * 0.2)
y_left_main += 5 * numpy.exp(-0.5 * ((x - 50) / 0.5) ** 2)

y_right = 0.5 * numpy.cos(x * 0.2)
y_right += 200 * numpy.exp(-0.5 * ((x - 85) / 0.2) ** 2)

y_left_extra = 0.3 * numpy.cos(x * 0.5)
y_left_extra += 8 * numpy.exp(-0.5 * ((x - 50) / 0.4) ** 2)

plot.addCurve(x, y_left_main, legend="left_main", yaxis="left")
plot.addCurve(x, y_left_extra, legend="left_extra", yaxis="left")
plot.addCurve(x, y_right, legend="right_hidden_spike", yaxis="right")

plot.show()

print("")
print("=== REPRO STEPS (#179) ===")
print("1. Zoom X to [40, 60]")
print("2. Disable X autoscale (lock X axis)")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- Y limits based ONLY on visible region (x in [40, 60])")
print("- Spike at x≈85 must NOT influence autoscale")
print("")
print("BUG (before fix):")
print("- Y axis explodes due to hidden spike at x≈85")
print("")

sys.exit(app.exec())
```


src/silx/gui/plot/PlotWidget.py
```diff
@@ -3304,7 +3364,11 @@ def _forceResetZoom(
             If None (the default), use margins from :meth:`getDataMargins`.
         """
         # Get data range
-        ranges = self.getDataRange()
+        if self._xAxis.isAutoScale():
+            ranges = self.getDataRange()
+        else:
+            ranges = self.getVisibleDataRange()
```
woutdenolf: Only take visible ranges into account when X-axis is not auto scaled.

src/silx/gui/plot/items/axis.py
```diff
@@ -601,4 +601,4 @@ def setAutoScale(self, flag: bool = True):
     @docstring(Axis)
     def _getDataRange(self) -> tuple[float, float] | None:
         ranges = self._getPlot().getDataRange()
-        return ranges.y2
+        return ranges.y
```
woutdenolf: Bug that became visible when implementing this feature.

src/silx/gui/plot/PlotWidget.py
```diff
@@ -892,6 +892,66 @@ def getDataRange(self) -> _PlotDataRange:
             self._updateDataRange()
         return self._dataRange
 
+    def getVisibleDataRange(self) -> _PlotDataRange:
```
woutdenolf: Name of the function `getVisibleDataRange` is technically not correct.

When the X-axis is fixed and we reset the zoom: the `Y` limits span the data range of `Y[mask]` with `mask = X>=Xmin & X<=Xmax` instead of `Y[:]`. That in fact the entire point of this PR.

When the Y-axis is fixed and we reset the zoom: the `X` limits span the data range of `X[:]` and not `X[mask]` with `mask = Y>=Ymin & Y<=Ymax`. Of course we _could_ do that as well like we do for Y. Should we?

So perhaps this function should be private. And the name should be something like `_getAutoscaleDataRange`? I don't like that name either.

woutdenolf: Especially for scatter plots we might want to have a fix Y axis to affect the reset of the X axis. This was not part of the original issue but I think it makes sense.

woutdenolf: In addition instead of doing the logic here

```python
if isinstance(item, items.PointsBase):
    ...
else:
    bounds = item.getBounds()
    if bounds is None:
        continue
```

we might want to move it to the Item classes, implement it for PointsBase and fallback to getBounds by default so that

```python
bounds = item.getAutoScaleBounds()
if bounds is None:
    continue
```

src/silx/gui/plot/PlotWidget.py
```diff
@@ -892,6 +892,66 @@ def getDataRange(self) -> _PlotDataRange:
...
+        xMin, xMax = xAxis.getLimits()
+
+        xMinOut = numpy.inf
+        xMaxOut = -numpy.inf
+
+        yMinLeft = numpy.inf
+        yMaxLeft = -numpy.inf
+
+        yMinRight = numpy.inf
+        yMaxRight = -numpy.inf
```
payno: `_updateDataRange` uses `float("nan")` instead of numpy.inf. 

Using `numpy.Inf` seems more logic on a first approach. But all the functions are using `nanmin`, `nanmax`... Under the wood I guess `numpy.isfinite` is used anyway but this require a bit more knowledge to understand.

And I think if you have an empty plot you can end up with `_PlotDataRange` containing `numpy.inf` values. Not sure this is expected / handle downstream.

Another question I have is: this function looks **very close** to the `_updateDataRange(self):` function. Could the two be merged or a part reused ?

woutdenolf: Ok. I will fix while I'm moving the logic to the Items like proposed in https://github.com/silx-kit/silx/pull/4633#discussion_r3432887848

src/silx/gui/plot/items/axis.py
t20100: Should be:
```suggestion
        return ranges.yright
```

https://github.com/silx-kit/silx/blob/da89bb416f348ba779c02f13eb67bfa232b6091b/src/silx/gui/plot/PlotWidget.py#L84

src/silx/gui/plot/PlotWidget.py
t20100: +1 for private

t20100: +1 for per-Item implementation

woutdenolf:
I added `Item.getAutoScaleBounds` in analogy to `Item.getBounds`. Then `PlotWidget` uses `getAutoScaleBounds` when resetting the zoom with at least one fixed axis.

Another example, this time a scatter plot:

Peek 2026-06-18 11-15


```python
import sys
import numpy

from silx.gui import qt
from silx.gui.plot import PlotWindow


app = qt.QApplication(sys.argv)

plot = PlotWindow()

# visible clusters
x = 50 + 7 * numpy.random.randn(80)
y = 1.0 + 5 * numpy.random.randn(80)
values = numpy.random.rand(len(x))
item = plot.addScatter(x, y, values, legend="A", colormap=Colormap("red"))

x = 65 + 2 * numpy.random.randn(80)
y = 2.0 + 10 * numpy.random.randn(80)
values = numpy.random.rand(len(x))
item = plot.addScatter(x, y, values, legend="B", colormap=Colormap("green"))

# offscreen cluster after zoom
x = 85 + 3 * numpy.random.randn(40)
y = 50 + 2 * numpy.random.randn(40)
values = numpy.random.rand(len(x))
item = plot.addScatter(x, y, values, legend="C", colormap=Colormap("blue"))

plot.resetZoom()

plot.show()
sys.exit(app.exec())
```

src/silx/gui/plot/test/test_bounds.py
```diff
@@ -0,0 +1,237 @@
...
+    for x, y, item in items:
+        expected = _expected_autoscale_bounds(
+            x,
+            y,
+            fixed_xmin=fixed_xmin,
+            fixed_xmax=fixed_xmax,
+            fixed_ymin=fixed_ymin,
+            fixed_ymax=fixed_ymax,
+        )
+        print(item.getName(), expected)
```
t20100:
```suggestion
```

t20100 (approved):
Fix in YRightAxis._getDataRange should be ranges.yright instead of ranges.y, otherwise LGTM!

Haven't saw the draft mode, removing approval

src/silx/gui/plot/test/test_bounds.py
woutdenolf: I decided to go the full way:

- do it for both X and Y
- move the logic to the items: currently called `getResetBounds` in analogy to `getBounds`, takes limits of fixed axes into account when determining the limits of autoscaling axes as opposed to be the full data bounds.
- introduced names tuple `Bounds` returned by `getBounds` and `getResetBounds` (funky stuff with NaN and None).
- add unit tests

I did `PointsBase` -> curves, scatters, ...

I'm now doing `Histogram`, `ImageBase`, `BoundingRect` and `_BaseExtent` (the last one I'm not sure).

woutdenolf:
Another example for histograms

Peek 2026-06-18 20-40

```python
import sys
import numpy

from silx.gui import qt
from silx.gui.plot import PlotWindow

app = qt.QApplication(sys.argv)

plot = PlotWindow()

rng = numpy.random.default_rng(42)

# Main population around x=50 (visible after zoom)
data_visible = rng.normal(loc=50, scale=2, size=5000)

# Strong population around x=85 (outside zoom window)
data_hidden = rng.normal(loc=85, scale=0.5, size=30000)

plot.addHistogram(
    histogram=numpy.histogram(data_visible, bins=80)[0],
    edges=numpy.histogram(data_visible, bins=80)[1],
    legend="visible_peak",
)

plot.addHistogram(
    histogram=numpy.histogram(data_hidden, bins=80)[0],
    edges=numpy.histogram(data_hidden, bins=80)[1],
    legend="hidden_peak",
)

plot.show()

print("")
print("=== HISTOGRAM RESETZOOM TEST ===")
print("1. Zoom X to [40, 60]")
print("2. Disable X autoscale")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- Y autoscale uses only visible histogram bins")
print("- Peak around x≈85 must not contribute")
print("")

sys.exit(app.exec())
```

woutdenolf:
Another example for images

Peek 2026-06-18 20-50


```python
import sys
import numpy

from silx.gui import qt
from silx.gui.plot import PlotWindow

app = qt.QApplication(sys.argv)

plot = PlotWindow()

image = numpy.random.normal(size=(300, 300))

# bright hotspot in center
image[140:160, 140:160] += 10

# huge hotspot far away
image[250:260, 250:260] += 100

plot.addImage(image)

plot.show()

print("")
print("=== IMAGE AUTOSCALE TEST ===")
print("1. Zoom around the center hotspot")
print("2. Disable X autoscale")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- Y limits follow visible image region")
print("- Far hotspot should not affect autoscale")
print("")

sys.exit(app.exec())
```

woutdenolf:
Another example for BoundRect (not sure how to visualize)

```python
import sys

from silx.gui import qt
from silx.gui.plot import PlotWindow
from silx.gui.plot.items.shape import BoundingRect

app = qt.QApplication(sys.argv)

plot = PlotWindow()

# Visible bounding region
rect_visible = BoundingRect()
rect_visible.setBounds((45, 55, 10, 20))
plot.addItem(rect_visible)

# Hidden bounding region
rect_hidden = BoundingRect()
rect_hidden.setBounds((80, 90, 100, 200))
plot.addItem(rect_hidden)

plot.show()

print("")
print("=== BOUNDINGRECT RESETZOOM TEST ===")
print("1. Zoom X to [40, 60]")
print("2. Disable X autoscale")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- Y limits become [10, 20]")
print("- BoundingRect at x=[80,90] must not contribute")
print("")

sys.exit(app.exec())
```

woutdenolf:
Another example for axis extents (not sure how to visualize)

```python
import sys

from silx.gui import qt
from silx.gui.plot import PlotWindow
from silx.gui.plot.items import XAxisExtent, YAxisExtent

app = qt.QApplication(sys.argv)

plot = PlotWindow()

# Visible extents
x_visible = XAxisExtent()
x_visible.setRange(45, 55)
plot.addItem(x_visible)

y_visible = YAxisExtent()
y_visible.setRange(45, 55)
plot.addItem(y_visible)

# Hidden extents
x_hidden = XAxisExtent()
x_hidden.setRange(80, 90)
plot.addItem(x_hidden)

y_hidden = YAxisExtent()
y_hidden.setRange(80, 90)
plot.addItem(y_hidden)

plot.show()

print("")
print("=== EXTENT RESETZOOM TEST ===")
print("")
print("CASE A: X fixed")
print("1. Zoom X to [40, 60]")
print("2. Disable X autoscale")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- Y limits become [45, 55]")
print("- YAxisExtent [80,90] ignored")
print("")
print("CASE B: Y fixed")
print("1. Zoom Y to [40, 60]")
print("2. Disable Y autoscale")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- X limits become [45, 55]")
print("- XAxisExtent [80,90] ignored")
print("")

sys.exit(app.exec())
```

woutdenolf:
@t20100 @payno The `Item.getBounds()` and the new `Item.getResetBounds()`  return NaN on the y limits for `XAxisExtent` and NaN on the x limits for `YAxisExtent`.

When resetting the zoom we now call `Item.getResetBounds()` instead of `Item.getBounds()` but I have I need a way to handle the NaNs because we set all 4

```python
class BackendBase:
    def setLimits(self, xmin, xmax, ymin, ymax, y2min=None, y2max=None):
```

In this case it seems that instead of `[nan, nan]` we have `[1, 100]` somehow.

I can either do this: uncomment https://github.com/silx-kit/silx/pull/4633/changes#diff-73bf8890a30feb0887556785fbfd8cbdc73ea4aca5e81db9de86907c4e42dd3bR942-R943

Or `BackendBase.setLimits` should handle NaN. What do you think?

t20100:
- `BoundRect` and `?AxisExtent` are invisible items that contributes to the "data ranges" to influence reset zoom. With them, one can ensure a given plot area is visible when reseting the zoom even if no data is display there, e.g. for a 2d scatter updating live which you know the final size from the start. AFAIR they were added for flint.

- `[1, 100]` is the fallback axis range when limits are not valid.

> I can either do this: uncomment Or BackendBase.setLimits should handle NaN. What do you think?

How it's done in  `PlotWidget._updateDataRange` (https://github.com/silx-kit/silx/blob/71cb9b864d4002b936de27c469c28f2bc1087bc0/src/silx/gui/plot/PlotWidget.py#L841) is similar to uncommenting the 2 lines in `pack`, so I would go this way.

BTW, it should be possible to put some code in common between `_updateDataRange` and `_getResetDataRange` (I prefer the implementation in `_getResetDataRange`).

src/silx/gui/plot/PlotWidget.py
woutdenolf: This is no longer true, we do take autoscale into account.

The difference between `resetZoom` and `_forceResetZoom` is not really clear from the docstrings. I'm not entirely sure about the original purpose either. As far as I understand

- When at least one axis is autoscaling, `resetZoom` calls `_forceResetZoom` and afterwards forces limits, except if both are autoscaling. Weird.
- When none of the axes are autoscaling, `resetZoom` does nothing.

src/silx/gui/plot/items/core.py
```diff
@@ -221,18 +265,38 @@ def _setPlot(self, plot):
...
+        - Fixed axes limits are applied to the resulting bounds
+
+        :returns: (xmin, xmax, ymin, ymax) or None
+        :rtype: Union[Bounds, None]
+        """
+        return self._getResetBounds()
+
+    def _getResetBounds(self) -> Bounds | None:
+        """:meth:`getResetBounds` implementation to override by sub-class."""
+        return self.getBounds()
```
woutdenolf, line 270: TODO: I think we should fix the limits for axis not in autoscale.

```python
xmin, xmax, ymin, ymax = self.getBounds()

xAxis = plot.getXAxis()
if not xAxis.isAutoScale():
    xmin, xmax = xAxis.getLimits()

yAxis = plot.getYAxis(self._getYAxis())
if not yAxis.isAutoScale():
    ymin, ymax = yAxis.getLimits()

return Bounds(xmin, xmax, ymin, ymax)
```

We do this in the items that override `_getResetBounds` but not in the default implementation.

Or better, this should be in `getResetBounds` and the implementations of `_getResetBounds` don't need to take care of this.

src/silx/gui/plot/PlotWidget.py
woutdenolf: If I were to describe the new behavior of `_forceResetZoom` it would be this

```
Reset the plot limits and redraw the plot.

The plot limits after a reset depend on whether axes have

- fixed limits (autoscale=False) or
- auto-scaling limits (autoscale=True).

See :meth:`getXAxis`, :meth:`getYAxis` and :meth:`Axis.setAutoScale`.

Axes that have fixed limits keep their limits after a reset.

Axes that have auto-scaling limits will have new limits determined by

- the data range within the limits of the axes with fixed limits or
- the full data range if all axes have auto-scaling limits.

Extra margins can be added around the data inside the plot area
(see :meth:`setDataMargins`).
Margins are given as one ratio of the data range per limit of the
data (xMin, xMax, yMin and yMax limits).
For log scale, extra margins are applied in log10 of the data.
```

This is exactly the same for `resetZoom` except that when non of the axes are auto-scaling, nothing happens. It also does a thing related to log-scale an limits `>=0`. I'm wondering whether this should not go to `_forceResetZoom` or even `Item.getResetBounds()`.

src/silx/gui/plot/items/core.py
woutdenolf: I did it in `getResetBounds`.

src/silx/gui/plot/PlotWidget.py
woutdenolf: @t20100 You created `_forceResetZoom` in https://github.com/silx-kit/silx/pull/1312. Not clear why. It used to be `resetZoom` only.

woutdenolf: I would be inclined to just have

```python
def resetZoom(self, dataMargins = None):
    return self._resetZoom(dataMargins=dataMargins, force=False)

def _resetZoom(self, dataMargins = None, force = True):
    ...
```

and all the logic happens in `_resetZoom` (currently `_forceResetZoom`), including to forcing and log-scale handling.

woutdenolf: I'm not sure why `force` would not always be True.

When all axes have autoscale=False we could just skip for performance reasons. But since "force" exists there must be other things that could influence the resetting even when autoscale=False for all. Something related to changing the aspect ratio fixing.

woutdenolf: Ok these is also `_notifyLimitsChanged`. This code is really convoluted.

woutdenolf: I refactored by keeping `resetZoom` and `_notifyLimitsChanged`.

I removed the logic that sets the limits of fixed axes again after `setLimits()` because the limits returned by `Item.getResetBounds()` already take fixed limits into account since those fixed limits are already needed to calculate the limits of variable axes.

src/silx/gui/plot/PlotWidget.py
```diff
@@ -3286,9 +3311,29 @@ def _forceResetZoom(
         self,
         dataMargins: tuple[float, float, float, float] | None = None,
     ):
-        """Reset the plot limits to the bounds of the data and redraw the plot.
+        """Force :meth:`resetZoom` without emitting Qt signals."""
+        self.resetZoom(dataMargins=dataMargins, force=True, notify=False)
```
woutdenolf: Not sure why Qt signals should not be emitted when forcing the zoom. I kept that behavior but I suspect it's a bug.

src/silx/gui/plot/items/core.py
```diff
@@ -221,18 +263,55 @@ def _setPlot(self, plot):
...
         """
         return self._getBounds()
 
-    def _getBounds(self):
+    def _getBounds(self) -> Bounds | None:
         """:meth:`getBounds` implementation to override by sub-class"""
         return None
 
+    def getResetBounds(self) -> Bounds | None:
+        """Returns the bounding box of this item used for resetting the axis ranges.
```
payno: This is a detail but I would keep the naming convention here to `bounds`.

Because we could have in the future bounding-box classes (that usually are reserved for being the smallest rectangle counting a set of points and comes with a different API like top-left, bottom-right, `contains`...)


```suggestion
        """Returns the bounds of this item used for resetting the axis ranges.
```

src/silx/gui/plot/items/core.py
```diff
@@ -319,28 +398,32 @@ def setInfo(self, info, copy=True):
...
+        xmin, xmax = xAxis.getLimits()
+        xmin, xmax = numpy.clip(bounds[:2], xmin, xmax)
+        ymin, ymax = yAxis.getLimits()
+        ymin, ymax = numpy.clip(bounds[2:], ymin, ymax)
 
         if xmin == xmax or ymin == ymax:  # Outside the plot area
             return None
         else:
-            return xmin, xmax, ymin, ymax
+            return Bounds.from_values(xmin, xmax, ymin, ymax)
```
payno: I am puzzled here. I find the `Bounds` API more convenient.

Nevertheless this is a break on the public API for at least `getBounds` and `getVisibleBounds` of the `Item` class.

Should we wait for the next major ? Or we have hope no-one is using this API ?

Fixing possible break would be fast. Nevertheless it might break somehwere...

src/silx/gui/plot/items/shape.py
```diff
@@ -243,22 +283,76 @@ def getRange(self) -> tuple[float, float]:
...
+
+            if xAxis.isAutoScale():
+                xmin, xmax = lmin, lmax
+            else:
+                # Expand to fixed limits if item bounds are within them
+                if xmin >= lmin and xmax <= lmax:
+                    xmin, xmax = lmin, lmax
+                elif xmin > lmax or xmax < lmin:
+                    # No overlap
+                    return None
```
payno: The 'clamping' operation seems to be repeated at several places. Maybe it could be grouped into one function.

src/silx/gui/plot/items/core.py
woutdenolf: `Bounds` is still a tuple so we do not break any API.

src/silx/gui/plot/items/shape.py
woutdenolf: Done. Logic in `_adjust_bounds_for_axis` is used both by invisible shape and invisible axis.

src/silx/gui/plot/items/core.py
woutdenolf: `Bounds.from_values` exists to handle None and NaN. It use to be done in code everywhere, now it is a helper function associated to the `Bounds` tuple.

woutdenolf:
Self review:

1. Plot and items have a awkward reference of each other (plot ref in items is weak). This is bad design in the first place and this PR increases the coupling. Instead we could pass the axis limits to `getResetBounds` instead of having the item get it from the plot

```python
class Item:
    def getResetBounds(self, limits: Bounds) -> Bounds | None:
         ...
```

And then the plot with pass the axis limits

```python
def getAutoScaleLimits(xAxis, yAxis) -> Bounds:
    if xAxis.isAutoScale():
        xmin, xmax = xAxis.getLimits()
        if xAxis._isLogarithmic():
            xmin = max(0, xmin)
            xmax = max(0, xmax)
    else:
        xmin, xmax = None, None

    if yAxis.isAutoScale():
        ymin, ymax = yAxis.getLimits()
        if yAxis._isLogarithmic():
            ymin = max(0, ymin)
            ymax = max(0, ymax)
    else:
        ymin, ymax = None, None

    return Bounds.from_values(xmin, xmax, ymin, ymax)
```

2. I'm not sure about `BoundingRect` and `_BaseExtent`. See `test_bounds._expected_reset_bounds` for the expected reset bounds based on the `x` and `y` values of the data and the fixed axis limits (i.e. axes that are not auto-scaling).

@t20100 or @payno Let me know whether you agree or not with 1 and 2 and I'll work on it.

payno:
> Plot and items have a awkward reference of each other (plot ref in items is weak). This is bad design in the first place and this PR increases the coupling. 

I am not sure I get this one. I might misunderstand but IMO this is an expected design. The item needs to interact with the plot and cannot keep a strong reference on the plot because there is child / parent relation.

src/silx/gui/plot/PlotWidget.py
t20100: It looks to have been done this way to avoid some issues when switching an axis to log scale or when enabling keep aspect ratio in order to by-pass the autoscale disabled.


+1 for refactoring

t20100:
Ideally `Item`s should have limited dependency on the `PlotWidget` they belong to, as a Qt widget has a parent but mostly does not rely on it.

Practically, since rendering backends are monoliths owned by the `PlotWidget`, the items depends on the `PlotWidget` for rendering&picking and they need to know the state of the `PlotWidget` for some features like `getVisibleBounds`.

I am +1 for not adding more coupling and go to something like:

```
class Item:
    def getResetBounds(self, limits: Bounds) -> Bounds | None:
```

loichuder:
> Ideally Items should have limited dependency on the PlotWidget they belong to, as a Qt widget has a parent but mostly does not rely on it.

The problem is that once the ref is here, it is very easy to use it and rely on it.

woutdenolf:
Changes:

- `Item.getResetBounds`: avoid access to `PlotWidget` via the weakref `self._getPlot()`, pass `AxesInfo` argument instead
- Isolate new named tuples in a `types` module.

src/silx/gui/plot/items/core.py
payno: Indeed my bad I misread.

woutdenolf:
Changed: harmonize the items with only independent variables (image, boundrect and axis extent). When such an item is outside a fixed axis limit, resetting does not take it into account when autoscaling the variable axis.  I put the helpers in src/silx/gui/plot/items/_bound_utils.py

image


```python
import sys
import numpy

from silx.gui import qt
from silx.gui.plot import PlotWindow

app = qt.QApplication(sys.argv)

plot = PlotWindow()

height, width = 100, 100


def make_test_image1(h, w):
    y, x = numpy.mgrid[:h, :w]

    # Spiral-like pattern
    cx, cy = w / 2, h / 2
    r = numpy.hypot(x - cx, y - cy)
    theta = numpy.arctan2(y - cy, x - cx)
    img = numpy.sin(r / 3 + 4 * theta)

    return img


def make_test_image2(h, w):
    y, x = numpy.mgrid[:h, :w]

    # Interference pattern
    img = (
        numpy.sin(x / 5) + numpy.sin(y / 7) + numpy.sin(numpy.hypot(x - 20, y - 70) / 2)
    )

    return img


def make_test_image3(h, w):
    y, x = numpy.mgrid[:h, :w]

    # Gaussian + ripples
    cx, cy = w / 2, h / 2
    r = numpy.hypot(x - cx, y - cy)
    img = numpy.exp(-(r**2) / (2 * 20**2))
    img += 0.3 * numpy.sin(r)

    return img


image = make_test_image1(height, width)
item = plot.addImage(image, origin=(0, 0), scale=(1, 1), legend="IN")

image = make_test_image2(height, width)
item = plot.addImage(image, origin=(width, height), scale=(1, 1), legend="HALF")

image = make_test_image3(height, width)
item = plot.addImage(image, origin=(2 * width, 2 * height), scale=(1, 1), legend="OUT")

plot.show()

print("")
print("=== IMAGE AUTOSCALE TEST ===")
print("1. Zoom in on one image")
print("2. Disable X autoscale")
print("3. Click Reset Zoom")
print("")
print("EXPECTED:")
print("- Y limits follow visible image region")
print("- Images outside FOV are not show")
print("")

sys.exit(app.exec())
```

woutdenolf:
I covered the self-review. Please verify the changes.

src/silx/gui/plot/PlotWidget.py
t20100: `_forceResetZoom` was added as a workaround for issues with log scale and matplotlib (#1312).
From a quick test with the oldest currently supported matplotlib (3.6), there is no issue using `notify=True` here: It emits 3 "limits changed" events instead of 2, which is already one too much anyway).
I would not add a `notify` argument to `resetZoom` in the public API (the reason for a private `_forceResetZoom` method at the time I guess) since this is generally a bad idea to block signals and look at multiple emitted signals in a follow-up PR.

src/silx/gui/plot/items/_bound_utils.py
```diff
@@ -0,0 +1,24 @@
+from .types import ItemBounds, AxesInfo, AxisInfo
+
+
+def bounds_outside_fixed_limits(bounds: ItemBounds, axesInfo: AxesInfo) -> bool:
+    """
+    Exlude when at least one axis has fixed limits and
```
t20100:
```suggestion
    Exclude when at least one axis has fixed limits and
```

src/silx/gui/plot/items/axis.py
```diff
@@ -341,6 +342,15 @@ def setRangeConstraints(self, minRange=None, maxRange=None):
             plot.setLimits(xMin, xMax, yMin, yMax, y2Min, y2Max)
         return updated
 
+    def getInfo(self) -> AxisInfo:
```
t20100: It would be best not to add this method to the public API: the information it provides is already available and this is used for now only in `resetZoom` + see comment on `getResetBounds`

src/silx/gui/plot/items/types.py
```diff
@@ -0,0 +1,72 @@
...
+            return None
+
+        return cls(*values)
+
+
+class AxisInfo(NamedTuple):
+    vmin: float
+    vmax: float
+    auto: bool
+    log: bool
```
t20100: Since we started to add a new axis scale ("asinh"), to be future-proof it would be best to pass `scale: "linear" | "log" | "asinh"`

src/silx/gui/plot/items/core.py
```diff
@@ -221,18 +222,52 @@ def _setPlot(self, plot):
...
+        :rtype: Union[ItemBounds, None]
         """
         return self._getBounds()
 
-    def _getBounds(self):
+    def _getBounds(self) -> ItemBounds | None:
         """:meth:`getBounds` implementation to override by sub-class"""
         return None
 
+    def getResetBounds(self, axesInfo: AxesInfo) -> ItemBounds | None:
```
t20100: Since `AxisInfo.auto = True` could be conveyed as undefined limits, I was thinking of a more generic method:
- `getBoundsWithin(self, bounds: ItemBounds) -> ItemBounds | None`
- or `getBoundsWithin(self, xlimits: tuple[float, float] | None, ylimits: tuple[float, float] | None) -> ItemBounds | None`

The only issue would be with `__minMaxDataWithError` which requires to know the axes scale... but that could also be passed as an axis limits.

t20100 (commented):
Thanks for digging in this long standing issue!
This looks good to me!

The new API looks OK for me, but I'm wondering if there is a way to avoid exposing the axis state through `getInfo` which provides a second way to access already available in the API, and if `AxisInfo` does not boil down to an axis limit in the end.

src/silx/gui/plot/items/core.py
woutdenolf: > getBoundsWithin(self, xlimits: tuple[float, float] | None, ylimits: tuple[float, float] | None) -> ItemBounds | None

I introduced `ItemBounds` precisely to avoid raw-dogging tuples.

I could replace `AxesInfo` and `AxisInfo` by

```python
class AxisBounds(NamedTuple):
    vmin: float
    vmax: float
    scale: Literal["linear", "log", "asinh"]

    @staticmethod
    def isnan(value: float) -> bool:
        return isnan(value)

    @property
    def auto(self) -> bool:
        return self.isnan(self.vmin) or self.isnan(self.vmax)

    def limits(self) -> tuple[float, float]:
        return self.vmin, self.vmax


class AxesBounds(NamedTuple):
    x: AxisBounds
    y: AxisBounds
```

And then use it like

```python
getBoundsWithin(self, bounds: AxesBounds) -> ItemBounds | None
```

Not really an improvement imo. It is called "bounds" now but it contains _scale_ which has nothing to do with bounds.

> ... but that could also be passed as an axis limits.

How can I do that without inevitable having some sort of AxisInfo?

t20100: Yes that doesn't change the problem.
For me, plot axis scale is the kind of thing that the item needs to be aware of since it changes what the item displays and so its state.
BTW, not for this PR, but to reduce the dependency of the item on the plot, the axis scales could be set by the plot on all of its items. 

I propose to either leave `getResetBounds` as it is or to go for `getBoundsWithin(self, xlimits: tuple[float, float] | None, ylimits: tuple[float, float] | None) -> ItemBounds | None` and access the axis scale state from the plot for now.

woutdenolf: I will leave it as it and make things private as much as possible. I could even make `getResetBounds` private. Shall I?

src/silx/gui/plot/items/axis.py
woutdenolf: Made private. But `silx.plot.gui.items.types` is still public. Should I split this module in a public and private one?

src/silx/gui/plot/items/types.py
woutdenolf: Used `AxisScaleType`

src/silx/gui/plot/items/core.py
t20100: Yes please, it's best for now IMO: it's easier to move from private to public than the other way around.

src/silx/gui/plot/items/axis.py
t20100: If we stick to rules for modules, then yes.
However, I'm not sure we need to be that picky for types: A type is only part of the public API if it is actually used by it.

What do you think?

woutdenolf: Made private

src/silx/gui/plot/items/types.py
woutdenolf: Done

src/silx/gui/plot/items/core.py
woutdenolf: Made private

woutdenolf:
I changed this to make all new things private and isolate types in a public and private module

| Old | New |
|-----|-----|
| `silx.gui.plot.items.axis.Axis.getInfo` | `silx.gui.plot.items.axis.Axis._getInfo` |
| `silx.gui.plot.PlotWidget.PlotWidget.resetZoom` (with `force`/`notify` params) | `silx.gui.plot.PlotWidget.PlotWidget._resetZoom` |
| `silx.gui.plot.PlotWidget.PlotWidget.resetZoom` (without `force`/`notify` params) | same (original)|
| `silx.gui.plot.items.axis.AxisScaleType` | `silx.gui.plot.items.types.AxisScaleType` |
| `silx.gui.plot.items.core.Item.getResetBounds` | `silx.gui.plot.items.core.Item._getResetBounds` |
| `silx.gui.plot.items.core.Item._getResetBounds` | `silx.gui.plot.items.core.Item._computeResetBounds` |
| `silx.gui.plot.items.core.PointsBase._getResetBounds` | `silx.gui.plot.items.core.PointsBase._computeResetBounds` |
| `silx.gui.plot.items.histogram.Histogram._getResetBounds` | `silx.gui.plot.items.histogram.Histogram._computeResetBounds` |
| `silx.gui.plot.items.image.ImageBase._getResetBounds` | `silx.gui.plot.items.image.ImageBase._computeResetBounds` |
| `silx.gui.plot.items.shape.BoundingRect._getResetBounds` | `silx.gui.plot.items.shape.BoundingRect._computeResetBounds` |
| `silx.gui.plot.items.shape._BaseExtent._getResetBounds` | `silx.gui.plot.items.shape._BaseExtent._computeResetBounds` |
| `silx.gui.plot.items.types.AxisInfo` | `silx.gui.plot.items._types.AxisInfo` |
| `silx.gui.plot.items.types.AxesInfo` | `silx.gui.plot.items._types.AxesInfo` |

src/silx/gui/plot/items/_types.py
```diff
@@ -0,0 +1,25 @@
...
+    def limits(self) -> tuple[float, float]:
+        return self.vmin, self.vmax
+
+    def has_limits(self) -> bool:
+        return not self.isnan(self.vmin) and not self.isnan(self.vmax)
+
+    @property
+    def is_logarithmic(self) -> bool:
+        return self.log == "log"
+
```
t20100, line 21:
```suggestion
class AxisInfo(NamedTuple):
    vmin: float
    vmax: float
    auto: bool
    scale: AxisScaleType

    def limits(self) -> tuple[float, float]:
        return self.vmin, self.vmax

    def has_limits(self) -> bool:
        return not self.isnan(self.vmin) and not self.isnan(self.vmax)

    @property
    def is_logarithmic(self) -> bool:
        return self.scale == "log"

```

src/silx/gui/plot/items/axis.py
```diff
@@ -335,6 +334,15 @@ def setRangeConstraints(self, minRange=None, maxRange=None):
...
 
+    def _getInfo(self) -> _types.AxisInfo:
+        vmin, vmax = self.getLimits()
+        log = self.getScale() == self.LOGARITHMIC
+        if log:
+            vmin = max(0, vmin)
+            vmax = max(0, vmax)
+        auto = self.isAutoScale()
+        return _types.AxisInfo(vmin=vmin, vmax=vmax, auto=auto, log=log)
+
```
t20100, line 345: There was an issue here: passing a `bool` as log and returning `self.log == "log"` in `is_logarithmic`
```suggestion
    def _getInfo(self) -> _types.AxisInfo:
        vmin, vmax = self.getLimits()
        scale = self.getScale()
        if scale == self.LOGARITHMIC:
            vmin = max(0, vmin)
            vmax = max(0, vmax)
        auto = self.isAutoScale()
        return _types.AxisInfo(vmin=vmin, vmax=vmax, auto=auto, scale=scale)

```

src/silx/gui/plot/items/core.py
```diff
@@ -1658,40 +1711,96 @@ def __minMaxDataWithError(
...
+            x, y, xerror, yerror = data
+
+        if mask is not None:
+            x = x[mask]
+            y = y[mask]
+
+            if xerror is not None and numpy.size(xerror) > 1:
+                xerror = xerror[mask]
+            if yerror is not None and numpy.size(yerror) > 1:
+                yerror = yerror[mask]
```
t20100: The error array can be (2, N) for separated -/+ errors.

```suggestion
            if xerror is not None and numpy.size(xerror) > 1:
                if xerror.ndim == 2:
                    xerror = numpy.array([xerror[0][mask], xerror[1][mask]])
                else:
                    xerror = xerror[mask]
            if yerror is not None and numpy.size(yerror) > 1:
                if yerror.ndim == 2:
                    yerror = numpy.array([yerror[0][mask], yerror[1][mask]])
                else:
                    yerror = yerror[mask]
```

Example:

```python
import sys
import numpy

from silx.gui import qt
from silx.gui.plot import PlotWindow

app = qt.QApplication(sys.argv)

plot = PlotWindow()

x = numpy.linspace(0, 100, 5000)

xerror = numpy.ones((2, 5000))
yerror = numpy.ones((2, 5000)) 

y_left_main = numpy.sin(x * 0.2)
y_left_main += 5 * numpy.exp(-0.5 * ((x - 50) / 0.5) ** 2)

plot.addCurve(x, y_left_main, xerror=xerror, yerror=yerror, legend="left_main", yaxis="left")

plot.show()
sys.exit(app.exec())
```

t20100 (approved):
Looks good, thanks for moving new API private.
Long pending feature, thanks for looking into it.

I found two issues: I propose some fixes in the comments.

Once fixed, LGTM!

src/silx/gui/plot/items/core.py
woutdenolf: I went for

```python
xerror = xerror[..., mask]
```

and added unit tests for curves with 1D and 2D errors on all axes. Ok?

t20100: Even better, LGTM

PR #4636, opened 2026-06-18
647 threadpoolpushbutton exception while closing the application

### LINKED ISSUE #647: [threadpoolpushbutton] Exception while closing the application
If the process is still working while we are closing the application, it raises an exception.
```
  File "/home/valls/.local/lib/python2.7/site-packages/silx/gui/widgets/ThreadPoolPushButton.py", line 64, in run
    holder.finished.emit()
RuntimeError: wrapped C/C++ object of type ThreadPoolPushButton has been deleted
```

payno:
Closes #647 

### PR summary

Fix exception if the `ThreadPoolPushButton` executable still alive after `ThreadPoolPushButton` destruction.

#### AI Disclosure
- [ ] No AI used
- [x] AI tool(s) used to check for alternatives.

### DIFF
--- a/src/silx/gui/widgets/ThreadPoolPushButton.py
+++ b/src/silx/gui/widgets/ThreadPoolPushButton.py
@@ -48,7 +48,7 @@ class _Wrapper(qt.QRunnable):
         self.__kwargs = kwargs
 
         # connect signal / slots
-        signalHolder.destroyed.connect(self.clear_signal_holder)
+        signalHolder.destroyed.connect(self._clearSignalHolder)
 
     def _getSignalHolder(self):
         if self.__signalHolder is None:
--- a/src/silx/gui/widgets/ThreadPoolPushButton.py
+++ b/src/silx/gui/widgets/ThreadPoolPushButton.py
@@ -63,33 +63,34 @@ class _Wrapper(qt.QRunnable):
         if not holder:
             return
 
-        holder.started.emit()
+        self._emitIfHolder("started")
         try:
             result = self.__callable(*self.__args, **self.__kwargs)
-            holder = self._getSignalHolder()
-            if holder:
-                holder.succeeded.emit(result)
+            self._emitIfHolder("succeeded", result)
         except Exception as e:
             module = self.__callable.__module__
             name = self.__callable.__name__
             _logger.error(
                 "Error while executing callable %s.%s.", module, name, exc_info=True
             )
-            holder = self._getSignalHolder()
-            if holder:
-                holder.failed.emit(e)
+            self._emitIfHolder("failed", e)
         finally:
             holder = self._getSignalHolder()
-            if holder:
-                holder.finished.emit()
-        holder = self._getSignalHolder()
-        if holder:
-            holder._sigReleaseRunner.emit(self)
+            self._emitIfHolder("finished")
+
+        self._emitIfHolder("_sigReleaseRunner", self)
 
     def autoDelete(self):
         """Returns true to ask the QThreadPool to manage the life cycle of
         this QRunner."""
         return True
+
+    def _emitIfHolder(self, emit_func, *args, **kwargs):
+        """Emit signal only if holder still exists."""
+        holder = self._getSignalHolder()
+        if not holder:
+            return
+        getattr(holder, emit_func).emit(*args, **kwargs)
 
 
 class ThreadPoolPushButton(WaitingPushButton):
--- a/src/silx/gui/widgets/ThreadPoolPushButton.py
+++ b/src/silx/gui/widgets/ThreadPoolPushButton.py
@@ -75,7 +75,6 @@ class _Wrapper(qt.QRunnable):
             )
             self._emitIfHolder("failed", e)
         finally:
-            holder = self._getSignalHolder()
             self._emitIfHolder("finished")
 
         self._emitIfHolder("_sigReleaseRunner", self)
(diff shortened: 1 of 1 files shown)


src/silx/gui/widgets/ThreadPoolPushButton.py
```diff
@@ -41,33 +42,56 @@ class _Wrapper(qt.QRunnable):
...
         """Constructor"""
         super().__init__()
-        self.__signalHolder = signalHolder
+        self.__signalHolder = weakref.ref(signalHolder)
         self.__callable = function
         self.__args = args
         self.__kwargs = kwargs
 
+        # connect signal / slots
+        signalHolder.destroyed.connect(self.clear_signal_holder)
```
payno: This is the `real fix` of the error.
To 'clarify' / improve life cycle management I also moved the  `__signalHolder` to use a weakref.

From what I understood using a [`Qt.QPointer`](https://doc.qt.io/qt-6/fr/qpointer.html) might have be enough. But those are not exposed from any of the PyQtX or Pyside...

src/silx/gui/widgets/ThreadPoolPushButton.py
```diff
@@ -41,33 +42,56 @@ class _Wrapper(qt.QRunnable):
...
         """Returns true to ask the QThreadPool to manage the life cycle of
         this QRunner."""
         return True
 
+    def _emitIfHolder(self, emit_func, *args, **kwargs):
+        """Emit signal only if holder still exists."""
+        holder = self._getSignalHolder()
+        if not holder:
+            return
+        getattr(holder, emit_func).emit(*args, **kwargs)
```
payno: I am not a big fan of using the getattr (but really not...)
Nevertheless without it the code gets very complex.

We could also create some lambda function or others early at the beginning of the `run` function and retrieving the different `holder.failed.emit`... but this would be very unclear IMO. Better ideas / alternative (very welcomed ;))

src/silx/gui/widgets/ThreadPoolPushButton.py
loichuder:
```suggestion
        # connect signal / slots
        signalHolder.destroyed.connect(self._clearSignalHolder)
```

src/silx/gui/widgets/ThreadPoolPushButton.py
loichuder: Use a function for each signal?

``` python
def emitSucceeded(self, result):
    holder = self._getSignalHolder()
    if not holder:
            return
    holder.succedeed.emit(result)

def emitFailed(self, error):
 ...
```

src/silx/gui/widgets/ThreadPoolPushButton.py
```diff
@@ -41,33 +42,56 @@ class _Wrapper(qt.QRunnable):
...
             name = self.__callable.__name__
             _logger.error(
                 "Error while executing callable %s.%s.", module, name, exc_info=True
             )
-            holder.failed.emit(e)
+            self._emitIfHolder("failed", e)
         finally:
-            holder.finished.emit()
-        holder._sigReleaseRunner.emit(self)
+            holder = self._getSignalHolder()
```
t20100: unused:
```suggestion
```

src/silx/gui/widgets/ThreadPoolPushButton.py
t20100: +1 to avoid `getattr`

payno: Let's go for one function per signal it will still be better than the `getattr` even if this add a lot of code for not much :(

loichuder: Yeah but it is way more readable.

src/silx/gui/widgets/ThreadPoolPushButton.py
```diff
@@ -41,33 +42,85 @@ class _Wrapper(qt.QRunnable):
...
+        self._emitSigReleaseRunner(self)
 
     def autoDelete(self):
         """Returns true to ask the QThreadPool to manage the life cycle of
         this QRunner."""
         return True
 
+    # Signal emission under condition of holder existing
+
+    def _emitSucceeded(self, result):
```
loichuder:
```suggestion
    def _emitSucceeded(self, result: Any):
```

src/silx/gui/widgets/ThreadPoolPushButton.py
```diff
@@ -41,33 +42,85 @@ class _Wrapper(qt.QRunnable):
...
+        holder.started.emit()
+
+    def _emitFinished(self):
+        """Emit 'finished' signal only if holder still exists."""
+        holder = self._getSignalHolder()
+        if not holder:
+            return
+        holder.finished.emit()
+
+    def _emitFailed(self, error):
```
loichuder:
```suggestion
    def _emitFailed(self, error: Exception):
```

src/silx/gui/widgets/ThreadPoolPushButton.py
```diff
@@ -41,33 +42,85 @@ class _Wrapper(qt.QRunnable):
...
+        """Emit 'failed' signal only if holder still exists."""
+        holder = self._getSignalHolder()
+        if not holder:
+            return
+        holder.failed.emit(error)
+
+    def _emitSigReleaseRunner(self, runner):
```
loichuder: Can't we remove the `runner` arg?

```suggestion
    def _emitSigReleaseRunner(self):
```

and then do:
```
self._sigReleaseRunner.emit(self)
```

loichuder (approved):
```from typing import Any```

and then, you're all good.

Thanks! 👍

t20100 (approved):
LGTM

PR #4640, opened 2026-06-19
silx.gui.plot.ImageStack: fix call to 'setNPrefetch'

payno:
### PR summary

Fix condition when `setNPrefetch` is called. Now `_urlData.maxsize` can be None.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/ImageStack.py
+++ b/src/silx/gui/plot/ImageStack.py
@@ -302,7 +302,10 @@ class ImageStack(qt.QMainWindow):
         """
         if n < 0:
             raise ValueError("'n' should be positive")
-        if self._urlData.maxsize < self.__n_prefetch:
+        if (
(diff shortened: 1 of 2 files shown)


src/silx/gui/plot/ImageStack.py
```diff
@@ -302,7 +302,10 @@ def setNPrefetch(self, n: int) -> None:
         """
         if n < 0:
             raise ValueError("'n' should be positive")
-        if self._urlData.maxsize < self.__n_prefetch:
+        if (
+            self._urlData.maxsize is not None
+            and self._urlData.maxsize < self.__n_prefetch
+        ):
```
payno, line 308: Dummy bug found when testing the examples.

t20100 (approved):

PR #4641, opened 2026-06-19
silx.gui.data.NXdataWidgets: Move ArrayCurvePlot in its own file

loichuder:
- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

### PR summary

Preparatory work for #4428 

#### AI Disclosure
- [x] No AI used

PR #4646, opened 2026-06-19
silx.gui.data.DataViews: give the possibility to change the X axis when viewing NXdata as curves and add support for aux signal errors

loichuder:

### PR summary

For #4428 

image
My first idea for this was to refactor `ArrayCurvePlot` which is why I prepared the way in #4641. However, I realized that I had to change the behaviour of the widget quite extensively so it was difficult to go through a deprecation cycle. 

After discussion with @t20100 , we chose to create a completely new widget (`NxCurvePlot`) to leave `ArrayCurvePlot` untouched. **I therefore suggest for the review to look at a meld of `ArrayCurvePlot.py` and `NxCurvePlot.py` to see the changes**. 

In essence
`setCurvesData` was changed to refer to `signals` and `axes` rather than `x` and `y`. The internal behaviour can now handle X axis changes, changing the value, the error and the scale dynamically.

Some parts were removed: 
- `getPlot` was unused
- special treatment of 1 or 2-value array for X was removed
- the handling of the active curve via `_setYLabelFromActiveLegend` will be refactored in a future PR
- the dynamic connection of  `_axesSelector` via `__axes_selector_is_connected` was removed as well

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/data/_NxCurvePlot.py
+++ b/src/silx/gui/data/_NxCurvePlot.py
@@ -58,15 +58,35 @@ class NxCurvePlot(qt.QWidget):
         title: str | None = None,
     ):
         self.__signals = list(signals)
+        if len(signal_names) != len(signals):
+            raise ValueError("Signal names must match the length of signals")
         self.__signals_names = list(signal_names)
-        self.__signal_errors = list(signal_errors) if signal_errors else None
+        if signal_errors:
+            if len(signal_errors) != len(signals):
+                raise ValueError("Signal errors must match the length of signals")
+            self.__signal_errors = list(signal_errors)
+        else:
+            self.__signal_errors = None
         self.__signal_scale = signal_scale or "linear"
         self.__axes = list(axes) if axes else None
-        self.__axes_names = list(axes_names) if axes_names else None
-        self.__axes_errors = list(axes_errors) if axes_errors else None
-        self.__axes_scales = (
-            [scale or "linear" for scale in axes_scales] if axes_scales else None
-        )
+        if axes_names:
+            if axes and len(axes) != len(axes_names):
+                raise ValueError("Axis names must match the length of axes")
+            self.__axes_names = list(axes_names)
+        else:
+            self.__axes_names = None
+        if axes_errors:
+            if axes and len(axes) != len(axes_errors):
+                raise ValueError("Axis errors must match the length of axes")
+            self.__axes_errors = list(axes_errors)
+        else:
+            self.__axes_errors = None
+        if axes_scales:
+            if axes and len(axes) != len(axes_scales):
+                raise ValueError("Axis scales must match the length of axes")
+            self.__axes_scales = list(scale or "linear" for scale in axes_scales)
+        else:
+            self.__axes_scales = None
 
         with blockSignals(self._axesSelector):
             self._axesSelector.clear()
--- a/src/silx/gui/data/_NxCurvePlot.py
+++ b/src/silx/gui/data/_NxCurvePlot.py
@@ -1,3 +1,5 @@
+from typing import Sequence
+
 import numpy
 
 from silx.gui import qt
@@ -22,7 +24,7 @@ class NxCurvePlot(qt.QWidget):
 
         self.__signals: list[numpy.ndarray] | None = None
         self.__signals_names: list[str] | None = None
-        self.__signal_errors: numpy.ndarray | None = None
+        self.__signal_errors: list[numpy.ndarray] | None = None
         self.__signal_scale: AxisScaleType = "linear"
         self.__axes: list[numpy.ndarray] | None = None
         self.__axes_names: list[str] | None = None
@@ -45,23 +47,23 @@ class NxCurvePlot(qt.QWidget):
 
     def setCurvesData(
         self,
-        signals: list[numpy.ndarray],
-        signal_names: list[str] | None = None,
-        signal_errors: numpy.ndarray | None = None,
+        signals: Sequence[numpy.ndarray],
+        signal_names: Sequence[str],
+        signal_errors: Sequence[numpy.ndarray] | None = None,
         signal_scale: AxisScaleType | None = None,
-        axes: list[numpy.ndarray] | None = None,
-        axes_names: list[str] | None = None,
-        axes_errors: list[numpy.ndarray | None] | None = None,
-        axes_scales: list[AxisScaleType | None] | None = None,
+        axes: Sequence[numpy.ndarray] | None = None,
+        axes_names: Sequence[str] | None = None,
+        axes_errors: Sequence[numpy.ndarray | None] | None = None,
+        axes_scales: Sequence[AxisScaleType | None] | None = None,
         title: str | None = None,
     ):
-        self.__signals = signals
-        self.__signals_names = signal_names
-        self.__signal_errors = signal_errors
+        self.__signals = list(signals)
+        self.__signals_names = list(signal_names)
+        self.__signal_errors = list(signal_errors) if signal_errors else None
         self.__signal_scale = signal_scale or "linear"
-        self.__axes = axes
-        self.__axes_names = axes_names
-        self.__axes_errors = axes_errors
+        self.__axes = list(axes) if axes else None
+        self.__axes_names = list(axes_names) if axes_names else None
+        self.__axes_errors = list(axes_errors) if axes_errors else None
         self.__axes_scales = (
             [scale or "linear" for scale in axes_scales] if axes_scales else None
         )
@@ -84,10 +86,9 @@ class NxCurvePlot(qt.QWidget):
         self._updateCurve()
 
     def _updateCurve(self):
-        if self.__signals is None:
+        if self.__signals is None or self.__signals_names is None:
             return
 
-        legends = self.__signals_names or [None] * len(self.__signals)
         self._plot.clear()
 
         axes_selection = self._axesSelector.selection()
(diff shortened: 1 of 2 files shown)


woutdenolf:
How do you end up with such a view?

image


```python
import sys
import numpy as np
import h5py
from silx.app.view.main import main


shape = (40, 50, 60, 70)

signal = np.random.random(shape)
signal_err = 0.05 * np.ones(shape)

ax0 = np.arange(shape[0])
ax1 = np.linspace(0, 1, shape[1])
ax2 = np.linspace(10, 20, shape[2])
ax3 = np.linspace(-5, 5, shape[3])

ax0_err = 0.1 * np.ones_like(ax0)
ax1_err = 0.01 * np.ones_like(ax1)
ax2_err = 0.2 * np.ones_like(ax2)
ax3_err = 0.05 * np.ones_like(ax3)

filename = "nxdata_4d.nxs"

with h5py.File(filename, "w") as f:
    f.attrs["default"] = "entry"
    entry = f.create_group("entry")
    entry.attrs["NX_class"] = "NXentry"

    entry.attrs["default"] = "data"
    data = entry.create_group("data")
    data.attrs["NX_class"] = "NXdata"

    data.create_dataset("signal0", data=signal)
    data.create_dataset("signal1", data=signal)
    data.create_dataset("signal2", data=signal)
    data.create_dataset("signal3", data=signal)

    data.create_dataset("signal0_errors", data=signal_err)
    data.create_dataset("signal1_errors", data=signal_err)
    data.create_dataset("signal2_errors", data=signal_err)
    data.create_dataset("signal3_errors", data=signal_err)

    data.create_dataset("axis0", data=ax0)
    data.create_dataset("axis1", data=ax1)
    data.create_dataset("axis2", data=ax2)
    data.create_dataset("axis3", data=ax3)

    data.create_dataset("axis0_errors", data=ax0_err)
    data.create_dataset("axis1_errors", data=ax1_err)
    data.create_dataset("axis2_errors", data=ax2_err)
    data.create_dataset("axis3_errors", data=ax3_err)

    data.attrs["signal"] = "signal0" 
    data.attrs["axes"] = ["axis0", "axis1", "axis2", "axis3"]
    data.attrs["auxiliary_signals"] = ["signal1", "signal2", "signal3"]


sys.argv = ["silx-view", filename]

main(sys.argv)
```

**Edit**: Sorry, my bad, I was not on main.

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -0,0 +1,139 @@
...
+        layout.addWidget(self._plot)
+        layout.addWidget(self._axesSelector)
+
+        self.setLayout(layout)
+
+    def setCurvesData(
+        self,
+        signals: list[numpy.ndarray],
+        signal_names: list[str] | None = None,
+        signal_errors: numpy.ndarray | None = None,
```
woutdenolf:
```suggestion
        signal_errors: list[numpy.ndarray | None] | None = None,
```

I'm assuming you meant this?

Note that all of of these also take numpy arrays. For these cases I often use

```python
signals: Sequence[numpy.ndarray]
```

That works for a numpy array and any other sequence (e.g. list, tuple) of numpy arrays.

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -0,0 +1,139 @@
...
+        signal_scale: AxisScaleType | None = None,
+        axes: list[numpy.ndarray] | None = None,
+        axes_names: list[str] | None = None,
+        axes_errors: list[numpy.ndarray | None] | None = None,
+        axes_scales: list[AxisScaleType | None] | None = None,
+        title: str | None = None,
+    ):
+        self.__signals = signals
+        self.__signals_names = signal_names
+        self.__signal_errors = signal_errors
```
woutdenolf: Lengths must match. When does it fail when they don't?

woutdenolf:
Seems like there is an issue with errors on auxiliary signals.

Signal is fine (errors on X and Y)

image

Aux signal is not fine (errors on X but not on Y)

image

```python
import sys
import numpy as np
import h5py
from silx.app.view.main import main


shape = (40, 50, 60, 70)

signal = np.random.random(shape)
signal_err = 0.1 * np.ones(shape)

ax0 = np.arange(shape[0])
ax1 = np.linspace(0, 1, shape[1])
ax2 = np.linspace(10, 20, shape[2])
ax3 = np.linspace(-5, 5, shape[3])

ax0_err = 0.5 * np.ones_like(ax0)
ax1_err = 0.3 * np.ones_like(ax1)
ax2_err = 0.8 * np.ones_like(ax2)
ax3_err = 0.4 * np.ones_like(ax3)

filename = "nxdata_4d.nxs"

with h5py.File(filename, "w") as f:
    f.attrs["default"] = "entry"
    entry = f.create_group("entry")
    entry.attrs["NX_class"] = "NXentry"

    entry.attrs["default"] = "data"
    data = entry.create_group("data")
    data.attrs["NX_class"] = "NXdata"

    data.create_dataset("signal0", data=signal)
    data.create_dataset("signal1", data=signal)
    data.create_dataset("signal2", data=signal)
    data.create_dataset("signal3", data=signal)

    data.create_dataset("signal0_errors", data=signal_err)
    data.create_dataset("signal1_errors", data=signal_err)
    data.create_dataset("signal2_errors", data=signal_err)
    data.create_dataset("signal3_errors", data=signal_err)

    data.create_dataset("axis0", data=ax0)
    data.create_dataset("axis1", data=ax1)
    data.create_dataset("axis2", data=ax2)
    data.create_dataset("axis3", data=ax3)

    data.create_dataset("axis0_errors", data=ax0_err)
    data.create_dataset("axis1_errors", data=ax1_err)
    data.create_dataset("axis2_errors", data=ax2_err)
    data.create_dataset("axis3_errors", data=ax3_err)

    data.attrs["signal"] = "signal0" 
    data.attrs["axes"] = ["axis0", "axis1", "axis2", "axis3"]
    data.attrs["auxiliary_signals"] = ["signal1", "signal2", "signal3"]


sys.argv = ["silx-view", filename]

main(sys.argv)
```

src/silx/gui/data/_NxCurvePlot.py
woutdenolf: Ok after running the test script below I see that not errors are supported on auxiliary signals. Is that intentional?

loichuder: Yeah sorry, I should have underlined this in the PR description: I only added the possibility to change the X axis without adding other features.

The previous `ArrayCurvePlot` did not support auxiliary signal errors so I didn't add it for now. I plan to do it in a follow-up PR.

src/silx/gui/data/_NxCurvePlot.py
loichuder: > When does it fail when they don't?

Probably when adding the curves.

But you are right, better to be explicit: I'll add checks in `setCurvesData`.

src/silx/gui/data/_NxCurvePlot.py
loichuder: But I see your point. Since I change the widget completely, I might as well add the support for auxiliary signals. That's less confusing.

src/silx/gui/data/_NxCurvePlot.py
loichuder: I added the checks in the crudest fashion for now. I think a heavier refactoring is needed to make the code less clunky.

Something like storing signals and axes in dataclasses with automatic fallbacks for example.

loichuder:
Now with support of auxiliary signal errors: 

image

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -0,0 +1,154 @@
...
+        signals: Sequence[numpy.ndarray],
+        signal_names: Sequence[str],
+        signal_errors: Sequence[numpy.ndarray] | None = None,
+        signal_scale: AxisScaleType | None = None,
+        axes: Sequence[numpy.ndarray] | None = None,
+        axes_names: Sequence[str] | None = None,
+        axes_errors: Sequence[numpy.ndarray | None] | None = None,
+        axes_scales: Sequence[AxisScaleType | None] | None = None,
+        title: str | None = None,
+    ):
```
woutdenolf, line 59: I would expect the type annotations of signal_errors and axes_errors to be the same: `Sequence[numpy.ndarray | None] | None`. Just like axes, not every signal might have errors.

woutdenolf (changes requested):

src/silx/gui/data/_NxCurvePlot.py
woutdenolf: `NXdata.get_axis_errors` can return `None`. As an aside, the name of that function is not very convenient now.

Note that in NeXus every dataset can have errors, this is not just an Nxdata thing: https://manual.nexusformat.org/datarules.html#uncertainties-or-errors

loichuder: > I would expect the type annotations of signal_errors and axes_errors to be the same:

Indeed, nice catch :fishing_pole_and_fish: 

> As an aside, the name of that function is not very convenient now.

Yes, agreed. This is the kind of side-effect that made me want to do it in a separate MR, me thinks.

loichuder: Added support for `None` in signal errors in 4284b56eb2b0301aba94ef112ee09e8eb9ba841b

I'm adding the refactoring of the `get_axis_errors` to the list of refactorings I'll do in the follow-up MR

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -0,0 +1,153 @@
...
+            x = numpy.arange(signalLength)
+        if self.__axes_errors is not None:
+            x_errors = self.__axes_errors[xIndex]
+        else:
+            x_errors = None
+
+        if self.__signal_errors is None:
+            y_errors = [None] * len(self.__signals)
+        else:
+            y_errors = self.__signal_errors
```
t20100, line 129: There is already `if else` block in `setCurvesData`, what about handling the `None` case directly there so those one are no longer needed?

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -0,0 +1,153 @@
...
+                xerror=x_errors,
+                yerror=y_error[axes_selection] if y_error else None,
+            )
+        self._plot.getYAxis().setScale(self.__signal_scale)
+        self._plot.setActiveCurve(self.__signals_names[0])
+
+        if self.__axes_names:
+            self._plot.setGraphXLabel(self.__axes_names[xIndex])
+        if self.__axes_scales:
+            self._plot.getXAxis().setScale(self.__axes_scales[xIndex])
```
t20100, line 147: Same here + if `self.__axes_names` or `self.__axes_scales` is `None` (which the method signature allows but is not the case as called in `DataViews`), the label and scale previously set remains (silx view recycle the widgets from one NXdata to the next).

loichuder: Same, treated it in #4651

src/silx/gui/data/_NxCurvePlot.py
loichuder: Good thinking. In the end, I treated it in #4651

woutdenolf (approved):
Thanks!

Btw, I've been using silx view from the main branch and it is so much better!

t20100 (approved):

PR #4651, opened 2026-06-24
‎silx.gui.data._NxCurvePlot.py: use dataclasses to store signals and axes

loichuder:

### PR summary

The early implementation of #4646 felt a bit clunky because:
- There was a lot of checks of lengths since `setCurvesData` arguments are sequences that are expected to have the same length https://github.com/silx-kit/silx/pull/4646#discussion_r3459630381
- These checks are even more clunky to write since some of the sequences are optional (`names`, `scales`, etc.)
- Also, because some sequences are optional, there was a lot of checking code in the functional part to see if those could be used or not, with some duplication https://github.com/silx-kit/silx/pull/4646#discussion_r3468007683 and https://github.com/silx-kit/silx/pull/4646#discussion_r3468037508

In this PR, I tried to address this by storing the sequences in dataclasses (`Signal` for signals and `Axis` for axes). **The idea is to handle the length mismatch and  fallbacks/optional values at a single place, that is when constructing the dataclasses.**

This also avoids forgetting to handle optional sequence members when the optional sequence itself was handled (like #4657) 

The functional part can then consume the dataclass sequences without having to worry about fields not being there.

There are still a few friction points that I'll highlight in the comments below but overall, I think this is an improvement. Let me know if you think the same. 

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/data/_NxCurvePlot.py
+++ b/src/silx/gui/data/_NxCurvePlot.py
@@ -3,11 +3,12 @@ from typing import Sequence
 import numpy
 
 from silx.gui import qt
-from ..plot.items import Curve
-from ..plot.items.axis import AxisScaleType
 
 from ..plot import Plot1D
+from ..plot.items import Curve
+from ..plot.items.axis import AxisScaleType
 from ..utils import blockSignals
+from ._models import Axis, Signal
 from .NumpyAxesSelector import NumpyAxesSelector
 
 
@@ -23,15 +24,9 @@ class NxCurvePlot(qt.QWidget):
     def __init__(self, parent: qt.QWidget | None = None):
         super().__init__(parent)
 
-        self.__signals: list[numpy.ndarray] | None = None
-        self.__signals_names: list[str] | None = None
-        self.__signal_errors: list[numpy.ndarray | None] | None = None
-        self.__signal_scale: AxisScaleType = "linear"
+        self.__signals: list[Signal] = []
+        self.__axes: list[Axis] = []
         self.__signal_curves: list[Curve] = []
-        self.__axes: list[numpy.ndarray] | None = None
-        self.__axes_names: list[str] | None = None
-        self.__axes_errors: list[numpy.ndarray | None] | None = None
-        self.__axes_scales: list[AxisScaleType] | None = None
 
         self._plot = Plot1D(self)
         self._plot.setGraphGrid(True)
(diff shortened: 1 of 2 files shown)


src/silx/gui/data/_models.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
+
+import numpy
+
+from ..plot.items.axis import AxisScaleType
+
+
+@dataclass(kw_only=True)
```
loichuder, line 8: `kw_only` is needed to be able to use required keys in `dataclass` inheriting a base `dataclass` with non-required keys.

https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -109,68 +105,49 @@ def setCurvesData(
...
-        xIndex = self._axesSelector.getIndicesOfNamedAxes()["X"]
-        if self.__axes:
-            x = self.__axes[xIndex]
+        axis = self.__axes[self._axesSelector.getIndicesOfNamedAxes()["X"]]
+        if axis.values is not None:
+            x = axis.values
         else:
-            signalLength = len(self.__signals[0][axes_selection])
+            signalLength = len(self.__signals[0].values[axes_selection])
             x = numpy.arange(signalLength)
```
loichuder, line 119: As mentioned in the description, most fallbacks are dealt with during the dataclass construction.

The only fallback that I could not move to the construction is the one on `axis.values` since it needs the length of the signal on the current slice to be computed.

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -109,68 +105,49 @@ def setCurvesData(
...
                 self._plot.addCurve(
                     x,
-                    signal[axes_selection],
-                    legend=legend,
-                    xerror=x_errors,
-                    yerror=y_error[axes_selection] if y_error else None,
+                    signal.values[axes_selection],
+                    legend=signal.name,
+                    xerror=axis.errors,
+                    yerror=signal.get_errors(axes_selection),
```
loichuder, line 130: The error fallback is now handled in `get_errors`

src/silx/gui/data/_NxCurvePlot.py
loichuder, line 121: This part is the reason of #4657. It is therefore fixed by the refactoring

loichuder:
CICICICICICICICI

Segfault for 3.14 that I thought was due to #4656 but apparently not since Cython is 3.2.6 now?

Flaky test for 3.10 that keeps failing #4660 

I'll investigate #4660 but since the 3.12 pipeline did go through I'd say this is ready for review

loichuder:
Still a segfault for Python 3.14: https://github.com/silx-kit/silx/issues/4668

3.10 and 3.12 are ok.

src/silx/gui/data/_models.py
```diff
@@ -0,0 +1,32 @@
...
+class _BaseEntity:
```
payno: Maybe a better name than 'Entity' could be found. like `_BaseNXattrField` or `_BaseNXField` or `_BaseDataField` ?

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -23,15 +24,9 @@ class NxCurvePlot(qt.QWidget):
     def __init__(self, parent: qt.QWidget | None = None):
         super().__init__(parent)
 
-        self.__signals: list[numpy.ndarray] | None = None
-        self.__signals_names: list[str] | None = None
-        self.__signal_errors: list[numpy.ndarray | None] | None = None
-        self.__signal_scale: AxisScaleType = "linear"
+        self.__signals: list[Signal] | None = None
+        self.__axes: list[Axis] | None = None
```
payno: Looking at the code I think we could even simplify here and have

```suggestion
        self.__signals: list[Signal] = []
        self.__axes: list[Axis] = []
```

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -109,68 +105,49 @@ def setCurvesData(
 
     def _updateCurve(self):
         """Updates X and Y data. Called when the plotted dimension changes"""
-        if self.__signals is None or self.__signals_names is None:
+        if self.__signals is None or self.__axes is None:
```
payno: If we go for having an empty list we could have instead

```suggestion
        if not self.__signals or not self.__axes:
```

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -109,68 +105,49 @@ def setCurvesData(
...
-        if self.__axes_scales:
-            self._plot.getXAxis().setScale(self.__axes_scales[xIndex])
+        self._plot.getXAxis().setScale(axis.scale)
+        self._plot.setGraphXLabel(axis.name)
         self._plot.resetZoom()
 
     def _updateYData(self):
         """Update Y data **only**. Called when the slicing indices change"""
-        if self.__signals is None or self.__signals_names is None:
+        if self.__signals is None:
```
payno: Still about having self.__signals always being a list
```suggestion
        if not self.__signals:
```

payno (commented):
Nice improvement congrats 💪 . My 'biggest' concern is on the name given to `_BaseEntity`. I think something better can be found.

src/silx/gui/data/_models.py
loichuder: Yeah, honestly I didn't think a lot about the name of this :sweat_smile: 

`_BaseNXField` seems good to me :+1:

src/silx/gui/data/_NxCurvePlot.py
loichuder: Hm, this means we can't distinguish between "data was not set" and "data was set but was empty". But perhaps the distinction is not needed.

I'll give it a try :+1:

loichuder: ebde2e23455f4be303f91356ad44ae50b25301fa looks good! Thanks for the suggestion!

src/silx/gui/data/_NxCurvePlot.py
loichuder: This check is actually not needed anymore since `_updateYData` will be a no-op if `self.__signals` is empty

payno (approved):
Great. I guess the CI should be fix on edna-site is back.

PR #4653, opened 2026-06-25
silx.gui.data._NxCurvePlot.py: Update only Y data when updating slicing indices

loichuder:

### PR summary

Before this PR, `_updateCurve` (that clears and redoes completely the plot) was connected both to the signal of changing the slicing indices (`selectionChanged`) and the signal of changing axes (`selectedAxisChanged`).

However, when slicing, only the Y data needs to change. Not only this makes less operations but it does not reset the state of the curves (e.g. if a curve was hidden by the user).

This PR introduces a new function `_updateYData` that is connected to `selectionChanged`. `_updateCurve` remains unchanged and connected to `selectedAxisChanged` since in this case, the plot needs to be regenerated from scratch.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/data/_NxCurvePlot.py
+++ b/src/silx/gui/data/_NxCurvePlot.py
@@ -3,7 +3,8 @@ from typing import Sequence
 import numpy
 
 from silx.gui import qt
-from silx.gui.plot.items.axis import AxisScaleType
+from ..plot.items import Curve
+from ..plot.items.axis import AxisScaleType
 
 from ..plot import Plot1D
 from ..utils import blockSignals
@@ -26,6 +27,7 @@ class NxCurvePlot(qt.QWidget):
         self.__signals_names: list[str] | None = None
         self.__signal_errors: list[numpy.ndarray | None] | None = None
         self.__signal_scale: AxisScaleType = "linear"
+        self.__signal_curves: list[Curve] = []
         self.__axes: list[numpy.ndarray] | None = None
         self.__axes_names: list[str] | None = None
         self.__axes_errors: list[numpy.ndarray | None] | None = None
@@ -35,7 +37,7 @@ class NxCurvePlot(qt.QWidget):
         self._plot.setGraphGrid(True)
 
         self._axesSelector = NumpyAxesSelector(self)
-        self._axesSelector.selectionChanged.connect(self._updateCurve)
+        self._axesSelector.selectionChanged.connect(self._updateYData)
         self._axesSelector.selectedAxisChanged.connect(self._updateCurve)
 
         layout = qt.QVBoxLayout()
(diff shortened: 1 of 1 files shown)


src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -106,11 +107,10 @@ def setCurvesData(
         self._updateCurve()
 
     def _updateCurve(self):
+        """Updates X and Y data. Called when the plotted dimension changes"""
         if self.__signals is None or self.__signals_names is None:
             return
 
-        self._plot.clear()
```
loichuder, line 112: Moved closer to the actual use

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -147,6 +148,27 @@ def _updateCurve(self):
             self._plot.getXAxis().setScale(self.__axes_scales[xIndex])
         self._plot.resetZoom()
 
+    def _updateYData(self):
+        """Update Y data but not X. Called when the slicing indices change"""
+        if self.__signals is None or self.__signals_names is None:
+            return
+
+        curves: list[Curve] = self._plot.getAllCurves(withhidden=True)
```
loichuder: `withhidden` is very important here to keep `curves`, `signals` and `errors` lists aligned.

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -147,6 +148,27 @@ def _updateCurve(self):
...
+
+        if self.__signal_errors is None:
+            y_errors = [None] * len(self.__signals)
+        else:
+            y_errors = self.__signal_errors
```
loichuder, line 163: Yet another clunky `if`. This will be fixed in #4651

loichuder:
CI is in shambles but it seems unrelated. I'll investigate

src/silx/gui/data/_NxCurvePlot.py
```diff
@@ -147,6 +148,27 @@ def _updateCurve(self):
             self._plot.getXAxis().setScale(self.__axes_scales[xIndex])
         self._plot.resetZoom()
 
+    def _updateYData(self):
+        """Update Y data but not X. Called when the slicing indices change"""
```
payno:
```suggestion
        """Update Y data **only**. Called when the slicing indices change"""
```

src/silx/gui/data/_NxCurvePlot.py
payno: 👍 I was about to add a comment on it 😇

payno (approved):
Nice PR thanks 👍 
I let you check the CI

src/silx/gui/data/_NxCurvePlot.py
t20100: Best to keep references to added curve items: by grabbing all curves you can get curves added by other tools (e.g. Fit action).

loichuder: Good to know, thanks!

loichuder:
> CI is in shambles but it seems unrelated. I'll investigate

#4656

t20100 (approved):
LGTM!

Issue #4648, opened 2026-06-22
silx.opencl.codec: the last block is no longer treated


loichuder:
It seems the last block is no longer treated: the decompressed array only contains 0 at the end.

Perhaps discarding the last block is too brutal. Perhaps the last block is smaller than the other ones and we have to take it into account?

_Originally posted by @loichuder in https://github.com/silx-kit/silx/issues/4644#issuecomment-4767177473_

loichuder:
To reproduce:

``` python
import struct

import bitshuffle
import numpy
from silx.opencl.codec.bitshuffle_lz4 import BitshuffleLz4

dtype = "uint16"

RANDOM_STATE = numpy.random.RandomState(42)
shape = (743, 647)
ref = RANDOM_STATE.poisson(100, size=shape).astype(dtype)
raw = (
    struct.pack(">Q", ref.nbytes) + b"\x00" * 4 + bitshuffle.compress_lz4(ref).tobytes()
)
bs = BitshuffleLz4(len(raw), numpy.prod(shape), dtype=dtype)
res = bs.decompress(raw).get()
assert numpy.array_equal(res, ref.ravel())
```

Context:
- pyopencl: `<pyopencl.Context at 0x2b593ce0 on <pyopencl.Device 'cpu-haswell-Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz' on 'Portable Computing Language' at 0x2b5430f0>>`
- bitshuffle version: 0.5.2
- numpy version: 2.1.3
- silx commit: 36ed9e114c0584a40410f50cd71d0ef516b86bfc

kif:
I take it but depends when I will have the time to debug.
`bs.decompress` has a new kwarg to use the OpenCL code for the pre-screening (slower, but apparently better tested) instead of the new cython module.

kif:
It is a fairly nasty bug ! Btw, I found that the decompression of "float64" images are likely to give inexact decompression as well but is is apparently decoupled bug.

kif:
I managed to make it somehow more robust in:
* stopping the scan when the block-size is larger than a valid LZ4 block (hard-coded at 8k)
* prevent out-of-buffer read in the subsequent kernel by passing an extra kernel argument 
But this does not explain why it works in OpenCL (regardless to the driver) and not with the code stupidly translated into Cython !
I had Claude thinking for more than an hour on the topic and it found the two codes exactly the same, but I can confirm they are not giving exactly the same results.

[...]

PR #4662, opened 2026-06-26
opencl.codec: prevent out of buffer read in bitshuffle-LZ4

kif:

#### AI Disclosure
- [x] No AI used

Related to #4648

### DIFF
--- a/src/silx/opencl/codec/ext/_unblock_bslz4.pyx
+++ b/src/silx/opencl/codec/ext/_unblock_bslz4.pyx
@@ -113,10 +113,10 @@ def unblock_bslz4(bytes src, uint8_t item_size=4):
                     logger.error(f"Read ({end}) beyond the end of the stream ({size}) -> invalid")
                 break
             block_size = load32_BE(buffer, pos)
-            if block_size>=block_max_size:
+            if block_size>block_max_size:
                 block_start = block_start[:block_idx]
                 with gil:
-                    logger.error(f"Invalid block size ({block_size}>={block_max_size}) ... since it is larger than the compressed block size")
+                    logger.error(f"Invalid block size ({block_size}>{block_max_size}) ... since it is larger than the compressed block size")
                 break
             block_start[block_idx] = end
             pos = end + block_size
--- a/src/silx/opencl/codec/ext/_unblock_bslz4.pyx
+++ b/src/silx/opencl/codec/ext/_unblock_bslz4.pyx
@@ -10,7 +10,7 @@
 
 __authors__ = ["Jérôme Kieffer"]
 __license__ = "MIT"
-__date__ = "05/06/2026"
+__date__ = "26/06/2026"
 
 
 from libc.stdint cimport uint8_t, uint32_t, uint64_t
@@ -61,11 +61,11 @@ def unblock_bslz4(bytes src):
     :return: list of start of block in the input stream (as numpy array)
     """
     cdef:
-        uint32_t size = len(src)
+        uint64_t size = len(src)
         uint64_t total_nbytes = load64_at(src, 0)
         uint32_t block_nbytes = load32_at(src, 8)
         const uint8_t[::1] buffer = src
-        uint32_t block_max = min(size//4+1, 1<<32-1)
+        uint32_t block_max = min(size//4+1, (1<<32)-1)
         uint64_t[::1] block_start = numpy.empty(block_max, dtype=numpy.uint64)
         uint32_t block_idx = 0
         uint64_t pos = 12
@@ -75,11 +75,13 @@ def unblock_bslz4(bytes src):
     with nogil:
         while ((end < size) and (block_idx < block_max)):
             block_size = load32_at(buffer, pos)
+            if block_size>8230:
+                # This is an invalid block ... since it is larger than the compressed block size
+                # the actual (uncompressed) block size is 8184
+                # but compressed block can be 5% larger in case of incompressible data
+                break
             block_start[block_idx] = end
             block_idx +=1
             pos = end + block_size
             end = pos + 4
-        # Remove last block if it overflows
-        if end >= size:
-            block_idx -= 1
     return numpy.asarray(block_start[:block_idx])
--- a/src/silx/opencl/codec/ext/_unblock_bslz4.pyx
+++ b/src/silx/opencl/codec/ext/_unblock_bslz4.pyx
@@ -10,12 +10,18 @@
 
 __authors__ = ["Jérôme Kieffer"]
 __license__ = "MIT"
-__date__ = "30/06/2026"
+__date__ = "02/07/2026"
 
 
 from libc.stdint cimport uint8_t, uint32_t, uint64_t
 from libc.math cimport ceil
 import numpy
+
+# Define some constants for bitshuffle
+cdef:
+    const uint32_t BSHUF_MIN_RECOMMEND_BLOCK = 128
+    const uint32_t BSHUF_BLOCKED_MULT = 8            # Block sizes must be multiple of this.
+    const uint32_t BSHUF_TARGET_BLOCK_SIZE_B 8192
 
 
 cdef inline uint64_t load64_BE(const uint8_t[::1] src,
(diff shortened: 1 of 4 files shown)


kif:
We investigated with @t20100 and it comes out the bitshuffle-LZ4 is doing some "specific" stuff for the very last bytes:
* The last block should be treated in a specific way:
* One can estimate better the number of blocks as the uncompressed blocksize is in the header [8,12] and the total uncompressed data is also in those header [0:8], both big-endian.
* This occurs mostly when the number of items is odd (the bytes/item is even)

kif:
@t20100 @loichuder 
I tested this code for integer random data from 9x9 to 256x256 without crash anymore ... of course it does not a proof that there is no bugs ... since we know there are some when considering float64. But for integer it does work.

kif:
```
#!/usr/bin/python3
import sys
import numpy
import random
import struct
import bitshuffle
from itertools import product
from silx.opencl.common import ocl, pyopencl
from silx.opencl.codec.bitshuffle_lz4 import BitshuffleLz4
from silx.opencl.codec._unblock_bslz4 import unblock_bslz4
from silx.opencl.codec.test.test_bitshuffle_lz4 import TestBitshuffleLz4

dtypes = ["uint16", "int16", "uint32", "int32", "uint64", "int64"]
# TODO inlcude all the others, "float64"]
rng = numpy.random.default_rng(42)
shape = (269, 271)
# niter = 100000
blocksize = 1<<13
print(blocksize)

for height, width in product(range(9,shape[0]+1),range(9,shape[1]+1)):
        shape = (int(height), int(width))
        for dtype in dtypes:

            ref = rng.poisson(100, size=shape).astype(dtype)
            compressed_block = bitshuffle.compress_lz4(ref).tobytes()
            print(shape, dtype, rng.bit_generator.state, len(compressed_block)%blocksize)
            raw = struct.pack(">Q", ref.nbytes) + struct.pack(">I", blocksize) + compressed_block
            bs = BitshuffleLz4(len(raw), numpy.prod(shape), dtype=dtype)
            res = bs.decompress(raw, force_unblock_on_device=True).get()
            ocl_version = bs.cl_mem["block_position"].get()[:bs.cl_mem["nb_blocks"].get()[0]]
            cyt_version = unblock_bslz4(raw)
            error = False
            if not  numpy.allclose(res, ref.ravel()):
                print("Decompression on GPU failed !")
                print("OpenCL", ocl_version, len(raw))
                print("Cython", cyt_version, len(raw))
                wrong=numpy.where(res-ref.ravel())[0]
                print(f"Number of wrong: {len(wrong)}: {wrong}")
                print(ref.ravel())
                print(res)
                print(bitshuffle.decompress_lz4(numpy.frombuffer(raw[12:], "uint8"),
                    shape,
                    dtype=numpy.dtype("float64")).ravel())
                error = True
                # sys.exit()
            if len(ocl_version) != len(cyt_version) or (ocl_version!=cyt_version).any():
                print("Cython and OpenCL do not agree on blocks")
                print("OpenCL", ocl_version, len(raw))
                print("Cython", cyt_version, len(raw))
                error = True
            if error:
                # Try to decompress in hybrid mode and check the validity of the output
                res2 = bs.decompress(raw, force_unblock_on_device=False).get()
                print(numpy.where(res2-ref.ravel()))
                print(ref.ravel())
                print(res2)
                print(bitshuffle.decompress_lz4(numpy.frombuffer(raw[12:], "uint8"),
                    shape,
                    dtype=numpy.dtype("float64")).ravel())
                sys.exit(0)

```

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
```diff
@@ -53,33 +54,47 @@ cdef inline uint32_t load32_at(const uint8_t[::1] src,
...
+        uint32_t block_max, block_max_size
+        uint64_t[::1] block_start
         uint32_t block_idx = 0
         uint64_t pos = 12
         uint64_t end = pos + 4
         uint64_t block_size
 
+    if block_nbytes == 0:
+            block_nbytes = 1<<13  # 8k
+            # This is a simplified version. One should take into account the item_size
```
t20100: Reference implementation: https://github.com/kiyo-masui/bitshuffle/blob/526440a16baff44bd405e0741ebd285858a5408d/src/bitshuffle_core.c#L2038-L2046

Called here in the filter: https://github.com/kiyo-masui/bitshuffle/blob/526440a16baff44bd405e0741ebd285858a5408d/src/bshuf_h5filter.c#L123

Since this is hard to track issues, I propose to implement it (can be a follow-up PR)

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
```diff
@@ -53,33 +54,47 @@ cdef inline uint32_t load32_at(const uint8_t[::1] src,
     return result
 
 
-def unblock_bslz4(bytes src):
+def unblock_bslz4(bytes src, uint8_t item_size=4):
```
t20100, line 80: Maybe best to not define a default value for `item_size` to make sure the correct value is passed.
Would be good to add to docstring to advertise it expects a number of bytes.

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
```diff
@@ -53,33 +54,47 @@ cdef inline uint32_t load32_at(const uint8_t[::1] src,
...
     with nogil:
-        while ((end < size) and (block_idx < block_max)):
-            block_size = load32_at(buffer, pos)
+        for block_idx in range(block_max):
+            if end >= size:
+                # This would read beyond the end of the stream -> invalid
+                block_start = block_start[:block_idx]
+                break
+            block_size = load32_BE(buffer, pos)
+            if block_size>=block_max_size:
```
t20100: I don't think it is in the reference implementation. Is it still needed?
This should not happen, if kept, probably good to log an error or even raise an exception?

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
```diff
@@ -53,33 +54,47 @@ cdef inline uint32_t load32_at(const uint8_t[::1] src,
...
+    # but compressed block can be 0.5% larger in case of incompressible data
+
+    block_max = (total_nbytes + block_nbytes - 1) // block_nbytes
+    block_start = numpy.empty(block_max, dtype=numpy.uint64)
     with nogil:
-        while ((end < size) and (block_idx < block_max)):
-            block_size = load32_at(buffer, pos)
+        for block_idx in range(block_max):
+            if end >= size:
+                # This would read beyond the end of the stream -> invalid
```
t20100: this should not happen, best to log an error IMO or even raise an exception?

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
```diff
@@ -53,33 +54,47 @@ cdef inline uint32_t load32_at(const uint8_t[::1] src,
...
         uint64_t block_size
 
+    if block_nbytes == 0:
+            block_nbytes = 1<<13  # 8k
+            # This is a simplified version. One should take into account the item_size
+    block_max_size = int(ceil(block_nbytes * 1.0046))
+    # the actual (uncompressed) block size is 8184
+    # but compressed block can be 0.5% larger in case of incompressible data
+
+    block_max = (total_nbytes + block_nbytes - 1) // block_nbytes
```
t20100: `block_max` -> `nblocks` maybe

This gives a wrong value if there is N full blocks + some left over uncompressed bytes:
block_nbytes = 512
total_nbytes = 2*512+2
nblocks = 3 while there is only 2 blocks + 2 left over uncompressed bytes

t20100 (commented):
I think there is still an issue in a specific case (see comment)

Comments for the cython implementation also applies to the opencl one.

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
kif: For now, this is not implemented at all ... there is comment telling it is far from being perfect.

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
kif: For now, `item_size` is neither used, nor documented ... I just created the entry point but to make it consistent with the OpenCL part it would require modification on the OpenCL part as well.

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
kif: Makes sense to log a warning

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
kif: +1 to log a warning.

src/silx/opencl/codec/ext/_unblock_bslz4.pyx
kif: I tried and no ... 
actually the way it is implemented in OpenCL is that the leftover bytes are treated by the last block, which is different from the `bitshuffle` implementation where it is some extra code executed after the end of the parallel section.

t20100 (approved):
LGTM, thanks for the fix!

PR #4520, opened 2026-03-12
Add pygfx/WGPU backend for 2D/3D plotting

physwkim:
OpenGL has been deprecated on macOS. This PR adds a new rendering backend based on pygfx/WGPU, enabling native Metal (macOS), Vulkan (Linux), and DirectX (Windows) support.

  Drop-in alternative to the existing OpenGL backend for both 2D and 3D plotting.

  2D (silx.gui.plot)

  - Curves, images, scatter, markers, shapes, error bars, color bar
  - GPU colormap pipeline (scalar texture + shader)
  - GPU compute shaders for min/max reduction and histogram
  - Async compute for non-blocking streaming updates
  - Efficient image streaming update path

  3D (silx.gui.plot3d)

  - SceneWindow with backend="pygfx" support
  - Mesh, scatter, volume (isosurface via marching cubes, cut planes), image, heightmap
  - Object parameters tree (SceneModel reuse via GroupItem)
  - 3D axes (pygfx Ruler) and bounding box wireframe
  - ClipPlane support for GroupItem

  Key changes

  - BackendPygfx.py — 2D backend implementation
  - Plot3DWidgetPygfx.py, SceneWidgetPygfx.py — 3D backend implementation
  - _PlotFrameCore.py — shared plot frame logic between OpenGL and pygfx
  - _pygfx_sync.py — Item3D to pygfx object sync layer
  - Parametrized tests for both backends

---

# Benchmarks

Two benchmarks were added to evaluate the backend performance.

---

## Benchmarks

Benchmarks were run on a **MacBook M4 Pro**.

- `compareBackendsFPS.py` for live line update performance
- `imageStreamingBenchmark.py` for image streaming throughput

## Image streaming benchmark

Example result for **4096 × 4096** streaming images:

```text
Size   Norm    FPS     plot_ms   other   total
1024   linear  507.3   1.56      0.02    1.58
2048   linear  308.1   2.64      0.03    2.67
4096   linear  125.7   6.77      0.04    6.81

```

The backend maintains **>120 FPS for 4K images** during continuous streaming updates.

image

---

## Curve update benchmark

Live curve updates comparison between backends:

image

matplotlib  : ~136 FPS
OpenGL      : ~120 FPS
pygfx/WGPU  : ~244 FPS

The pygfx backend shows ~2× higher performance than the OpenGL backend for dynamic curve updates.

---

# Platform Status

| Platform | Status |
|--------|--------|
| macOS (Metal) | Fully supported |
[...]

PR #4665, opened 2026-06-26
silx.gui.plot: Added pygfx/WGPU rendering backend for 2D plotting

physwkim:
Split out of #4520 to provide the **2D plot backend only**, as suggested in the
review (one feature per PR). #4520 stays open for the 2D/3D demo; the 3D backend
will follow in a separate PR.

This adds a pygfx/WGPU rendering backend for `silx.gui.plot`, enabling
native Metal (macOS), Vulkan (Linux) and DirectX (Windows).
It is a drop-in alternative to the OpenGL backend, selected with
`PlotWidget(backend="pygfx")`.

- Curves, images, scatter, markers, shapes, error bars, color bar
- Images colormapped through silx's `Colormap` (`applyToData`), so all
  normalizations (linear, log, sqrt, gamma, arcsinh) and the NaN color reuse
  the shared, tested code path
- Efficient image streaming update path (image item reuse)
- Native Wayland via the `bitmap` present method

Optional dependency: `pip install silx[pygfx]`
(pygfx >=0.16,<0.17 · rendercanvas >=2.6.1).

### Updates addressing the first review round
- Reduced the backend id to `pygfx` (dropped the `wgpu` alias)
- Removed the unused GPU compute code (min/max reduction, histogram)
- Removed an unused image-update helper
- Colormap now goes through `Colormap.applyToData` instead of a GPU LUT
  pipeline (this also fixes log/sqrt/gamma and the NaN color for scalar images)
- Reused the shared aspect-ratio helpers (`findDimToKeep`/`ensureAspectRatio`)
- Save via the shared `saveImageToFile`, which adds basic SVG export
- Clamp bogus screen DPI as in `silx.gui._glutils.OpenGLWidget`
- Native Wayland handled by falling back to the `bitmap` present method,
  so the `DISPLAY`/Wayland availability guards are no longer needed

### AI disclosure
This PR was prepared with Opus 4.8.


src/silx/gui/plot/PlotWidget.py
```diff
@@ -531,6 +531,28 @@ def __getBackendClass(self, backend: BackendType) -> BackendBase:
...
+                if sys.platform.startswith("linux"):
+                    if not os.environ.get("DISPLAY", ""):
+                        raise RuntimeError(
+                            "pygfx backend is not available: "
+                            "DISPLAY environment variable not set"
+                        )
+                    if os.environ.get("XDG_SESSION_TYPE", "") == "wayland":
+                        raise RuntimeError(
+                            "pygfx backend is not available: "
+                            "Wayland sessions are not supported"
```
t20100: Just a quick comment, do you have some link to an issue to track this upstream?

Since wayland is used in recent versions of linux and qt6 is picking it if available, this is quite critical.

physwkim: Yes, it's being tracked upstream here:

https://github.com/pygfx/rendercanvas/issues/36
https://github.com/pygfx/wgpu-py/issues/688
https://github.com/pygfx/rendercanvas/issues/66

physwkim: I verified that rendering works on native Wayland when using present_method="bitmap"

t20100: Thanks for the references!

So with wayland, pygfx is missing some system pointer to bypass qt and access the window system.

PySide6 looks to have recently added [QWaylandApplication](https://doc.qt.io/qtforpython-6/PySide6/QtGui/QNativeInterface.html#PySide6.QtGui.QNativeInterface.QWaylandApplication).
It was requested in [pyqt6 some time ago](https://riverbankcomputing.com/pipermail/pyqt/2025-August/thread.html#46308).
So hopefully this will move forward.

pyproject.toml
```diff
@@ -52,8 +52,12 @@ opencl = [
     'pyopencl', # For silx.opencl
     'Mako',     # For pyopencl reduction
 ]
+pygfx = [
+    'pygfx >=0.16.0,<0.17.0',   # For silx.gui.plot (pygfx backend)
```
t20100, line 56: Why pinning to `<0.17.0`? API is often broken?

src/silx/gui/plot/PlotWidget.py
```diff
@@ -531,6 +531,29 @@ def __getBackendClass(self, backend: BackendType) -> BackendBase:
                     _logger.debug("Backtrace", exc_info=True)
                     raise RuntimeError("OpenGL backend is not available")
 
+            elif backend in ("pygfx", "wgpu"):
```
t20100: I propose to only use one id: `pygfx`, there could be other wrappers WGPU

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+    # vmin should map to index ~1 out of 256
+    # index = (val - newVmin) / (newVmax - newVmin) * 255
+    # For val=vmin, index=1: 1 = (vmin - sentinel) / (newVmax - sentinel) * 255
+    # newVmax = sentinel + (vmin - sentinel) * 255
+    newVmax = sentinel + (vmax - sentinel) * 256.0 / 255.0
+
+    return scalar_data, (float(newVmin), float(newVmax))
+
+
+# WGSL Compute shaders ########################################################
```
t20100: The compute part does not seems to be used in this PR, can you remove it for now?

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+        vertices = numpy.array(vertices, dtype=numpy.float32)
+        indices = numpy.array(indices, dtype=numpy.int32)
+
+        fillColor = gfx.Color(color.r, color.g, color.b, alpha * 0.5)
+        geom = gfx.Geometry(positions=vertices, indices=indices)
+        mat = gfx.MeshBasicMaterial(color=fillColor, side="both")
+        return gfx.Mesh(geom, mat)
+
+
+def _fastColormapRange(data, colormap):
```
t20100: There is no need to add extra wrapper/fast-path, it's simpler to maintain a unique implementation of it

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+
+    # Adjust clim so sentinel maps to ~index 0 and vmin maps to ~index 1+
+    # LUT[0] = nanColor, LUT[1..255] = original LUT[0..254]
+    newLut = numpy.empty_like(lut)
+    newLut[0] = nanColor
+    # Remap: compress original 256 entries into indices 1..255
+    indices = numpy.linspace(0, 255, 255).astype(numpy.int32)
+    newLut[1:] = lut[indices]
+
+    lut[:] = newLut
```
t20100: `lut` is modifed in-place, this would be more readable if not. Also why compacting the LUT ratehr than extending it to add the NaN color?

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+
+    elif normalization == "sqrt":
+        scalar = numpy.sqrt(numpy.clip(scalar, 0, None))
+        clim = (float(numpy.sqrt(max(vmin, 0))), float(numpy.sqrt(max(vmax, 0))))
+        return scalar, clim, 1.0
+
+    elif normalization == "arcsinh":
+        scalar = numpy.arcsinh(scalar)
+        clim = (float(numpy.arcsinh(vmin)), float(numpy.arcsinh(vmax)))
+        return scalar, clim, 1.0
```
t20100: If there is no on GPU normalization (it seems there is only linear and gamma correction available), it sound similar to computing the RGBA texture and sending it to the GPU.
The benefit of doing so is that it can rely on existing functions and there would be no need for change here when adding a new normalization: Look at `silx.gui.colors.Colormap.applyToData`

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+            self._imageObj.material.clim = (vmin, vmax)
+            if cmapTex is not None:
+                self._imageObj.material.map = cmapTex
+
+        ox, oy = origin
+        sx, sy = scale
+        self._imageObj.local.position = (ox, oy, 0)
+        self._imageObj.local.scale = (sx, sy, 1)
+
+    def updateData(self, data, clim=None):
```
t20100: This is never called, can you remove it?

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+        :param data: numpy array
+        :param data_min: histogram lower bound (in normalized space)
+        :param data_max: histogram upper bound (in normalized space)
+        :param num_bins: number of bins
+        :param norm_mode: 0=linear, 1=log10, 2=sqrt, 3=arcsinh
+        :returns: (counts, bin_edges) or None
+        """
+        ac = self._getAsyncCompute()
+        ac.submit_histogram(data, data_min, data_max, num_bins, norm_mode)
+        return ac.get_histogram()
```
t20100: This compute code does not seem to be used, please remove it from this PR

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+
+    def getDevicePixelRatio(self):
+        return self.devicePixelRatioF()
+
+    def getDotsPerInch(self):
+        screen = self.screen()
+        if screen is not None:
+            return screen.logicalDotsPerInch() * self.getDevicePixelRatio()
+        return 92
+
```
t20100, line 832: We had some issue with screen DPI reported by Qt, see https://github.com/silx-kit/silx/blob/c2344a0f600b312083fbd794a6a25bfe551b709f/src/silx/gui/_glutils/OpenGLWidget.py#L354

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+        if fileFormat not in ["png", "ppm", "svg", "tif", "tiff"]:
+            raise NotImplementedError("Unsupported format: %s" % fileFormat)
+
+        # Force a synchronous render
+        self._draw()
+        snapshot = self._renderer.snapshot()
+
+        # snapshot is (H, W, 4) RGBA uint8
+        from PIL import Image as PILImage
```
t20100: Best to avoid lazy import unless necessary.

I propose for now to do as in the OpenGL backend and use `silx.gui.plot.backends.glutils.PlotImageFile.saveImageToFile` so there is basic SVG support.
In another PR, we can consider using pillow (since it is already a dependency of `fabio`  which silx depends on, so it is OK to use it).

src/silx/gui/plot/backends/BackendPygfx.py
```diff
@@ -0,0 +1,2814 @@
...
+        elif keepDim == "x":
+            dataH = (xMax - xMin) * plotHeight / float(plotWidth)
+            yCenter = 0.5 * (yMin + yMax)
+            yMin = yCenter - 0.5 * dataH
+            yMax = yCenter + 0.5 * dataH
+            y2Center = 0.5 * (y2Min + y2Max)
+            y2Min = y2Center - 0.5 * dataH
+            y2Max = y2Center + 0.5 * dataH
+        else:
+            raise RuntimeError("Unsupported dimension to keep: %s" % keepDim)
```
t20100: why changing it compared to the implementation in `BackendOpenGL`?

src/silx/gui/plot/PlotWidget.py
```diff
@@ -531,6 +531,29 @@ def __getBackendClass(self, backend: BackendType) -> BackendBase:
...
+                    ):
+                        raise RuntimeError(
+                            "pygfx backend is not available: "
+                            "neither DISPLAY nor WAYLAND_DISPLAY is set"
+                        )
+                    # Both XWayland (xcb) and the native "wayland" Qt platform are
+                    # supported: BackendPygfx uses the fast "screen" present on
+                    # X11/XWayland and falls back to "bitmap" present on native
+                    # Wayland, where "screen" would conflict with Qt's compositing.
+
```
t20100: There is no need checking `DISPLAY`, it should already be an issue before reaching this point.
Is checking `WAYLAND_DISPLAY` really needed since there is a fallback to `bitmap`?

t20100 (commented):
Thanks for your contribution and your patience! Sorry for the delay in reviewing this.

First round of general comments and partial review.

Do not hesitate to add your name and/or employer as copyright owner of the pygfx backend part! As long as it is under the MIT license, it is welcomed!

We will certainly need help to maintain this backend (new features, bug fixes,...), is it OK for you to help us over time?

`_PlotFrameCore.py` is mostly a copy of `GLPlotFrame.py` and `BackendPygfx.py` has a lot in common with `BackendOpenGL.py` which completely makes sense for a first integration.
We'll refactor this after merging this PR to avoid maintaining duplicated code. To ease this refactoring, it would be best to keep the duplicated code as close as possible to the one of the OpenGL backend so the parts to put in common can be clearly identified (e.g. `_ensureAspectRatio`).

physwkim:
Thank you for the thorough review, and no worries about the delay!

silx has been a huge help for us at PAL, so of course I'm happy to help maintain this backend over time and to contribute wherever I can be useful.

On copyright: I've added Pohang Accelerator Laboratory as a copyright holder alongside ESRF on the pygfx backend, and credited you and @loichuder in the authors list.

I've pushed a round of changes addressing the inline comments below: dropped the `wgpu` alias, removed the unused GPU compute code, reworked colormapping to go through `Colormap.applyToData`, and reused the shared aspect-ratio / save-image / DPI helpers. Details are in each thread.

pyproject.toml
physwkim: pygfx is still pre-1.0, so a minor bump (0.16 -> 0.17) can carry breaking changes. I'd rather keep the upper bound and validate against each new minor before relaxing it, than let it float. Happy to raise it in a follow-up once 0.17 is verified.

src/silx/gui/plot/PlotWidget.py
physwkim: Done -- dropped the `wgpu` alias; the backend is now selected only with `PlotWidget(backend="pygfx")`.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Removed -- the GPU min/max reduction and histogram compute code was never wired into the backend, so I dropped it along with the now-unused `wgpu` import.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Removed the extra wrapper -- colormapping now goes through a single `Colormap.applyToData` path (see the LUT thread below).

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Reworked: scalar images are now colormapped on the CPU via `Colormap.applyToData` and uploaded as an RGBA texture (the same path as RGBA images), so the in-place LUT and the compacting are both gone. This reuses the shared, tested colormap code.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Agreed. Rather than reimplement normalizations on the GPU, scalar images now go through `Colormap.applyToData`, which handles all normalizations (linear, log, sqrt, gamma, arcsinh) and the NaN color on the CPU. As a side effect this fixes log/sqrt/gamma and NaN handling for scalar images, which the old LUT path only did for linear.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Removed -- it was an unused duplicate; streaming updates go through the reusable-item pool in `addImage`.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Removed together with the rest of the compute code.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Thanks for the pointer. `getDotsPerInch` now mirrors `silx.gui._glutils.OpenGLWidget`: it uses `physicalDotsPerInch()` and clamps values below 55 / above 1000 to a sane default, logged once.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: Done -- `saveGraph` now uses `silx.gui.plot.backends.glutils.PlotImageFile.saveImageToFile`, like the OpenGL backend (no more lazy PIL import). It also gives basic SVG export. Happy to consider pillow separately.

src/silx/gui/plot/backends/BackendPygfx.py
physwkim: No good reason -- reverted to the shared helpers: `_ensureAspectRatio` now uses `findDimToKeep` / `ensureAspectRatio` from `backends.utils`, matching the OpenGL backend.

src/silx/gui/plot/PlotWidget.py
physwkim: The guard now accepts `WAYLAND_DISPLAY` in addition to `DISPLAY` and no longer rejects native Wayland: BackendPygfx detects the native Wayland Qt platform and falls back from the `screen` to the `bitmap` present method there, while X11/XWayland keep `screen`. I kept a minimal linux display check for a clearer early error, but I'm happy to drop it entirely if you'd prefer.

src/silx/gui/plot/PlotWidget.py
physwkim: Closing the loop: this is now implemented in the backend -- native Wayland sessions automatically use the `bitmap` present method, while X11/XWayland keep the faster `screen` present. The upstream issue (rendercanvas#36) still tracks native `screen` support on Wayland.

src/silx/gui/plot/PlotWidget.py
t20100: Why having this tests?
```
if sys.platform.startswith("linux"):
                    if not os.environ.get("DISPLAY", "") and not os.environ.get(
                        "WAYLAND_DISPLAY", ""
                    ):
```

It should be the case, or I would just let the app fail, the other backends would not work as well if `DISPLAY` is not defined.

t20100 (commented):
Thanks al lot for addressing the review!
I'll have a second close look at it.

> silx has been a huge help for us at PAL

Good to hear! Happy if it can help!

> so of course I'm happy to help maintain this backend over time and to contribute wherever I can be useful.

Excellent!

src/silx/gui/plot/PlotWidget.py
physwkim: Good call — dropped it entirely. The `pygfx` branch now looks like the other two.

The Wayland note that comment carried already lives in `BackendPygfx.__init__`, where the present method actually gets picked.

PR #4666, opened 2026-06-29
silx.gui.widgets.test.test_threadpoolpushbutton: move to pytest

loichuder:
- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

### PR summary

A stab at fixing another segfault occurring in the CI:  https://github.com/silx-kit/silx/actions/runs/28374949973/job/84062007669

As always, as a first step, I convert the culprit test to pytest.

#### AI Disclosure
- [x] No AI used

PR #4667, opened 2026-06-30
silx.gui.utils.testutils: Added `waitAsLongAs` helper to `qapp_utils` and `TestCaseQt` 

t20100:

### PR summary

! Merge PR #4666 first !

Move `waitForPendingOperations` to a more general helper in `qapp_utils`.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/hdf5/test/test_hdf5.py
+++ b/src/silx/gui/hdf5/test/test_hdf5.py
@@ -63,15 +63,6 @@ def create_NXentry(group, name):
     return node
 
 
-def waitForPendingOperations(qapp_utils: TestCaseQt, model):
-    for _ in range(20):
-        if not model.hasPendingOperations():
-            break
-        qapp_utils.qWait(200)
-    else:
-        raise RuntimeError("Still waiting for a pending operation")
-
-
 def testErrorInInsertFilenameAsync(tmp_path, caplog, qapp_utils):
     filename = str(tmp_path / "corrupt.h5")
     with open(filename, "wb") as h5file:
@@ -82,7 +73,7 @@ def testErrorInInsertFilenameAsync(tmp_path, caplog, qapp_utils):
     model.insertFileAsync(filename)
 
     assert qt.silxGlobalThreadPool().waitForDone(2000)  # Needed for PySide6
-    waitForPendingOperations(qapp_utils, model)
+    qapp_utils.waitAsLongAs(model.hasPendingOperations)
     assert len(caplog.records) == 1
     assert "can't be read as HDF5" in caplog.records[0].getMessage()
 
@@ -148,7 +139,7 @@ class TestHdf5TreeModel(TestCaseQt):
             self.assertIsInstance(
                 model.nodeFromIndex(index), hdf5.Hdf5LoadingItem.Hdf5LoadingItem
             )
-            waitForPendingOperations(self, model)
+            self.waitAsLongAs(model.hasPendingOperations)
             index = model.index(0, 0, qt.QModelIndex())
             self.assertIsInstance(model.nodeFromIndex(index), hdf5.Hdf5Item.Hdf5Item)
         finally:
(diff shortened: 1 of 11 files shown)


loichuder:
Can you rebase? Thanks!

src/silx/gui/hdf5/test/test_hdf5.py
```diff
@@ -148,7 +139,7 @@ def testInsertFilenameAsync(self):
             self.assertIsInstance(
                 model.nodeFromIndex(index), hdf5.Hdf5LoadingItem.Hdf5LoadingItem
             )
-            waitForPendingOperations(self, model)
+            self.waitUntil(lambda: not model.hasPendingOperations())
```
t20100: Could be `waitAsLongAs(model.hasPendingOperations)` 🤷

t20100 (commented):
rebased

src/silx/gui/hdf5/test/test_hdf5.py
loichuder: What about

```
waitForPendingOperations(model)
```

and move the `hasPendingOperations` call in the util?

Or is there another usecase that I missed?

t20100: I found quite a few more places where this pattern is used. It's mostly to call `hasPendingOperations` (not only on `model`), but there's a few places that would benefit a more generic helper:

https://github.com/silx-kit/silx/blob/29ab4cd4050d3bfde4c5d43b59c7e553c132c163/src/silx/conftest.py#L213-L216

https://github.com/silx-kit/silx/blob/29ab4cd4050d3bfde4c5d43b59c7e553c132c163/src/silx/gui/plot/test/utils.py#L71-L74

https://github.com/silx-kit/silx/blob/29ab4cd4050d3bfde4c5d43b59c7e553c132c163/src/silx/gui/utils/test/test_async.py#L92-L96

https://github.com/silx-kit/silx/blob/29ab4cd4050d3bfde4c5d43b59c7e553c132c163/src/silx/gui/utils/test/test_async.py#L117-L121

loichuder: Ok then keep the generic one. It is already quite an improvement :+1:

src/silx/gui/utils/test/test_async.py
```diff
@@ -89,13 +89,7 @@ def testFromPythonThread(self):
...
         thread = threading.Thread(target=self._threadedTest)
         thread.start()
-        for i in range(100):  # Loop over for 10 seconds
-            self.qapp.processEvents()
-            thread.join(0.1)
-            if not thread.is_alive():
-                break
-        else:
-            self.fail("Thread task still running")
+        self.waitUntil(lambda: not thread.is_alive())
```
t20100: it wait move from join to qWait but it waits all the same

src/silx/gui/utils/test/test_async.py
```diff
@@ -114,10 +108,4 @@ def autoDelete(self):
...
         runner = Runner(self._threadedTest)
         threadPool.start(runner)
-        for i in range(100):  # Loop over for 10 seconds
-            self.qapp.processEvents()
-            done = threadPool.waitForDone(100)
-            if done:
-                break
-        else:
-            self.fail("Thread pool task still running")
+        self.waitUntil(lambda: threadPool.waitForDone(10))
```
t20100: same wait partly moves from waitForDone to qWait but it should wait all the same

loichuder (approved):
It seems inverting the callable condition would avoid a lot of `not` (using `waitAsLongAs` as you suggested) but fine to keep it like this.

t20100:
> It seems inverting the callable condition would avoid a lot of not (using waitAsLongAs as you suggested) but fine to keep it like this.

Done, it was almost a regexp away.
I also spotted one extra loop that can be replaced: [a5ad65b](https://github.com/silx-kit/silx/pull/4667/commits/a5ad65b01b580a1e52dd5b4d0b2d93994b80249d)

@loichuder I added new commits if you want to have a look at the last modifications.
If you are OK, I'll squash it all before merging.

PR #4669, opened 2026-07-03
silx.gui.plot.PlotWidget: deprecate `histogram` arg from `addCurve` and add new args to `addHistogram`

loichuder:

### PR summary

I noticed when working on #4651 that there are two ways of adding an histogram to a `PlotWidget`:
- use `addHistogram`
- use `addCurve` with the parameter `histogram` 

This is confusing and inconsistent for several reasons:
1. The level of functionality is not the same. For example, adding via `addCurve` allows to change the `yaxis` while adding via `addHistogram` does not
2. Some of the parameters of `addCurve` are ignored when using `histogram` since they do not apply to histograms. Some of them log warnings when they are unused but not all, since it is not really tractable to track all arguments of `addCurve`.
3. This makes the return type of `addCurve` needlessly inconsistent. If I don't use `histogram`, I know that `addCurve` will return a `Curve` but my IDE will still show an error as the return type is (rightly so) `Curve | Histogram`.

I propose therefore the deprecate the `histogram` parameter of `addCurve` in favour of using `addHistogram`.

In this PR, I deprecate the argument while also matching the functionality of `addHistogram` with what can be done by using `histogram` in `addCurve`. For this, I added several arguments to `addHistogram` (e.g. the `yaxis` mentionned above).

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/plot/PlotWidget.py
+++ b/src/silx/gui/plot/PlotWidget.py
@@ -1123,17 +1123,7 @@ class PlotWidget(qt.QMainWindow):
                                 (Default: True)
         :param fill: True to fill the curve, False otherwise (default).
         :param resetzoom: True (the default) to reset the zoom.
-        :param histogram: if not None then the curve will be draw as an
-            histogram. The step for each values of the curve can be set to the
-            left, center or right of the original x curve values.
-            If histogram is not None and len(x) == len(y)+1 then x is directly
-            take as edges of the histogram.
-            Type of histogram::
-
-            - None (default)
-            - 'left'
-            - 'right'
-            - 'center'
+        :param histogram: DEPRECATED, use `addHistogram` with `align` instead.
         :param copy: True make a copy of the data (default),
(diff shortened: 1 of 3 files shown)


src/silx/gui/plot/PlotWidget.py
```diff
@@ -1123,7 +1123,8 @@ def addCurve(
                                 (Default: True)
         :param fill: True to fill the curve, False otherwise (default).
         :param resetzoom: True (the default) to reset the zoom.
-        :param histogram: if not None then the curve will be draw as an
+        :param histogram: DEPRECATED, use `addHistogram` with `align` instead.
```
t20100, line 1126: `:param histogram:` could be completely removed from the docstring otherwise I would just keep:
```:param histogram: DEPRECATED, use `addHistogram` with `align` instead.```

t20100 (approved):
Makes sense to me.

PR #4613, opened 2026-06-15
Utils: add an helper to increase opened file soft limit

maxenceprog:
- create a new module in silx.utils
- call the new function in silx.app.view.main

close #4608 

- [x] The PR title is formatted as: `<Module or Topic>: <Action> <Summary>` (see [contributing guidelines](https://github.com/silx-kit/silx/blob/main/doc/source/contribute/development.rst#pull-request-title-format))

#### AI Disclosure
- [x] No AI used
- [ ] AI tool(s) ... used for ...

PR #4679, opened 2026-07-28
silx.utils.files: add helper increase_opened_files_limit (only works …

### LINKED ISSUE #4608: Provide a helper to set the limit number of opened file to the maximum
This is useful for accessing hdf5 virtual dataset which links to too many files since the VDS keeps them all opened.
Typical limit is 1024 concurrently opened files.

This code sets this limit to the maximum:
https://github.com/silx-kit/silx/blob/137ffc9424ada7b1b1613586b5d84c55364f9f22/src/silx/app/view/main.py#L105-L118

Instead of copying it around, let's make a small helper

loichuder:
…on Unix)

### PR summary

Reimplementation of #4613

Fixes https://github.com/silx-kit/silx/issues/4608 

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/utils/files.py
+++ b/src/silx/utils/files.py
@@ -1,34 +1,45 @@
-# /*##########################################################################
-#
-# Copyright (c) 2019 European Synchrotron Radiation Facility
-#
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
-# furnished to do so, subject to the following conditions:
-#
-# The above copyright notice and this permission notice shall be included in
-# all copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


src/silx/utils/files.py
```diff
@@ -1,35 +1,46 @@
...
-__authors__ = ["V. Valls"]
-__license__ = "MIT"
-__date__ = "19/09/2016"
-
+import logging
 import os.path
 import glob
 
+try:
+    from fabio.utils.cli import relax_ulimit
```
loichuder: Depending on the installed version of `fabio`, `relax_ulimit` may not be available.

Especially for Python 3.10 as the `fabio` version providing `relax_ulimit` is only supporting 3.11+

src/silx/utils/files.py
```diff
@@ -53,3 +64,7 @@ def expand_filenames(filenames):
         else:
             result.append(filename)
     return result
+
+
+def increase_opened_files_limit():
+    relax_ulimit()
```
loichuder: Do we wish to rename the function (https://github.com/silx-kit/silx/pull/4613#discussion_r3414517268) ?

`relax_ulimit` is not very explicit IMO

t20100: Maybe worth renaming in `fabio`, attn @kif 

```suggestion
    _relax_ulimit()
```

IMO, it is worth a docstring and checking that it lands in the doc

src/silx/utils/files.py
t20100: It would have been better to have fabio for python3.10 as well...

```suggestion
    from fabio.utils.cli import relax_ulimit as _relax_ulimit
```

src/silx/utils/files.py
```diff
@@ -1,35 +1,46 @@
...
+    from fabio.utils.cli import relax_ulimit
+except ImportError:
+    try:
+        import resource
+    except ImportError:
+        resource = None
+
+    _logger = logging.getLogger(__name__)
+
+    def relax_ulimit():
```
t20100:
```suggestion
    def _relax_ulimit():
```

t20100 (approved):
I would have prefer it to be exposed at a single place, but it also makes sense to have it in this module.

LGTM

src/silx/utils/files.py
loichuder: Added docstring and the function to the doc:

image


I did not expose the whole module in `autodoc` since we don't want to advertise `expand_filenames`

Issue #4233, opened 2025-03-21
plot: Arcsinh for y-axis


vallsv:
Related to https://github.com/silx-kit/silx/issues/2906

Matplotlib now provides a `asinh` for axis.

https://matplotlib.org/stable/api/scale_api.html#matplotlib.scale.AsinhScale

It could be very convenient to expose it in silx.

Usually people use log scale, but anyway can have negative value, which result in plot with missing data.
The `arcsinh` would solve such problem.

t20100:
This has been implemented in the matplotlib backend in #4529.
Use interaction, changing plot limits and OpenGL backend still need to handle arcsinh.

@EdgarGF93 I assigned myself for the remaining work.

loichuder:
Fixed by https://github.com/silx-kit/silx/pull/4681

Should be part of 3.1.0

PR #4680, opened 2026-07-28
silx.gui.plot: Fixed PlotWidget user interaction and limits for asinh scaled axes

t20100:

### PR summary

This PR is a follow-up of #4529 which left user interaction and axes limits unmanaged for asinh axis scale.
It refactors the previously hard-coded plot axis scale `if isLog: ... else: # linear` by using some `apply/revert` helpers taking the scale as argument.

TODO:
- [x] Check handling of erroneous values in `apply/revert`
- [x] Fix negative min limit issue
- [x] More testing

attn @loichuder 

Related to  #4233

#### AI Disclosure
- [x] Claude Code used to generate tests

### DIFF
--- a/src/silx/gui/plot/PlotInteraction.py
+++ b/src/silx/gui/plot/PlotInteraction.py
@@ -266,20 +266,24 @@ class Pan(_PlotInteractionWithClickEvents):
         previous: float,
     ) -> tuple[float, float]:
         axisScale = axis.getScale()
-        # TODO handle apply/revert errors
         currentScaled = axis_scale.apply(axisScale, current)
         previousScaled = axis_scale.apply(axisScale, previous)
         delta = currentScaled - previousScaled
 
         axisMin, axisMax = axis.getLimits()
-        newMin = axis_scale.revert(
-            axisScale, axis_scale.apply(axisScale, axisMin) - delta
-        )
-        newMax = axis_scale.revert(
-            axisScale, axis_scale.apply(axisScale, axisMax) - delta
-        )
+        try:
+            newMin = axis_scale.revert(
+                axisScale, axis_scale.apply(axisScale, axisMin) - delta
+            )
+            newMax = axis_scale.revert(
+                axisScale, axis_scale.apply(axisScale, axisMax) - delta
+            )
+        except (ValueError, OverflowError):
+            return newMin, newMax
 
-        if axis_scale.inSafeRange(newMin) and axis_scale.inSafeRange(newMax):
+        if axis_scale.inSafeRange(axisScale, newMin) and axis_scale.inSafeRange(
+            axisScale, newMax
+        ):
             return newMin, newMax
         else:
             return axisMin, axisMax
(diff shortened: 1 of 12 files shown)


loichuder:
I added https://github.com/silx-kit/silx/pull/4680/commits/5e92eb0b51e0664d862c7c288677d2a9b5d9a099 to handle the `apply/revert` erroneous values. In essence,

- `apply` never throws but returns `nan` if the value is not valid. We just need to check that the result is then in the `safeRange` since `nan` will never be. Note that this is done differently depending on the context, we may want to have a consistent way of checking this.
- `revert` can throw `OverflowError`. I therefore added `try/catch` blocks where needed. These also catch `ValueError` since it was like that before but I think we only need to catch `OverflowError`.

I also generated some basic units tests for `axis_scale` with Claude (updated the AI policy accordingly).

src/silx/gui/plot/_utils/__init__.py
```diff
@@ -28,19 +28,53 @@
...
+    rangeLimit = maxScaled - minScaled
+    try:
+        min_ = axis_scale.revert(scale, minScaled - minMargin * rangeLimit)
+        max_ = axis_scale.revert(scale, maxScaled + maxMargin * rangeLimit)
+    except (ValueError, OverflowError):
+        return minLimit, maxLimit
+    else:
+        if axis_scale.inSafeRange(scale, min_) and axis_scale.inSafeRange(scale, max_):
+            return min_, max_
+        return min_, max_
```
t20100: There's something fishy here:

```suggestion
            return min_, max_
        return minLimit, maxLimit
```

src/silx/gui/plot/_utils/panzoom.py
```diff
@@ -157,58 +151,52 @@ def applyZoomToPlot(
...
-    :param float min_: Min value of the data range to pan.
-    :param float max_: Max value of the data range to pan.
-                       Must be >= min.
-    :param float panFactor: Signed proportion of the range to use for pan.
-    :param bool isLog10: True if log10 scale, False if linear scale.
+    :param min_: Min value of the data range to pan.
+    :param max_: Max value of the data range to pan.
+                 Must be >= min.
+    :param panFactor: Signed proportion of the range to use for pan.
+    :param axisScale:
```
t20100:
```suggestion
    :param axisScale:
    :param min_: Min value of the data range to pan.
    :param max_: Max value of the data range to pan.
                 Must be >= min.
    :param panFactor: Signed proportion of the range to use for pan.
```

src/silx/gui/plot/PlotInteraction.py
```diff
@@ -265,72 +259,49 @@ def _pixelToData(self, x, y):
...
+        axisMin, axisMax = axis.getLimits()
+        try:
+            newMin = axis_scale.revert(
+                axisScale, axis_scale.apply(axisScale, axisMin) - delta
+            )
+            newMax = axis_scale.revert(
+                axisScale, axis_scale.apply(axisScale, axisMax) - delta
+            )
+        except (ValueError, OverflowError):
+            return newMin, newMax
```
t20100:
```suggestion
            return axisMin, axisMax
```

t20100 (commented):
Spotted a few typos.
Error handling logic looks good to me!

loichuder (approved):

PR #4681, opened 2026-07-28
silx.gui.plot.PlotWidget: Added asinh axis scale support to OpenGL backend 

t20100:

### PR summary

~~Merge PR #4680 first !~~

~~Merge PR #4683 first!~~

This PR is a follow-up of #4529 and #4680 and adds support of asinh axis scale to the OpenGL backend.

TODO:
- [x] Rework to use `axis_scale` helpers
- [x] More testing

attn @loichuder

Related to #4233

#### AI Disclosure
- [x] No AI used


src/silx/gui/plot/backends/glutils/GLPlotCurve.py
```diff
@@ -1238,11 +1239,11 @@ def deduce_baseline(baseline):
...
                     _baseline = numpy.repeat(_baseline, len(self.xData))
-                if isYLog is True:
-                    with numpy.errstate(divide="ignore", invalid="ignore"):
-                        log_val = numpy.log10(_baseline)
-                        _baseline = numpy.where(_baseline > 0.0, log_val, -38)
-                return _baseline
+
+                scaledBaseline = axis_scale.apply(_baseline)
+                if yScale == "log":
+                    scaledBaseline[_baseline <= 0.0] = -38
```
t20100, line 1221: Specific tweak, could be changed to use apply(min(sage range))

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -251,6 +252,12 @@ def ticks(self):
             self._ticks = tuple(self._ticksGenerator())
         return self._ticks
 
+    def applyScale(self, data: float | numpy.ndarray) -> float | numpy.ndarray:
+        return axis_scale.apply(self.scale, data)
+
+    def revertScale(self, data: float | numpy.ndarray) -> float | numpy.ndarray:
+        return axis_scale.revert(self.scale, data)
```
t20100: Wrappers to make it simpler: @loichuder , your suggestion of having `AxisScale` classes would make it neat here: Expose the `AxisScale` instance instead of the helpers and the `scale` string.

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -361,53 +369,130 @@ def _frange(start, stop, step):
...
             start += step
 
+    def _defaultFormatTicks(
+        self, min_: float, max_: float, tickValues: Sequence[float]
+    ) -> tuple[str, list[str]]:
+        """Returns offset/order text and list of tick labels"""
+        self._tickFormatter.axis.set_view_interval(min_, max_)
+        self._tickFormatter.axis.set_data_interval(min_, max_)
+        tickLabels = self._tickFormatter.format_ticks(tickValues)
+        return self._tickFormatter.get_offset(), tickLabels
```
t20100: This is wrapping matplotlib formatter: reused in asinh fallback

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -361,53 +369,130 @@ def _frange(start, stop, step):
...
+            )
             dataMin = 1.0
             if dataMax < dataMin:
                 dataMax = 1.0
 
         if dataMin != dataMax:  # data range is not null
             (x0, y0), (x1, y1) = self.displayCoords
 
-            if self.isLog:
+            if self.scale == "asinh":
```
t20100: First implementation reused log10 tick positionner... only works for ranges >=1 ;)
So implemented a new tick positioner for asinh.

For the future, I would seriously consider relying on matplotlib machinery for all axes scales (see https://github.com/matplotlib/matplotlib/blob/4f746106634ea5d24402a34a9be8ca53e290b95a/lib/matplotlib/ticker.py) but I wanted to finish the asinh feature first.

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -361,53 +369,130 @@ def _frange(start, stop, step):
...
+                    )
+                uniqueLog10Ticks = set(
+                    pos
+                    for pos in log10Ticks
+                    if numpy.isfinite(pos) and dataMin <= pos <= dataMax
+                )
+                if dataMin * dataMax < 0:  # crossing zero: ensure 0
+                    uniqueLog10Ticks.add(0.0)
+
+                tickPositions = numpy.array(sorted(uniqueLog10Ticks))
```
t20100: Inspired from https://github.com/matplotlib/matplotlib/blob/4f746106634ea5d24402a34a9be8ca53e290b95a/lib/matplotlib/ticker.py#L2814
We should switch to it at some point.

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -361,53 +369,130 @@ def _frange(start, stop, step):
...
+
+                if dataMin * dataMax < 0:
+                    # Remove ticks too close to 0
+                    scaledTicks = self.applyScale(tickPositions)
+                    distanceToZero = axisLengthInches * abs(
+                        scaledTicks / (scaledMax - scaledMin)
+                    )
+                    tickPositions = tickPositions[
+                        numpy.logical_or(distanceToZero == 0, distanceToZero >= 0.5)
+                    ]
```
t20100: custom removal of ticks too close to the zero tick based on on-screen distance

src/silx/gui/plot/_utils/panzoom.py
```diff
@@ -194,7 +194,7 @@ def applyPan(
 
     newMin = revert(axisScale, scaledMin + scaledOffset)
     newMax = revert(axisScale, scaledMax + scaledOffset)
-    if isValid(newMin) and isValid(newMax):
+    if isValid(axisScale, newMin) and isValid(axisScale, newMax):
```
t20100, line 197: Fix bug from previous PR on asinh support

src/silx/gui/plot/backends/glutils/GLPlotImage.py
```diff
@@ -40,6 +40,9 @@
 from .GLPlotItem import GLPlotItem
 
 
+_SHADER_AXIS_SCALES: dict[str, int] = {"linear": 0, "log": 1, "asinh": 2}
```
t20100: For completeness, OpenGL backend supports image display with log (and now asinh) axes scale.

This is currently disabled; to enable it, apply this patch to `/src/silx/gui/plot/items/image.py`:
```
@@ -148,8 +148,7 @@ class ImageBase(DataItem, LabelsMixIn, DraggableMixIn, AlphaMixIn):
     def _isPlotLinear(self, plot):
         """Return True if plot only uses linear scale for both of x and y
         axes."""
-        xAxis, yAxis = self._getAxisInstances(plot)
-        return xAxis.getScale() == xAxis.LINEAR and yAxis.getScale() == yAxis.LINEAR
+        return True
 
     def _getBounds(self) -> types.ItemBounds | None:
         plot = self.getPlot()
```

IMO we need to decide whether we enable it or we remove the code supporting non-linear axes scale for images.

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
loichuder: I would suggest to extract the tick generation into utils since it is becoming quite lengthy

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
```diff
@@ -1150,37 +1209,18 @@ def transformedDataY2ProjMat(self):
...
     def dataToPixel(self, x, y, axis="left"):
         """Convert data coordinate to widget pixel coordinate."""
         assert axis in ("left", "right")
 
         trBounds = self.transformedDataRanges
 
-        xDataTr = self.__applyLog(x, self.xAxis.isLog)
-        if xDataTr is None:
+        xDataTr = self.xAxis.applyScale(x)
+        if isinstance(xDataTr, numbers.Real) and not numpy.isfinite(xDataTr):
```
loichuder, line 724: This changes the behaviour when `xDataTr` is `NaN`

src/silx/gui/plot/backends/glutils/GLPlotImage.py
loichuder: I can't think of a situation where this could be used.

Let's merge it like that and we can remove it if we want afterwards.

src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -1126,36 +1129,36 @@ def addImage(self, data, origin, scale, colormap, alpha):
         else:
             raise RuntimeError(f"Unsupported data shape {data.shape}")
 
-        # TODO is this needed?
-        if self._plotFrame.xAxis.isLog and image.xMin <= 0.0:
-            raise RuntimeError("Cannot add image with X <= 0 with X axis log scale")
-        if self._plotFrame.yAxis.isLog and image.yMin <= 0.0:
-            raise RuntimeError("Cannot add image with Y <= 0 with Y axis log scale")
+        if not axis_scale.isValid(self._plotFrame.xAxis.scale, image.xMin):
```
loichuder: This changes the behaviour if `xMin` is `NaN` (raises a error while it did not previously)

src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -1126,36 +1129,36 @@ def addImage(self, data, origin, scale, colormap, alpha):
...
     ):
         x = numpy.asarray(x)
         y = numpy.asarray(y)
 
-        # TODO is this needed?
-        if self._plotFrame.xAxis.isLog and x.min() <= 0.0:
-            raise RuntimeError("Cannot add item with X <= 0 with X axis log scale")
-        if self._plotFrame.yAxis.isLog and y.min() <= 0.0:
-            raise RuntimeError("Cannot add item with Y <= 0 with Y axis log scale")
+        if not axis_scale.isValid(self._plotFrame.xAxis.scale, x.min()):
```
loichuder: Same as above

src/silx/gui/plot/backends/BackendOpenGL.py
```diff
@@ -1604,14 +1592,23 @@ def isXAxisInverted(self) -> bool:
...
-        if self._plotFrame.xAxis.isLog or self._plotFrame.yAxis.isLog:
-            return False
-        else:
-            return self._keepDataAspectRatio
+        return False if self._hasNonLinearScales() else self._keepDataAspectRatio
 
     def setKeepDataAspectRatio(self, flag: bool):
-        if flag and (self._plotFrame.xAxis.isLog or self._plotFrame.yAxis.isLog):
+        if flag and self._hasNonLinearScales():
             _logger.warning("KeepDataAspectRatio is ignored with log axes")
```
loichuder:
```suggestion
            _logger.warning("KeepDataAspectRatio is ignored with non linear axes")
```

src/silx/gui/plot/backends/glutils/GLPlotFrame.py
loichuder: Nevermind, it's fine. I am adding a test specifically for this though,

loichuder (approved):

PR #4682, opened 2026-07-29
silx.gui.data: Added support of asinh axis scale

t20100:

### PR summary

~~Merge PR #4680 first!~~

Small follow-up PR of #4680 adding use of `AxisScaleType` in `silx.gui.data`

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/gui/data/NXdataWidgets.py
+++ b/src/silx/gui/data/NXdataWidgets.py
@@ -28,7 +28,7 @@ __license__ = "MIT"
 __date__ = "12/11/2018"
 
 import logging
-from typing import Literal
+from typing import Sequence
 
 import h5py
 import numpy
@@ -46,6 +46,7 @@ from silx.io.commonh5 import Dataset
 from silx.math.calibration import ArrayCalibration, LinearCalibration, NoCalibration
 
 from ._utils import getAxesCalib, setImageCoords
+from ..plot.items.types import AxisScaleType
 from ..utils import blockSignals
 from ...utils.deprecation import deprecated, deprecated_warning
 from .ArrayCurvePlot import ArrayCurvePlot as _ArrayCurvePlot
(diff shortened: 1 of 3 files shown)


t20100:
This PR makes it less safe to erroneous scale names.
Should we handle this here (not the best IMO), early when getting information from NXdata or in `setScale` which currently contains an `assert`?

loichuder:
> Should we handle this here (not the best IMO), early when getting information from NXdata or in setScale which currently contains an assert?

I'd say both in `NXdata` and `setScale`.

I think we always want to handle erroneous scale names when scalling `setScale` for any axis. Ideally, it should be more explicit than a simple `assert`.

Also, the validity of NXdata is tested when parsing it so I think we should also handle erroneous scale names there as well. That would avoid errors down the line whose origin may be difficult to track down.

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -129,34 +130,34 @@ def getPlot(self):
...
+        x: numpy.ndarray,
+        values: Sequence[numpy.ndarray],
+        yerror: numpy.ndarray | None = None,
+        xerror: numpy.ndarray | None = None,
+        ylabel: numpy.ndarray | None = None,
+        xlabel: numpy.ndarray | None = None,
+        title: str = "",
+        scatter_titles: Sequence[str] | None = None,
+        xscale: AxisScaleType | None = None,
+        yscale: AxisScaleType | None = None,
```
loichuder, line 143: 💪

loichuder (approved):

loichuder:
> > Should we handle this here (not the best IMO), early when getting information from NXdata or in setScale which currently contains an assert?
> 
> I'd say both in `NXdata` and `setScale`.
> 
> I think we always want to handle erroneous scale names when scalling `setScale` for any axis. Ideally, it should be more explicit than a simple `assert`.
> 
> Also, the validity of NXdata is tested when parsing it so I think we should also handle erroneous scale names there as well. That would avoid errors down the line whose origin may be difficult to track down.

I'll do a follow-up PR

t20100:
> I think we always want to handle erroneous scale names when scalling setScale for any axis. Ideally, it should be more explicit than a simple assert.

:+1: 

> I'll do a follow-up PR

Thanks!

PR #4685, opened 2026-07-30
silx.io.nxdata.parse.NXdata: add support for asinh scale type in SILX_style

loichuder:

### PR summary

In https://github.com/silx-kit/silx/pull/4682#issuecomment-5131006726, I wanted to add checks in the NXdata parsing but it turns out that they are already there. I just needed to add support for the `asinh` scale type.

I also added related tests.

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/io/nxdata/parse.py
+++ b/src/silx/io/nxdata/parse.py
@@ -1,30 +1,7 @@
-# /*##########################################################################
-#
-# Copyright (c) 2017-2022 European Synchrotron Radiation Facility
-#
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
-# furnished to do so, subject to the following conditions:
-#
-# The above copyright notice and this permission notice shall be included in
-# all copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


src/silx/io/nxdata/parse.py
```diff
@@ -41,11 +41,13 @@
...
 
 import json
-from typing import Any
+from typing import Any, Literal, get_args
 
 import h5py
 import numpy
 
+# from ...gui.plot.items.types import AxisScaleType <-- circular import
+AxisScaleType = Literal["linear", "log", "asinh"]
```
loichuder: Not sure how to fix this without changing too much so I duplicated the type for now.

src/silx/io/nxdata/parse.py
```diff
@@ -128,23 +131,20 @@ def __init__(self, nxdata):
...
                 )
             else:
                 self._signal_scale_type = scale_type
 
-    axes_scale_types = property(
-        lambda self: self._axes_scale_types,
-        doc="Tuple of NXdata axes scale types (None, 'linear' or 'log'). List[str]",
-    )
+    @property
+    def axes_scale_types(self) -> tuple[AxisScaleType | None, ...]:
```
loichuder: Converted those properties to add typing

src/silx/io/nxdata/parse.py
t20100: adding a `types` module at the root of `silx`... but IMO it also makes sense to keep both definitions separated since they are in 2 different top packages and actually represents different types even if they match.

src/silx/io/nxdata/parse.py
```diff
@@ -66,6 +68,8 @@
 __license__ = "MIT"
 __date__ = "24/03/2020"
 
+_SCALES = get_args(AxisScaleType)
```
t20100: didn't knew this, great!

t20100 (approved):

src/silx/io/nxdata/parse.py
loichuder: Ok I'll update the comment to reflect this

PR #4687, opened 2026-08-03
silx.opencl.common: Fixed allow 0 in get_platform() and get_device()

takluyver:
### PR summary

It appears that 0 is valid as a platform and device ID, but the `get_platform()` and `get_device()` methods check `> 0` and thus reject it. This changes `>` to `>=`.

#### AI Disclosure
- [x] No AI used


loichuder (approved):
How was this unnoticed for so long lol

Thanks for the fix 👍 

attn @kif

PR #4691, opened 2026-08-06
Doc: fix warning when sphinx-autodoc-typehints parses `numpy.trapezoid`

loichuder:

### PR summary

Latest documentation build had a new warning (that get treated as an error):

```
 /opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/numpy/lib/_function_base_impl.py:4922: WARNING: Cannot resolve forward reference in type annotations of "silx.gui.plot.CurvesROIWidget.trapezoid" (module numpy.lib._function_base_impl): name 'ScalarT' is not defined [sphinx_autodoc_typehints.forward_reference]
```

It seems `sphinx_autodoc_typehints` cannot parse properly the new typing of `numpy.trapezoid`. Both libraries are dependencies so this is not something we can fix ourselves.

What we can do, however, is to avoid exposing `trapezoid` when importing it in `silx.gui.plot.CurvesROIWidget.`. This way, it stays invisible to `sphinx_autodoc_typehints` and does not try to parse it. 

#### AI Disclosure
- [x] Claude code used to investigate the issue and produce the solution

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -82,8 +82,6 @@ if importlib.util.find_spec("sphinx_autodoc_typehints"):
 
     always_document_param_types = True
 
-# Ignore warnings about type hints that numpy's stubs can't resolve (e.g. trapezoid's ScalarT).
-suppress_warnings = ["sphinx_autodoc_typehints.forward_reference"]
 
 autodoc_member_order = "bysource"
 
--- a/src/silx/gui/plot/CurvesROIWidget.py
+++ b/src/silx/gui/plot/CurvesROIWidget.py
@@ -39,10 +39,12 @@ import functools
 import numpy
 from typing import Union
 
+# `sphinx-autodoc-typehints` can't parse the trapezoid signature for recent numpy versions (2.5+)
+# So don't expose it so it is not picked up by `sphinx-autodoc-typehints`
 try:
-    from numpy import trapezoid
+    from numpy import trapezoid as _trapezoid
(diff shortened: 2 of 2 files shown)


doc/source/conf.py
```diff
@@ -82,8 +82,6 @@
 
     always_document_param_types = True
 
-# Ignore warnings about type hints that numpy's stubs can't resolve (e.g. trapezoid's ScalarT).
```
loichuder, line 85: Revert the tentative fix I pushed on main: 9ba46e2977ccbb49d6156d79581bae59a26fd0da

Claude discovered that warnings can't be suppressed when they are treated as errors as we do in our documentation build:

> I don't have a doc page that spells this out in prose — Sphinx's own docs for -W/--exception-on-warning and for suppress_warnings don't mention this interaction. What I have is the source itself, which is the actual evidence:
>
> [sphinx/util/logging.py](https://www.sphinx-doc.org/en/master/_modules/sphinx/util/logging.html) — the setup() function:
>
> ```python
> warning_handler = WarningStreamHandler(SafeEncodingWriter(warning))
> if app._exception_on_warning:
>    warning_handler.addFilter(_RaiseOnWarningFilter())
> warning_handler.addFilter(WarningSuppressor(app))
> ```
> _RaiseOnWarningFilter is registered first and raises immediately on any warning; 
> WarningSuppressor (which reads suppress_warnings) never runs.

payno (approved):
This is a nice one 👍

PR #4703, opened 2026-08-26
silx.gui.data.NXdataWidgets.ArrayImagePlot: fix ValueError for images with irregular axes

### LINKED ISSUE #4701: NXDataWidgets ArrayImagePlot fails calling setScale since 3.1.0
Since upgrading to version 3.1.0, we get an error when viewing NXdata for images from some HDF5 files: 

`ERROR:silx.app.view.main:<class 'ValueError'> None is not one of the supported scales: {'linear', 'log', 'asinh'}`.

<details>
<summary>Click to see detail</summary>

The error occurs as soon as the file is opened, and it tries to load the default view.  

We see a blank image.  The _HDF5_, and _Curve_ views still work fine.

Image

```
$ silx view VB_250eV_PE20eV_scan-3427\ 0001.h5
ERROR:silx.app.view.main:An error occured in silx view:
ERROR:silx.app.view.main:<class 'ValueError'> None is not one of the supported scales: {'linear', 'log', 'asinh'}   File "/opt/conda/envs/silx/lib/python3.14/site-packages/silx/gui/data/DataViewer.py", line 298, in __setDataInView
    self.__currentView.setData(self.__displayedData)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/silx/lib/python3.14/site-packages/silx/gui/data/DataViews.py", line 1328, in setData
    widget.setImageData(
    ~~~~~~~~~~~~~~~~~~~^
        [nxd.signal] + nxd.auxiliary_signals,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        title=nxd.title,
        ^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/conda/envs/silx/lib/python3.14/site-packages/silx/gui/data/NXdataWidgets.py", line 321, in setImageData
    self._updateImageAxes()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/conda/envs/silx/lib/python3.14/site-packages/silx/gui/data/_BaseImagePlot.py", line 82, in _updateImageAxes
    self._addItemToPlot(xAxis, yAxis)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/opt/conda/envs/silx/lib/python3.14/site-packages/silx/gui/data/NXdataWidgets.py", line 360, in _addItemToPlot
    self._plot.getXAxis().setScale(xAxisScale)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/opt/conda/envs/silx/lib/python3.14/site-packages/silx/gui/plot/items/axis.py", line 224, in setScale
    raise ValueError(
        f"{scale} is not one of the supported scales: {self._SCALES}"
    )
...
```

</details>

With version 3.0.1 (and 2.2.2) these kind of files were loading without any problems.  Example below.

<details>
<summary>Click to see screenshots from 3.0.1</summary>

Image

Image

</details>

I see there was a 3.1.1 release, but it isn't on conda-forge yet, so I haven't tried it.  The changes between 3.1.0 and 3.1.1 don't seem to touch this area of the code, though.

---
P.S. Thanks for this useful tool!

loichuder:

### PR summary

Fix #4701 

This comes from an untested conditional branch of `ArrayImagePlot` when the image has defined but non-affine axes. In this case, the image get plotted as a scatter and the scales get applied to the X and Y axes. Since I was not testing if the scales were `None`, it raised a `ValueError` in case the scales were not defined.

I added a check for `None` and a non-regression test. 

#### AI Disclosure
- [x] Claude code used to generate an early version of the regression test that I tuned by hand.

### DIFF
--- a/src/silx/gui/data/NXdataWidgets.py
+++ b/src/silx/gui/data/NXdataWidgets.py
@@ -352,13 +352,11 @@ class ArrayImagePlot(BaseImagePlot):
             if self._axesScales:
                 xAxisIndex, yAxisIndex = self._getXYIndices()
                 xAxisScale = self._axesScales[xAxisIndex]
+                if xAxisScale is not None:
+                    self._plot.getXAxis().setScale(xAxisScale)
                 yAxisScale = self._axesScales[yAxisIndex]
-            else:
-                xAxisScale = None
-                yAxisScale = None
-
-            self._plot.getXAxis().setScale(xAxisScale)
-            self._plot.getYAxis().setScale(yAxisScale)
+                if yAxisScale is not None:
+                    self._plot.getYAxis().setScale(yAxisScale)
 
             xScatter, yScatter = numpy.meshgrid(xAxis, yAxis)
             self._plot.addScatter(
--- a/src/silx/gui/data/test/test_nexus_dataviewer.py
+++ b/src/silx/gui/data/test/test_nexus_dataviewer.py
@@ -65,6 +65,26 @@ def test_rgb_image_with_interpretation(qapp, qWidgetFactory, tmp_path):
         assert isinstance(plot.getImage("rgb"), ImageRgba)
 
 
+def test_image_with_non_affine_axis_becomes_scatter(qapp, qWidgetFactory, tmp_path):
+    widget: DataViewer = qWidgetFactory(DataViewer)
+    with h5py.File(tmp_path / "test.h5", "w") as h5file:
+        h5file.attrs["NX_class"] = "NXdata"
+        h5file.attrs["signal"] = "signal"
+        h5file.create_dataset(name="signal", data=numpy.random.random((10, 10)))
+        h5file.create_dataset(name="non_affine_x", data=numpy.logspace(0, 1, 10))
+        h5file.create_dataset(name="y", data=numpy.arange(10))
+        h5file.attrs["axes"] = ["y", "non_affine_x"]
+
+        widget.setData(h5file["/"])
+
+        qapp.processEvents()
+
+        imageView = widget.currentAvailableViews()[0]
+        plot: Plot2D = imageView.getWidget().getPlot()
+        assert plot.getImage() is None
+        assert plot.getScatter() is not None
+
+
 @pytest.mark.skipif(
     sys.version_info.major == 3 and sys.version_info.minor == 14,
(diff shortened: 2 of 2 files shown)


loichuder:
@payno This is a bugfix. Should I target the `3.1` branch instead of `main`?

woutdenolf (approved):

src/silx/gui/data/NXdataWidgets.py
```diff
@@ -352,13 +352,11 @@ def _addItemToPlot(self, xAxis, yAxis):
...
                 xAxisScale = self._axesScales[xAxisIndex]
+                if xAxisScale is not None:
+                    self._plot.getXAxis().setScale(xAxisScale)
                 yAxisScale = self._axesScales[yAxisIndex]
-            else:
-                xAxisScale = None
-                yAxisScale = None
-
-            self._plot.getXAxis().setScale(xAxisScale)
-            self._plot.getYAxis().setScale(yAxisScale)
```
woutdenolf, line 361: By removing this we are assuming that `setScale(None)` is no-op.

woutdenolf: Ok it is worse than a no-op, it is the cause of the ValueError we are fixing in this PR.

https://github.com/silx-kit/silx/blob/645343a2318c134d893a9830dcf14adfc2fb7f88/src/silx/gui/plot/items/axis.py#L223-L226

I'm wondering whether there are other places in the code where we could pass a None.

woutdenolf: At least when looking for `Axis().setScale(` this is the only instance.

woutdenolf:
I would say target `main` and then a back-port PR if needed.

src/silx/gui/data/NXdataWidgets.py
loichuder: > I'm wondering whether there are other places in the code where we could pass a None.

I checked all instances of `setScale` (not only `Axis().setScale`) and this is the only place I could find where `None` could be passed

ajoubertza:
Thanks for the fix @loichuder 🙂

PR #4711, opened 2026-09-01
silx.gui.icons, silx view, silx compare: Fixed support of qt6<6.5

### LINKED ISSUE #4709: silx.gui.icons requires qt>=6.8  which is not what is advertised
Error on ubuntu24.04 (Qt 6.4.2):
'QStyleHints' object has no attribute 'colorScheme' File "/usr/local/lib/python3.12/dist-packages/silx/gui/icons.py"

The problem is here: https://github.com/silx-kit/silx/blob/98c0aa6445ce1af724586dde348e0852c85b6681/src/silx/gui/icons.py#L125

Sounds simple to check the version of Qt to enable color scheme management.

t20100:

### PR summary

This PR avoids calling colorscheme API when the version of qt6 does not support it.

closes #4709

#### AI Disclosure
- [x] No AI used

### DIFF
--- a/src/silx/app/compare/main.py
+++ b/src/silx/app/compare/main.py
@@ -28,6 +28,7 @@ import sys
 import logging
 import argparse
 import silx
+from packaging.version import Version
 from silx.gui import qt
 from silx.app.utils import parseutils
 from silx.app.compare.CompareImagesWindow import CompareImagesWindow
@@ -84,7 +85,7 @@ def mainQt(options):
         silx.config.DEFAULT_PLOT_BACKEND = "opengl"
 
     app = qt.QApplication([])
-    if qt.BINDING != "PyQt5":
+    if Version(qt.qVersion()) >= Version("6.8.0"):
         app.styleHints().setColorScheme(qt.Qt.ColorScheme.Light)
 
     window = CompareImagesWindow(backend=backend, settings=settings)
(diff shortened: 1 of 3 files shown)


src/silx/app/compare/main.py
```diff
@@ -84,7 +85,7 @@ def mainQt(options):
         silx.config.DEFAULT_PLOT_BACKEND = "opengl"
 
     app = qt.QApplication([])
-    if qt.BINDING != "PyQt5":
+    if qt.BINDING != "PyQt5" and Version(qt.qVersion()) >= Version("6.8.0"):
```
loichuder: Is not the first condition included in the second?

src/silx/gui/icons.py
```diff
@@ -39,6 +39,12 @@
 import silx.resources
 from silx.utils import weakref as silxweakref
 
+from packaging.version import Version
+
+_QT_HAS_COLOR_SCHEME = qt.BINDING != "PyQt5" and Version(qt.qVersion()) >= Version(
```
loichuder: Why making a variable here instead of inlining the condition like the other files?

t20100: It's called once at the other places while it is called for each icon repaint here.

src/silx/app/compare/main.py
t20100: Indeed, I'll remove it

loichuder (approved):

